PR #46, opened 2019-11-08
[WIP] C++ backend

cmnrd:
This is a C++ backend for lingua franka. While there are still some open issues, the backend is pretty complete and covers most aspects of the language. I adapted most of the C test programms to show that the new C++ works.

The C++ backend is a pure code generator and does not provide a runtime library by itself. It relies
instead on [enactor](https://github.com/tud-ccc/enactor), a standalone framework for the creation of reactor networks in C++. This dependency is automatically pulled in when building the target.

Please have alook at this and let me know what you think.

These are some of the more important open issues:
- Support for preambles is somewhat limited. The C++ backend generates a source and a header file for each reactor and currently the preamble goes into the header. This works well with includes, but causese problems, e.g., with function definitions as those should be placed in a source file.
- While local deadlines (on reactions) are supported by this backend, deadlines on input ports are not supported. I am somewhat confused by the deadline's ability to schedule another action at the same logical time. This needs dome clarification.
- While there are test programs, they are not yet integrated with the test system of LF.

PR #56, opened 2019-11-14
Updates in the C++ backend

cmnrd:
This includes the following updates to the C++ bakend:
- Some fixes so that all tests succeed
- the cmake issue mentioned in #46 should now be resolved
- I renamed the external framework that the code generator depends on to [reactor-cpp](https://github.com/tud-ccc/reactor-cpp)


cmnrd:
Apparently the gcc version that the default vm of Travis CI uses was too old for my C++14 code. I tried to set the distribution to bionic to get a more recent version. This makes all C++ tests pass, but now the C test fails with a weired error.

cmnrd:
C++ tests do now succeed without any changes to the travis config. But there is something broken in the build of the C target.

lhstrh:
Thanks for the fixes! I'll look into the build problem.

PR #105, opened 2020-02-15
ReflexGame: ensure a deterministic display

jeanguyomarch:
Without an explicit flush of the standard output, the message
  "Hit Return or Enter!"
may not be visible, because the stdio buffers are not necessarily
flushed. We now ensure that this message is always visible.


lhstrh:
Ah, good catch... Thanks for the patch!

PR #116, opened 2020-03-12
Changed analyzeResource and refreshProject to work for windows.

mkuo005:
The Hello World example in https://github.com/icyphy/lingua-franca/wiki/Downloading-and-Building#building-the-ide does not seem to work in windows.

There are compile errors: null pointer and pattern matching errors in /org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend when executing in windows environment. 

Suggest changing the following
1) `private def analyzeResource(Resource resource)` function
use File object to get file name and directory rather than using finding the last '/' so that the code works on all platforms.

2) `protected def refreshProject()` function
Test for the File.seprator to determine if it is running on Unix based or Windows based systems. If the system is running on windows add an extra File.separator to each File.separator (i.e., "\\") since the path separator used by windows "\\" is an escape character.


lhstrh:
Thanks for these fixes! None of the active developers work on Windows, which increases the likelihood of these kinds of bugs creeping in...

Issue #140, opened 2020-04-06
Kieler fails to install


cxbrooks:
I'm having problems installing Kieler under Eclipse.

I created https://rtsys.informatik.uni-kiel.de/jira/browse/KIPRA-2035, but thought I should also report the issue here.

After the installation failed, I reproduced the problem as follows:

 First, I did:
`rm -rf ~/.eclipse ~/.p2 ~/.m2`

Then I installed Eclipse 2020-03 from the installer and selected "Eclipse for Java and DSL Developers," which includes Xtext.

Then,  followed the instructions at https://github.com/icyphy/lingua-franca/wiki/diagrams#enabling-diagram-synthesis 
Go to Help -> Install New Software -> Add Update Site
Add http://rtsys.informatik.uni-kiel.de/~kieler/updatesite/ 
Selected 
* Kieler Lightweight Diagrams Generic View
* Kieler Lightweight Diagrams Runtime
* Kieler Vector Graphics Export (now called KIELER Lightweight Diagrams Vector Graphics Export Powered by FreeHEP 2.1.1)

The installation failed, the following message dialog appeared:

> An error occurred while collecting items to be installed
> session context was:(profile=_Users_cxh_eclipse_dsl-2020-03_Eclipse.app_Contents_Eclipse, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
> No repository found containing: osgi.bundle,org.eclipse.elk.alg.common,0.6.1
> No repository found containing: osgi.bundle,org.eclipse.elk.alg.layered,0.6.1
> No repository found containing: osgi.bundle,org.eclipse.elk.core,0.6.1
> No repository found containing: osgi.bundle,org.eclipse.elk.core.service,0.6.1
> No repository found containing: org.eclipse.update.feature,org.eclipse.elk.feature,0.6.1
> No repository found containing: osgi.bundle,org.eclipse.elk.graph,0.6.1

I went back to Help -> Install New Software and in the Works with drop down, selected Elk 6.1, which corresponds with http://download.eclipse.org/elk/updates/releases/0.6.1/

It is not clear which things should be installed, so I selected:
* Eclipse Layout Kernel (Incubation) - Core components
* Eclipse Layout Kernel (Incubation) - Layout Algorithms
* Eclipse Layout Kernel (Incubation) - SDK

The following dialog appeared:

> An error occurred while collecting items to be installed
[...]

PR #141, opened 2020-04-06
Dependency Update for Diagrams

a-sr:
These changes update the diagrams plugin to the API of the latest Klighd/ELK release.

Note that, as soon as the PR is merged, everyone has to update their development eclipse to get the latest release. Otherwise there will be compilation errors.

I do not know if the diagram plugin is build in CI and which update site is used there but the build might break.


a-sr:
I just created issue #142 to discuss how to handle such a change in dependencies in the future.

a-sr:
Issue #140 revealed that the new Klighd version cannot be installed via the current update site.
Hence, if this PR is merged everyone might need to register the new Klighd update site (https://kieler.github.io/KLighD/) and remove all the old ones (as described [here](https://github.com/icyphy/lingua-franca/issues/140#issuecomment-610448114)) to be able to successfully update Klighd and ELK.

PR #156, opened 2020-05-04
Multi Ports and Instances

cmnrd:
After thinking about last week's discussion on multi ports, I decided to try it out and come up with a syntax and basic C++ support. This PR is a work in progress and intended to show what I came up with. There are still missing bits and the syntax is up for discussion.

I decided to use the array syntax to declare multi ports which is different from what we discussed. 
```
input[42] in:unsigned;
output42] in:unsigned;
```
A similar syntax is supported for instantiations:
```
foo = new[42] Foo();
```
The advantage of the array syntax is that it allows to have fixed size and variable sized arrays. For variable sized array, new ports could be created as needed at runtime, for instance, when a mutation
creates a new connection.

When connecting two multi ports, all the port instances are automatically connected (see [ArrayPorts.lf)](https://github.com/icyphy/lingua-franca/blob/2a54c6c337fb3d438b80a94a3633ca6ecaab6369/test/Cpp/ArrayPorts.lf). Similarly, a multi port can be connected to the input port of an instance array like so:
``` 
target Cpp;

reactor Source {
    output[4] out:unsigned;
    
    reaction (startup) -> out {=
        for (unsigned i; i < out.size(); i++) {
            out[i].set(i);
        }
    =}
}

reactor Sink(id:unsigned(0)) {
    input in:unsigned;
    
    reaction (in) {=
        std::cout << "Received " << *in.get() << '\n';
        if (*in.get() != id) {
            std::cerr << "Error: expected " << id << "!\n";
            exit(1);
        }
    =}
}

main reactor ArrayReactorInstances {
    source = new Source();
    sink = new[4] Sink();
    source.out -> sink.in;
}
```
In this example the Sink reactor is instantiated 4 times. Note that the `id` parameter is assigned automatically, giving each instance its own id. There is only one instance of the source reactor which has an array output port. When connecting, the array port instances get automatically connected to the input ports of the different sink instances. Similar examples exist for array instance to array instance connections ([ArrayReactorInstances.lf](https://github.com/icyphy/lingua-franca/blob/2a54c6c337fb3d438b80a94a3633ca6ecaab6369/test/Cpp/ArrayReactorInstances.lf)) and array instance to array input connections ([ArrayPortToArrayInstance.lf](https://github.com/icyphy/lingua-franca/blob/2a54c6c337fb3d438b80a94a3633ca6ecaab6369/test/Cpp/ArrayPortToArrayInstance.lf)). Of course, the validator needs to check that the array size on the left side of the connection would match the array size on the right size. It would also be possible to support individual connections (`source.out[3] -> sink[0].in`) or range based connections (`source.out[2-3]` -> sink[0-1].in).

In the C++ target, the array ports are simply placed in a fixed size container. This ensures that there are still independent port instances while providing the possibility to iterate over all ports in reactions. For instance, one can do this:
```
input[42] in:unsigned;
reaction(in) {
    for (auto& i : in) {
        if (i.is_present()) {
            unsigned value = *i.get();
            // do something useful
        }
    }
}
```
We could also allow reactions to individual instances:
```
input[42] in:unsigned;
reaction(in[0]) {
    unsigned value = *in.get();
    // do something useful
}
reaction(in[1]) {
    unsigned value = *in.get();
    // do something useful
}
// ....
```

I was also able to simplify the [Philosopher Benchmark](https://github.com/icyphy/lingua-franca/blob/2a54c6c337fb3d438b80a94a3633ca6ecaab6369/benchmark/Cpp/Savina/PhilosophersSingleReaction.lf) with this approach.
Philosopher Of course, the diagram view should somehow indicate that there are multiple instances of `Philosopher` and that all the connections to/from the philosophers are 'multi connections'.

What do you think of this?

### DIFF
--- a/benchmark/Cpp/Savina/PhilosophersSingleReaction.lf
+++ b/benchmark/Cpp/Savina/PhilosophersSingleReaction.lf
@@ -9,7 +9,6 @@ reactor Arbitrator(count:unsigned(10000), verbose:bool(false)) {
     public preamble {=
         constexpr unsigned num_philosophers = 20;
         using ForkArray = std::array<bool, num_philosophers>;
-        using PhilosopherArray = std::array<Philosopher*, num_philosophers>;
     =}
     
     private preamble {=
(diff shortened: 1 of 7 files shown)


lhstrh:
Very nice! I like this proposal quite a bit. I do think there is an advantage in being able to express bounds on the multiplexing, which is the main difference with the approach we had discussed so far.

One thing I think would improve clarity, is to use `()` instead of `[]` (to avoid confusion with ordinary arrays), but I'm not 100% sure about this. If we use `()`, then perhaps it is not abundantly clear that `input()` denotes a multiplex port (an unbounded one). Maybe it should be `input(*)` or `input(:)` in that case? Not sure. WDYT?

I'm also thinking that perhaps it would be better to adopt MatLAB's array indexing syntax, in which case we would use `:` to express ranges rather than `-`, which could be misunderstood to denote subtraction...

edwardalee:
I like this!  I think we should stick with [...] rather than (...) because, at least in the C target, an array is an accurate description and the C code will access everything using [...].  Also, an array-typed multi-input would become an array of arrays in C:  in[id][index] and in_is_present[id], which seems natural to me.

I presume that an unspecified number of inputs would be given as:
```
    input[] in:int;
```
right?

lhstrh:
If we were to keep the `[]` but use `:` for ranges that would be a variation on the MatLAB syntax because MatLAB uses `()`. That's OK -- just a thing to note. 

I agree that is seems natural from a C perspective to use `[]`, but that would actually be a reason for me to argue _against_ using this notation, simply because multiplex ports are a Lingua Franca construct, not a target-language construct. Therefore, using `()` would set multiplex ports apart nicely from ordinary arrays in most target languages (even though they would be accessed in target code exactly the way you describe...)

edwardalee:
I don't understand the proposal for : instead of -. I don't see any - in the proposal. What are you proposing exactly?

lhstrh:
I'm specifically referring to the following text that Christian wrote:
> It would also be possible to support individual connections (source.out[3] -> sink[0].in) or range based connections (source.out[2-3] -> sink[0-1].in).

edwardalee:
Ah, OK, I missed that. The Matlab syntax would also support having a stride:
```
    source.out[1:10:2] -> sink[1:5];
```
(I'm not sure whether the stride in Matlab is the second or third term.)

lhstrh:
It is `start:step:end` in MatLAB. I think it's the other way around in Python (the step size is an optional third). We could also adopt the Python style, which does use square brackets...

edwardalee:
I like the Python style better.

lhstrh:
OK, I think it's most important that we adopt a style that is already known. It's also easier to explain that way. I'd be fine with adopting the Python syntax. Shall we go ahead, merge this into master, and add the fancy indexing to our to do list?

edwardalee:
Sure.

cmnrd:
I see the `[]` already as an LF construct as it is used in types. `int[]` is a list of integers. In C this translates to an int array. In C++, this becomes a `std::vector<int>` and in other languages it might produce some other list type. Therefore, it seems best to me to stick to the `[]` syntax rather than introducing a new one. I also think that `[]` is very common across languages and therefore is widely understood. About ranges: I also like the python syntax best.

PR #178, opened 2020-06-23
TypeScript Federates

mew2ub:
These changes make it possible to write federated reactors in TypeScript

- Most of the new code is in TypeScriptGenerator. Non-federated reactors should be unaffected. If the LF program has a federated reactor the code generator will produce a separate .js and .ts file for each federate (i.e. top level reactor of the federated reactor). Following the C Generator's naming convention for federated reactors, these files are generated to projectDir/<LF_file_name>/src/<name_of_the_federate>.ts and projectDir/<LF_file_name>/dist/<name_of_the_federate>.js. The TypeScriptGenerator also generates a C program for the RTI, which can be found at projectDir/src-gen/<LF_file_name>_RTI.c. The RTI's executable is compiled to projectDir/bin/<LF_file_name>_RTI.

- I refactored the code used to generate and compile the RTI out of CGenerator and into GeneratorBase so it could also be called by TypeScriptGenerator. I think this is a pretty good solution for multiple targets using the same C RTI code.

- This LF repo branch points to the "Federate" branch of the reactor-ts submodule repo. If the "Federate" branch is merged into master, this should be updated.

- I added an example federated TypeScript program to lingua-franca/example/DistributedTS, based off the lingua-franca/example/Distributed program.


mew2ub:
Hmmm. I'm not sure why the TS tests are failing in the build. They work locally.

lhstrh:
It says there is a merge conflict. You have to pull in changes from `master`, resolve the conflicts, and then push to this branch.

mew2ub:
Haha that's so obvious. No idea why I didn't think of that.

PR #181, opened 2020-06-25
TypeScript federates

mew2ub:
These changes make it possible to write federated reactors in TypeScript

- Most of the new code is in TypeScriptGenerator. Non-federated reactors should be unaffected. If the LF program has a federated reactor the code generator will produce a separate .js and .ts file for each federate (i.e. top level reactor of the federated reactor). Following the C Generator's naming convention for federated reactors, these files are generated to projectDir/<LF_file_name>/src/<name_of_the_federate>.ts and projectDir/<LF_file_name>/dist/<name_of_the_federate>.js. The TypeScriptGenerator also generates a C program for the RTI, which can be found at projectDir/src-gen/<LF_file_name>_RTI.c. The RTI's executable is compiled to projectDir/bin/<LF_file_name>_RTI.

- I refactored the code used to generate and compile the RTI out of CGenerator and into GeneratorBase so it could also be called by TypeScriptGenerator. I think this is a pretty good solution for multiple targets using the same C RTI code.

- This LF repo branch points to the "Federate" branch of the reactor-ts submodule repo. If the "Federate" branch is merged into master, this should be updated.

- I added an example federated TypeScript program to lingua-franca/example/DistributedTS, based off the lingua-franca/example/Distributed program.


test/TS/MovingAverage.lf
```diff
@@ -3,8 +3,7 @@
 // four inputs and produces that as output.  The source is a counting
 // sequence.
 target TypeScript {
-    timeout: 1 sec,
-    fast: true
```
lhstrh, line 7: What's the reason for this change?

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -137,6 +145,11 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
      */
     protected var iResource = null as IResource
     
+    /**
```
lhstrh, line 148: To my knowledge, `ReactorInstance` is only used in the C code generator and doesn't belong in `GeneratorBase`. If this changed, what is the TypeScript code generator now using `ReactorInstance` for? This class is meant to unfold the entire instantiation hierarchy and carry out analysis that the C++ and TS target perform at runtime.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -603,6 +632,205 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+    
+    // //////////////////////////////////////////
+    // Protected methods for code generation
+    // of the RTI.
+    
+    // FIXME: Allow target code generators to specify the directory
+    // structure for the generated C RTI?
+    /** Create the runtime infrastructure (RTI) source file.
+     */
+    def generateFederateRTI() {
```
lhstrh, line 645: I don't think this belongs in `GeneratorBase`. This is all C-specific codegen stuff. If the point is to make it accessible for other code generators perhaps it can remain in `Cgenerator` and be made static or be moved into a separate file. This class is intended to provide the basic framework for generating code for various target languages; there should be no target-specific code in it.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -603,6 +632,205 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+        }
+    }
+    
+    /** Invoke the compiler on the generated RTI */
+    def compileRTI() {
+        var fileToCompile = filename + '_RTI'
+        executeCommand(compileCommand(fileToCompile, false), directory)
+    }
+    
+    /** Return a command to compile the specified C file.
```
lhstrh, line 790: Also C-specific. Doesn't belong.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/TypeScriptGenerator.xtend
```diff
@@ -63,6 +63,17 @@ import static extension org.icyphy.ASTUtils.*
...
 class TypeScriptGenerator extends GeneratorBase {
 
+
+    ////////////////////////////////////////////
+    //// Private variables
+
+    new () {
+        super()
+        // set defaults for federate compilation
+        this.targetCompiler = "gcc"
```
lhstrh, line 73: A reference to GCC in TypeScriptGenerator?

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/TypeScriptGenerator.xtend
```diff
@@ -236,6 +283,36 @@ class TypeScriptGenerator extends GeneratorBase {
...
+            // This will overwrite previous versions.
+            var files = newArrayList("rti.c", "rti.h", "federate.c", "reactor_threaded.c", "reactor.c", "reactor_common.c", "reactor.h", "pqueue.c", "pqueue.h", "util.h", "util.c")
+
+            for (file : files) {
+                copyFileFromClassPath(
+                    File.separator + "lib" + File.separator + "C" + File.separator + file,
+                    cSrcGenPath + File.separator + file
+                )
+            }
+            compileRTI()
```
lhstrh, line 314: This should be handled in a different class. All TypeScriptGenerator should care about is that an RTI gets produced. It should not have to deal with it's dependencies or compilation details. This information should not be scattered around the code base.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
edwardalee: The intent of these is that all targets should be using the same RTI. So either you have to design your own (likely incompatible) RTI, or need to adopt a common one.  How do you suggest dealing with that?  Should the TS code generator just generate incomplete code, with no RTI?  How would you generate the RTI?

lhstrh: I'm not questioning the intent, I'm questioning one very specific aspect of the implementation: the leaking of target-specific code into a generic base class and into code generators with different targets.

test/TS/MovingAverage.lf
mew2ub: This is probably a mistaken commit. I don't remember why I changed this, and the C test on which this is based has `fast: true`, so I don't know what's going on. I'll see if anything goes wrong when I change it back.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
mew2ub: When I refactored the RTI generation functions out of CGenerator and into GeneratorBase, I performed a pretty mechanical operation of moving all of the variables used for RTI generation into GeneratorBase. `main` is currently used on line 826 of the function `compileCommand` in GeneratorBase. The TypeScript generator invokes compileCommand.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
mew2ub: If the same C RTI is used for all targets, doesn't that mean the functions used to generate the RTI aren't target-specific? As far as I see, there's no leaking going on here.

lhstrh: I don't know what else I can do you illuminate the problem other than what I've already stated... The file is called `GeneratorBase`, not `RTIGenerator`...

mew2ub: I agree it would be an improvement to separate the RTI specific code from GeneratorBase (and I'll start looking into that), but I maintain that GeneratorBase should provide all target language generators with the capability to produce an RTI. Target language generators need to do the following things:

- Determine the location where the RTI code and binary should be placed 
- Compile the RTI at the appropriate time
- Report errors about the RTI generation along with other compilation errors

Since this is shared among all target language generators, GeneratorBase should provide these capabilities. But maybe GeneratorBase can pass off the implementation of these functions to an RTIGenerator (right now there is only a CRTIGenerator but there could be more and RTIGenerator would be an abstract class).

lhstrh: I never suggested the GeneratorBase shouldn't offer the capabilities that you enumerate. All I'm saying is that GeneratorBase shouldn't have any C (or any other target) code or compiler configuration in it. Yes, this requires some careful thought, but that's necessary for the sake of keeping a clean and maintainable code base.

mew2ub:
This latest commit removed the code for generating the RTI from `GeneratorBase`, but that change required signifcant refactoring to `GeneratorBase`.

I created a new class called `RTIGenerator` which provides utilities for creating and compiling an RTI with the functions: `generateFederateRTI` and `compileRTI`. `GeneratorBase` creates an instance of `RTIGenerator` if there are federates and gives target generators access to those functions to use as they will.

Since `GeneratorBase` and `RTIGenerator` have significant "code generator" nature in common, I made them both subclasses of a new abstract class called `CodeGenerator`. `CodeGenerator` doesn't have anything new, but I refactored into it the capabilities that used to be in `GeneratorBase` that are used for code generation but are not specific to generating Lingua Franca programs. For example:

- the `code` string buffer
- executeCommand
- refreshProject
- reportError

This change was necessary because I couldn't pull the RTI generation code out of `GeneratorBase` without giving that code access to these common generation functions. What I left in `GeneratorBase` were the functions used specfically to analyze Lingua Franca programs and to facilitate target language generators.

This new class structure works nicely if we were to implement different kinds of code generators to complement target language generators in the future. Also, if we were to write RTI generators for different languages, it would be easy to change this use an abstract `RTIGeneratorBase` extended by specific target language RTIGenerators.

I didn't want to rename `GeneratorBase` in this commit because I thought that would make it confusing to everyone else what I had changed, but I'm thinking a better name for it after the refactoring might be something like `ReactorGeneratorBase`. `CodeGenerator` may not be the best name either.

mew2ub:
> I don't quite understand why you did this. GeneratorBase served as the base class for all code generators and now you've created a base class for this base class. Why? Specifically, why can't RTIGenerator just extend GeneratorBase?

GeneratorBase contains a number of functions that are unrelated to RTI generation like:

- generatePreamble
- generateNetworkSenderBody
- processImports

It doesn't make sense for an RTIGenerator to have these functions.

lhstrh:
This refactoring doesn't look good to me for a number of reasons. It also doesn't look like an improvement over the previous revision of this PR. Note that this won't mesh well with https://github.com/icyphy/lingua-franca/pull/188 either, which is almost ready to merge into master. I suggest you revert 1c5e221 (it failed to build, anyway). 

We're better off merging up to cae1489 and refactoring this later.

mew2ub:
Okay. I see two different ways to think about this.

1. RTI generation with a C RTI is a capability of all target generators. The functions for RTI generation go in GeneratorBase.
2. The language for RTI generation my not be C. GeneratorBase provides RTI generation to target generators indirectly. The functions for RTI generation need to go in a class with access to code generator functions like `pr` or be rewritten to exclude those.

I'm happy to go with 1. for now.

lhstrh:
Note that this PR has conflicts that need to be resolved before it can be merged...

PR #191, opened 2020-07-04
Refactored the C target to begin the support for Lingua Franca Common Core

Soroosh129:
This also includes the beginning of a C/C++ target based on the C core


Soroosh129:
The checks now pass after rerunning the Travis CI. It seems there is a certain degree of non-determinism when testing federates.  This non-determinism is most likely not caused by these restructuring.

The original failed test was `DistributedCount.lf`. The results can be seen [here](https://travis-ci.com/github/icyphy/lingua-franca/builds/174268589).

edwardalee:
I will merge this, but then change the CCpp src-gen directory name to desired-src-gen.  We should not check in any generated files.  I will add a README that this directory is temporary.

Soroosh129:
Just as a suggestion maybe we could add a `.experimental` folder for this sort of thing that includes proof of concepts.

edwardalee:
Good idea...

Edward

------------
Edward A Lee
Professor
UC Berkeley

On Jul 4, 2020, at 9:17 AM, Soroush Bateni <notifications@github.com> wrote:

Just as a suggestion maybe we could add a .experimental folder for this
sort of thing that includes proof of concepts.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/icyphy/lingua-franca/pull/191#issuecomment-653784891>,
or unsubscribe

PR #195, opened 2020-07-08
Common core (work in progress)

Soroosh129:
The idea behind this merge is to separate user-facing SET and schedule APIs from the core runtime functionality of Lingua Franca.


edwardalee:
This is clearly a work in progress, but merging with the master is OK with me.

lhstrh:
One major concern that I have is the huge amount of code replication in this PR. If the goal is to have a C-core-based C++ target, then it should be possible to reuse all the cpp tests. We could introduce a flag to `lfc` that allows us to indicate which runtime to use... I think this makes more sense than having a dedicated `CCpp` target that basically replicates everything. Perhaps you can come up with a concrete strategy for this together with @cmnrd.

edwardalee:
I think in this case it makes more sense to make the CCpp target able to import and run C reactors.  This will make it much easier to reuse the C core and presumably the same technique could be used for a Python target.  Being able to write C reactors in a Python target could be a huge win.

Soroosh129:
> One major concern that I have is the huge amount of code replication in this PR. If the goal is to have a C-core-based C++ target, then it should be possible to reuse all the cpp tests. We could introduce a flag to `lfc` that allows us to indicate which runtime to use... I think this makes more sense than having a dedicated `CCpp` target that basically replicates everything. Perhaps you can come up with a concrete strategy for this together with @cmnrd.

I think this is a very valid point. What I would like to point out is that the `CCpp` target was not developed to replace the C and C++ runtimes and code generators. In fact, I expect it to be removed in the near future since the C target already has most of the capabilities. The `CCpp` target was an experimental exercise in order to move to other C/C++-friendly languages. First step was to figure out which parts of the C lib were language-specific APIs and which parts are actually runtime semantics and the tests were a huge help. I don't see any benefits to them beyond that.

Soroosh129:
> I think in this case it makes more sense to make the CCpp target able to import and run C reactors. This will make it much easier to reuse the C core and presumably the same technique could be used for a Python target. Being able to write C reactors in a Python target could be a huge win.

@edwardalee This functionality is already enabled in the CCppGenerator by including `C` as an acceptable target. I have verified it using the `test/CCpp/Import.lf` regression test by setting the targets of `test/CCpp/Imported.lf` and `test/CCpp/ImportedAgain.lf` to `C`.

PR #197, opened 2020-07-22
TypeScript protocol buffer support

mew2ub:
I added protocol buffer support to the TypeScript target. The test files test/ProtoNoPacking.lf and test/PersonProtocolBuffers.lf show how you can use it. These tests are not part of test-manifest because the corresponding C versions don't work on travis yet (although the TS versions do pass on travis).

The main change to TypeScriptGenerator is it now compiles all imported ".proto" files as its first step.

I changed the default package.json file, which means your local TypeScript tests, benchmarks, and examples may not work if you don't delete the old package.json. package.json files are not replaced automatically.


xtext/org.icyphy.linguafranca/src/org/icyphy/generator/TypeScriptGenerator.xtend
```diff
@@ -326,6 +362,29 @@ class TypeScriptGenerator extends GeneratorBase {
...
+    /** Open a non-Lingua Franca import file at the specified URI
+     *  in the specified resource set. Throw an exception if the
+     *  file import is not supported. This class imports .proto files
+     *  and runs, if possible, the protoc protocol buffer code generator
+     *  to produce the required .d.ts and .js files.
+     *  @param importStatement The original import statement (used for error reporting).
+     *  @param resourceSet The resource set in which to find the file.
+     *  @param resolvedURI The URI to import.
+     */
+    override openForeignImport(Import importStatement, ResourceSet resourceSet, URI resolvedURI) {
```
Soroosh129, line 375: This might change in the future if we use `files` in target properties to import .proto files. Nonetheless, this is backwards compatible with the current import mechanism.

Soroosh129:
This pull request looks good to me. We might need to change how `openForeignImport` acts in the future though.

PR #201, opened 2020-07-30
New Import System

lhstrh:
- New syntax for import that references a file and has to declare which particular reactors to import.
- The scope provider ensures that declared imports are limited to file referenced in import statement.
- Files are currently resolved relative to the compiled file's location, but this can be expanded.
- Imported reactors can be renamed (the C code generator already supports this).
- Various checks, including:
  - unused imports; and
  - instantiation cycles inside of imported reactors.
- Grouped graph classes in a graph package.
- Made various improvements to classes in graph package.
- Updated all tests to adopt new import syntax.
- Simplified validation by letting Xtext handled duplicate names.
- Added a `files` and `protobufs` directive.


xtext/org.icyphy.linguafranca/src/org/icyphy/generator/Main.xtend
```diff
@@ -81,47 +115,51 @@ class Main {
...
-		]
-		
-		generator.generate(resource, fileAccess, context)
-		System.out.println('Code generation finished.')
-	}
+
+        val resource = set.getResource(URI.createFileURI(fileName), true)
+
+        // Read the code
+        val code = new StringBuilder();
```
lsk567, line 139: Looks like the `code` variable is unused in code generation.

lhstrh: Ah, you're right! Looks like an old remnant that is obsolete. I'll remove it.

PR #205, opened 2020-08-12
Multiports

edwardalee:
This branch realizes parallel connections in the C and Cpp targets. In the C target, the widths of the multiports and banks of reactors can be parameterized, but not in Cpp, where they still have to be given as numeric constants. I have ported most of the tests for multiports and banks to Cpp, except where they are specifically testing parameterization. Some of the tests in Cpp have FIXMEs where the corresponding C test uses parameters.


xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CppGenerator.xtend
```diff
@@ -266,20 +266,20 @@ class CppGenerator extends GeneratorBase {
                 return port.widthSpec.terms.get(0).width
             }
         }
-        throw new Exception("Only integer valued multiport widths are supported for now.")
+        throw new Exception("Only multiport widths with literal integer values are supported for now.")
```
cmnrd, line 269: Shouldn't this be checked by the validator rather than throwing an exception here?

edwardalee: Good idea. Done. The TS target now fails to validate if you try to create multiports or banks. The Cpp target fails to validate if you try to set the width using a parameter.

xtext/org.icyphy.linguafranca/src/org/icyphy/ASTUtils.xtend
```diff
@@ -1167,8 +1174,95 @@ class ASTUtils {
...
     }
+        
+    /**
+     * If the parameter has a literal integer value, then return that value.
+     * Otherwise, return -1.
+     * @param parameter The parameter.
+     * @return The parameter's integer value or -1
+     *  if the value cannot be determined or it is not an integer.
+     */
+    def static int intValue(Parameter parameter) {
```
lhstrh, line 1185: Maybe call this `widthValue` or `portWidth` to make clear that this attempts to retrieve a _positive_ integer and returns -1 if it fails to?

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CGenerator.xtend
```diff
@@ -943,8 +941,8 @@ class CGenerator extends GeneratorBase {
                     echo "Copying source files to host «federate.host»"
                     scp «filename»_«federate.name».c ctarget.h «federate.host»:«path»/src-gen
                     popd > /dev/null
-                    echo "Compiling on host «federate.host» using: «this.targetCompiler» -O2 src-gen/«filename»_«federate.name».c -o bin/«filename»_«federate.name» -pthread"
-                    ssh «federate.host» 'cd «path»; «this.targetCompiler» -O2 src-gen/«filename»_«federate.name».c -o bin/«filename»_«federate.name» -pthread'
+                    echo "Compiling on host «federate.host» using: «this.targetCompiler» «this.targetCompilerFlags» src-gen/«filename»_«federate.name».c -o bin/«filename»_«federate.name» -pthread"
```
lhstrh, line 944: Shouldn't the `-pthread` flag depend on the `threads` target property?

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CppGenerator.xtend
```diff
@@ -671,43 +673,126 @@ class CppGenerator extends GeneratorBase {
...
-                for (unsigned i = 0; i < «leftContainer.name».size(); i++) {
-                  «leftContainer.name»[i].«leftPort.name».bind_to(&«rightContainer.name»[i].«rightPort.name»);
-                }
-            '''
-        } else if (leftContainer !== null && leftContainer.arraySpec !== null &&
-            rightPort.widthSpec !== null) {
-            return '''
-                for (unsigned i = 0; i < «leftContainer.name».size(); i++) {
-                  «leftContainer.name»[i].«leftPort.name».bind_to(&«c.rightPort.name»[i]);
+    def generate(Connection c) {
```
lhstrh, line 709: It seems this function can easily be broken down into smaller parts. It's rather long now.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/ReactorInstance.xtend
```diff
@@ -665,6 +605,79 @@ class ReactorInstance extends NamedInstance<Instantiation> {
...
     
+    /**
+     * For the specified parameter, if its value is a positive integer, return
+     * that value. Otherwise, report an error and return 0.
+     * @param parameter The parameter.
+     * @param errorReportingObject If an error occurs, report the error
+     *  on this object in the AST.
+     */
+    def int intParameterValue(Parameter parameter, EObject errorReportingObject) {
+        // For some reason (?) parameter values can be tuples.
```
lhstrh, line 616:
```suggestion
        // Parameter values can be tuples.
```

test/C/BankToBank.lf
```diff
@@ -1,39 +1,40 @@
...
 	output out:int;
 	state s:int(0);
 	reaction(t) -> out {=
         SET(out, self->s);
-   	    self->s++;
+   	    self->s += self->bank_position;
 	=}
 }
-reactor Destination {
+reactor Destination(bank_position:int(0)) {
```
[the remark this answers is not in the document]
edwardalee: I prefer full words, and "index" seems too vague to me.  How about "instance"? That's what Ptolemy II uses.

test/C/MultiportToPort.lf
```diff
@@ -37,6 +37,5 @@ main reactor MultiportToPort {
     a = new Source();
     b1 = new Destination();
     b2 = new Destination(expected = 1);
-    a.out -> b1.in;  // Channel 0 of a.out to b1.in.
-    a.out -> b2.in;  // Channel 1 of a.out to b2.in.
+    a.out -> b1.in || b2.in;
```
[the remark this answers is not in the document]
edwardalee: I will replace `||` with comma.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CGenerator.xtend
edwardalee: Federated execution requires pthreads regardless of the threads target property because it requires that schedule() be thread safe.

xtext/org.icyphy.linguafranca/src/org/icyphy/ASTUtils.xtend
edwardalee: This function is no longer used. I have removed it.

test/C/BankToBank.lf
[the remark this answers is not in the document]
lhstrh: Yes, "instance" looks good to me.

PR #210, opened 2020-09-07
Refactor command execution and unify it across targets

### LINKED ISSUE #206: Don't compile twice if target compiler returns non-zero exit code
We should only fall back to `bash --login -c` when the executable cannot be found on the path, not when the executable was found but exited with a non-zero exit code.

cmnrd:
This change refactors the way we execute commands across targets. The old way we used in GeneratorBase had two major downsides:
1. It was not general enough to support the C++ target. For instance, it did not provide a mechanism for setting environment variables.
2. The retry mechanism used to get things to work on Mac OS was buggy (see #206) 

With this change, executing a command becomes a two-step process. First, `createComand(cmd, args)` is used to create a new `ProcessBuilder` object. `createCommand()` makes sure that the specified command is actually available by running `which cmd`. If it is not available, it tries again using the bash shell mechanism. In either case, it returns the proper `ProcessBuilder` object that can be modified as needed (e.g. set the directory or environment variables). Second, `executeCommand(builder)` executes the previously created command. The function can be given an optional parameter in order to record the command's outputs. This is important to process and report any errors as, for instance, the C target does.

Before we merge this, I think it is important to test this change on various platforms. Edward, could you test if this works on your 
Mac? Travis is currently doing his part, but I am not sure if these tests also cover RTI generation and proto buffers. If not, it should also be tested if the commands for these features work as expected.

closes #206

### DIFF
--- a/xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
+++ b/xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
@@ -1025,9 +1025,32 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
      * 
      * This method makes sure that the given command is executable,
      * It first tries to find the command with 'which cmake'. If that
-     * fails, it tries again with bash. n case this fails again,
+     * fails, it tries again with bash. In case this fails again,
      * it returns null. Otherwise, a correctly constructed ProcessBuilder
      * object is returned. 
+     * 
+     * A bit more context:
+     * If the command cannot be found directly, then a second attempt is made using a
+     * Bash shell with the --login option, which sources the user's 
+     * ~/.bash_profile, ~/.bash_login, or ~/.bashrc (whichever
+     * is first found) before running the command. This helps to ensure that
+     * the user's PATH variable is set according to their usual environment,
+     * assuming that they use a bash shell.
+     * 
+     * More information: Unfortunately, at least on a Mac if you are running
+     * within Eclipse, the PATH variable is extremely limited; supposedly, it
+     * is given by the default provided in /etc/paths, but at least on my machine,
+     * it does not even include directories in that file for some reason.
+     * One way to add a directory like
+     * /usr/local/bin to the path once-and-for-all is this:
+     * 
+     * sudo launchctl config user path /usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
+     * 
+     * But asking users to do that is not ideal. Hence, we try a more hack-y
+     * approach of just trying to execute using a bash shell.
+     * Also note that while ProcessBuilder can configured to use custom
+     * environment variables, these variables do not affect the command that is
+     * to be executed but merely the environment in which the command executes.
      * 
      * @param cmd The command to be executed
      * @return A ProcessBuilder object if the command was found or null otherwise.
@@ -1039,7 +1062,7 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
     /**
      * Creates a ProcessBuilder for a given command and its arguments.
      * 
-     * This method makes sure that the given command is executable,
+     * This method ensures that the given command is executable,
      * It first tries to find the command with 'which cmake'. If that
      * fails, it tries again with bash. n case this fails again,
      * it returns null. Otherwise, a correctly constructed ProcessBuilder
@@ -1071,6 +1094,9 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
             return new ProcessBuilder(#["bash", "--login", "-c", '''«cmd» «args.join(" ")»'''])
         }
         println("FAILED")
+        reportError("The command " + cmd + " could not be found.\n" +
+                    "Make sure that your PATH variable includes the directory where " + cmd + " is installed.\n" +
+                    "You can set PATH in ~/.bash_profile on Linux or Mac.")
         return null as ProcessBuilder
     }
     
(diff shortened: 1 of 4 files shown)


cmnrd:
Travis seems to be happy now :)

edwardalee:
Looks good to me!

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -918,74 +988,56 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+        // Try running with bash.
+        // The --login option forces bash to look for and load the first of
+        // ~/.bash_profile, ~/.bash_login, and ~/.bashrc that it finds.
+        print('''--- Trying again with bash ... ''')
+        val bashCommand = #["bash", "--login", "-c", '''which «cmd»''']
+        val bashBuilder = new ProcessBuilder(bashCommand)
+        val bashReturn = bashBuilder.start().waitFor()
+        if (bashReturn == 0) {
+            println("SUCCESS")
+            val builder = new ProcessBuilder(#["bash", "--login", "-c", '''«cmd» «args.join(" ")»'''])
```
lhstrh, line 1032: Why invoke the compiler using Bash? If we invoke `which` in Bash it will give us the full path to the command that we can execute directly. Why not use that? It will be less confusing when the actual compile command is shown upon encountering an error.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -918,74 +988,56 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+     * object is returned. 
+     * 
+     * @param cmd The command to be executed
+     * @param args A list of arguments for the given command
+     * @return A ProcessBuilder object if the command was found or null otherwise.
+     */
+    protected def createCommand(String cmd, List<String> args) {
+        // Make sure the command is found in the PATH.
+        print('''--- Looking for command «cmd» ... ''')
+        val whichBuilder = new ProcessBuilder(#["which", cmd])
```
lhstrh, line 1014: Does `which` work on Windows? Perhaps @Soroosh129 can comment on this.

lhstrh (changes requested):
Thanks for doing this refactoring; much better indeed to first resolve the command rather then speculatively executing it...

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
Soroosh129: In cmd, the `which` equivalent is ingeniously called `where`.
Although if [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) is installed, `bash` is added to `C:\Windows\System32` and can be used natively.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
cmnrd: Right. Good idea! That thought simply did not occur to me. I will give that a try

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
cmnrd: Alright, so we should check if we are on Windows and then use `where` for the first try and `bash ... which` as the fallback. I will change that

cmnrd: I pushed some new changes. @Soroosh129 could you check how that works on Windows?

lhstrh (approved):
Looks good!

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
Soroosh129: I can confirm the new change works on Windows. There are still some bugs related to the IDE + Cpp/TS on Windows that I think are unrelated to this.

cmnrd: Thanks! Could you document these bugs in a new issue? Or do we already have an issue on that?

Soroosh129:
A problem with this new command execution scheme is that you cannot prepend the command with anything. For example, I need to set a few variables before running `python3`:

    CC=gcc LDFLAGS="-Wl,-allow-multiple-definition" python3 -m pip install .

cmnrd:
Actually that is a major advantage of the new scheme. It was not possible before. The ProcessBuilder does not behave like a shell, where you can prepend assignments of environment variables, but it provides another mechanism instead. Simply call `environment()` on the ProcessBuilder object that `createCommand("python3", ...)` returns. That gives you a map, where you can set your environment variable prior to calling `executeCommand()`. See the C++ code generator for an example of how this works: https://github.com/icyphy/lingua-franca/blob/master/xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CppGenerator.xtend#L1026

Soroosh129:
> Actually that is a major advantage of the new scheme. It was not possible before. The ProcessBuilder does not behave like a shell, where you can prepend assignments of environment variables, but it provides another mechanism instead. Simply call `environment()` on the ProcessBuilder object that `createCommand("python3", ...)` returns. That gives you a map, where you can set your environment variable prior to calling `executeCommand()`. See the C++ code generator for an example of how this works: https://github.com/icyphy/lingua-franca/blob/master/xtext/org.icyphy.linguafranca/src/org/icyphy/generator/CppGenerator.xtend#L1026

Great, It works!

PR #219, opened 2020-09-30
Update LICENSE.md

tanaymishra:


lhstrh:
When a single year is used in a copyright notice, this year refers to the first year of the publication. This statement legally protects the copyright owner from potential infringement from that year and onward. Changing 2019 to 2020, as suggested in this PR, could in theory nullify a year's worth to said protection. We could also list multiple years to indicate years in which substantial revisions occurred, or write a range, but that seems overly verbose to me and will require maintenance.

edwardalee:
I think you have to indicate a range if the content has been modified after the initial date.
Otherwise, for the purposes of establishing prior art, for example, you would be claiming a date of conception earlier than the actual date, which could invalidate the establishment of prior art.

The git record, of course, could set the record straight, but at considerable effort.

Edward



———————
Edward A. Lee
EECS, UC Berkeley
http://eecs.berkeley.edu/~eal

lhstrh:
Indicating a range is fine, but simply replacing `2019` with `2020` would be wrong, I think. By indicating a range you run the risk of forgetting to update it, and I see no practical advantage to doing this. Is there any?

edwardalee:
Yes, I think this has to be done once a year in order for the copyright notice to remain valid. You can stop doing it when you stop updating the code.

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 The Lingua-Franca toolkit is licensed under the BSD 2-Clause License:
 
-Copyright (c) 2019, Industrial Cyberphysical Systems (iCyPhy). All rights reserved.
+Copyright (c) 2020, Industrial Cyberphysical Systems (iCyPhy). All rights reserved.
```
lhstrh, line 3:
```suggestion
Copyright (c) 2019-2020, Industrial Cyberphysical Systems (iCyPhy). All rights reserved.
```

lhstrh (changes requested):

Issue #212, opened 2020-09-10
Optimization: bypassing the reaction queue


lhstrh:
Under some circumstances, newly triggered reactions could be executed immediately by the same worker that was executing the triggering reaction (bypassing the reaction queue altogether). This could significantly improve the performance in situations where there is a lot of contention to obtain the mutex that protects the reaction queue. The Sieve of Eratosthenes benchmark is an example where this happens. Reactions eligible for such treatment could simply be marked as such.

It appears that a reaction `Y` is eligible for immediate execution if:
 - there exists only one reaction or mutation, let's call it `X`, whose effect(s) trigger(s) `Y`; and
 - there exist no reactions of mutations (other than `X` itself) that must happen before `Y`; and
 - `Y` has no dependencies on actions or timers

Extra care must be taken to avoid that this execution policy undermines EDF. `Y` should not bypass the reaction queue if the first-in-line reaction on the reaction queue has a deadline earlier than `Y`.

edwardalee:
I have pushed a change to the physical-connection branch that implements a variant of this optimization. The way that it works is that there is now a new field in reaction_t:
```
    reaction_t* last_enabling_reaction; // The last enabling reaction, or NULL if there is none. Used for optimization. INSTANCE.
```
This is calculated statically (or after a mutation) and points to an upstream reaction if there is a single, unique, last upstream reaction on which this reaction depends.  This reaction can depend on multiple reactions, but if those depend on each other, then only the last one is set.  The field will be null if there is no such reaction.

The threaded runtime now checks: After running a reaction `r`, if any downstream reaction has this field pointing to `r`, then this is a candidate for running immediately.  There may be multiple such candidates. The code looks for the candidate with the least deadline and puts all other candidates on the reaction queue.  After so scanning all downstream reactions, it executes the candidate without going through the reaction queue.

This optimization reduces the execution time of PingPongThreaded by about 37%, to 168 msec 1,000,000 ping-pongs.  This is closer to the 97 msec that the unthreaded version takes.

edwardalee:
Unfortunately, we aren't out of the woods yet. If you create a variant PingPongThreaded that has two disconnected instances a Ping-Pong pair, the execution time on multiple threads explodes to more than 4 seconds from 168 msec.  I.e., twice as much work in 23 times as much time. Ugh.  The same model runs without threads in 270 msec.

PR #223, opened 2020-10-11
Physical connection

edwardalee:
Soroush and I have made some changes to fix federated execution and quite a few improvements in debugging and logging outputs. It would be better to continue these improvements in the master branch. So I'm proposing merging this branch with master.

This branch also implements an optimization where a reaction can be executed without going through the reaction queue. See issue #212 .


test/C/PingPongDistributed.lf
```diff
@@ -23,16 +23,17 @@
  * @author Edward A. Lee
  */
 target C;
+
+preamble {=
+    // #define VERBOSE
```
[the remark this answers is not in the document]
lhstrh: I agree. Create a ticket?

benchmark/C/Savina/PingPongDistributed.lf
```diff
@@ -12,7 +12,7 @@
  * 
  * Unthreaded: 97 msec
  * Threaded: 265 msec
- * Distributed: 53 seconds
+ * Distributed: 6 seconds
```
lhstrh, line 15: While it's a good idea to write these numbers down, this is probably not the best place to put them. We should probably create a shared spreadsheet (or even a spot on the wiki) where we can more carefully document which optimizations were involved, which platform it was executed on, what hardware, etc.

benchmark/C/Savina/PingPongThreaded.lf
```diff
@@ -0,0 +1,68 @@
...
+ * to break the causality loop.
+ * 
+ * To get a sense, some (informal) results for 1,000,000 ping-pongs
+ * on my Mac:
+ * 
+ * Unthreaded: 97 msec
+ * Threaded: 265 msec  (168 msec with optimization of executing immediately)
+ * 
+ * There is no parallelism in this application, so it does not benefit from being
+ * being threaded, just some additional overhead.
```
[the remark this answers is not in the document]
lhstrh: As per my earlier comment, I think this discussion should be moved elsewhere.

test/C/PingPongThreaded.lf
```diff
@@ -0,0 +1,68 @@
...
+ * to break the causality loop.
+ * 
+ * To get a sense, some (informal) results for 1,000,000 ping-pongs
+ * on my Mac:
+ * 
+ * Unthreaded: 97 msec
+ * Threaded: 265 msec  (168 msec with optimization of executing immediately)
+ * 
+ * There is no parallelism in this application, so it does not benefit from being
+ * being threaded, just some additional overhead.
```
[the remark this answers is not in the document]
lhstrh: Why are we duplicating benchmarks as tests?

test/C/PingPongDistributed.lf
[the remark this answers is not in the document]
Soroosh129: Done. #224

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
```diff
@@ -48,74 +48,220 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        port++;
+        server_fd.sin_port = htons(port);
+        result = bind(
+                socket_descriptor,
+                (struct sockaddr *) &server_fd,
+                sizeof(server_fd));
+    }
+    if (result != 0) {
+        if (specified_port == 0) {
+            error_print_and_exit("ERROR on binding the socket for federate %d. Cannot find a usable port. Consider increasing PORT_RANGE_LIMIT in federate.c",
```
lhstrh, line 210: Exceeding acceptable line width.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
```diff
@@ -48,74 +48,220 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                                 _lf_my_fed_id);
+        } else {
+            error_print_and_exit("ERROR on binding socket for federate %d. Specified port is not available. Consider leaving the port unspecified",
```
lhstrh, line 213: Exceeding acceptable line width.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
```diff
@@ -149,31 +295,269 @@ void send_via_rti_timed(unsigned int port, unsigned int federate, size_t length,
...
+    int received_federates = 0;
+    pthread_t thread_ids[_lf_number_of_inbound_physical_connections];
+    while (received_federates < _lf_number_of_inbound_physical_connections) {
+        // Wait for an incoming connection request.
+        struct sockaddr client_fd;
+        uint32_t client_length = sizeof(client_fd);
+        int socket_id = accept(_lf_server_socket, &client_fd, &client_length);
+        // FIXME: Error handling here is too harsh maybe?
+        if (socket_id < 0) {
+            return NULL;
```
lhstrh, line 331: Print/log error message?

xtext/org.icyphy.linguafranca/src/lib/core/reactor_threaded.c
```diff
@@ -485,6 +485,141 @@ reaction_t* first_ready_reaction() {
...
 volatile bool __advancing_time = false;
 
+/**
+ * For the specified reaction, if it has produced outputs, insert the
+ * resulting triggered reactions into the reaction queue.
+ * This procedure assumes the mutex lock is not held and grabs
+ * the lock only when it actually inserts something onto the reaction queue.
+ * @param reaction The reaction that has just executed.
+ */
+void schedule_output_reactions(reaction_t* reaction) {
```
lhstrh, line 495: Considering how long this function has become, I'd suggest to factor out the common code, put it back in `reactor_common.c`, and factor out the threaded/non-threaded-specific code into a separate function to be called from `schedule_output_reactions` in `reactor_common.c`. This way, the comments that are currently written inline can be pulled to the top of this new function so that it's also visible in the generated documentation. This gives is less code duplication, better documentation, all for the price of just one function call.

lhstrh (changes requested):
I didn't have the bandwidth to do a super in-depth review, but this PR generally looks good and ready for merging into master. I've left a few comments. Particularly `federate.c` needs some more careful code formatting with attention to the max line width. I've also proposed a code refactoring of the optimization implemented in `reactor_threaded.c` that I think should happen before doing the merge.

test/C/PingPongThreaded.lf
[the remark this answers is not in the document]
edwardalee: Because if we don't, the benchmarks will stop working.

xtext/org.icyphy.linguafranca/src/lib/core/rti.c
```diff
@@ -497,34 +581,74 @@ void* federate(void* fed) {
     // Send back to the federate the maximum time plus an offset.
     // Start by sending a timestamp marker.
     unsigned char message_marker = TIMESTAMP;
-    write_to_socket(my_fed->socket, 1, &message_marker);
+    int bytes_written = write_to_socket2(my_fed->socket, 1, &message_marker);
+    if (bytes_written < 1) {
+        fprintf(stderr, "ERROR sending timestamp to federate %d.", my_fed->id);
```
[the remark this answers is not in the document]
edwardalee: I think the RTI should not exit when a federate fails. It should print something and continue executing. See https://github.com/icyphy/lingua-franca/wiki/Termination

test/C/PingPongThreaded.lf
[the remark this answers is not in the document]
lhstrh: That's not a reason for duplicating code. If we want benchmarks to be run on Travis, we can configure it to do so. I'd suggest to remove the duplicated code and create a ticket for configuring Travis to run the benchmarks.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
Soroosh129: This is a very good point. With the introduction of new `write_to_socket` and `read_from_socket` functions, the line widths will generally be too long. Is there a way to automatically handle/detect these situations (e.g., by at least producing a warning somewhere)?

lhstrh:
There as been a need to augment the test manifest for a long time. In the manifest we should allow the specification of parameters and target properties so that benchmark code can be tested without spending too much wall clock time. If we duplicate code, we will soon have out-of-sync versions of the same stuff floating around, and we'll have tests that pass that represent benchmarks that could fail.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
lhstrh: Consoles usually line wrap. I was merely commenting on the code exceeding 80 chars. If you want to build a long string literal you can just write `"first part " "second part"` in C.

edwardalee:
This is a nontrivial change. The C target does not currently support changing parameters nor target parameters on the command line.  Plus we have to write a parser for the test manifest, handle the more complicated directory structure. I suggest opening an enhancement request and moving forward with duplicated code for now.  Not having the tests is not a good solution, even in the short term.

Soroosh129 (commented):
[This](7bda1d189791daedd60df7bbb41219ffe22d8f9a) commit shows the dangers of producing an empty `DEBUG_PRINT.` I suggest we should go back to having an `if` statement inside the macro before [this](fdc90689c32911ab58b91ebfffe4affdcb74cf4e) change.

edwardalee:
I agree to going back if there is no measurable impact on performance. Can you run the TimeLimit test with and without the empty DEBUG_PRINT (perhaps after replacing all the `// printf("DEBUG: ...` statements)?

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
Soroosh129: I think 80 chars might be too restrictive especially given the indentation of the code here. Maybe we should try to adhere to 125 characters (which is the GitHub's line length limit) instead?

lhstrh:
I agree it's nontrivial, and it shouldn't be part of this PR. I'm personally convinced that code duplication should always be nipped in the bud and should never be considered an acceptable temporary solution. If not kept in check, it tends to quickly spiral out of control. Take our test suite, for example, which now has a formidable amount of code duplication (many redefinitions of sources, sinks, print and test reactors). Because this wasn't addressed early on, we now have the problem multiplied across targets, and the problem keeps growing as we add more targets and tests... Perhaps we should do a big "cleanup" in December...

edwardalee:
Well, it's not acceptable to me to remove these tests. If you must, please remove the benchmark directory then. Might as well remove the whole directory. I'll put all my benchmark work in the test directory. This will have the unfortunate side effect of making the tests run much longer...

lhstrh:
To clarify, what I suggested was to allow further accumulation of duplicate code for now and do a big clean up in December. I'm well aware that cleaning up the test suite is going to be a dreadful task, so I'm not going to press others to do this while I'm unavailable to help.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
lhstrh: IDK, I've always stuck to 80 chars for as long as I remember. We should write an actual style guide and settle on something that is agreeable to most contributors. I mostly flagged this because some lines were actually cut off while I was reviewing...

edwardalee:
Oops... Didn't mean to close this.

Soroosh129:
> I agree to going back if there is no measurable impact on performance. Can you run the TimeLimit test with and without the empty DEBUG_PRINT (perhaps after replacing all the `// printf("DEBUG: ...` statements)?

Here are the average results for 800 runs of `TimeLimit.lf` after _all_ `// printf("DEBUG: ...` statements were converted:
Empty `DEBUG_PRINT` (no `VERBOSE`): 781,319,221 ns
Conditional `DEBUG_PRINT` (no `VERBOSE`): 788,356,250 ns

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
edwardalee: I like the idea of settling on 125 characters.  The 80 character limit dates back to punched cards, which had 80 columns. In Ptolemy, we used 80 characters so we could print out the code in two column format for code reviews. But we aren't printing code anymore and our screens can easily handle 125 characters.

edwardalee:
This seems like a negligible performance hit (0.1%) so I'm OK with reverting to the if statement. Was compiler optimization turned on? (e.g., -O2)

Soroosh129:
> This seems like a negligible performance hit (0.1%) so I'm OK with reverting to the if statement. Was compiler optimization turned on? (e.g., -O2)

Yes I think this is the default behavior of the CGenerator. Here is the compile command verbatim: `gcc -O2 src-gen/TimeLimit.c -o bin/TimeLimit`

benchmark/C/Savina/PingPongDistributed.lf
edwardalee: I have started a Benchmarks wiki page and put pointers to it in the benchmarks: 
https://github.com/icyphy/lingua-franca/wiki/Benchmarks

Soroosh129: We do have a [NVIDIA Jetson AGX Xavier](https://www.nvidia.com/en-us/autonomous-machines/embedded-systems/jetson-agx-xavier/) if you think benchmarks on a board like that could also be interesting. The C target at least works fine.

xtext/org.icyphy.linguafranca/src/lib/core/federate.c
Soroosh129: This should be fixed in 755c2d77108667ba5bc22ff800fd94ef3fac9dd9. I don't know why GitHub is not showing this comment as outdated.

benchmark/C/Savina/PingPongDistributed.lf
edwardalee: Ah, yes, that would be quite interesting.  My guess is we would have to tune the code to get good performance...

xtext/.gitignore
```diff
@@ -11,3 +11,4 @@ src-gen/
 xtend-gen/
 /*/target/
 /.gradle
+org.icyphy.linguafranca/.launch
```
lhstrh, line 14: I guess this is better put in `xtext/org.icyphy.linguafranca/.gitignore`. I've noticed that all `.gitignore` files in the repo refer to local directories, which makes things easier to find.

lhstrh (approved):
Looks good!

Soroosh129:
Should we merge this branch? I think this branch achieves the goal of "establishing physical-connections" as its name suggests.

lhstrh:
Yes, we should, but the following comment has remained unaddressed: https://github.com/icyphy/lingua-franca/pull/223#discussion_r504143970

Can we address this comment before merging?

xtext/org.icyphy.linguafranca/src/lib/core/reactor_threaded.c
edwardalee: This is done.  It has the side effect of applying the same optimization now in the unthreaded case, which I thought wouldn't make much difference, but it improves the performance of the TimeLimit test by about 15%!

lhstrh: Awesome!

Soroosh129: On AGX, I also see a 15% improvement for the unthreaded version of TimeLimit. However, the threaded version got about 11% worse.

edwardalee: I believe the only difference is an extra function call when things have to get queued on the reaction queue... I doubt this would add 11%, so there must be something else going on...

Soroosh129: I ran the benchmarks again. The 11% reduction in performance appears to have been an anomaly.

PR #243, opened 2020-12-07
Savina benchmarks for Cpp and benchmark runner.

hnnskl:
Overview:
- I added implementations of Savina benchmarks for the C++ target in benchmark/Cpp/Savina.
- For the benchmarks I used subdirectories with the same names as in the original Savina repository https://github.com/shamsimam/savina
- I deleted the old benchmarks in that folder that are the same or a similar implementation. The other benchmarks were moved into the subdirectories, these are only different versions of the dining philosophers.
- Benchmarks files with the name suffix "Generator" in their name contain code for the code generator Cog in comments. They can be used as normal LF files but are also parametrized.
- The general design of the benchmarks, their output and control through the BenchmarkRunner.lf is all based on the original Savina implementations.
- An extensible benchmark runner to automatically run the benchmarks and create graphs from the results can be found in util/BenchmarkRunner. A Readme explains how to setup and use it.

Let me know if you want any modifications.


cmnrd:
Thanks @hnnskl! :+1:

 Should we merge this in? I am happy with it. I will probably make some adjustments to the benchmark runner in the future though.

lhstrh:
I agree with you, @cmnrd; we should merge this into master. Great work, @hnnskl! This is a real leap forward.

I was thinking that it might be useful to discuss the workings of `BenchmarkRunner` during a code review session.

PR #249, opened 2020-12-17
A flexible benchmark runner

cmnrd:
This adds a flexible benchmark runner. The runner is written in python and is based on [hydra](https://hydra.cc/docs/intro), a tool for dynamically creating hierarchical configurations by composition. This allows to significantly reduce the complexity of the runner and to provide a flexible infrastructure that can be controlled simply by YAML files and command line overrides.

A complete description of its usage can be found in the [wiki](https://github.com/icyphy/lingua-franca/wiki/Running-Benchmarks).

Note that the runner has only been tested on Linux and does not provide configurations for all implemented benchmarks yet.

Note that this PR is based on #248 which should be merged first.


cmnrd:
I am not sure why the tests fail. But since this PR only adds new code and does not modify existing files, I assume there is some kind on nondeterministic error in the tests.

edwardalee:
The Python tests are passing for me.  The offending test seems to be ActionIsPresent.  The test suite reports no reason for failure.

Soroosh129:
> The Python tests are passing for me. The offending test seems to be ActionIsPresent. The test suite reports no reason for failure.

I have been grappling with this test for two days now. I have tried different Python versions, environments, GitHub action settings, etc. I cannot replicate the segmentation fault. It is a strange phenomenon.

PR #264, opened 2021-01-20
Clock synchronization

edwardalee:
For federated execution, by default, if a federate is running on a different host (or a differently named host, even if it's the same physical machine), then it will perform initial clock synchronization during federation startup to synchronize its clock with the RTI. If, in addition, you set the target property clock-sync to ON, then it will also perform runtime clock synchronization during execution. If you set clock-sync to OFF, then all clock synchronization is disabled.

The status of clock synchronization is that initial clock synchronization is performing very well. The runtime clock synchronization not so well. The default, therefore, should be OK for many applications as long as either the application does not run very long or the clocks don't drift much.

This branch contains a lot of improvements to the C target code base that have little to do with clock synchronization.


xtext/org.icyphy.linguafranca/src/lib/core/federate.c
```diff
@@ -368,6 +388,13 @@ void send_timed_message(interval_t additional_delay,
         // should be (get_logical_time(), get_microstep())
     }
 
+    if (compare_tags2(current_message_timestamp,
```
lhstrh, line 391: What does `compare_tags2` do? Not very descriptive.

lhstrh (approved):
I was only able to do a cursory review, but this looks good to me. There is a fair amount of code replication in the examples, which we should reduce, but that can also be done after the merge...

PR #265, opened 2021-01-20
C benchmarks

lhstrh:
This is a work-in-process toward porting existing C++ benchmarks to C. While still in progress, some of the benchmarks have revealed some issues in our compiler that have been addressed in this branch and should find their way into master preferably sooner than later.


benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
```diff
@@ -182,7 +183,7 @@ reactor ManagerReactor(bufferSize:size_t(50), numProducers:size_t(40), numConsum
     =}
     
     reaction(consume) -> outConsumers, outProducers {=
-        printf("This reaction is not invoked.\n");
+        printf("consume reaction is not invoked.\n");
```
edwardalee, line 191: Shouldn't this be "is invoked"?

benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
```diff
@@ -231,6 +234,7 @@ reactor ManagerReactor(bufferSize:size_t(50), numProducers:size_t(40), numConsum
     =}
     
     reaction(inConsumers) -> consume {=
+        printf("inConsumers reaction is not invoked.\n");
```
edwardalee, line 246: Shouldn't this be "is invoked"?

There are a bunch more of these...
Note also that benchmarks probably should not have printf() in them because it adds to the execution time.
Perhaps you want instead DEBUG_PRINT()?  This way, the overhead is negligible unless you turn on logging.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/ReactionInstance.xtend
```diff
@@ -100,19 +101,20 @@ class ReactionInstance extends NamedInstance<Reaction> {
...
+                        this.effects.add(multiportInstance)
                         multiportInstance.dependsOnReactions.add(this)
                     }
                 } else {
-                    this.dependentPorts.add(portInstance)
+                    this.effects.add(portInstance)
                     portInstance.dependsOnReactions.add(this)
                 }
             } else {
+                // FIXME: Unused.
```
edwardalee, line 112: This commit seems to comment out construction of some graph information that we might need in the future (dependencies across actions, for example). I guess this is OK, but is there a reason for this? Does it create a significant performance hit?  It will probably be needed again when we calculate safe-to-process offsets.

lhstrh: The concern was not performance. There were a lot of `dependsOn` and `dependentOn` variables and I had a hard time remembering what they were being used for. It turned out that these ones weren't used at all. I can reinstate them, but I think it would be better to lump them in with `sources` and `effects` (as they are also in the language) and use filters to separate ports from actions and timers. If you agree, I'll carry out this refactoring before the merge.

benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
lhstrh: Looks like debug code. I'll remove them, but these benchmarks still need to be scrutinized, anyway...

benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
lhstrh: Same as above.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/ReactionInstance.xtend
edwardalee: OK, sounds reasonable.

benchmark/C/Savina/piprecision/PiPrecisionBenchmarkGeneratorNoRunner2.lf
```diff
@@ -0,0 +1,267 @@
...
+   
+    typedef struct WorkMessage {
+      int scale;
+      int term;
+    } WorkMessage;
+    
+    void printBenchmarkInfo(char* benchmarkId) {
```
Soroosh129, line 33: `printBenchmarkInfo()` and `printSystemInfo()` are repeated a lot in various benchmarks. Should we abstract them into a `benchmark_util.h/c`?

benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
```diff
@@ -0,0 +1,393 @@
...
+ */
+ 
+target C {
+    flags: "-lm",
+    logging: DEBUG,
+    keepalive: true,
+    timeout: 2 secs
+};
+
+import BenchmarkRunner from "../BenchmarkRunner.lf";
```
Soroosh129, line 16: Some benchmarks rely on the benchmark runner and some don't. Does the latest version of `BenchmarkRunner` have an adverse effect on performance? If not, I think we can use `BenchmarkRunner` consistently in all the benchmarks.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/ReactionInstance.xtend
lhstrh: This is done in https://github.com/icyphy/lingua-franca/commit/15ff7abc03d195998b6fcdfbcb6c4e74f87ee0ac.

benchmark/C/Savina/bndbuffer/ProdConsBenchmark.lf
lhstrh: We're still in the process of evaluating whether there is any merit to have a benchmark runner in the C tests. We'll remove whatever we don't use later on.

benchmark/C/Savina/piprecision/PiPrecisionBenchmarkGeneratorNoRunner2.lf
lhstrh: Good idea! As stated above, these benchmarks are still in a pretty rough shape. The key reason for the PR are other fixes that are in this branch, not so much the benchmarks themselves. We'll improve them later.

test/C/MultiportToBankStruct.lf
```diff
@@ -0,0 +1,72 @@
+ // Check multiport output to bank of recipients when type
```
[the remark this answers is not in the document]
lhstrh: If this other test covers that case, then we don't need it, I agree.

PR #273, opened 2021-02-02
changes for windows issues

soyerefsane:
Corrected the windows changes to compile C code.


xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -866,10 +874,14 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
      *  will never have a `-c` flag.
      */
     protected def compileCCommand(String fileToCompile, boolean doNotLinkIfNoMain) {
-        val cFilename = getTargetFileName(fileToCompile);            
-        val relativeSrcFilename = "src-gen" + File.separator + cFilename;
-        val relativeBinFilename = "bin" + File.separator + fileToCompile;
-
+    	val isBash = findCommandEnv(config.compiler)
```
lhstrh, line 877: A better name for the variable `isBash` would be `env` or `environment`. Also, I noticed you indent with tabs. Please use 4 spaces per tab instead.

lhstrh (commented):
Looks good to me!

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -1022,34 +1034,31 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
      * @return A ProcessBuilder object if the command was found or null otherwise.
      */
     protected def createCommand(String cmd) {
-        return createCommand(cmd, #[])
+        return createCommand(cmd, #[], findCommandEnv(cmd))
     }
     
     /**
-     * Creates a ProcessBuilder for a given command and its arguments.
+     * Tries to find the command.
```
edwardalee, line 1041: Better documentation for this would be:

Return ExecutionEnvironment.NATIVE if the specified command is directly executable on the current host
and ExecutionEnvironment.BASH if the command must be executed within a bash shell.
The latter occurs, for example, if the specified command is not in the native path but is in the path
specified by the user's bash configuration file.
If the specified command is not found in either the native environment nor the bash environment,
then this reports and error and returns null.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
edwardalee: I agree with these.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -325,6 +325,14 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
      */
     var indentation = new LinkedHashMap<StringBuilder, String>()
     
+    /**
+     * Defines the execution environment that is used to execute binaries.
```
edwardalee, line 329: Perhaps a more detailed comment would help. Suggest adding:

A given command may be directly executable on the host (NATIVE) or it may need to be executed
within a bash shell (BASH). On Unix-like machines, this is typically determined by the PATH variable
which could be different in these two environments.

edwardalee (commented):
Looks good, modulo the small comments.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/TypeScriptGenerator.xtend
```diff
@@ -264,7 +264,7 @@ class TypeScriptGenerator extends GeneratorBase {
                 "--js_out=import_style=commonjs,binary:" + outPath,
                 "--ts_out=" + srcGenPath)
             protocArgs.addAll(config.protoFiles.fold(newLinkedList, [list, file | list.add(file); list]))
-            val protoc = createCommand("protoc", protocArgs)
+            val protoc = createCommand("protoc", protocArgs, findCommandEnv("protoc"))
```
cmnrd, line 267: It appears that `findCommandEnv()` is called for every call to `createCommand()`. Also the first argument of `creatCommand()` seems to be always identical to the argument passed to `findCommandEnv()` (unless I missed something). So why not call `findCommandEnv()` directly in `createCommand()`? This would make the usage much simpler.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -1047,9 +1057,8 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
         val whichReturn = whichBuilder.start().waitFor()
         if (whichReturn == 0) {
             println("SUCCESS")
-            val builder = new ProcessBuilder(#[cmd] + args)
-            builder.directory(new File(directory))
-            return builder
+
+            return ExecutionEnvironment.NATIVE
```
cmnrd, line 1061: Why the indirection with the `ExecutionEnvironment` enum? Couldn't this function simply return the command to be executed directly, i.e. `#[cmd]` or `#["bash", "--login", "-c", "\"", cmd]`? The enum requires common knowledge of the meaning of its values, which doesn't appear like a convenient design.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -1062,13 +1071,38 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+     * @param env is the type of the Execution Environment.
+     * @return A ProcessBuilder object if the command was found or null otherwise.
+     */
+    protected def createCommand(String cmd, List<String> args, ExecutionEnvironment env) { 
+        if (env == ExecutionEnvironment.NATIVE) {
+            val builder = new ProcessBuilder(#[cmd] + args)
+            builder.directory(new File(directory))
+            return builder
+        } else if (env == ExecutionEnvironment.BASH) {
+            val newCmd = #["bash", "--login", "-c", "\"", cmd]
```
cmnrd, line 1100: What is the `"\""` doing? It looks suspicious.

soyerefsane: In order for the command to work in windows we need to add quotation marks to gcc command. That's why I added \" as opening quotations.

lhstrh (approved):

PR #275, opened 2021-02-04
Docker

edwardalee:
Dockerfile generation is now working for both federated and unfederated execution.  How to use the docker files is now documented in the [Containerized Execution page on the wiki](https://github.com/icyphy/lingua-franca/wiki/Containerized-Execution).

I have merged with master and there are no conflicts.


Soroosh129:
The `DistributedNetworkOrderDecentralized.lf` test failing could be a cause of concern. The output from the failed test reads:
```
Received 1 at tag (10000000, 0) with intended tag (10000000, 0).
Received 2 at tag (10000000, 0) with intended tag (10000000, 0).
```
while the expected output is:
```
Received 1 at tag (10000000, 0) with intended tag (10000000, 0).
Received 2 at tag (10000000, 1) with intended tag (10000000, 0).
```
Essentially, it appears that the same reaction is called twice in the same tag.
This could point to a non-determinism somewhere in our message/event handling stack. This problem does not seem to be caused by any specific changes in this branch though. I just wanted to point it out.

Soroosh129:
This is pretty cool! I was able to replicate the instructions on the wiki. Some things I found:

- On my machine, running TimeLimit on Docker seems to have about a 13% impact on performance (much less for the threaded version (about 4%)).
- The `at` syntax with an alias does not seem to work on my machine. For example, the following fails:
```
federated reactor DistributedCount at distributedcount_rti {
    c = new Count();
    p = new Print();
    c.out -> p.in after 200 msec;
}
```
with error
```
Aborting. Unable to validate resource.
WARNING:Invalid host name or fully qualified domain name. (file:/home/soroush/lingua-franca/git/lingua-franca/test/C/DistributedCount.lf line : 38 column : 39)
```
- It also seems to be not possible to close the RTI once it has been started using `docker run` (`docker stop foo` does work). I guess this is a limitation of Docker.

lhstrh:
@Soroosh129: What you put in the at clause should be a valid URI. No underscores are allowed in hostnames.

edwardalee:
Yes, the overall control is still pretty limited. There are two obvious next steps: One is to create the .yml file for Docker compose. This will encode the information that is on the wiki in an executable file. The second, which is a bigger step than I want to take right now, is to embrace Kubernetes for deployment. This can handle things like shutdown and failures of nodes.  The Docker compose step looks pretty easy, though, so we should just do that... We can follow the same strategy of just generating the yaml file but not executing anything.

Soroosh129:
> Yes, the overall control is still pretty limited. There are two obvious next steps: One is to create the .yml file for Docker compose. This will encode the information that is on the wiki in an executable file. The second, which is a bigger step than I want to take right now, is to embrace Kubernetes for deployment. This can handle things like shutdown and failures of nodes. The Docker compose step looks pretty easy, though, so we should just do that... We can follow the same strategy of just generating the yaml file but not executing anything.
> 

Looking at the Docker compose stuff, it certainly does look promising and relatively straightforward to generate. 

One thing to note is that, at least on Linux and Windows, it is not installed alongside Docker. It appears to be a separate install.
Another thing to consider is that it appears Docker Compose is for launching multiple Docker containers on the same machine. It seems that [Docker Swarm](https://docs.docker.com/engine/swarm/) is their tool to launch containers on multiple hosts (more akin to Kubernetes).


> Also, I saw the warning too, but it's just a warning. Docker doesn't seem to mind the underscore in the name. I guess we can replace it with something else.

For me, `lfc` refuses to generate code after the warning.

test/C/HelloWorld.lf
```diff
@@ -1,6 +1,7 @@
 target C {
     tracing: true,
-    logging: error
+    logging: error,
+    docker: {FROM: "alpine:latest"}
```
lhstrh, line 4: Shouldn't this be `docker-options`? The value of `docker` should be a `boolean` according to TargetProperty.java.

That said, we could also make this a type union of "boolean" or "docker options", in which case we can do with a single target property `docker` (one would not specify any docker options if they weren't interested in using it, and `true` would be a sensible value to use when a default configuration is desired...

xtext/org.icyphy.linguafranca/src/org/icyphy/TargetProperty.java
```diff
@@ -108,6 +108,34 @@
                 config.compiler = ASTUtils.toText(value);
             }),
     
+    /**
+     * Directive to let the execution engine allow logical time to elapse
+     * faster than physical time.
+     */
+    DOCKER("docker", PrimitiveType.BOOLEAN, Arrays.asList(Target.ALL),
```
lhstrh, line 115: If all targets support this, then why is `writeDockerFile` in `CGenerator`?

lhstrh (commented):
Left some comments that might be worth addressing before merging this.

test/C/HelloWorld.lf
edwardalee: This is done. There is now just one docker target option that can be a boolean or a list of options.  We should do the same for clock-sync options.

PR #295, opened 2021-03-23
Add some Luos tests code example

nicolas-rabault:


nicolas-rabault:
Agree. Can I be of any help regarding this work?

What kind of timer do you need? Do you need one able to interrupt main program or a nanosecond systick could do the job?

edwardalee:
Well, what needs to happen is some pretty low-level Luos programming, something we are not prepared to do here.  A good place to start would be to take the C code generated from one of the simple tests, such as [CountTest.lf](https://github.com/icyphy/lingua-franca/blob/master/test/C/src/CountTest.lf) and see what changes are needed to get it to compile and run with Luos.  I suspect this would start with replacing `printf` with something simpler, like flashing an LED. But the next step will be to provide implementations for the functions required by the `#include` statements that are not available. Some of these will be timer functions.

edwardalee:
@Soroosh129 reminded me that we have a platform-support branch that factors out the functions that have to be implemented for code generated by Lingua Franca to run as-is on a given platform. See [https://github.com/icyphy/lingua-franca/tree/platform-abstraction/xtext/org.icyphy.linguafranca/src/lib/core/platform] for a set of files for a variety of platforms.  I good starting point would be to provide Luos implementations for the functions in this file: [https://github.com/icyphy/lingua-franca/blob/platform-abstraction/xtext/org.icyphy.linguafranca/src/lib/core/platform/lf_linux_support.c].

@Soroosh129 : Should we merge master into this branch and do a pull request and review?

Soroosh129:
Yes absolutely. I'm on it.

nicolas-rabault:
After some discussions with @lsk567, perhaps we should try to keep our tech were there are good. Luos is good to do some low level access to ressources and interactions between boards. So we could use Luos to generate some data to feed the software trigger you have on reactors. By thinking it this way a reactor is not directly a Luos container having to deal with Luos messages to interact with other reactors but reactors can use Luos to read the actual hardware state.
Nobody have to make big changes here, LF use Luos to get access to physical ressources and Luos use LF to have deterministic and easy to make applications : 
Dessin sans titre

As we discuss with @lsk567, currently in this configuration Luos just destroy the determinism of LF because there is no precise timestamp on the actual hardware measurement, so it's not interesting for both of us right now.

As you spotted @edwardalee actually the Luos weak point around determinism is time.

But at Luos we are planning to fix it and make a ns hardware based clock synchronization between nodes allowing to get a precise distant (or local) event date. Perhaps using this timestamped information we could succeed to have both benefits with a minimal work.

On the LF side there is nothing to do instead of making something work on an actual board. But as i understand this is the plan you already have. The Luos specific code simply can be done into the user code.
On the Luos side we have to be able to give you timestamp on data, and this feature fit with our plan. The LF code is writen to describe a program behavior on one or several containers.

Issue #292, opened 2021-03-22
Getting more precise timing on Linux


edwardalee:
This is not a high priority, but Linux offers the ability for a thread to sleep until an absolute time via the clock_nanosleep() system call. With the TIMER_ABSTIME flag, an absolutely timed sleep is possible. This is rather like the PRET "deadline" instruction and could lead to significantly more precise timing on Linux.

Soroosh129:
I want to point out that this is currently implemented in the `platform-abstraction` branch [here](https://github.com/icyphy/lingua-franca/blob/platform-abstraction/xtext/org.icyphy.linguafranca/src/lib/core/platform/lf_linux_support.c).

edwardalee:
Ah, I had forgotten about the platform-abstraction branch!  That implementation, doesn't quite do what I had in mind. The goal is explained nicely in this [blog by Björn Brandenburg](https://sigbed.org/2020/09/05/liu-and-layland-and-linux-a-blueprint-for-proper-real-time-tasks/). I think we would need to abstract the wait at a somewhat higher level.

lhstrh:
Duplicate of #220?

PR #304, opened 2021-03-31
Platform abstraction

Soroosh129:
This branch abstracts away few platform-specific functionalities related to time and thread management by providing a set of APIs in [`lib/core/platform.h`](https://github.com/icyphy/lingua-franca/blob/platform-abstraction/xtext/org.icyphy.linguafranca/src/lib/core/platform.h).

Eventually, platform abstraction should also include APIs for inter-process communication (e.g., using sockets) to enable more versatile federated programs, but that was not the original intention for this branch.


lhstrh:
We resolved to merge this branch into master, but there is still an issue related to it: #292. Is there anything that needs to happen before we can close this issue?

edwardalee:
That issue should remain open, but it need not block the merge.

PR #310, opened 2021-04-01
"trace" variable declared as extern since already in trace_util.c

jdeantoni:
"trace" variable declared as extern otherwise it does not compile at least on Linux with gcc 10.2.1:
`gcc -o trace_to_chrome trace_to_chrome.o trace_util.o
/usr/lib64/gcc/x86_64-suse-linux/10/../../../../x86_64-suse-linux/bin/ld: trace_util.o:(.bss+0x420): multiple definition of 'trace'; trace_to_chrome.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
make: *** [makefile:17: trace_to_chrome] Error 1
`

### DIFF
--- a/util/tracing/trace_to_chrome.c
+++ b/util/tracing/trace_to_chrome.c
@@ -41,7 +41,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define PID_FOR_UNKNOWN_EVENT 2000000
 
 /** Buffer for reading trace records. */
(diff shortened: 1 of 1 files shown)


edwardalee (commented):
Wouldn't it be better to remove this declaration from trace_to_chrome.c altogether?
The extern declaration is already in trace_util.h, which is #included in trace_to_chrome.c.

jdeantoni:
I hesitated between both approach. Removing it from trace_to_chrome.c works fine also. Should I do another PR ?

edwardalee:
Sure, thanks.

PR #316, opened 2021-04-09
Distributed cycles

Soroosh129:
This branch adds preliminary support for zero-delay cycles in a distributed system both using the centralized coordination and decentralized coordination. This is done by strategically inserting control reactions that will block the execution until it is safe to proceed.

There is still a huge caveat for decentralized applications where control reactions fail: it is possible for federates to incorrectly commit to a tag in certain circumstances.


edwardalee (commented):
This looks quite nice!  The java classes in federated look like a really good start towards a more modular architecture.  Note that these classes need some docs. But that can happen after a merge.  I realize this is a work in progress, but it's good progress! I suggest we merge.

I suggest renaming federated/ASTUtils.java to federated/FedASTUtils.java, or some such. It is confusing to have two classes in the tree with the same name.

Also, there are two directories with failing federated tests: tests/C/federated/failing and tests/C/federated/failing/knownfailed. Is that deliberate?

Note that I have added to GeneratorBase a method generatePortRef(), which is more specific than generateVarRef() and explicitly handles multiports and banks.

Soroosh129:
@edwardalee Thank you so much for the constructive comments. I've tried to address each one in individual commits as much as possible.

xtext/org.icyphy.linguafranca/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,9 +1,17 @@
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
```
lhstrh, line 3: I'm not sure this should be checked in. I keep coming back to this issue of folks pushing random Eclipse configuration data to the repo, which I think is a really bad idea, but I recall there being an unresolved discussion about this with @a-sr about this...

xtext/org.icyphy.linguafranca/src/org/icyphy/ASTUtils.xtend
```diff
@@ -517,12 +517,37 @@ class ASTUtils {
             rightFederate,
             rightBankIndex,
             rightChannelIndex,
-            action.inferredType
+            action.inferredType,
```
lhstrh, line 520: Seems like `makeCommunication` belongs in `FedASTUtils`, not `ASTUtils`?

xtext/org.icyphy.linguafranca/src/org/icyphy/federated/CGeneratorExtension.java
```diff
@@ -0,0 +1,239 @@
...
+
+import org.icyphy.generator.CGenerator;
+import org.icyphy.generator.FederateInstance;
+import org.icyphy.generator.ReactorInstance;
+import org.icyphy.linguaFranca.Input;
+import org.icyphy.linguaFranca.Port;
+import org.icyphy.linguaFranca.Reactor;
+import org.icyphy.linguaFranca.ReactorDecl;
+import org.icyphy.linguaFranca.VarRef;
+
```
lhstrh, line 13: Comments missing. What is this class for?

xtext/org.icyphy.linguafranca/src/org/icyphy/federated/FedASTUtils.java
```diff
@@ -0,0 +1,331 @@
...
+import org.icyphy.linguaFranca.Reactor;
+import org.icyphy.linguaFranca.Type;
+import org.icyphy.linguaFranca.Value;
+import org.icyphy.linguaFranca.VarRef;
+import org.icyphy.linguaFranca.Variable;
+
+/**
+ * @author Soroush Bateni {soroush@utdallas.edu}
+ *
+ */
```
lhstrh, line 53: Comments missing.

xtext/org.icyphy.linguafranca/src/org/icyphy/generator/GeneratorBase.xtend
```diff
@@ -1257,6 +1289,63 @@ abstract class GeneratorBase extends AbstractLinguaFrancaValidator {
...
+        int sendingBankIndex,
+        int sendingChannelIndex
+    ) {
+        throw new UnsupportedOperationException("This target does not support direct connections between federates.")        
+    }  
+    
+    /**
+     * 
+     */
+    def isFederatedAndDecentralized() {
```
lhstrh, line 1331: Comments. Also, maybe a better name would be `isDecentralizedFederation`? Same with the centralized version below.

lhstrh (commented):
Looks good overall! I browsed the federated C code, but there are too many new/experimental concepts floating around for me to make sense of it. I do like use of the term `STP` instead of `tardy`. I think that a holistic in-depth review of the federated runtime code (after we've worked out the fundamental issues that we were discussing in our last meeting) would be good to have at some point.

lhstrh:
@edwardalee @Soroosh129 I will take care of the merge with master if you don't mind. Please push whatever you have in the pipeline (whether this is incomplete or not) and let me know when this is done. After merging with master I expect you might need to do a fresh Oomph install to get everything working properly again.

edwardalee:
OK, I've pushed my latest (incomplete) changes to the distributed-cycles branch.
Let me know when I can start working again...

Edward

———————
Edward A. Lee
EECS, UC Berkeley
http://eecs.berkeley.edu/~eal

lhstrh:
OK, this is now up-to-date with master. The way to proceed is most likely to do a clean Oomph install. I've verified that this procedure now works again.

Issue #294, opened 2021-03-23
Fix C++ tests on Windows


cmnrd:
The new test infrastructure implemented in #289 enables platform independent testing. In principle, this allows running C++ test on Windows in CI. However, there appear to be still a few bugs in the tests or the build system which need to be fixed first.

lhstrh:
Related to the C++ CI workflows: is there a way we can speed up the build process? The build time of C++ compared to other targets is rather excessive. It would be interesting to give the C target `cmake` support (as we've discussed earlier), but I would be reluctant to use it if this means that build times become very large as a result.

cmnrd:
It is not necessarily cmake that slows things down. Of course cmake adds another step, but I found that its negligible compared to the time the C++ linker takes. Note that you are comparing apples and pears here. Compiling the C programs is much faster because it only compiles a single source file and thus linking is trivial. Splitting the generated code in multiple files and using a multi-step build process will likely also slow things down for C (independent of the build tool). Also note that compiling and linking C++ programs usually takes longer than for C programs.

To answer your question: yes it is possible to speed cmake up, but only by further refactoring the way we build LF projects. If we separate code generation from compilation, i.e. generate code for all LF files in the project and run compile only once, then cmake can be smarter about the build process by running compile commands for different binaries in parallel and also reusing shared dependencies (i.e. two LF files both importing the same library LF file).

PR #318, opened 2021-04-13
Fix MacOS and Windows tests, cleanup path calculations

### LINKED ISSUE #293: Fix C++ tests on Mac OS
#289 disabled CI tests for C++ on Mac OS. This is due to an unresolved linker problem, which is likely caused by the reorganization of the generated directory tree implemented in #289. This linker problem should be fixed and the Mac OS test re-enabled.

cmnrd:
This PR cleans up some remaining problems after the changes introduced in #289. In particular, this PR fixes and re-enables the CI tests for MacOS and Windows. This PR also cleans up path calculations (mostly in C++ generator, also a few changes in BaseGenerator) and removes deprecated methods from the FileConfig class.

Closes #293 and #294


lhstrh (approved):
Excellent! These changes look great. Thanks for the cleanup!

PR #319, opened 2021-04-14
Reorganize

lhstrh:
Simplification of paths/package structure.


cmnrd:
I fixed a small bug which rendered `lfc`'s rebuild mechanism useless. We should probably add tests that verify the correct operation of our scripts. I opened #320.

PR #321, opened 2021-04-14
Add lfc and run-lf-tests powershell scripts and fix a few portability issues

### LINKED ISSUE #230: Scripts under bin/ should run on Windows
This is closely related to #225 but I think this discussion is complex enough to deserve its own issue.

I installed the LF IDE in my Windows VM and after fixing a few things it turned out to work very well. Due to the magic of cmake, we can compile C++ programs from the IDE using MSVC (The native C/C++ compiler on Windows). The Windows Subsystem for Linux (WSL) is not required to get this working. However, now I am stuck because I can only compile LF programs from the IDE. All of our scripts in `bin` do not run on Windows. Most important, I cannot use the script for running tests.

I think we should carefully consider how we can make our scripts work on Windows. As far as I can see, there are 3 options that all come with their own set of drawbacks:

1. Keep everything as is and use bash scripts. This implies that the Windows Subsystem for Linux would be one of our requirements. I think this is quite a big dependency and I don't think  it is safe to assume that everyone has access to it. For instance, I am stuck with an older Windows build that does not support WSL yet because the function upgrades are hold back by my administrator.
2. Provide separate powershell scripts. This is probably the simplest solution, but also doubles the effort required for writing and maintaining scripts.
3. Switch to a portable scripting language such as Python. This is my preferred option, but obviously it also adds a big dependency. However, I think python is easier to install and use on most machines than WSL (Python can even be installed without administrator privileges)

What do you think?

cmnrd:
This PR adds PowerShell implementations of the lfc and run-lf-tests scripts. It also fixes a few compatibility issues that turned up in the Main class and GeneratorBase when running lfc on Windows. I have, however, a few remaining questions because I am a bit unsure where to place the powershell scripts and which scripts need to be ported.

- [] Should the powershell scripts be located in the same directory as the bash scripts?
- [] Do we need a powershell implementation of the build-lfc script?
- [] Do we need the build-lfc script at all? Most of its functionality is already covered by the lfc script.

Also note that this PR is based on the `reorganize` branch because the changes here are affected by the changes introduced in `reorganize`. Thus, this PR should only be merged after #319 is merged.

Closes #230.


org.lflang/src/org/lflang/MainConflictChecker.java
```diff
@@ -80,6 +80,7 @@ public MainConflictChecker(FileConfig fileConfig) {
          */
         @Override
         public FileVisitResult visitFile(Path path, BasicFileAttributes attr) {
+            path = path.normalize();
```
lhstrh, line 83: :+1:

lhstrh:
Looks good!

PR #326, opened 2021-04-28
Add CLI option to clean artifacts before building

cmnrd:
Essentially this change allows running `lfc --clean <file>` which will delete all existing build artifacts before generating code and invoking the target compiler. For the C++ target, this is quite useful as changing build related target properties might cause conflicts with already existing artifacts. While ultimately the goal should be to make the build process more robust, having the clean option is very convenient for now.

There are a few things to note:
- This changes the meaning of the `-c`  CLI flag. In the past `-c` was used as a shortcut for the `--target-compiler` option. However, I would argue that using clean is more common than changing the target compiler. Also, I think using `-c` for clean is more intuitive to users.
- This only works in the standalone mode and is not supported in the Eclipse IDE. We should probably hook the newly introduced `doClean()` method to the Eclipse clean button somehow.
- This seems to work well for C++, but it hasn't been tested for other targets. Also, other targets should override `doClean()` if they create directories besides `bin` and `src-gen`.


org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -440,6 +443,41 @@ abstract class GeneratorBase extends AbstractLFValidator {
         // use the result.
     }
 
+    /**
+     * Check if a clean was requested from the standalone compiler and perform the clean step.
+     */
+    protected def void cleanIfNeeded(IGeneratorContext context) {
```
lhstrh, line 449: Seem like a better place to put this would be `FileConfig`. Note that `context` is also accessible in that class.

org.lflang/src/org/lflang/generator/Main.java
```diff
@@ -195,7 +195,8 @@ public static String redAndBold(String s) {
      * @author Marten Lohstroh <marten@berkeley.edu>
      */
     enum CLIOption {
-        COMPILER("c", "target-compiler", true, false, "Target compiler to invoke.", true),
+        COMPILER(null, "target-compiler", true, false, "Target compiler to invoke.", true),
+        CLEAN("c", "clean", false, false, "Clean before building.", true),
```
lhstrh, line 199: This adds the flag, but `context` is not set in response to option being enabled. Are you sure this even works yet?

lhstrh (changes requested):
Good idea; added comments.

org.lflang/src/org/lflang/generator/Main.java
cmnrd: I don't understand this comment. Yes it works :) Since the last boolean in the constructor (`passOn`) is set to true, some magic elsewhere in the Main class automatically adds the option to the context.

lhstrh: Ah, yes. I wrote that magic :-)

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
cmnrd: The actual cleaning is target dependent, so it wouldn't really make sense to factor this out. I think what could be factored out is the check whether cleaning is needed or not. I am not sure if `FileConfig` is really the right place for it, but I could add a flag `needsClean` or something similar there. Probably the `deleteDirectory` should also be moved to `FileConfig`.

lhstrh: Yes, my suggestion would be to move all of these new methods to `FileConfig`. Note that we can have target-specific extensions of `FileConfig`; the TypeScript generator so far is the only one that needed it, but we can easily create one for Cpp as well.

cmnrd: Can you point me to an example of how this extension mechanism works? I did not find it in `FileConfig.java` nor in `TypeScriptGenerator.xtend`.

lhstrh: [`FileScriptFileConfig.java`](https://github.com/icyphy/lingua-franca/blob/master/org.lflang/src/org/lflang/generator/TypeScriptFileConfig.java) is a specialized version of `FileConfig`, tailored to TypeScript. `TypeScriptGenerator` [overrides](https://github.com/icyphy/lingua-franca/blob/9bbb747d6e32db816c3420d3895a0842bdfb6d3b/org.lflang/src/org/lflang/generator/TypeScriptGenerator.xtend#L1370) `setFileConfig` in order to assign an instance of `TypeScriptFileConfig` to the `fileConfig` class variable specified in `GeneratorBase`.

cmnrd: I moved the code over to `FileConfig`

lhstrh (approved):
Very clean! Thanks!!

PR #331, opened 2021-05-13
Ability to generate an RCP client with Maven

lhstrh:


lhstrh (commented):
This includes some IntelliJ/Eclipse configuration files that I'm not sure belong in the repo. We can remove them later if it turns out they shouldn't be there.

PR #345, opened 2021-05-21
Port the C++ Generator to Kotlin

cmnrd:
This branch ports the C++ code generator from Xtend to Kotlin. This effort is still a work in progress. I am opening this PR, so that everyone can see what is going on and get involved in a discussion about which "style" or patterns to use.

I configured this PR to compare against the branch `kotlin-with-gradle`. This way we can focus on the C++ related changes here. However, we will also need to merge the `kotlin-with-gradle` branch to make this PR work.


org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+        fsa.generateFile(relSrcGenPath.resolve("CMakeLists.txt").toString(), generateCmake())
+
+        // copy static library files over to the src-gen directory
+        val genIncludeDir = srcGenPath.resolve("__include__")
+        copyFileFromClassPath("${libDir}/lfutil.hh", genIncludeDir.resolve("lfutil.hh").toString())
+        copyFileFromClassPath("${libDir}/time_parser.hh", genIncludeDir.resolve("time_parser.hh").toString())
+        copyFileFromClassPath("${libDir}/3rd-party/CLI11.hpp", genIncludeDir.resolve("CLI").resolve("CLI11.hpp").toString())
+    }
+
+    @Suppress("LocalVariableName") // allows us to use capital S as variable name below
```
oowekyala, line 146: You could suppress this for just the relevant statement I think. Like

```kotlin
@Suppress("LocalVariableName")
val S = '$'
```

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+    private fun generateCmake(): String {
+        val runtimeVersion = targetConfig.runtimeVersion ?: "26e6e641916924eae2e83bbf40cbc9b933414310"
```
oowekyala, line 148: Where does this default version comes from ? Would be nice to put this in a named `const val` in the companion object

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+
+        // FIXME Is there a way to simplify this line?
+        val includeFile = if (targetConfig.cmakeInclude.isNullOrBlank()) null else FileConfig.toUnixString(fileConfig.srcPath.resolve(targetConfig.cmakeInclude))
```
oowekyala, line 151: You could write

```kotlin
targetConfig.cmakeInclude
            ?.takeIf { it.isNotBlank() }
            ?.let { fileConfig.srcPath.resolve(it).toUnixString() }
```
It's more straightforward I think, but if you don't find this clearer, I would suggest just writing the `if` and `else` on separate lines to make it easier to read (and using your `toUnixString` extension)

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+                |   set_target_properties(reactor-cpp PROPERTIES IMPORTED_LOCATION "$S{REACTOR_CPP_LIB_DIR}/$S{REACTOR_CPP_LIB_NAME}")
+                |endif()
+                |
+                |if (APPLE)
+                |   file(RELATIVE_PATH REL_LIB_PATH "$S{REACTOR_CPP_BIN_DIR}" "$S{REACTOR_CPP_LIB_DIR}")
+                |   set(CMAKE_INSTALL_RPATH "@executable_path/$S{REL_LIB_PATH}")
+                |else ()
+                |   set(CMAKE_INSTALL_RPATH "$S{REACTOR_CPP_LIB_DIR}")
+                |endif ()
+            """.trimIndent()
```
oowekyala, line 220: I think this trimIndent is unnecessary, also the one in the other `if` branch

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+            |    "$S{PROJECT_SOURCE_DIR}"
+            |    "$S{PROJECT_SOURCE_DIR}/__include__"
+            |)
+            |target_link_libraries($S{LF_MAIN_TARGET} reactor-cpp)
+            |
+            |install(TARGETS $S{LF_MAIN_TARGET})
+        ${
+            if (includeFile == null) "" else """
+                |
+                |include($includeFile)""".trimIndent()
```
oowekyala, line 242: same for this `trimIndent`.
Also this dollar block could be written `${ includeFile?.let { "|include($it)" }.orEmpty() }`

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+    private val Resource.preambleSourcePath get() = this.genDir.resolve("_lf_preamble.cc")
+
+    /** Path to the header file corresponding to this reactor */
+    private val Reactor.headerPath get() = this.eResource().genDir.resolve("${this.name}.hh")
+
+    /** Path to the implementation header file corresponding to this reactor (needed for generic reactors) */
+    private val Reactor.headerImplPath get() = this.eResource().genDir.resolve("${this.name}_impl.hh")
+
+    /** Path to the source file corresponding to this reactor (needed for non generic reactors)  */
+    private val Reactor.sourcPath get() = this.eResource().genDir.resolve("${this.name}.cc")
```
oowekyala, line 105: typo: sourcePath

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+import org.lflang.Target
+import org.lflang.lf.Action
+import org.lflang.lf.Reactor
+import org.lflang.lf.VarRef
+import java.nio.file.Path
+
+/* *******************************************************************************************
+ * The following definitions are shortcuts to access static members of FileConfig and ASTUtils
+ *
+ * TODO these should likely be moved to a common place in the future
```
oowekyala, line 42: :+1:

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+ * This will produce the expected output:
+ * ```
+ * class Bar {
+ *     void foo() {
+ *         // do something useful
+ *     }
+ * };
+ * ```
+ */
+operator fun String.rangeTo(str: String) = str.prependIndent(this)
```
oowekyala, line 82: Should this be a call to `replaceIndent` instead? Otherwise the common indent of the `str` parameter is preserved. I'm not sure, this matters in the use cases you have below, but this could make a difference in the future

org.lflang/src/org/lflang/generator/CppGenerator.kt
```diff
@@ -0,0 +1,263 @@
...
+            """.trimIndent()
+        }
+            |
+            |set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
+            |
+            |set(LF_MAIN_TARGET ${topLevelName})
+            |
+            |add_executable($S{LF_MAIN_TARGET}
+            |    main.cc
+        ${" |    "..sourceFiles.joinToString("\n")}
```
oowekyala, line 229: This looks nice but I'm not sure how to pronounce it... Reading `..` as `rangeTo` sounds weird here. Maybe using an infix function would be more readable? Something like:
```kt
" |    " indenting sourceFiles.joinToString("\n")
```

oowekyala (commented):
This looks really good! I think you already made yourself at home with extensions, my comments are really just minor nitpicks

org.lflang/src/org/lflang/generator/CppGenerator.kt
cmnrd: Yes, I first used an infix function, but I could find a concise name, `indent` just seems too long... I really like the look of the `..` operator here, but yes, I am completely abusing it.

I guess this is something we should discuss with the others and find a solution we all agree on.

org.lflang/src/org/lflang/generator/CppGenerator.kt
cmnrd: Probably, yes. I didn't know about `replaceIndent`

org.lflang/src/org/lflang/generator/CppGenerator.kt
cmnrd: I am not so sure about that. I added it because `trimIndent()` also removes blank lines at the beginning and end of the string. I would assume that without the call, there would be an additional empty line in the resulting string, but I did not try what happens without `trimIndent()`.

org.lflang/src/org/lflang/generator/CppGenerator.kt
cmnrd: Oh, ok. I just trusted IntelliJ here to do the right thing

org.lflang/src/org/lflang/generator/CppGenerator.kt
cmnrd: Removing the `trimIndent()` indeed leads to insertian of an aditional blank line with some spaces in the resulting cmake file (basically line 220 gets inserted into the resulting file without rhe trimIndent). While this isn't critical, I think its cleaner with the `trimIndent()` in place.

cmnrd:
@lhstrh  All the C++ pass now. There is some weird error in one of the C tests where I don't really know where it comes from and whether it is because of one of my changes.

I am currently working on some improvements, mostly moving code around adding a separate error reporter class that works independent of GeneratorBase.

org.lflang/src/org/lflang/generator/IErrorReporter.java
```diff
@@ -0,0 +1,56 @@
...
+    String reportError(EObject obj, String message);
+
+
+    /**
+     * Report a warning on the receiving parse tree object.
+     *
+     * @param obj     The parse tree object.
+     * @param message The error message.
+     */
+    String reportWaring(EObject obj, String message);
```
oowekyala, line 55:
```suggestion
    String reportWarning(EObject obj, String message);
```

org.lflang/src/org/lflang/generator/cpp/CppActionGenerator.kt
```diff
@@ -0,0 +1,80 @@
...
+            " |"..reactor.actions.joinToString("\n", "// actions\n", "\n") { generateDeclaration(it) }
+        }
+            |// default actions
+            |reactor::StartupAction $startupName {"$startupName", this};
+            |reactor::ShutdownAction $shutdownName {"$shutdownName", this};
+        """.trimMargin()
+    }
+
+    /** Get all action initializiers */
+    fun geberateInitializers() =
```
oowekyala, line 78: typo: generate

org.lflang/src/org/lflang/generator/cpp/CppActionGenerator.kt
```diff
@@ -0,0 +1,80 @@
...
+            //action.reportError(
+            //    "minSpacing and spacing violation policies are not yet supported for logical actions in reactor-ccp!");
+        } else {
+            val time = action.minDelay?.toTime() ?: "reactor::Duration::zero()"
+            """, ${action.name}{"${action.name}", this, $time}"""
+        }
+    }
+
+    private fun initializePhysicalInitializer(action: Action): String {
+        return if (action.minDelay != null || action.minSpacing != null || !action.policy.isNullOrEmpty()) {
```
oowekyala, line 56: you could use the  expression body syntax for these methods. Also you could write them as eg
```kotlin
fun initializePhysicalInitializer(action: Action): String = with (action) {
   if (minDelay != null || minSpacing != null || ...)
    ...
}
```
this allows you to omit the `action.` prefix without making the function an extension.

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -0,0 +1,198 @@
...
+        // check if the dependency is a multiport
+        if (variable is Port && variable.widthSpec != null) {
+            return """
+                for (unsigned i = 0; i < ${antidependency.name}.size(); i++) {
+                  ${reaction.name}.declare_antidependency(&${antidependency.name}[i]);
+                }
+            """.trimIndent()
+        }
+        // treat as single antidependency otherwise
+        return if (variable is Action) "${reaction.name}.declare_scheduable_action(&${antidependency.name});"
```
oowekyala, line 80:
```suggestion
        return if (variable is Action) "${reaction.name}.declare_schedulable_action(&${antidependency.name});"
```
?

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /** Get a list of PortReferences for the given list of variables
+     *
+     * This checks whether the variable refers to a multiport and generated an instance of
+     * PortReferrence for each port instance in the multiport. If the port is containe in a
+     * multiport, the result includes instances PortReference for each pair of bank and multiport
+     * instance.
+     */
+    private fun enumarateAllPortsFromReferences(references: List<VarRef>): List<PortReference> {
```
oowekyala, line 148:
```suggestion
    private fun enumerateAllPortsFromReferences(references: List<VarRef>): List<PortReference> {
```

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -0,0 +1,198 @@
...
+            "${it.first.toCode()}.bind_to(&${it.second.toCode()});"
+        }
+    }
+
+    /**
+     * Generate the definition of the reactor's assemble() method
+     *
+     * The body of this method will declare all triggers, dependencies and antidependencies to the runtime.
+     */
+    fun generateDefinition() = with(prependOperator) {
```
oowekyala, line 189: :+1:

org.lflang/src/org/lflang/generator/cpp/CppExtensions.kt
```diff
@@ -0,0 +1,305 @@
...
+/* *******************************************************************************************
+ * The following definitions are shortcuts to access static members of FileConfig and ASTUtils
+ *
+ * TODO these should likely be moved to a common place in the future
+ */
+
+val Resource.name: String get() = FileConfig.getName(this)
+
+fun Path.toUnixString(): String = FileConfig.toUnixString(this)
+fun Path.createDirectories() = FileConfig.createDirectories(this)
```
oowekyala, line 43: Actually kotlin 1.5.0 added an identical extension to the standard library. I think I need some more config to make it available, as that part of the stdlib is JVM-specific and distributed in a separate dependency.

org.lflang/src/org/lflang/generator/cpp/CppMainGenerator.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.lflang.lf.Reactor
+
+/** C++ code generator responsible for generating the main file including the main() function */
+class CppMainGenerator(
+    private val main: Reactor,
+    private val targetConfig: TargetConfig,
+    private val fileConfig: CppFileConfig,
+) {
+
+    private fun generateParmeterParser(param: Parameter): String {
```
oowekyala, line 15:
```suggestion
    private fun generateParameterParser(param: Parameter): String {
```

oowekyala (commented):
Great work @cmnrd :) Especially splitting up the logic makes the code much easier to review.

org.lflang/src/org/lflang/AstExtensions.kt
```diff
@@ -281,3 +300,88 @@ fun <T> List<T>.tail() = subList(1, size)
...
+ * the width if it can be determined and -1 if not.
+ * It will not be able to be determined if either the
+ * width is variable (in which case you should use
+ * {@link inferPortWidth(VarRef, Connection, List<Instantiation>})
+ * or the list of instantiations is incomplete or missing.
+ * If there are parameter references in the width, they are
+ * evaluated to the extent possible given the instantiations list.
+ *
+ * The instantiations list is as in
+ * {@link initialValue(Parameter, List<Instantiation>}.
```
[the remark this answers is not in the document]
oowekyala: (These comments about javadoc are mostly about awareness, IMO they can be fixed later more thoroughly)

org.lflang/src/org/lflang/generator/ErrorReporter.kt
```diff
@@ -0,0 +1,150 @@
+package org.lflang.generator
+
+import org.eclipse.core.resources.IMarker
+import org.eclipse.core.resources.IResource
+import org.eclipse.emf.ecore.EObject
+import org.eclipse.emf.ecore.resource.Resource
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils
+import org.lflang.generator.cpp.CppFileConfig
+
+class ErrorReporter(private val fileConfig: CppFileConfig) : IErrorReporter {
```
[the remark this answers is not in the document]
lhstrh: Note that we've added an `ErrorReporter` interface in the `reaction-graph` branch that is about to get merged into `master`.

org.lflang/src/org/lflang/generator/IErrorReporter.java
lhstrh: See [this new interface](https://github.com/icyphy/lingua-franca/blob/reaction-graph/org.lflang/src/org/lflang/ErrorReporter.java).

org.lflang/src/org/lflang/AstExtensions.kt
```diff
@@ -281,3 +300,88 @@ fun <T> List<T>.tail() = subList(1, size)
...
+ * @return The inferred type, or "undefined" if none was declared.
+ */
+val Port.inferredType: InferredType get() = ASTUtils.getInferredType(this)
+
+/**
+ * Report whether a state variable has been initialized or not.
+ * @receiver The state variable to be checked.
+ * @return True if the variable was initialized, false otherwise.
+ */
+val StateVar.isInitialized :Boolean get() = (this.parens.size == 2)
```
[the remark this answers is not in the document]
cmnrd: It seems to work fine for me most of the time... Not sure why it breaks sometimes. I just run the formatter again on the file and it changed it.

org.lflang/src/org/lflang/generator/ErrorReporter.kt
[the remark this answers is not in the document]
cmnrd: Thanks for pointing to the ErrorReporter in reaction-graph! I will then remove the current error reporter interface and rename the `ErrorReporter` class in Kotlin to `GeneratorErrorReporter` implementing what `GeneratorBase` currently provides.

lhstrh (approved):
Amazing work! There are some comments missing, but this will become more visible when we generate KDoc documentation, so we can also address it then.

org.lflang/src/org/lflang/generator/ErrorReporter.kt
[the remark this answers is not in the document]
cmnrd: Done.

org.lflang/src/org/lflang/generator/ErrorReporter.kt
```diff
@@ -0,0 +1,150 @@
...
+     * @param obj     The parse tree object.
+     * @param message The error message.
+     */
+    override fun reportWaring(obj: EObject, message: String): String = report(message, IMarker.SEVERITY_WARNING, obj)
+
+    /** Report an error on the receiving parse tree object.
+     *  @receiver The parse tree object.
+     *  @param message The error message.
+     */
+    fun EObject.extReportError(message: String): String = report(message, IMarker.SEVERITY_ERROR, obj = this)
```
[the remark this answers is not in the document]
cmnrd: This is not relevant anymore because I removed the extension function from the interface. However, I had the problem that the java `IErrorReporter` interface (or `ErrorReporter`) now, defined the method `String reportError(EObject obj, String message)`. The Kotlin implementation of this interface overrides this method. Now Kotlin does not allow me to define an additional extension function with the same name. Hence, the `ext` prefix.

For now, I simply removed the extensions, because it doesn't seem possible to define them as part of the Java interface.

org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -18,51 +26,170 @@
...
+            default: {
+                throw new RuntimeException("Unexpected target!");
             }
         }
+    }
+
+    /**
+     * Create a C++ code generator
+     * 
+     * Since the CppGenerator class is implemented in Kotlin, the class is is
```
lhstrh, line 132:
```suggestion
     * Since the CppGenerator class is implemented in Kotlin, the class is
```

lhstrh (approved):

PR #356, opened 2021-06-04
Kotlin support

### LINKED ISSUE #337: Configure Maven to compile Kotlin code
Now that we rely on Maven for the production of our Eclipse RCA, we should make sure that it is capable of dealing with Kotlin code as well. Currently, the configuration in the `kotlin-with-gradle` branch is purely based on Gradle...

oowekyala:
This adds support for writing code in Kotlin. The output language of Xtext is now Java, not Xtend. The gradle and maven builds are updated to make the kotlin compiler run.  They're also patched to work around Xtext regenerating Java files. Namely there's a short list of those files [here](https://github.com/icyphy/lingua-franca/blob/kotlin-with-gradle/org.lflang/build.gradle#L36-L42) and [here](https://github.com/icyphy/lingua-franca/blob/kotlin-with-gradle/org.lflang/pom.xml#L102-L108)

The build process now runs like so:
1. generate Java from Xtend sources (with Java sources in sourcepath)
2. generate bytecode from Kotlin sources (with output of step 1 in sourcepath)
3. generate bytecode from Java sources produced by step 1 (with Kotlin bytecode in classpath)

This means 
* Xtend sources may use only Xtend and Java sources,
* Kotlin and Java sources may use all of Kotlin, Java, and Xtend sources.

Many files are ported from Xtend to Kotlin or Java. Note that #345 is built on these changes.

### TODO

* Tests are run during the maven build, but we could also disable them, since the CI only uses gradle. 
* Since 4c0f025eca83fc81c3f0b080cfd9fcd3800d4374 I'm not sure we need cca6fdc. But I'm not sure what generates the `.target` file either.. @a-sr could you take a look?

closes #337


cmnrd (approved):
This looks really good to me. It seems that the recent commits solved the build issues I had in the past. It works now out of the box, both in the eclipse oomph install and in the standalone gradle build.

It should be safe to merge this now.

cmnrd (changes requested):
I stand corrected... I didn't double-check and did not realize that my oomph installer checked out master instead of this branch. With the recent changes  (from `./gradlew eclipse) the eclipse build on my end still fails, now complaining that dependencies are not found. This is because, they are given in absolute paths.

org.lflang.ide/.classpath
```diff
@@ -53,7 +53,7 @@
 			<attribute name="gradle_used_by_scope" value="main,test"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry sourcepath="/home/clem/.gradle/caches/modules-2/files-2.1/com.google.inject/guice/3.0/d117c4481f64955cb1862ed38cb568add2f5fe05/guice-3.0-sources.jar" kind="lib" path="/home/clem/.gradle/caches/modules-2/files-2.1/com.google.inject/guice/3.0/9d84f15fe35e2c716a02979fb62f50a29f38aefa/guice-3.0.jar">
+	<classpathentry sourcepath="/home/clem/.gradle/caches/modules-2/files-2.1/com.google.inject/guice/5.0.1/2c1282746b8307a049a95e2374c855378c9f29e2/guice-5.0.1-sources.jar" kind="lib" path="/home/clem/.gradle/caches/modules-2/files-2.1/com.google.inject/guice/5.0.1/dae7556b441cada2b4f0a2314eb68e1ff423429/guice-5.0.1.jar">
```
lhstrh, line 56: These entries are problematic because they feature full paths that only exist on your machine... Also raised this with @a-sr and @cmnrd earlier today. In master, not `.classpath` files are tracked. Why are they in this branch?

org.lflang/build.gradle
```diff
@@ -81,3 +81,13 @@ task generateStandaloneCompiler() {
 }
 
 generateStandaloneCompiler.finalizedBy shadowJar
+
+task runLfc(type: JavaExec) {
```
cmnrd, line 85: Interesting. Do you know that the `lfc` script is capable of something similar? If you run `lfc -r`, it rebuilds itself via gradle before running.

cmnrd (approved):
It should be safe to merge this now!

Since this MR does not include any Kotlin files that are actually used, no one should notice a difference. Mostly, this MR changes configurations so that Kotlin files can be compiled in Gradle, Maven and (only in theory...) in Eclipse.

When merging this, we should notify everyone to do a fresh oomph install. I would also recommend to use a fresh git clone for this. This is because we removed several xtend files, but heir generated artifacts are not under version control. If the xtend-gen artifacts are not deleted manually (or by using a fresh clone), there will likely be build errors due to duplicated class definitions. This is also important to know when switching between branches.

While testing the new oomph setup, I noticed to issues. First, messages like _"Kotlin Analysis" has encountered a problem_ tend to appear sometimes. Second, my install always results in a state where there are errors shown for `org.lflang.web`. However, triggering recompilation of the files or cleaning this project removes the error messages.

org.lflang/build.gradle
oowekyala: Ah ok I didn't know. I added the gradle task to be able to run & debug it from intellij, I'm not sure it would be possible with a bash script

PR #358, opened 2021-06-04
CI/nightly build

lhstrh:
Workflow developed by @Feliix42 for a nightly build.


Feliix42:
I have tested the Action locally via `act`, and it appears that the job works: https://github.com/icyphy/lingua-franca/releases/tag/nightly
Once it ran for a few days we'll see if everything works as intended (i.e. the replacing of files works).

One nitpick though: The nightly release will be updated every time but the timestamp will still be the same one from about 2 hours ago when I created it initially. Also the file names for the archives from different days will be indistinguishable, because the Action I used can't delete files, but overwrite existing ones. So to avoid ending up with an ever-growing number of files attached to the release, I just rename them before uploading them. Maybe that will have to be changed in the future. Also the description of the nightly release could be changed if desired.

Anyway: The PR should be ready to merge now as soon as @lhstrh has set up the token for the GitHub API. :tada: Maybe a Squash Merge would be in order to keep the history clean, but do as you wish.

lhstrh:
I agree that it would be desirable to be able to remove files and replace when with filenames that have a date incorporated in them.

Otherwise it becomes hard to track the origin of downloaded artifacts.

My guess this (as well as addressing the timestamp issue) can be addressed by removing and recreating the release in its entirety?

We should also produce `lfc.tar.gz` and `lfc.zip` and attach them to the release, just like we did with the alpha releas.

lhstrh:
I just noticed this actions that might be useful: 
- https://github.com/marketplace/actions/remove-a-release-asset
- https://github.com/marketplace/actions/recreate-release

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,56 @@
...
+          artifacts: "upload_data/*"
+          body: "## Epoch IDE Nightly Build
+
+This is the continuously updated Nightly release of the Lingua Franca IDE, Epoch.
+Expect instabilities and bugs; use at your own discretion."
+          commit: master
+          tag: nightly
+          name: "Epoch IDE Nightly Release"
+          prerelease: true
+          token: ${{ secrets.GITHUB_TOKEN }}
```
lhstrh, line 56:
```suggestion
          token: ${{ secrets.NIGHTLY_BUILD }}
```

Feliix42:
Ok, I added the `lfc` build and the asset deletion/reupload. There may be an issue still with the `sed` shenanigans I did in the `.github/scripts/package_lfc.sh`, but I can only debug these when the script is running on github, so maybe we could merge the PR and I fix what's left open on `master` (or in a separate PR)?

.github/scripts/package_lfc.sh
```diff
@@ -0,0 +1,32 @@
...
+outname="lfc_nightly_$(date '+%Y%m%d-%H%M%S')"
+
+# build lfc compiler
+./gradlew generateStandaloneCompiler
+
+# assemble the files in a separate directory
+mkdir -p $outname/bin
+mkdir -p $outname/lib
+
+mv org.lflang/build/libs/org.lflang-0.1.0-SNAPSHOT-all.jar $outname/lib
```
cmnrd, line 16: Should this better be `org.lflang/build/libs/org.lflang-*-SNAPSHOT-all.jar` to account for future version updates? I think the yml file below does something similar for the RCA artifact.

cmnrd (approved):
This looks good! Apart from the small comment I made, I think this is ready to go. I also think that any bug fixes can be done directly on master should this be needed.

.github/scripts/package_lfc.sh
Feliix42: Good point, I will fix that.

PR #365, opened 2021-06-10
LF filetype Vim plugin

HugoForrat:
This is a little Vimscript code to add a Lingua Franca filetype to (neo)vim that I've written that others may benefit from.
It brings basic syntax highlighting and rudimentary text-object support for the target code blocks.

It can be installed by copying the three directories (`ftplugin`, `ftdetect`, `syntax`) and their content into:

- `~/.vim/` if you're using Vim
- `~/.config/nvim/` if you're using Neovim
- Or really anywhere in your Vim `'runtimepath'`

(One can of course, simply copy the content of the directories if they already exist).

### DIFF
--- a/util/lingua-franca.vim/ftdetect/linguafranca.vim
+++ b/util/lingua-franca.vim/ftdetect/linguafranca.vim
@@ -1,0 +1,3 @@
+" ftdetect/linguafranca.vim
+
+autocmd BufNewFile,BufRead *.lf setfiletype linguafranca
--- a/util/lingua-franca.vim/ftplugin/linguafranca.vim
+++ b/util/lingua-franca.vim/ftplugin/linguafranca.vim
@@ -1,0 +1,26 @@
+" ftplugin/linguafranca.vim
+" Arbitrary linguafranca related vim code
+
+setlocal commentstring=//%s
+setlocal formatoptions=jcroq
+" 'comments' configuration stolen from the C file, let's see how it works
+setlocal comments=sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,://
+
+" function! CursorInTargetBlock()
+"   " TODO
+"   " See: :help searchpair(), searchpairpos(), `n` flag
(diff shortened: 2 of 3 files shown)


cmnrd (approved):
This is great! @lhstrh should we merge this?

Soroosh129:
I noticed that this really great feature is undocumented. Would it be possible to document this under `Downloading and Building` on the Wiki, alongside Eclipse and IntelliJ?

I think `vim` + `lfc` is a very viable development flow. What do you guys think @lhstrh @cmnrd @HugoForrat?

HugoForrat:
>I think `vim` + `lfc` is a very viable development flow.

Yes absolutely, this has been very convenient especially when working within Autoware's ADE image.
I can write some documentation which is indeed lacking. This should amount to little more than what's in the first message of this PR.

However, as @Feliix42 said in Slack, having this in it's own repository in the LF organisation could make the install process even more straightforward, provided that the plugin repository is public. But I've noticed that all repos there are private; is this a conscious decision or a simple coincidence?

Feliix42:
Regarding the latter: I already prepped a repository and moved the code in there with some rudimentary documentation. If there's no opposition, I can create a public repository in the LF organization and push the code. :)

Feliix42:
The repository is there now; just let me know when we can make it public, then I'll also remove the plugin from the util folder.

edwardalee:
The lfc+vim plugin should be documented in [Downloading and Building](https://github.com/icyphy/lingua-franca/wiki/Downloading-and-Building), at a minimum with a pointer to documentation in the new repo.

HugoForrat:
I've edited the [Downloading and Building](https://github.com/icyphy/lingua-franca/wiki/Downloading-and-Building) wiki page, making the assumption that the plugin repository will be made public. We'll need to correct the page if we choose to keep it private.

cmnrd:
Thanks @Feliix42! I don't think there is any reason to keep it private ;) Did you already remove the plugin within the LF repo?

PR #370, opened 2021-06-14
Show printed messages in the console view of the LF IDE

cmnrd:
As the title says, this change redirects all messages printed via `println()` to the internal Console view of our IDE. Effectively, this makes all messages appear in the internal view of our IDE and no longer in the host IDE if running from Eclipse (or on the command line when running the product).

I am not sure where the best place is to put this code. The `LFUiModule` class was the best place I could find, and it seems to work well. We probably should also consider introducing some logging mechanism that lets us distinguish between messages we want to show the user in the Command view and other messages that are only interesting to developers.

### DIFF
--- a/org.lflang.ui/META-INF/MANIFEST.MF
+++ b/org.lflang.ui/META-INF/MANIFEST.MF
@@ -19,7 +19,8 @@ Require-Bundle: org.lflang,
  org.eclipse.xtext.xbase.lib;bundle-version="2.14.0",
  org.eclipse.xtend.lib;bundle-version="2.14.0";resolution:=optional,
  org.eclipse.ui.intro,
- org.eclipse.ui.intro.universal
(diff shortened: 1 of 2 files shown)


edwardalee:
Cool!  Next step would be to make it simple to run the generated code and view its output on this console.

cmnrd:
Definitely! I am not sure how we can invoke the binary, but for printing its output I think it is even better to create a new console. Consoles are named and you can switch between them in the IDE. Currently, there is only the one "LF output", but we could add new consoles for each binary we run to make the output distinct from our LF output.

lhstrh (approved):
Very cool! Yes, I think `LFUiModule` is a good place to put it. That said, given that this is a generated file, I'm surprised that it's still an Xtend file; it should be a Java file...

PR #371, opened 2021-06-14
Better platform abstraction for lf_clock_gettime() and _LF_CLOCK

Soroosh129:
This makes _LF_CLOCK, which is platform-dependent, an internal property of the given platform's implementation files.

A consequence of that is a backward-incompatible change to lf_clock_gettime(). Any platform support library in the future would have to change (e.g., the RISCV bare-metal support).

This also makes get_physical_time() more platform independent, mainly by using the new lf_clock_gettime().

### DIFF
--- a/org.lflang/src/lib/core/platform.h
+++ b/org.lflang/src/lib/core/platform.h
@@ -159,7 +159,7 @@ extern int lf_cond_wait(lf_cond_t* cond, lf_mutex_t* mutex);
  * @return 0 on success, LF_TIMEOUT on timeout, and platform-specific error
  *  number otherwise.
  */
-extern int lf_cond_timedwait(lf_cond_t* cond, lf_mutex_t* mutex, long long absolute_time_ns);
+extern int lf_cond_timedwait(lf_cond_t* cond, lf_mutex_t* mutex, instant_t absolute_time_ns);
 
 #endif
 
--- a/org.lflang/src/lib/core/platform.h
+++ b/org.lflang/src/lib/core/platform.h
@@ -65,6 +65,24 @@ typedef _lf_thread_t lf_thread_t;        // Type to hold handle to a thread
 typedef _lf_time_spec_t lf_time_spec_t;  // Type to hold time in a traditional {second, nanosecond} POSIX format
 typedef _lf_clock_t lf_clock_t;          // Type to hold a clock identifier (e.g., CLOCK_REALTIME on POSIX)
 
+/**
+ * Time instant. Both physical and logical times are represented
+ * using this typedef.
+ * WARNING: If this code is used after about the year 2262,
+ * then representing time as a 64-bit long long will be insufficient.
+ */
+typedef long long instant_t;
+
+/**
+ * Interval of time.
+ */
+typedef long long interval_t;
+
+/**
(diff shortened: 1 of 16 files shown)


org.lflang/src/lib/core/platform/lf_linux_support.c
```diff
@@ -42,10 +42,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 /**
- * Fetch the value of clk_id and store it in tp.
+ * Fetch the value of _LF_CLOCK and store it in tp.
```
edwardalee, line 45: What is the return value?

org.lflang/src/lib/core/platform/lf_macos_support.c
```diff
@@ -42,10 +42,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 /**
- * Fetch the value of clk_id and store it in tp.
+ * Fetch the value of _LF_CLOCK and store it in tp.
```
edwardalee, line 45: Return value?

org.lflang/src/lib/core/platform/lf_windows_support.c
```diff
@@ -154,13 +154,13 @@ int lf_cond_timedwait(_lf_cond_t* cond, _lf_critical_section_t* critical_section
 }
 
 /**
- * Fetch the value of clk_id and store it in tp.
+ * Fetch the value _LF_CLOCK and store it in tp.
```
edwardalee, line 157: Return value?

edwardalee (approved):
Looks good to me! Minor documentation requests...

org.lflang/src/lib/core/platform.h
```diff
@@ -129,9 +129,15 @@ extern int lf_cond_timedwait(lf_cond_t* cond, lf_mutex_t* mutex, long long absol
...
 
 /**
- * Fetch the value of clk_id and store it in tp.
+ * Fetch the value of an internal (and platform-specific) physical clock and 
+ * store it in tp. 
+ * 
+ * Ideally, the underlying platform clock should be monotonic. However, the
+ * core lib tries to enforce monotonicity at higher level APIs (see tag.h). 
+ * Nonetheless, if the underlying clock is not monotonic, it can result in 
+ * identical timestamps being presented to the user programs that use those APIs.
```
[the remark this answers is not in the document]
Soroosh129: I think you are right. I will fix the description.

Soroosh129:
@efsanesoyer @lsk567 FYI, I have removed timespecs from `platform.h` APIs. Instead, they purely use instant_t. This to some degree reduces the overhead of converting back and forth between `timespec` and `instant_t`, although for UNIX-based systems, a one time translation is needed to use clock APIs. But, for bare-iron, I think no messing around with `timespec` is hopefully needed.

lsk567:
This is very helpful. Hopefully, this can resolve the error regarding `LF_CLOCK` that we are looking at right now. Thanks @Soroosh129 !

PR #372, opened 2021-06-15
Added STP offset to reduce the chance of test failure

Soroosh129:
This branch introduces a large STP offset to the DistributedStop.lf test to reduce the chance of test failure.

### DIFF
--- a/test/C/src/federated/DistributedStop.lf
+++ b/test/C/src/federated/DistributedStop.lf
@@ -60,7 +60,9 @@ reactor Sender {
     =}
 }
 
-reactor Receiver {
+reactor Receiver (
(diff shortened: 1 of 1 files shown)


lhstrh (commented):
Looks reasonable, but also a bit curious. Isn't shutdown coordinated by the RTI? In which case, why should this process be sensitive to physical delays?

Soroosh129:
> Looks reasonable, but also a bit curious. Isn't shutdown coordinated by the RTI? In which case, why should this process be sensitive to physical delays?

Because both the sender and the receiver call `request_stop()`.

edwardalee (approved):
looks OK to me.

PR #373, opened 2021-06-17
Carla Intersection Example

Soroosh129:
This branch includes the following:
 - A native Python-based simulation of the intersection example using Carla
 - Updates to the Python target, including a more robust error handling for reaction execution errors
 - Previously removed Python tests for multiports and banks
 - Some misc. updates to existing ROS examples to bring them up-to-date


edwardalee (approved):
Good idea to reorganize experimental directory by target language. Why are there still some programs in experimental/src?  Why not move all of them?

Good to restore the Multiport/Bank Python tests.

Intersection.lf and CarlaIntersection.lf are nicely architected.  The comments say "RTU" but it should be "RSU" and should be expanded once.  Should the Intersection.lf example be expanded to have some test inputs so it is a standalone executable that can run independent of Carla?

Soroosh129:
@edwardalee I have added a `Simulator` reactor to the Python version of [Intersection.lf](https://github.com/icyphy/lingua-franca/blob/cb80560a97140d3982a23d8c6c1cbb247c303f0f/experimental/Python/src/Intersection/Intersection.lf). It applies a simplified drag and acceleration profile based on the input vehicle commands and the initial velocity and GPS position. The starting GPS coordinates and starting velocity closely mimic the parameters that are used in the Carla version (as does the intersection position).

edwardalee:
The Intersection simulator looks great!  This should work well with a curses animation when we get that working (@lsk567, any progress on this?).  One small point: The adjustment for drag should probably occur before printing the current acceleration:
```
        current_acceleration -= self.drag_acceleration
```

lhstrh (approved):

Issue #339, opened 2021-05-18
A reaction can schedule an action without declaring it as an effect


lsk567:
In the current implementation, a reaction can schedule an action **without undeclaring it as an effect** when the same action is a trigger. Ideally, the validator should check for this. Currently, the following buggy snippet can compile in the C target.
```
logical action foo;
reaction(foo) {=
    schedule(foo, 0);
=}
```

Whereas the correct version should be
```
logical action foo;
reaction(foo) -> foo {=  // "foo" should be declared as an effect if it is scheduled in the body. 
    schedule(foo, 0);
=}
```

Soroosh129:
I think this is hard to prevent in the C runtime. One possibility is to add a new field to actions/triggers called `read_only`, which is true by default. The `schedule` functions then will check this field when they are called. The initialization code of the reaction can then set `read_only` to false if the trigger/action is an effect.

This would be a very shallow protection though. The `read_only` field can be overwritten in the body of the reaction.

lhstrh:
Shallow protection would be fine. The goal is merely to detect unintended missing dependencies.

PR #375, opened 2021-06-18
Properly implement scoping rules in the generated C++ code

### LINKED ISSUE #66: The C++ target should detect undeclared dependencies at compile time
Currently the C++ target relies on the runtime to detect undeclared dependencies. At compile time, 
access to undeclared dependencies is not checked. We should change the C++ code generation to
implement the scoping rules of Lingua Franca.

Also see tud-ccc/reactor-cpp#1

cmnrd:
This PR prevents target code from accessing timers, ports, or actions that have not been declared as triggers, sources, or effects of the reaction in LF code. Accessing undeclared triggers, sources, or effects results in compiler errors. Also, scheduling an action that is only referenced as a trigger or source (but not an effect) results in a compile error (this is related to #339). There is actually no trivial way for the C++ programmer to circumvent this mechanism. Thus, this change should significantly improve the reliability of reactors written in C++.

~~This is based on the `kotlin-cpp-generator` branch and requires #356 and #345 to be merged first. I will mark this PR as draft until the other two PRs are merged.~~

Closes #66.
Closes tud-ccc/reactor-cpp#1


cmnrd:
@lhstrh This is ready for merging now.

lhstrh (approved):
Excellent! Glad you managed to address this!

Issue #369, opened 2021-06-14
Separate code generation and building in the LF IDE


cmnrd:
Currently, saving a `*.lf` file in the IDE will trigger both code generation and then the target compiler. However, depending on the target compilation might take some time. This can be very annoying as the IDE becomes unusable after each save until the build process completed. This is particularly bad for the C++ target that currently has a long compilation time (see #355). While we should definitely make an effort to reduce compilation time, I think we should also decouple code generation from building/compilation.

I think it's fine to invoke the code generator on every save, but whether the target compiler is also invoked should at least be configurable. As an alternative to the current behavior, we could introduce a build button that allows the user to explicitly invoke the build process.

Soroosh129:
Can the compilation occur in a separate Java thread to keep the IDE responsive?

cmnrd:
I am not sure. We invoke the `doGenerate` method for each `*.lf` file in the project. This would mean that each call to `doGenerate` would need to spawn its own thread to handle the target compiler. Also it might not be safe to run to invoke the target compilation process in parallel. For instance, in the C++ target this would likely cause problems. We could, however, introduce a "compilation manager" that does this in a more coordinated way.

Independent of how the IDE could be made more responsive during the built, I think there should be a way to separate the code generation and error checking from the actual build process. Building can be very complex and users might want to disable auto-build while still profiting form error-checking and linting and being able to inspect generated code. I for myself expect from an IDE that it only performs checks and code linting on safe, and would only call a compiler or invoke other expensive operations only if I explicitly hit the build button.

PR #376, opened 2021-06-18
VS Code

lhstrh:
This is a draft pull request to track changes in the `vscode` branch.

The process for opening the extension should be straightforward. As stated by Alexander:
> Simply run: `gradlew startCode`.  You also need `mvn` on your path and if Java 11 is not your default JDK you need to prepend something like this: `JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/`

Features currently provided include:
* Highlighting
    * Includes: Syntax highlighting, semantic highlighting
    * Triggered by: Opening a non-empty file with the file extension `.lf`. Full highlighting requires a target specification at the top of the file.
* Building upon request by the user
  * Triggered by: Pressing Ctrl+Shift+P and entering the command "Build Lingua Franca" while an LF file is active in the editor
 * C++ validation
    * Triggered by: Opening a non-empty file with a C++ syntax error but no LF syntax errors and saving it, after a version of it has already been built. (It must be built beforehand because access to the CMake cache is required.)
* Diagrams
  * Triggered by: Opening a Lingua Franca source file and clicking the diagrams icon: image

The Xtext team has already implemented many useful services, including the following. (It is not yet clear if all of them work.)
* Content assist
* Folding ranges
* Find references
* Quick fixes (beta)
* Formatting
* Hover
* Highlight all references to the current symbol
* Rename symbols
* Get workspace symbols

Here is a list of current action items in this branch.

**Required before the first release of the extension:**
- [x] Correctly resolve dependencies that can only be retrieved using Gradle
- [x] Build in a way that allows faster development than fully generating the LDS jar
- [x] Syntax highlighting
- [x] Semantic highlighting
- [x] VS Code: Perform a full build iff the user requests it

For desired features that will not be implemented as part of this PR, see: https://github.com/lf-lang/lingua-franca/issues/614


petervdonovan:
Here is a 10,000-foot view of what I _think_ is the current design problem affecting target language validation (which the user recognizes as the red/yellow squiggles that appear under syntax errors).

As we know, target language validation involves code generation. Without that, there is no generated code to validate.

However, code generation currently seems to be intermingled with 3 distinct tasks:
1. Code generation (running `TSGenerator`, `CGenerator`, `CppGenerator`, etc.)
2. Code validation/compilation (running `LFValidatorImpl`, for example, or `tsc` for type checking, or `cmake`)
3. Library copying/installation (creating a `node_modules` directory, `core` directory, etc.)

For the purpose of target language validation, code generation really should be nothing more than code generation (Task 1).

Task 2 is not desired. This is because
* Validation that is performed in the process of code-generation is redundant if said code-generation is part of an ongoing validation process.
* We should not have code generation that involves validation that involves code generation, either.

Task 3 is not exactly desired. Although it must happen once, it should not happen every time the user presses `Ctrl+S` while editing a Lingua Franca file. This is because
* At best, Task 3 involves a lot of file I/O, which is risky (i.e., able to cause an `IOException`) and slow.
* It might involve downloading/installing packages, which is even slower.
* It is redundant. The exact same standard libraries are always created because the user never edits `package.json` (except maybe in some edge cases, like when the libraries change while the user is working).

This motivates an impulse to try to make these tasks as separate as possible, so that it is easy to only do Task 1 regardless of the target language. My early impression based on the TypeScript and C++ generators is that Task 2 is not totally separate from Task 1, and in general that Task 3 is not currently separate from Task 1 either.

edwardalee:
@petervdonovan 's points are good ones, but right now, I think that at least the C code generator skips some checks on the assumption that the validator has been run.

cmnrd:
Just to add the C++ perspective: The LF Validator is expected to be run before the code generation.
There are barely any checks in the C++ code generator, which makes the generator fairly simple.
It assumes that the given LF Program is valid and simply translates it into C++ code. 
I actually think that a valid LF program is a prerequisite for both code generation and for performing target language validation. 

The C++ code generator already separates the 3 concerns of validation, code generation, and compilation. However, in the current design of `LFGenerator` and `GeneratorBase`, `doGenerate()` currently covers all 3 concerns. I agree to @petervdonovan that this is a design flaw, and we should probably make an effort to extend the interface of `LFGenerator` so that we have `doValidate()`, `doGenerate()` and `doCompile()`. Of course, the requirement would be that validation is performed before code generation and code generation before compilation. But, if for instance only a reaction body changed, we don't need to run the LF validator again and can directly generate code. Separating the 3 steps could also help to improve the experience in Epoch. I find it pretty annoying that a file save starts a (potentially lengthy) compilation process. I think instead we should only validate on save and have a separate compilation button (see #369).

When it comes to target language validation, we might not even want to perform a complete code generation. For instance, validating the body of a C++ reaction, would only require all visible preambles, the inner reactor class (containing all state and parameters), and the function implementing the reaction. This is only a fraction of the code that would normally be generated.

a-sr:
I was just working on the Kotlin build problems in #437 and given these new insights, I am quite wary that the hack we discussed last week to provide the libraries for the kotlinx.* packages will actually work.
Even if these TS mapping classes are only supposed to work in the LS, the files are still located in the org.lflang package and thus will be compiled into the RCA. This means these libraries that are only available in Gradle have to be hacked into the Maven build, one that is not started from within Gradle, hence no dependencies downloaded, and that only uses Eclipse p2 dependencies. I am not sure if this will be possible in any way and even if I am not sure if it will improve our current state of build infrastructure.

petervdonovan:
Okay, noted. Thank you for the update, Alexander. I have already implemented the hack, and it seems to work for the language server, but it sounds like at some point we have to either
* Use a different serialization library (which I assume is the most likely option?), or
* Move the code that uses those dependencies out of `org.lflang` in the case that it is not used in the RCA

(I wonder if there is a way to start a Gradle task or Python script from within the Maven build, but in any case it is probably a moot point, since I agree that doing so might be bad for our build infrastructure.)

I admit to still being somewhat in a prototyping mindset, so this will be another of a few items that need to be fixed.

org.lflang/src/org/lflang/generator/ts/sourcemap/SourceMapSegment.kt
```diff
@@ -0,0 +1,205 @@
+package org.lflang.generator.ts.sourcemap
+
+import kotlinx.serialization.*
+
+/**
+ * Represents a source map segment that follows the specification provided in
```
cmnrd, line 6: What is "Source Map Revision 3 Proposal"? Is this a feature for mapping source code locations in target code back to locations in LF code? If so, is this applicable also to other languages?

cmnrd (commented):
This looks quite impressive! What is the current status of this PR. It looks like there are still tasks to do, but I am a bit worried about the size of the PR. It seems that this PR also touches a few critical parts of our codebase, that also overlap with other open PRs. Should we maybe aim at merging this soon and continue working in subsequent smaller PRs?

Also, how can I try the VS code plugin? :)

org.lflang/src/org/lflang/generator/c/CCmakeCompiler.java
```diff
@@ -45,6 +44,10 @@
  * @author Soroush Bateni <soroush@utdallas.edu>
  */
 class CCmakeCompiler extends CCompiler {
+    // FIXME: This does not override the compileCCommand
+    //  method because there is no single command that
+    //  compiles; however, it is not clear why
+    //  compileCCommand is public in the first place.
```
Soroosh129, line 50: Because it is also used in `org.lflang.federated.FedCLauncher`.

org.lflang/src/org/lflang/generator/c/CCmakeCompiler.java
```diff
@@ -71,68 +74,23 @@ public boolean runCCompiler(String file, boolean noBinary) throws IOException {
...
-     * This produces a C-specific compile command.
-     * 
-     * @param fileToCompile The C filename without the .c extension.
-     * @param noBinary If true, the compiler will create a .o output instead of a binary. 
-     *  If false, the compile command will produce a binary.
+     * Return a command to configure the specified C file
+     * using CMake.
+     * This produces a C-specific configure command.
      */
-    public LFCommand compileCmakeCommand(
```
Soroosh129, line 131: This will also be useful in `org.lflang.federated.FedCLauncher` once we iron out a few issues regarding CMake paths.

Soroosh129:
I also agree that this is impressive progress! 
This PR as @cmnrd pointed out is already really big. 
Is the VS Code plugin in a working state that can be described as a "beta"? :) 
If so, I am also advocating for it to be reviewed as is and merged in. I would suggest that you could create an issue linked to one of the projects containing the remaining tasks.

org.lflang/src/org/lflang/generator/ts/sourcemap/SourceMapSegment.kt
petervdonovan: 1. It is [the standard](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.djovrt4kdvga) for how to map between TS, JS, and uglified/compressed JS. I should clarify that comment.
2. Yes, that was the intention when I created these classes -- mapping from TS back to LF.
3. It might be applicable? Perhaps that should be discussed. We could use one way to establish mappings for C and C++ (line directives), another way for TS (source maps), and another way for Python and Rust. However, I have heard that we wish to report where errors come from in generated code in addition to knowing where they originate in the source, so maybe the special functionality from line directives (line numbers being adjusted by the compiler when it reports errors) isn't really useful to us. It also isn't clear yet how source maps will hook into existing debugging tools for TS, so it is possible that their special functionality isn't useful to us either. It therefore seems possible that we should choose a single way for establishing these mappings that can be reused across target languages, whether that is source maps, something like line directives, or something of our own design.

org.lflang/src/org/lflang/generator/c/CCmakeCompiler.java
petervdonovan: Oops, noted. I will look at that. In hindsight it was probably wrong to sneak these changes into this branch. I'll be more careful next time.

petervdonovan:
Given that the worst parts of this branch have recently been cleaned up or deleted, maybe it can be merged in as you suggest. However, @lhstrh has provided a lot of guidance for the development of this branch, including the issues that currently need to be solved. Maybe he has an opinion?

org.lflang/src/org/lflang/generator/ts/sourcemap/SourceMapSegment.kt
cmnrd: For C++ I have refrained from using the line directives. For one part this is due to aesthetic aspects, but for the other part it is also because I find that errors reported from the C++ compiler can easily become misleading. Especially, in larger projects where we might have several additional libraries written in target code. To debug certain compiler errors, I think its important to be able to investigate the generated source code (even as an end-user).

lhstrh: Yes, ideally we'd be reporting the error in the right location of the LF file while reporting also to which file and line number this corresponds in the generated code.

PR #379, opened 2021-06-24
FIXMEs and minor changes discussed during code review

lhstrh:
With notes from @oowekyala:

# Code review 6.24
_NOTE: line numbers are only more or less accurate bc marten added some comments_
### High-level comment
- Federates may have a different instant_t. Since the RTI needs to find max of timestamps received from each federates, it needs to assume that the message fits in some number of bytes
- So the federation may break if federates have different concrete data types for eg instant_t (probs also for int)
- instances of `sizeof(int)` and `sizeof(long long)` should be replaced with data types that have fixed width
### `handle_timestamp(fed_t)`
L1115: 2nd write to socket
 - receiver of the data may see just the message header, and block waiting for the body, if an error occurs between both `write_to_socket` calls. Should be consolidated into single write.
L1123:
 - we're holding the lock until the end of the function, including a write to a socket
### `handle_address_ad(ushort)`
- couple of renamings "ad" -> "advertisement", including constant `MSG_TYPE_ADDRESS_AD` -> `MSG_TYPE_ADDRESS_ADVERTISEMENT`
L1062:
  - move assert that checks that the server port is a valid number (< 2^16) into here, to check it before storing it.

### DIFF
--- a/org.lflang/src/lib/C/ctarget.h
+++ b/org.lflang/src/lib/C/ctarget.h
@@ -144,6 +144,7 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * Schedule an action to occur with the specified value and time offset
  * with no payload (no value conveyed).
  * See schedule_token(), which this uses, for details.
+ * 
  * @param action Pointer to an action on the self struct.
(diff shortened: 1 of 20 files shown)


lhstrh (commented):
Looks good to me!

PR #382, opened 2021-06-29
Support definition of methods similar to state variables

cmnrd:
This draft MR adds rudimentary support for the definition of methods in LF code. Methods are defined similarly to state variables. They can be invoked from reactions and allow encapsulating certain functionality and also enable code reuse between reactions. Currently, we only have preambles for defining functions, but this is not always sufficient as methods might need to access parameters and state variables or call other methods. In particular, we noticed that we need methods for porting complex ROS nodes to LF in a practical way. This MR also adds a test case, which also serves as a usage example.

This MR depends on #356, #345, and #375 and should only be merged once the other MRs are merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,7 +104,7 @@ crashlytics-build.properties
 fabric.properties
 
 ### Maven ###
-target/
+/target/
 pom.xml.tag
 pom.xml.releaseBackup
 pom.xml.versionsBackup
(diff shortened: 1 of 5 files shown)


HugoForrat (approved):

edwardalee (approved):
Looks good to me!  This needs documentation on the wiki.

cmnrd:
I documented it here: https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-Cpp#using-methods

edwardalee:
Shouldn't this be in the language specification? (Even if currently only supported by Cpp).

cmnrd:
Oh, right. I completely forgot about the language specification. I also added a description of methods there.

PR #384, opened 2021-06-30
Adding C benchmarks

mattchorlian:
Counting Benchmarks Results:

count_logs

Fork/Join Throughput Benchmark results:

fjthrput_results


cmnrd:
Just out of curiosity: Did you compare those numbers to the Akka implementation in Savina and/or to the C++ implementation in LF?

mattchorlian:
I will set up cmake and C++17 locally in order to compile C++ programs and I will get back to your with some results @cmnrd

mattchorlian:
C++ Counting Results:  
-Best: 439 ms
-Worst: 681 ms
-Median: 448 ms

C++ ForkJoin(Throughput) Results:
- Best: 829 ms
- Worst: 919 ms
- Median: 899 ms

cmnrd:
I noticed that you use the `BenchmarkRunner` reactor to control the benchmark execution. A while ago there was a discussion about this benchmark runner reactor and I think the dominating opinion was that the runner shouldn't be used for the C benchmarks. Instead, the same benchmark should just be run multiple times by invoking the binary multiple times. The python script for running the benchmarks already does that automatic repetition of the benchmark (see the ping pong benchmark). This, however, is in conflict with the benchmark runner reactor, because then all benchmarks will be run N^2 times.

I am fine with both options (using the `BenchmarkRunner` reactor or letting the runner script handle the repetition), but we should decide for one and use it consistently across all C benchmarks. This consistency is also important for parsing the results by the runner script.

Another side note: The benchmarks should be placed within a `src` directory, to allow the LF compiler to detect the project root (see the C++ implementation). This ensures that there are no name conflicts with benchmarks implemented in other languages and also makes the location in which binaries are produced well-defined.

mattchorlian:
I am going to hold off on committing any .yaml configuration files in benchmark/runner/conf/benchmark until we decide on a final directory structure for the benchmarks

benchmark/C/Savina/src/micro/count/CountingBenchmarkN.lf
```diff
@@ -0,0 +1,151 @@
+/**
```
cmnrd, line 1: What should we do with the two different Versions of this benchmark? I mean `CountingBenchmark.lf` and `CountingBenchmarkN.lf`. I think it would suffice to only have the cog based version. Do you agree?

Also I think we should simplify the names. Instead of `CountingBenchmark.lf` we can just call it `Counting.lf`. This is already done in the C++ benchmarks.

cmnrd (commented):
Looks good! Could you update and add the configuration (*.yaml) files? Then we could merge this PR.

benchmark/C/Savina/src/micro/count/CountingBenchmarkN.lf
mattchorlian: I agree, it only make sense to keep the cog version around.  I just pushed a commit with the .yaml configurations updated and the renamed Counting.lf file

benchmark/runner/conf/benchmark/savina_micro_count.yaml
```diff
@@ -23,3 +23,10 @@ targets:
     gen_args: null
     run_args:
       messages: ["--countTo", "<value>"]
+  lf-c:
+    copy_sources:
+      - "${lf_path}/benchmark/C/Savina/src/micro/count/Counting.lf"
```
cmnrd, line 28: The path here is wrong due to the 'count' directory.

benchmark/runner/conf/benchmark/savina_micro_fjthrput.yaml
```diff
@@ -27,3 +27,10 @@ targets:
       actors: ["-D", "numWorkers=<value>"]
     run_args:
       messages: ["--numMessagesPerReactor", "<value>"]
+  lf-c:
+    copy_sources:
+      - "${lf_path}/benchmark/C/Savina/src/micro/fjthrput/Throughput.lf"
```
cmnrd, line 32: Also this path is wrong

benchmark/runner/conf/benchmark/savina_micro_fjthrput.yaml
```diff
@@ -27,3 +27,10 @@ targets:
...
+    lf_file: "Throughput.lf"
+    binary: "Throughput"
+    gen_args:
+      messages: ["-D", "numWorkers=<value>"]
```
cmnrd, line 36: This maps the `messages` parameter to the `numWorkers` cog variable, which is wrong. Actually, there should be two variables, one for the number of messages and one for the number of workers.

benchmark/C/Savina/src/micro/Counting.lf
```diff
@@ -0,0 +1,151 @@
...
+
+    reaction(inRetrieveResult) -> outSendResult {=
+        SET(outSendResult,self->count);
+    =}
+}
+
+main reactor CountingBenchmarkN(numIterations:int(12), countTo:int(1000000)) {
+    
+    /* [[[cog
+    cog.outl(f"producer = new ProducerReactor(countTo = {countTo});")
```
cmnrd, line 135: This is only a nitpick, but I think it is a bit cleaner to use cog to only change the default value of the `countTo` parameter above, and then pass this parameter along to the other reactors.

benchmark/C/Savina/src/micro/Throughput.lf
```diff
@@ -0,0 +1,183 @@
...
+# To change the generated code in-line within this file run:
+# $ python -m cog -r this-file.lf
+# To generate a new file from this file stripping the generator code in the process run:
+# $ python -m cog -d -o output-file.lf this-file.lf
+#
+# Use the command line option -D to specify generator parameters, for example:
+# $ python -m cog -r -D parameter=100 this-file.lf
+#
+# Generator parameters used in this file:
+# -D numWorkers=60
```
cmnrd, line 33: This benchmark should have two parameters: number of workers  and number of messages. Note that in the C++ version only the number of workers is a cog parameter, because the number of messages is implemented as a parameter to the main reactor. In C++, such parameters can be set when invoking the binary via command line arguments. This, however, is not possible in C and all parameters must be implemented as cog parameters.

benchmark/C/Savina/src/micro/Throughput.lf
```diff
@@ -0,0 +1,183 @@
...
+    
+    input inStart:bool;
+    output outFinished:bool;
+  
+     /* [[[cog
+      cog.outl(f'output[{numWorkers}] outMessage:bool;')
+      cog.outl(f'input[{numWorkers}] inFinished:bool;')
+      cog.outl(f'output[{numWorkers}] outReset:bool;')
+    ]]] */
+    output[60] outMessage:bool;
```
cmnrd, line 109: This can be simplified a bit by using the reactor parameters. For instance, you can write `output[numConsumer] outMessage:bool`. Note that this is a relatively new feature, that did not exist when the C++ benchmarks where written. We are currently updating all the C++ benchmarks in #406.

benchmark/C/Savina/src/micro/Throughput.lf
```diff
@@ -0,0 +1,183 @@
...
+            SET(outFinished, true);
+        }
+    =}
+}
+
+
+main reactor (numIterations:int(12), numMessagesPerReactor:int(10000)) {
+    
+    /* [[[cog
+      cog.outl(f'producer = new ThroughputProducer(totalMessages=numMessagesPerReactor, numConsumer={numWorkers});')
```
cmnrd, line 164: Also here, you can make numConsumer a parameter to the main reactor. Then you can pass along the parameter and only need to generate a single line from cog.

cmnrd (commented):
Looks good! But unfortunately the configurations are not yet quite right and thus the benchmark script is not able to run them. You can find a description of how to run benchmarks via the benchmark runner script here: https://github.com/icyphy/lingua-franca/wiki/Running-Benchmarks

benchmark/C/Savina/src/micro/Counting.lf
mattchorlian: Does this mean cog it at the top of the file and use it like a global variable?

benchmark/C/Savina/src/micro/Throughput.lf
mattchorlian: Do you mean numWorkers?  Or does it not matter?

benchmark/C/Savina/src/micro/Counting.lf
cmnrd: I am not sure what you mean by "at the top of the file" here. What I meant is that you can cog the main reactor declaration, such that you overwrite the default values of any parameters to the main reactor. Then you have one central place where the code is changed by cog and not multiple.

Also, this would help to bring the C and C++ code more in line. In C++, cog is not needed because parameters to the main reactor can be overwritten on the command line without recompiling.

benchmark/C/Savina/src/micro/Counting.lf
```diff
@@ -0,0 +1,151 @@
...
+    reaction(inIncrement) {=
+        self->count++;
+    =}
+
+    reaction(inRetrieveResult) -> outSendResult {=
+        SET(outSendResult,self->count);
+    =}
+}
+
+main reactor CountingBenchmarkN(numIterations:int(12), countTo:int(1000000)) {
```
cmnrd, line 132: I just noticed that the main reactor is named. This should actually trigger an error in our compiler. The name should either be omitted, or be the same as the name of the file.

mattchorlian: Yes, just edited that and it should be reflected in the commit I'm about to push

cmnrd (approved):
This one is good to go in! Thanks @mattchorlian for all your adjustments. I only noticed that the older benchmarks currently don't work with the benchmark runner. Do you want to address this in this PR, or rework the existing benchmarks in another one? I am not sure which state they are in, and they might also require some code changes

I pushed one commit that fixed a bug in the runner python script. Now both benchmarks run without errors.

mattchorlian:
I think it makes sense to merge this PR so master at least has the updated C benchmark organization and a couple more benchmarks.  Then I can try to rework benchmarks and tackle the benchmark runner issue separately in a different PR.  Does that seem reasonable? @cmnrd

cmnrd:
Yes, absolutely!

lhstrh (approved):

Issue #386, opened 2021-07-01
Diagram synthesis fails with NullPointerException for Bank example


cmnrd:
The following minimal program results in a NullPointerException in the diagram synthesis:
```
target Cpp;
reactor Contained {
    output out:unsigned;
}
main reactor {
    c = new[4] Contained();
    reaction(startup) c.out {==}
}
```
It works fine when removing `[4]` to make `c` a normal reactor instance.

edwardalee:
This is fixed in the `banks-to-reaction` branch.  I have issued a pull request #391.

cmnrd:
closed by #391

Issue #377, opened 2021-06-23
Bank to reaction capability is missing


edwardalee:
Our syntax allows a reaction to be triggered by a contained bank, but this does not currently work, and it it's not entirely clear what it should mean. My proposal is that the reaction is triggered by any output from any of the instances of the bank. This means that the target language needs syntax for getting the width of the bank and iterating over it, something like the following (this does not currently compile):

<img width="566" alt="image" src="https://user-images.githubusercontent.com/8513334/123098987-96bf7280-d3e6-11eb-9e46-d7fe7ec6f120.png">

cmnrd:
I am currently revising the code C++ code generator to fix #66 and this has also become an issue there. I am taking a similar approach as you suggested (actually I assumed this already works in C). Iterating in C++ is a bit simpler, as the symbol `s` would be of type `std::array<...>` and you can use a for each loop.

This gets slightly more complicated if the contained port is also a multiport.

cmnrd:
#387 implements this for C++. I added two tests [ReadMultiportOutputOfContainedBank.lf](https://github.com/icyphy/lingua-franca/blob/5bbf33cc85dda48769d41f3ff018dd3d725f2e83/test/Cpp/src/multiport/ReadMultiportOutputOfContainedBank.lf) and [ReadOutputOfContainedBank.lf
](https://github.com/icyphy/lingua-franca/blob/5bbf33cc85dda48769d41f3ff018dd3d725f2e83/test/Cpp/src/multiport/ReadOutputOfContainedBank.lf).

Issue #343, opened 2021-05-20
In banks of reactors, bank_index is not a proper parameter


edwardalee:
Currently, `bank_index` is available for reactors to read inside reactions (see [Banks of Reactors](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors#banks-of-reactors)). However, it is not a proper parameter of the reactor in the bank. As a consequence, its value cannot be assigned to other reactors and cannot be passed down to contained reactors.  I would like to do something like:
```
    r = new[4] Foo(bar = bank_index);
```
which would assign the bank index to the parameter `bar` of reactor `Foo`.

cmnrd:
It appears that the C target diverged a bit from the original C++ implementation. In C++, `bank_index` is treated as a special parameter. For instance:
```
reactor Foo(bank_index:size_t(0)) {
}
main reactor {
  foo = new[4] Foo();
}
```
In that program, `Foo` will be instantiated 4 times and receive the corresponding `bank_index`. The instantiation automatically overwrites the bank_index parameter if the instantiation is a bank instantiation and the instantiated reactor as a parameter named `bank_index`. Otherwise, the bank_index will not be set. I guess the same could be done in C.

edwardalee:
Yes, the C target should be changed so that bank_index is a parameter. The current design does not support passing the bank_index to contained reactors, which becomes possible if it is a parameter.

cmnrd:
I think this was implemented for the C target (judging from the C tests), but I am not sure. Can we close this?

edwardalee:
Yes, this is fixed in the C target and also in TS. Both can pass bank_index parameters.

Issue #209, opened 2020-09-03
Banks, multi ports and special connection patterns


cmnrd:
This issue is closely related to #190. Since this issue grew to be quite extensive, I figured it would be best to break certain aspects down into a separate issue.

So while discussing the issue in #190 we often said that there is no use-case for referencing individual ports and that we should always connect entire banks and/or multi ports. However, as it turns out there is at least one use-case... meet the  All-Pairs Shortest Path (APSP) benchmark from the Savina suite. The benchmark essentially consists of a grid of workers, where each worker solves a part of the problem and sends updates to its neighbors. A grid of 4x4 workers looks like this: 
![rect834](https://user-images.githubusercontent.com/6460123/92093725-f440ef00-edd3-11ea-9201-3773f3ded328.png)
where each of the rectangle represents a worker actor/reactor and each worker is connected to its neighbors.

So how can such a grid be build in Lingua Franca elegantly? Of course, all the workers can be placed in a bank, but then there is no way to draw individual connections. Even if there would be such a mechanism, listing all the connections explicitly is not a real solution.

In order to solve this issues properly, I think we would need an expressive syntax in LF. But I don't think that's the way to go, as we will only get more problems (similar to the parameterized widths). Maybe it is a good alternative to use the expressiveness of our target languages instead. We could expose the connection mechanism in a callable function and allow drawing of connection in a startup mutation. Saroush also proposed, that the target code could generate an adjacency matrix, that then is respected while creating reactions.

edwardalee:
Yes, this could be another higher-order pattern similar to the ScatterGather example.  It could be realized today in a manner similar to ScatterGather, but a better approach would be to first abstract what ScatterGather does into an API.  Today, ScatterGather reaches directly into the guts of the C implementation.

lhstrh:
I've been building the API for mutations in the TypeScript runtime. I'll take a stab at implementing All-Pairs Shortest Path using a startup mutation.

cmnrd:
Yes, a higher level function could help to solve this. It just should be general enough so that various connection patterns and not just a plain grid can be created. I am not sure how this could be achieved. I think you are right, and we should take small steps and find a general solution for ScatterGather first.

cmnrd:
Closing this. #406 proves that our current connection syntax is sufficient for covering all benchmarks from the Savina suite that do not support mutations.

PR #387, opened 2021-07-01
Support parameterized multiports in C++

### LINKED ISSUE #349: Support paramterized multiports and banks through generics
The C target currently supports multiports and reactor banks with a variable sized width (provided by a parameter). This is possible, because the C code generator treats parameter as compile time constants. Thus, it can resolve the precise value of any width specification during compile-time. However, this is not possible in the C++ target as parameters are *runtime* constants. They are initialized when the application starts and may be overwritten by CLI parameters that the generated binary parses.

It occurred to me that generics present a potential solution for the problem as they are always resolved at compile time. We currently have syntax for this:
```
reactor Foo<T> {
  output out:T;
}
```
We could enhance it like this:
```
reactor Foo<T: type, w: width(4)> {
  output[w] out:T;
}
```
Where the types with `<>` would be restricted to either `type` or `width`. This would also provide a viable solution for the C target should it allow overwriting parameters via the CLI in the future.

cmnrd:
This PR adds support for parameterized multiport and bank widths for the C++ target. Since parameters are assigned at runtime in C++, this is achieved by resolving the precise widths at runtime and letting the generated code draw the connections as needed. For this to work, actually only a small library function was added to `lib/lfutil.hh` and no bigger changes in `reactor-cpp` where required.

~~This PR is based on #375, and should only be merged after merging #375.~~

Related to #386, #377, #343, #193, #209

Closes #349, #190


cmnrd:
After delays on multiport connections with (potentially) parameterized width turned out to be really tricky. The problem is, that the AST transformation for after delays does not infer the width of the generated bank of delays. It uses a "variable" width, which means that the precise width needs to be inferred by the generator. However, this is not trivial and I don't know how to infer the width at runtime without providing additional information.

I talked to @edwardalee yesterday about this problem. He suggested to not use AST transformations, but support after delays directly in the C++ runtime. This is something I actually wanted to try for a long time, so I did that today. However, this turned out to be far from trivial. The main problem is that we can have various kinds of connections in a chain (input to contained input, output to input, contained output to output) and all those connections can have associated delays. I didn't find a straight forward way to deal with all the different corner-cases and I think that supporting delayed connections might require a significant redesign of the C++ runtime. So I abandoned that plan for now...

I actually found another solution to the problem which is probably a bit hacky. Essentially, I introduced the `widthof` keyword, which can be used to get the width of a port like so `input[widthof(some.port)] in:int`. I think such an operator could be quite useful, but for now the validator actually forbids its usage in LF code. However, the AST transformation uses this mechanism to define the width of the delay bank, best on all right ports involved in the connection. This gives the C++ code generator all the information required for determining the width of the delay bank at runtime. It can simply some the actual widths of all ports (multiplied by the bank width if the port is a contained port in a bank). This mechanism is currently only activated for C++, all other targets still use the variable length width.

@edwardalee @lhstrh What do you think of this approach? Do you think its a viable solution to the problem?

edwardalee:
I'm not sure this will solve the problem in general.  What if you have the following widths:
```
   a.out: 2
   b.out: 2
   c.in: 3
   d.in: 1
```
And you connect them as follows:
```
    a.out, b.out -> c.in, d.in after 10 msec;
```
Then I think there will be three inserted delay reactors with widths 2, 1, and 1.
Also, what ports would be scope for the `widthof` operator?

cmnrd:
For your example, the LF code generated by the AST transformation would look like this:
```delay = new[widthof(c.in) + widthof(d.in)]```

I am not sure why you think that three delay reactors would be inserted. As far as I can see, the AST transformation always generates one bank of delay reactors per after connection, where the width of the delay bank is the summed width of all ports (and banks) on the right side of the connection. If the referenced port is a multiport within a bank, `widthof(bank.port)` would return the bank width multiplied by the port width.

The scope of the `widthof` operator is the same as for all other references to ports in LF (like triggers or effects). Basically, it can reference all ports within the current reactor or within any contained reactors (one level deep).

cmnrd:
I would like to merge this soon if possible. @edwardalee @lhstrh Do you think the `widthof` solution is viable for now?

edwardalee:
The `widthof` solution still feels a bit kludgey to me, but I'm not opposed to including it.  I'm not sure how this will play out for federated execution, but I guess we can cross that bridge when we come to it.

cmnrd:
Ok, I am also not too sure about `widthof`, but it seems to solve the issue for now. We should keep this in mind however and might want to find better solutions in the future. @edwardalee Could you approve this PR? I need the approval to merge due to new merging rules.

edwardalee:
Github does not offer me a button to approve this pull request, so I don't see how I could be blocking it.

edwardalee (approved):
OK, I found the hidden button...

PR #390, opened 2021-07-01
Comments from code review: Jul-1-2021

Soroosh129:
> From @petervdonovan :
> 
> Code review 7.01
> ==========
> core/rti.c
> Issues:
> 1. [L1359] Socket implementations vary, and here we assume that all pending messages will be sent before closing the socket. (There was additional discussion of the proper way for `handle_federate_resign` to close the socket.)
> 2. [L1383] The documentation of the shutdown API is not completely clear. However, perhaps the only reason why the documentation is lacking is that the authors assume that it is understood to be the same as the TCP IP.
> 3. [L1386] We need to make sure the `close` function is called once. This brings up a more general issue that the documentation for `handle_federate_resign` should perhaps specify all of its assumptions more clearly.
> 4. [L787] It might be better if we could use a finer-grained mutex. However, there is no obvious way to do this without risk of deadlock. It is possible that `select` would be more performant, but this is unclear depending on the operating system.
> 5. [L840] It might be unnecessary to send a TAG in response to a TAN. It is agreed that this actually is necessary, but we are at least two papers away from fully explaining/justifying this (?).
> 6. [L735] FIXME. This is another deviation from HLA that should be fully discussed in a paper.
> 7. [L736] There is no NET or TAN at shutdown (?) and startup is similarly an exceptional edge case.
> 8. [L747] This may be a bug -- it might be possible to write a test case that puts this in deadlock. Perhaps it should be "transitive."
> Smaller issues:
> 1. [L1357] There would be more consistency if it were possible to use annotation, "@mutex," but there may not be a way to do that.
> 2. [L1394] The word TAG can be confusing: it means Time Advance Grant, but we also use the word "tag" frequently to discuss superdense time.
> 3. [L799] "Appropriate" in the function name `send_tag_advance_if_appropriate` is a little vague. "Safe" may be preferred. Also, "tag_advance" may be redundant: It means "time advance grant advance."
> 4. [L805] The difference between the respective behaviors of `transitive_send_TAG_if_appropriate` and `send_tag_advance_if_appropriate` (L799) is not obvious from their names.
> 5. [L804] This empty initialization is said to be deprecated, but we need it to avoid creating a new function (since `transitive_send_TAG_if_appropriate` is recursive). We could initialize to 0 to avoid the warning.
> 6. [L827] Log and debug are functions, not macros, so they may add unnecessary overhead when not in debug mode. The compiler _should_ be able to always optimize it out according to experiments, but there may have been some question as to whether this optimization will always happen.
> Side notes:
> 1. This implementation of the RTI should perhaps be considered a reference implementation. In the future it might be re-implemented in Java or Kotlin. Or we could bootstrap, and implement it using Lingua Franca.
> 2. Not all benchmarks should imitate the Akka implementation in the Savina suite because LF allows a very different implementation that can be more performant.
> 3. We should perhaps have a documentation system such as Doxygen that might allow us to create links, since we are finding that we wish more detailed information such as that contained in rti.h were accessible elsewhere.
> Current progress: 
> * We finished looking at `handle_logical_tag_complete`, but we did not examine the algorithm in `send_tag_advance_if_appropriate`.
>


cmnrd:
Should we close this? Or do these comments still need to be addressed?

lhstrh:
AFAIK these still need to be addressed...

cmnrd (commented):
Why is this part of the code review PR? Shouldn't changes on the benchmarks be in a separate PR? It's quite difficult to keep track of what is going on with these multipurpose PRs.

Soroosh129:
> Why is this part of the code review PR? Shouldn't changes on the benchmarks be in a separate PR? It's quite difficult to keep track of what is going on with these multipurpose PRs.

The weird structure of this particular benchmark was raised during our code review meeting (although it is called benchmark meeting now). The issues in our federated runtime was also raised there, but I will create a new PR for it...

Soroosh129:
The changes in this branch should be short enough for a review, although I agree with @cmnrd 's comment that we should keep the PRs focused on a single topic.

org.lflang/src/lib/core/federated/RTI/rti.c
```diff
@@ -590,7 +594,7 @@ void send_provisional_tag_advance_grant(federate_t* fed, tag_t tag) {
  *
  * @return True if the TAG message is sent and false otherwise.
  */
-bool send_tag_advance_if_appropriate(federate_t* fed) {
+bool send_TAG_if_transitively_safe(federate_t* fed) {
```
lhstrh, line 597: Not sure that the word "transitive" is necessary here.

org.lflang/src/lib/core/federated/RTI/rti.c
```diff
@@ -689,7 +693,28 @@ bool send_tag_advance_if_appropriate(federate_t* fed) {
 }
 
 /**
- * Handle a logical tag complete (LTC) message.
+ * For all federates transitively downstream of the specified federate, determine
```
lhstrh, line 696: Also this sentence works without the word "transitive".  I would just describe this as a recursive function that visits downstream federates and sends them a TAG if it's safe for them to advance.

org.lflang/src/lib/core/federated/RTI/rti.c
```diff
@@ -689,7 +693,28 @@ bool send_tag_advance_if_appropriate(federate_t* fed) {
...
+ * @param visited An array of booleans used to determine whether a federate has
+ *  been visited (initially all false).
+ */
+void send_TAG_if_transitively_safe_to_all_downstream(federate_t* fed, bool visited[]) {
+	visited[fed->id] = true;
+	for (int i = 0; i < fed->num_downstream; i++) {
+		federate_t* downstream = &_RTI.federates[fed->downstream[i]];
+		if (visited[downstream->id]) continue;
+		send_TAG_if_transitively_safe(downstream);
+		send_TAG_if_transitively_safe_to_all_downstream(downstream, visited);
```
lhstrh, line 711: I think the names of these functions can be improved. I would call one `grant_advancement` (because it could also send `PTAG`, not just `TAG`) and the other `traverse_and_grant_advancement`. I don't think it is necessary to explain in the name the fact notices will only be sent granted that certain conditions are true. Comments are adequate for that.

lhstrh (commented):
Minor suggestions for improvement.

org.lflang/src/lib/core/federated/RTI/rti.c
Soroosh129: `transitive` refers to [this](https://github.com/lf-lang/lingua-franca/blob/bde2bf2e74871f9255dd34e256aaa771c61e5eb6/org.lflang/src/lib/core/federated/RTI/rti.c#L479) functionality of these functions.

org.lflang/src/lib/core/federated/RTI/rti.c
Soroosh129: For the comment, I agree that recursive is more accurate.

org.lflang/src/lib/core/federated/RTI/rti.c
Soroosh129: `grant_advancement` can be misleading since these functions sometimes don't issue a `TAG` or `PTAG`. Also advancement might be a bit vague. Does it mean tag advancement in all contexts?

Soroosh129: How about `grant_TAG_if_safe` and `grant_TAG_downstream_if_safe`?

org.lflang/src/lib/core/federated/RTI/rti.c
lhstrh: Transitivity is the property used to infer whether or not it is safe to advance. It's an implementation detail which doesn't matter to the user of the function. The user cares about signalling federates to advance once safe. Is there another function that's part of the API that does the same thing but ignores transitivity?

org.lflang/src/lib/core/federated/RTI/rti.c
lhstrh: It chose the term "advancement" because it generalizes the different kinds of grants that are issued (PTAG/TAG); stating that it only grants a TAG is inaccurate. The `if_safe` part strikes me as unnecessary. The function `malloc` also doesn't always allocate memory, yet we don't call it `malloc_if_memory_available`. That's just too verbose. Finally, the `downstream` aspect seems to convey another implementation detail that is not relevant to the caller of the API who just wants to traverse the graph of federates and grant advancement.

Soroosh129: In the case of `malloc`, it doesn't allocate memory if an error occurs. In this function, not sending a TAG/PTAG is not an error. This conditional sending of TAG/PTAG is part of the main logic of this function. I think it should be clear that calling this function will not necessarily yield a TAG/PTAG.

In terms of advancement, as I said, I think it is vague. What does it mean to grant an advancement? These functions don't even necessarily grant a tag advancement even if they send a TAG. If a PTAG has been sent to a federate, and then later this function sends a TAG, it will just unblock control reactions.

Soroosh129: Maybe `grant_advance_grant_if_safe`? I think that downstream is pretty important. Traverse doesn't indicate the direction of the traversal. Calling the second function will send TAG/PTAG downstream only if it is safe.

lhstrh: I'm fine with `send_advance _grant` and `send_advance_grants_downstream`, but I haven't changed my opinion on the `if_safe` part. IMO, to pursue safety is the default in programming; only doing something distinctly unsafe requires raising extra awareness to where I would factor it into the name of a function. I think this boils down to a difference in API design aesthetics that's not worth debating...

Soroosh129: So maybe going back to `_if_apropriate` is a better choice than `_is_safe`.

I think that it should be clear that this function is not going to send a TAG/PTAG just because it is called. It should be called later if the circumstances change.

org.lflang/src/lib/core/federated/RTI/rti.c
Soroosh129: Yes, we have `send_tag_advance_grant` and `send_provisional_tag_advance_grant`.

org.lflang/src/lib/core/federated/RTI/rti.c
lhstrh: I don't think `if_appropriate` is better. I also don't think there's anything that I can add to this discussion at this point, and it doesn't seem like we're able to agree on things, so perhaps someone else can comment on this. We can also declare this a non-issue and move on.

Soroosh129: But there is a logistical issue as well. We'll end up with the following four functions:
```
send_advance_grant
send_advance_grant_downstream
send_tag_advance_grant  // Sends TAG to a federate without any calculations
send_provisional_tag_advance_grant // Sends PTAG to a federate without any calculations
```
If you give me a concrete name for all of these, I will go ahead and rename them to your suggestion.

edwardalee: I think I'm with Soroush on this, though honestly I'm not completely sure. If you have a method called "send_advance_grant" and it doesn't send an advance grant, you would have to assume it is defective, right?  What this method does, I think, is to determine whether it is "safe" (in the "safe to process" sense) to send an advance grant, and, if so, send it.  So to me, send_advance_grant_if_safe seems like a pretty descriptive name.

lhstrh: In any case, I think that `send_advance_grant_if_safe` is an improvement over the current name, so let's go with that then. To be consistent, I suppose the recursive function should be called `send_downstream_advance_grants_if_safe`? IDK, I don't like these super long function names, but again, I think it's better than `send_TAG_if_transitively_safe_to_all_downstream`.

lhstrh: Instead of an `if_safe` suffix, a `try` prefix could also be used to convey the tentative nature of the operation. I think that's more common.

org.lflang/src/lib/core/federated/RTI/rti.c
```diff
@@ -40,6 +40,10 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * but we control both ends, and hence, for commonly used
  * processors, this will be more efficient since it won't have
  * to swap bytes.
+ * 
+ * This implementation of the RTI should be considered a reference
+ * implementation. In the future it might be re-implemented in Java or Kotlin.
+ * Or we could bootstrap, and implement it using Lingua Franca.
```
edwardalee, line 46: Comma is not needed on this line.

org.lflang/src/lib/core/federated/RTI/rti.c
edwardalee: "try" implies that there may be some reason outside the control of the system that it would fail, like network connectivity.

A meta issue: In an OO language, this function would be private.  It is designed for internal use. I personally prefer long descriptive names for such functions because it reduces fatigue reading the code. I spend less time jumping to the function to find its documentation.

org.lflang/src/lib/core/federated/RTI/rti.c
```diff
@@ -764,19 +775,21 @@ void handle_next_event_tag(federate_t* fed) {
     // If the federate has no upstream federates, then it does not wait for
     // nor expect a reply. It just proceeds to advance time.
     if (fed->num_upstream > 0) {
-        send_tag_advance_if_appropriate(fed);
+        send_TAG_if_transitively_safe(fed); // FIXME: Rename appropriate to 
+                                              // allowed or safe instead of appropriate
```
edwardalee, line 779: Remove FIXME here?

edwardalee (approved):
To me, this looks ready to merge.

benchmark/C/Savina/src/micro/Distributed/PingPongDistributed.lf
```diff
@@ -0,0 +1,28 @@
...
+ * There is no parallelism in this application, so it does not benefit from being
+ * being distributed.
+ * 
+ * These measurements are total execution time, including startup and shutdown, of
+ * all three programs.
+ * 
+ * @author Edward A. Lee
+ */
+target C;
+import Ping, Pong from "../PingPong.lf"
```
cmnrd, line 22: We should think about naming conventions for distributed savina benchmarks and also find a way to integrate this with the benchmark runner. But I guess we can do this further down the road when we have a few more benchmarks.

cmnrd (approved):
Looks good!

benchmark/C/Savina/src/micro/Distributed/PingPongDistributed.lf
Soroosh129: Maybe down the line, we can augment the benchmark runner to automatically run the C benchmarks as federated.

lhstrh (approved):
Looks good!

PR #391, opened 2021-07-01
Banks to reaction

edwardalee:
This pull request provides support for banks and multiports at the top-level of a federated program.


edwardalee:
I believe this branch (banks-to-reaction) is now ready to merge with master. Any objections? It provides support for banks of federates, multiports, and parallel connections.

Soroosh129 (approved):
This looks good to me!

I think we should at some point spend the effort and re-structure the CGenerator into smaller pieces to make it easier to work with.

lhstrh (approved):
I agree with @Soroosh129 that the multiport stuff should be factored out and `CGenerator` simplified, but that can be done at a later time.

PR #394, opened 2021-07-03
381 file handling in python target

### LINKED ISSUE #381: File handling with Python target
When compiling a Python target .lf program, it prints:
```
******** generated binaries: /Users/eal/lingua-franca-master/git/lingua-franca/experimental/Python/bin
```
But there are no generated binaries. Perhaps we should suppress this somehow?  And suppress creation of a bin directory?

Also, it would be useful if the compiler would print information about how to run the program. E.g.,
```
******** To run the generated program:
python /Users/eal/lingua-franca-master/git/lingua-franca/experimental/Python/src-gen/Intersection/Intersection/Intersection.py
```
A user should be able to just copy and paste this command, at least until we build a way to run the generated program from within the IDE.

Soroosh129:
Fixes #381.


edwardalee:
Looks great!

PR #401, opened 2021-07-09
Benchmarks

ith8:
Cigarette Smoker Results:
Execution - Summary:
Best Time:	 0.084 msec
Worst Time:	 0.107 msec
Median Time:	 0.165 msec

Sleeping Barber Results:
Execution - Summary:
Best Time:	 0.066 msec
Worst Time:	 0.055 msec
Median Time:	 0.078 msec

C++ Results:
Cigarette Smoker:
Execution - Summary:
          Best Time:     9.192 ms
          Worst Time:    11.647 ms
          Median:    10.886 ms

Sleeping Barber:
Execution - Summary:
          Best Time:   235.287 ms
          Worst Time:   264.893 ms
          Median:   249.625 ms

### DIFF
--- a/benchmark/C/Savina/Barber.lf
+++ b/benchmark/C/Savina/Barber.lf
@@ -59,9 +59,9 @@ reactor SleepingBarber(
 
     logical action done_cutting;
     logical action check_waitingroom;
+    logical action customers_arrive:int; // schedule customer arrival
     
-    
-    reaction(startup)-> customers_start {=
+    reaction(startup)-> customers_arrive {=
         self->chairs = calloc(self->num_chairs, sizeof(int));
         for (int i=0; i < self->num_chairs; i++){
             self->chairs[i] = -1;
@@ -69,16 +69,20 @@ reactor SleepingBarber(
         self->customer_served = 0;
         self->sleep = true;
     	self->customerID = -1;
-        //TODO: schedule random customer arrivals
+        // Schedule random customer arrivals
+        srand(0);
         for(int i = 0; i < self->num_customers; i++) {
-			SET(customers_start[i], true);
-			printf("Customer %d arrived.\n", i);
+			schedule_int(customers_arrive,rand()%(self->num_customers),i);
     	}
     =}
     
+    reaction(customers_arrive) -> customers_start {=
+        int i = customers_arrive->value;
+        SET(customers_start[i], true);
+    =}
     reaction(cut) -> done_cutting, wait, full {=
         for (int i = 0; i < self->num_customers; i++){
-            if (cut[i]){
+            if (cut[i]->is_present){
         		if (self->sleep){
             		self->sleep = false;
             		printf("*********** Barber wakes up at %lld\n", get_elapsed_logical_time());
@@ -86,17 +90,15 @@ reactor SleepingBarber(
             		self->customerID = i;
         		} else if (put_customer(self->chairs, self->num_chairs,i)){
             		SET(wait[i],true);
-            		printf("Customer %d is in the wait room.\n", i);
         		} else {
             		SET(full[i],true);
-            		printf("Waiting room full customer %d left.\n", i);
         		}
         	}
         }
     =}
     
     reaction(done_cutting)->check_waitingroom, done {=
-        printf("Barber is serving customer %d.\n", self->customerID);
+        printf("Barber is done serving customer %d.\n", self->customerID);
         SET(done[self->customerID],true);
         self->customerID = -1;
         self->customer_served++; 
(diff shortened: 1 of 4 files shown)


cmnrd:
Looks good! It seems that the results you report are obtained for very different problem sizes and are thus not comparible. This brings me to another point: The benchmarks should be integrated with the benchmark runner script in `benchmark/runner`. For this, the configuration files for those benchmarks would need to be updated so that they include the C configuration. Also the benchmarks need to be adapted, so that the parameters specifying the problem size can be varied via the cog code gnerator. See the [PingPongGenerator](https://github.com/icyphy/lingua-franca/blob/master/benchmark/C/Savina/PingPongGenerator.lf).

Finally, there is the problem with the BenchmarkRunner reactor that I also mentioned in #384. This should be discussed with the whole team, but as far as I know the consensus is to remove the BenchmarkRunner and have each invocation of the binray execute only a single iteration of the benchmark. (This is also what the runner script currently assumes for LF C programs)

cmnrd (changes requested):
I really like how you implemented those problems in a lean and simple way! However, unfortunately, this makes your implementation quite different from the original benchmarks, and in consequence this makes comparing the C version with the other versions very hard. Take for instance the Sleeping Barber benchmark. In the Akka implementation, there are two more actors: the waiting room and a customer factory. Also both the barber and the factory have a workload function that performs a (admittedly meaningless) computation. Since you removed those from the benchmark, we cannot really compare the performance between your version and the Akka implementation (as well as the LF C++ version).

In the meantime, we updated and simplified all the C++ implementations of the benchmarks. So those should give a good template. Could you make another pass over your benchmarks and  bring them more in line with the C++ versions? Of course, there are always trade-offs when converting an actor program to LF, and its not always obvious how to translate the benchmarks. The C++ implementations are by no means perfect in this regard. So if you have suggestions or think that something should be done differently, I am happy to have a talk about this! However, unless there is good reason for divergence, we should aim at having at least similar implementations in C and C++. In that regard, could you also see that the file names align between the C and C++ versions?

I made a few further comments on Barber.lf that are mostly related to cog and the integration with the benchmark runner. These are exemplary and also apply to the other 3 benchmarks.

edwardalee:
On the question of the differences between the benchmark designs, I think that in some cases we will want two versions of a benchmark to measure two different things:

A. Replicate as closely as possible the logic in the Akka implementation. We will have to pay careful attention to timing, which has to be added to the benchmark because all LF programs use logical time, and none do in Akka. In this case, sometime it will make sense to use physical connections (syntax ~>), resulting in nondeterministic models. 

B. Think about the problem that the benchmark is trying to solve and solve the problem using semantic features of Lingua Franca.

A is a better measure of the efficiency of our implementation.  B is a better measure of the suitability of our language.  These are measuring different things.

Perhaps we should set up a parallel directory structure for these two classes of benchmarks?

I have only really compared these strategies for the smokers benchmark, and even then, not fully.  Strategy A uses a centralized dispatcher.  Strategy B (which you can find in example/C) does not.  But I haven’t tried scaling these up nor measuring the actual performance differences.  This might be a good place to start.

cmnrd:
Yes, I fully agree.

I think the focus of this PR is A, and hence we should try to match it up with the C++ versions (this can also mean adjusting the C++ versions if there is good reason).

I am also very interested in B, and I think there are some interesting candidates to have a closer look at. Doing this in a separate directory is a good idea, and probably the example directory is a good place to start. Maybe Thee can add her version of the Sleeping Barber there as a starting point.

lhstrh:
Why not keep the benchmark where it is and rename it to `<name>_Optimized` or `<name>_Alt1` or something like that? I think we still want to do actual bench marking on this, but compare the results with the "canonical" Savina-based LF implementation rather than directly against C++/Akka/CAF.

edwardalee:
Yes, <name>_Atl1.lf sounds like a good idea, and yes, comparing against the type A LF benchmark seems like the right thing.

edwardalee:
We had a discussion on this in the call this morning (afternoon) and decided this:  Type A benchmarks will go in the benchmarks directory.  Type B benchmarks will go into the examples directory, and the corresponding type A benchmark will have a comment saying something like:

This implementation closely follows the realization of the Savina benchmark in Akka. An alternative implementation of this problem that takes advantage of reactor semantics can be found in `example/...`.

Then we can run performance comparisons between the type A and B benchmarks.

To have everything in one place, these are the definitions:

A. Replicate as closely as possible the logic in the Akka implementation. We will have to pay careful attention to timing, which has to be added to the benchmark because all LF programs use logical time, and none do in Akka. In this case, sometime it will make sense to use physical connections (syntax ~>), resulting in nondeterministic models.

B. Think about the problem that the benchmark is trying to solve and solve the problem using semantic features of Lingua Franca.

lhstrh:
@ith8, @cmnrd: what's the status of this PR? I would like to merge this as soon as possible. What's blocking us?

cmnrd:
See the comments I made in my review. As far as I know, nothing has changed since then. Is @ith8 working on this?

It would be ok for me to merge this unfinished, but we need to be aware that the benchmarks as they are now do not really compare to Savina and that they are not fully integrated yet with the benchmark runner script.

mattchorlian:
Should we merge this PR in with these benchmarks moved to the examples directory?  The consensus seems to be these fall under Category B.  Then, I can open a new PR with the Akka-based (Category A) implementations in the coming days.  How does this sound?  Once these are hooked up to the benchmark runner script are they ready to be merged? @cmnrd

mattchorlian:
On second thought, since I have the Akka-based SleepingBarber benchmark and config file ready to go I will add it here so it can get merged in with these changes.  This way, with CigSmok already merged, the new SleepingBarber added, and the dictionary benchmarks all set, we have covered all 4 benchmarks this PR aimed to address.

PR #402, opened 2021-07-10
CMake for the C target

Soroosh129:
This branch adds the option of compiling generated C target code using `cmake`. This can be enabled by using the `cmake: true` target property.

This branch is built on top of the `standalone-RTI` branch, which should be merged before this.


org.lflang/src/lib/core/federated/net_common.h
```diff
@@ -592,6 +600,35 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * bytes), delay (8 bytes)) for this federate's connection to upstream federates
+ * (by direct connection). The delay is the minimum "after" delay of all
+ * connections from the upstream federate.
+ *
+ * Depending on the second four bytes, the next bytes are fed IDs (2
+ * bytes each), of this federate's downstream federates (by direct connection).
+ *
+ * @note The upstream and downstream connections are transmitted on the same
+ *  message to prevent (at least to some degree) the scenario where the RTI has
+ *  information about one, but not the other (which is a critical error).
```
edwardalee, line 627: It seems that possible critical error is actually a different one: Where federate A claims that federate B is downstream of it, but federate B does not claim that federate A is upstream of it.  Perhaps federates should only inform the RTI of **downstream** federates.  The RTI would use this to populate the data structure for both referenced federates. Then the RTI could never have inconsistent data.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -192,6 +193,28 @@ public FileConfig(Resource resource, IFileSystemAccess2 fsa, IGeneratorContext c
         this.iResource = getIResource(resource);
     }
     
+    protected FileConfig(FileConfig fileConfig) throws IOException {
```
edwardalee, line 196: Document that this is a copy constructor?

org.lflang/src/org/lflang/federated/FedFileConfig.java
```diff
@@ -0,0 +1,50 @@
...
+package org.lflang.federated;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.lflang.FileConfig;
+
+/**
+ * A child class of @see FileConfig that extends the base functionality to add support
+ * for federated execution. Each federate should instantiate this class.
```
edwardalee, line 35: Perhaps replace the last sentence with: The code generator should create one instance of this class for each federate?

org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -13,6 +13,7 @@
 import org.lflang.ErrorReporter;
 import org.lflang.FileConfig;
 import org.lflang.Target;
+import org.lflang.generator.c.CGenerator;
```
edwardalee, line 16: Wait, why does LFGenerator now have a dependence on CGenerator?

org.lflang/src/org/lflang/generator/TypeScriptGenerator.xtend
```diff
@@ -335,7 +335,6 @@ class TypeScriptGenerator extends GeneratorBase {
 
         // If this is a federated execution, generate C code for the RTI.
         if (isFederated) {
-            createFederateRTI()
```
edwardalee, line 338: Why is the TypeScript generator copying all these C files?  Maybe this was there only for the RTI?  No longer relevant?

test/C/src/federated/DistributedCountDecentralized.lf
```diff
@@ -18,10 +18,10 @@ reactor Print {
         interval_t elapsed_time = get_elapsed_logical_time();
         printf("At tag (%lld, %u), received %d. "
                "The original intended tag of the message was (%lld, %u).\n",
-                elapsed_time,
+                (long long int)elapsed_time,
```
edwardalee, line 21: Is this cast needed? If so, we have *many* places that will need it.  Maybe there is a printf code for int64?

edwardalee (approved):
This looks great.  One major question: Should federates send only **downstream** federates to the RTI? Once the RTI has heard from all the federates, it will know which are upstream and downstream of which.  By having federates send both upstream and downstream, there is a risk of an inconsistency getting created.

org.lflang/src/lib/core/federated/RTI/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+# Usage:
+#
+# To compile with cmake, run the following commands:
+#
+# $> mkdir build && cd build
+# $> cmake ../
+# $> make
+# $> sudo make install
+#
+# This create a binary RTI in the current working directory. Please put this in
```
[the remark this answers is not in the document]
cmnrd: I am not sure in which context this cmake file is called. By default, cmake installs to the system root. Only if an install prefix is given when running cmake, it alternates the installation path. This is analogous to th way `configure` works. In case of the C++ LF target, the installation path is automatically adjusted to install in the project dir (or cwd if there is no src directory). However, for the RTI you might whant to have it available in the system root path.

org.lflang/src/org/lflang/FileConfig.java
cmnrd: Why is a copy constructor needed? In my view, an instance of FileConfig should only exist once and be passed around by reference.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -374,6 +397,45 @@ public static void createDirectories(Path path) {
...
+     * 
+     * @param source The source file path.
+     * @param dest The destination file path.
+     * @throws RuntimeException if copy fails.
+     */
+    public static void copyFile(Path src, Path dest) {
+        try {
+            Files.copy(src, dest, java.nio.file.StandardCopyOption.REPLACE_EXISTING);
+        } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
```
cmnrd, line 436: Why is RuntimeException thrown here and not IOException as in copyDirectory()? Also catching all types of exceptions is a code smell, as it has the potential to mask problems.

org.lflang/src/org/lflang/generator/LFGenerator.java
cmnrd: LFGenerator is responsible for instantiating the target generators. Thus it needs to import all generators. However, the imports don't appear here, because LFGenerator and most of the target generators are in the same package. But the CGenerator is now in the c subpackage and thus the import is needed.

org.lflang/src/org/lflang/federated/FedFileConfig.java
```diff
@@ -0,0 +1,50 @@
...
+ * 
+ * @author Soroush Bateni
+ *
+ */
+public class FedFileConfig extends FileConfig {
+    
+    public FedFileConfig(FileConfig fileConfig, String federateName) throws IOException {
```
cmnrd, line 42: Is this class used somewhere? I couldn't find it.

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -0,0 +1,206 @@
...
+            } else {
+                cMakeCode.append("set(CMAKE_C_COMPILER "+targetConfig.compiler+")\n");
+            }
+            
+            // cMakeCode.append("set(CMAKE_CXX_COMPILER "+targetConfig.compiler+")\n");
+        }
+        
+        // Set the compiler flags
+        // We can detect a few common libraries and use the proper target_link_libraries to find them            
+        for (String compilerFlag : targetConfig.compilerFlags) {
```
cmnrd, line 168: This looks a bit dangerous to me. I think it would be better to forbid the usage of the compiler flags target property if cmake is enabled and use `cmake-include` instead. Note that the default case in the code below is not portable and would only work with gcc/g++ or clang.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -394,88 +406,61 @@ class CGenerator extends GeneratorBase {
...
+            fileConfig = oldFileConfig;
             startTimeStepIsPresentCount = 0
             startTimeStepTokens = 0
             
-            
             // If federated, append the federate name to the file name.
             // Only generate one output if there is no federation.
             if (isFederated) {
                 topLevelName = baseFilename + '_' + federate.name // FIXME: don't (temporarily) reassign a class variable for this
+                fileConfig = new FedFileConfig(fileConfig, federate.name);
```
cmnrd, line 462: Ah, here it is used. I guess in this case also the copy constructor makes sense. However, I think the copy constructor should not be placed in `FileConfig`, but in `FedFileConfig` so that we can create federated configs from a file config.

org.lflang/src/lib/core/federated/RTI/CMakeLists.txt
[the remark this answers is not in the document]
Soroosh129: The cmake command is invoked by the user manually in a `build` directory in this folder. Just running `make` will produce an RTI binary in that folder (i.e., `org.lflang/src/lib/core/federated/RTI/build/RTI`). 

As @cmnrd pointed out, I think we need the `RTI` to be accessible from everywhere. Therefore, by default, `sudo make install` will put the binary in `/usr/local/bin/`. This behavior can always be overridden by passing the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html) variable to cmake (e.g., `cmake -DCMAKE_INSTALL_PREFIX=/usr ../`). This is I think standard CMake procedure as best as I can tell. The default installation place is documented as "`/usr/local` on UNIX and `c:/Program Files` on Windows".

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: @edwardalee Okay, thank you for catching that!

@cmnrd There is a quirk to the way .c files are generated for federated programs because multiple programs (federates) would need to be created from a single .lf file. The quirk is that each program will be called the same name as its variable (e.g., `a = new Count()` will be called federate a) and will have its own generated `.c` file. Therefore, each federate needs its own CMakeLists.txt. To avoid any conflicts, the generated files for each federate would need to be put in their own separate folder.

For example, for [DistributedCount.lf](https://github.com/icyphy/lingua-franca/blob/master/test/C/src/federated/DistributedCount.lf), the generated folder structure will look like:
```
DistributedCount % tree -L 2
.
├── c
│   ├── DistributedCount_c.c
│   ├── core
│   └── ctarget.h
└── p
    ├── DistributedCount_p.c
    ├── core
    └── ctarget.h

4 directories, 4 files
```

The normal FileConfig does not understand this structure, and assumes that the `src-gen` directory for DistibutedCount.lf is `DistributedCount`. This federated FileConfig will spring into action only if the program is federated and will change the `src-gen` directory to correctly reflect the federate's.

test/C/src/federated/DistributedCountDecentralized.lf
Soroosh129: Yes. Unfortunately, `int64_t` on my 64-bit machines is a `long int`, not a `long long int`. In this case, `%ld` would be fine on most 64-bit machines. 

However, there is a syntax for printing `int64_t` ([source](https://stackoverflow.com/questions/9225567/how-to-print-a-int64-t-type-in-c), [documentation](https://en.cppreference.com/w/cpp/types/integer)):
```
#include <inttypes.h>
int64_t my_int = 999999999999999999;
printf("%" PRId64 "\n", my_int);
```

I wouldn't expect the end user to know this. Maybe we should revert back to `long long int` for `instant_t`/`interval_t`? This would mostly be a problem for the RTI, I think, which can be resolved by some careful casting. The `long long int` specification states an integer that is "at least" 64-bit, which can cause unexpected behavior I'm sure in the future. However, 128-bit machines are probably a few years away.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: I think a lot of variables assigned in the copy constructor are private. I didn't want to convert them to protected unnecessarily. Do you think that would be a better solution?

org.lflang/src/org/lflang/generator/TypeScriptGenerator.xtend
Soroosh129: Ah, great catch! This piece of code seemed to only be relevant for the RTI.

lhstrh: Yes, Hokeun and I have plans to completely clean up `TypeScriptGenerator`. After this pull request has been merged in we'll be able to get rid of RTI-related code in that class.

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
Soroosh129: Ah, good point. I would be hesitant to remove this completely though (i.e., in the validator) because of backwards compatibility. If CMake was enabled by default, but wasn't installed for example, we might want to revert back to using the native gcc compiler. Using `cmake-include` will not help in that case. I will add a warning here for now.

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: This was to avoid checked [exceptions](https://www.baeldung.com/java-lambda-exceptions#handling-checked-exceptions) in a lambda expression. I have pushed a fix.

org.lflang/src/lib/core/federated/net_common.h
Soroosh129: I have created an issue to track this problem.

test/C/src/federated/DistributedCountDecentralized.lf
Soroosh129: I have create an issue (#426) to track this.

Soroosh129:
I believe this branch is ready to merge. I have changed the default build system to be CMake and added CMake to REQUIREMENTS.md. Is there any other documentation that needs to be modified?

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -96,10 +96,20 @@
...
     CMAKE_INCLUDE("cmake-include", PrimitiveType.STRING,
-            Arrays.asList(Target.CPP), (config, value) -> {
+            Arrays.asList(Target.CPP, Target.C), (config, value) -> {
                 config.cmakeInclude = ASTUtils.toText(value);
             }),
     
+    /**
+     * Directive to enable and disable the use of CMake.
+     * 
+     * The default is disabled.
```
lhstrh, line 106:
```suggestion
     * The default is enabled.
```

org.lflang/src/org/lflang/federated/FedCLauncher.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Utility class that can be used to create a launcher for federated LF programs
+ * that are written in C.
+ * 
+ * @author Soroush Bateni <soroush@utdallas.edu>
+ */
+public class FedCLauncher extends FedLauncher {
+
+    /**
+     * @param targetConfig The current target configuration.
```
lhstrh, line 43: Missing comment.

org.lflang/src/org/lflang/federated/FedLauncher.xtend
```diff
@@ -0,0 +1,394 @@
...
+     *     openssl version
+     * 
+     * @param coreFiles The files from the core directory that must be
+     *  copied to the remote machines.
+     * @param federates A list of federate instances in the federation
+     * @param federationRTIProperties Contains relevant properties of the RTI.
+     *  Can have values for 'host', 'dir', and 'user' 
+     * 
+     */
+    def createLauncher(
```
lhstrh, line 126: I would suggest to break this up into multiple functions.

org.lflang/src/org/lflang/generator/FederateInstance.xtend
```diff
@@ -369,6 +369,7 @@ class FederateInstance {
     
     /** The generator using this. */
     var generator = null as GeneratorBase
+    def getGenerator() { return generator; }
```
lhstrh, line 372: Undocumented function.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -0,0 +1,189 @@
...
+import org.lflang.Mode;
+import org.lflang.TargetConfig;
+import org.lflang.generator.GeneratorCommandFactory;
+import org.lflang.util.LFCommand;
+
+/**
+ * Responsible for creating and executing the necessary native command to compile code that is generated
+ * by the CGenerator. This class invokes a compiler directly.
+ * 
+ * @author Soroush Bateni <soroush@utdallas.edu>
```
lhstrh, line 44: I'm glad that we're factoring things out of `GeneratorBase`, but we should be mindful of the authorship of code that gets factor out; if this code gets put in a different class, then the original authors should also be listed as a coauthor of the new class. Granted, it's sometimes a bit difficult to determine who originally authored code, but we should try to be mindful of this. This is a general comment that doesn't only apply to this class, of course.

test/C/src/federated/HelloWorldFederatedCPPNoCMake.lf
```diff
@@ -0,0 +1,31 @@
+/**
+ * This test is a variant of HelloWorld.lf that checks the compatibility
+ * of the federated C runtime with g++ and no CMake. This test should not pass if it does not compile.
+ */
+target C {
+    tracing: true,
+    compiler: "g++",
+    flags: "-Wno-write-strings", // Suppress a ridiculous number of warnings.
```
lhstrh, line 8: Why not address the warnings?

lhstrh (approved):
Great progress toward a more modular code base! I left a few minor comments...

org.lflang/src/org/lflang/generator/FederateInstance.xtend
Soroosh129: Should getters and setter be documented? The document would just say `Returns generator`.

org.lflang/src/org/lflang/generator/c/CCompiler.java
Soroosh129: This `@author` thing is something that Eclipse adds for me automatically whenever I create a Java class. Perhaps authors of the original file (`GeneratorBase`) in this case should automatically also be included here in the same order?

test/C/src/federated/HelloWorldFederatedCPPNoCMake.lf
Soroosh129: These warnings are due to a behavior that is undefined under `C++` but perfectly legal in `C`. There is a discussion about this [here](https://stackoverflow.com/questions/20944784/why-is-conversion-from-string-constant-to-char-valid-in-c-but-invalid-in-c). This warning can go away if we explicitly cast string assignments (e.g., `="foo"`) to `char*` (e.g., `=(char*)"foo"`). However, I don't think we should care too much about C++ in the C target though, although we can add these casts and be done with it. Nonetheless, I think this change might merit its own pull request perhaps.

org.lflang/src/org/lflang/federated/FedLauncher.xtend
Soroosh129: This is a great suggestion. I will do that as part of the federated Python developement. I partly factored out the federated launcher here to facilitate a `FedPythonLauncher` in a modular way. Breaking this function up will make this even more modular.

edwardalee:
The example given in [Writing-Reactors-in-C](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#the-c-target-specification) assumes cmake is not being used. I think this needs to be updated.

Soroosh129:
> The example given in [Writing-Reactors-in-C](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#the-c-target-specification) assumes cmake is not being used. I think this needs to be updated.

Okay, thank you for pointing that out. I have added the CMake-related target parameters to the text. For now, it includes a mention to the variable `LF_MAIN_TARGET`, which should be renamed to `LF_MAIN_REACTOR`. I think we need to do that in unison with the C++ target to avoid confusion, perhaps. What do you think @cmnrd?

org.lflang/src/org/lflang/federated/FedLauncher.xtend
lhstrh: OK, sounds good to me.

org.lflang/src/org/lflang/generator/c/CCompiler.java
lhstrh: If in doubt, `git blame` can also be a useful tool.

test/C/src/federated/HelloWorldFederatedCPPNoCMake.lf
lhstrh: OK, I understand. Let's not address this as part of PR then.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
lhstrh: I see no objection in making those variables protected. Shall we just do that and implement @cmnrd's suggestion?

org.lflang/src/org/lflang/generator/c/CCompiler.java
Soroosh129: Okay, thanks for the pointer!

org.lflang/src/org/lflang/generator/FederateInstance.xtend
lhstrh: It's not a big deal. Active annotation can be used for this type of thing in Xtend. Perhaps using that will generate better documentation automatically. It looks something like:
```
 @Accessors(PUBLIC_GETTER, PROTECTED_SETTER) int age
```

lhstrh:
> > The example given in [Writing-Reactors-in-C](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#the-c-target-specification) assumes cmake is not being used. I think this needs to be updated.
> 
> Okay, thank you for pointing that out. I have added the CMake-related target parameters to the text. For now, it includes a mention to the variable `LF_MAIN_TARGET`, which should be renamed to `LF_MAIN_REACTOR`. I think we need to do that in unison with the C++ target to avoid confusion, perhaps. What do you think @cmnrd?

How about we create an issue for this and just continue with the merge?

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: I think a copy constructor in `FileConfig` is a useful feature to have. I don't agree necessarily that we should remove this copy constructor. Can leaving it as is lead to a dangerous situation?

org.lflang/src/org/lflang/generator/FederateInstance.xtend
Soroosh129: I think using this type of annotation requires adding one more dependency: `org.eclipse.xtend.lib.annotations`. 
I will just add a comment here, which I think will be less painful when we port this class to Kotlin or Java.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
lhstrh: I don't see it being a problem, no. Invoke the copy constructor of `FileConfig` inside the copy constructor of `FedFileConfig`? That way you also don't have to modify the visibility.

Soroosh129:
> > > The example given in [Writing-Reactors-in-C](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#the-c-target-specification) assumes cmake is not being used. I think this needs to be updated.
> > 
> > 
> > Okay, thank you for pointing that out. I have added the CMake-related target parameters to the text. For now, it includes a mention to the variable `LF_MAIN_TARGET`, which should be renamed to `LF_MAIN_REACTOR`. I think we need to do that in unison with the C++ target to avoid confusion, perhaps. What do you think @cmnrd?
> 
> How about we create an issue for this and just continue with the merge?

Okay sounds good. I have created #429.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: Okay. Done in ebc62defca50d3d407245c1e9b112a0a64d75531.

PR #406, opened 2021-07-12
Rework C++ benchmarks

cmnrd:
This is a work in progress merge request for keeping track of progress in a current effort to improve the C++ benchmarks. Apart from beautification and removing bottlenecks, this overhaul is also needed because merging #375 actually broke many of the benchmarks. This is because they circumvented the scoping rules in some places to better handle banks/multiports. With #387, this should not be required anymore and we need to adjust all the benchmarks.

This PR also adjusts the directory structure of the C++ benchmarks as discussed in #410.
Additionally, the renaming from `instance` to `bank_index` has not been propagated into all benchmarks. Benchmarks with this problem have been annotated.

Current Status:
*  micro benchmarks
    - [x] Big
    - [x] Chameneos
    - [x] Count
    - [x] Fork Join
    - [x] Ping Pong
    - [x] Thread Ring
* concurrency benchmarks
    - [x] Banking
    - [x] Barber
    - [x] Bounded Buffer
    - [x] Cigarette Smokeres
      (fix `instance` reference)
    - [x] Concurrent Dictionary
      (fix `instance` reference)
    - [x] Concurrent Sorted List
      (fix `instance` reference)
    - [x] Logistic Map
      (fix `instance` reference)
    - [x] Philosophers
* parallelism benchmarks
    - [x] APSP
    - [x] ASTAR
      (fix `instance` reference)
    - [x] Filter Bank
      (fix `instance` reference)
    - [x] N Queens K Solutions
      (fix `instance` reference)
    - [x] Pi Precise
    - [x] Radix Sort
    - [x] Rec Matrix Multiplication
      (fix `instance` reference)
    - [x] Trapezoid
      (fix `instance` reference)
* other steps
    - [x] check that all occurrences of "master" are renamed
    - [x] add copyright and author notes

This is based on ~~#387, #382, #399, #416 and #420~~. Therefore, this MR also includes changes from the other MRs and should only be merged once the other MRs are merged.

In spirit of #423, this PR also renames all reactors called "Master"


cmnrd:
This PR is now ready to be merged!

benchmark/Cpp/Savina/src/BenchmarkRunner.lf
```diff
@@ -1,135 +1,83 @@
...
  *         printSystemInfo();
- *         runner.inStart.set();
  *     =}
  * 
- * @param numIterations How many times to execute the kernel of the benchmark to measure.
- * @param useInit Benchmarks needs initialization and handles the corresponding signals.
- * @param useCleanupIteration Benchmark needs cleanup after each iteration and handles the corresponding signals.
- * 
  * @author Hannes Klein
+ * @autho Christian Menard
```
Feliix42, line 29: There's a typo here :wink:

Feliix42 (commented):
Although there's a typo in the docs I think it's overall looking good!

Feliix42 (approved):

benchmark/Cpp/Savina/src/Matrix.hh
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2020 TU Dresden
+
+// A simple classes for strocing a matrix compactly and providing convenient access
```
lhstrh, line 3:
```suggestion
// A simple class for storing a matrix compactly and providing convenient access.
```

benchmark/Cpp/Savina/src/concurrency/Banking.lf
```diff
@@ -0,0 +1,201 @@
...
+ * receiving any new messages.
+ * 
+ * The initial generation of all the credit messages is an interesting problem in LF.
+ * In the Akka implementation, the teller actor will just send `numBankings` randomly 
+ * generated messages to the account actors, which then process one message after the 
+ * other. However, this is not easily doable in LF, as a port can only hold a single 
+ * value. In order to send multiple messages to the same account, the teller needs to
+ * use an action and advance in logical time before sending the next message. This is
+ * currently implemented such that the Teller creates an initial queue per account
+ * which contains all credit messages that should be send to this account. Then it
```
lhstrh, line 47:
```suggestion
 * which contains all credit messages that should be sent to this account. Then it 
```

benchmark/Cpp/Savina/src/concurrency/Banking.lf
```diff
@@ -0,0 +1,201 @@
...
+                double amount = *port.get();
+                // increase the balance
+                balance += amount;
+                reactor::log::Info() << "Account " << bank_index << " received " << amount;                
+            }
+        }
+    =}
+}
+
+// FIXME default value of numAccounts should be 1000
```
lhstrh, line 181: Why this FIXME?

benchmark/Cpp/Savina/src/concurrency/BoundedBuffer.lf
```diff
@@ -0,0 +1,227 @@
...
+ * This benchmark models a bounded buffer that is used to synchronize between a
+ * group of consumers and a group of producers. In the Akka implementation of
+ * this benchmark, all producers, consumers and the manager operate fully
+ * asynchronously. However, in the LF implementation everything is synchronous.
+ * Thus, strictly speaking, the buffer isn't actually needed as we can design
+ * the program so that every tick a producer produces data and a consumer
+ * consumes it. But, the number of consumers and producers can be seleced
+ * arbitrarily. Therefore, we still need some logic that takes care of or
+ * orchestrating the producers according to the rate at which the consumers
+ * consume. Also, we could implement the consumer such that they don't become
```
lhstrh, line 13: I don't understand the "could" in this sentence.  Should it be "must"?

benchmark/Cpp/Savina/src/concurrency/CigaretteSmoker.lf
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Since actors operate asynchronously, several smokers can actually smoke in
+ * parallel. In theory, this allows for a fast execution, as the workload of
+ * each smoker can be processed in parallel by many smokers. Achieving this in
+ * LF, however, is non-trivial. In LF, all the communication is synchronous and
+ * instantaneous. When the arbitrator tells a smoker to smoke, the smoker will
+ * start smoking at the same tag and the execution will wait until this smoker
+ * finished. Only then the runtime will forward logical time to the next tag and
+ * the arbitrator will tell the next smoker to smoke. Obviously, there is no
+ * parallelism in this solution.
```
lhstrh, line 28: The point of this challenge problem to achieve coordinated mutual exclusion (i.e., without deadlock) between concurrent processes. If parallel smoking is desired, the coordinator can put more than one ingredient on the table and thus signal multiple smokers to smoke simultaneously. This is easy to do in LF...

benchmark/Cpp/Savina/src/concurrency/Dictionary.lf
```diff
@@ -0,0 +1,192 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ * 
+ * This is a relatively simple benchmarks where multiple workers interact concurrently with
+ * a central dictionary. The original Akka implementation does not make an effort to synchronize and order the incoming requests. It simply processes requests in the order they are delivered to the dictinary actor by the runtime.
```
lhstrh, line 5: Please format; exceeding maximum line width.

benchmark/Cpp/Savina/src/concurrency/Dictionary.lf
```diff
@@ -0,0 +1,192 @@
...
+ * The only synchroniztion applied, is that the workers wait for a response from the dictionary before sending the next request.
+ * 
+ * In the LF implementation, all components operate synchronously. This means at each logical time step all workers send a request to the dictionary. The dictionary processes the requests in a fixed order which makes the whole application deterministic.
```
lhstrh, line 8:
```suggestion
 * In the LF implementation, all components operate synchronously. This means that at each logical time step all workers send a request to the dictionary. The dictionary processes the requests in a fixed order which makes the whole application deterministic.   
```

benchmark/Cpp/Savina/src/concurrency/LogisticMap.lf
```diff
@@ -0,0 +1,123 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ *
```
lhstrh, line 3: The comments you wrote for the other benchmarks are great and super useful! They are missing from this benchmark.

benchmark/Cpp/Savina/src/concurrency/Philosopher.lf
```diff
@@ -1,58 +1,20 @@
 /**
- * Concurrency benchmark from the Savina benchmark suite.
- * See https://shamsimam.github.io/papers/2014-agere-savina.pdf.
+ * Copyright (C) 2020 TU Dresden
  * 
  * This implementation stays close to the Savina implementation
```
lhstrh, line 4: These comments could be improved.

benchmark/Cpp/Savina/src/concurrency/SleepingBarber.lf
```diff
@@ -0,0 +1,279 @@
...
+ * Copyright (C) 2020 TU Dresden
+ * 
+ * In this benchmark, there are three main components: The barber, a waiting room and the customers.
+ * The customers are produced and controlled by a CustomerFactory.
+ * This implementation is close to the Savina Akka implementation
+ * in terms of the work that is done in each actor/reactor and
+ * that for each message sent in Savina there is a message send
+ * in this implementation. But this implementation does not
+ * show the basic idea of the problem in that the waiting
+ * room is basically never because the customers arrival is
```
lhstrh, line 11: Wonky sentence.

benchmark/Cpp/Savina/src/concurrency/SleepingBarber.lf
```diff
@@ -0,0 +1,279 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ * 
+ * In this benchmark, there are three main components: The barber, a waiting room and the customers.
+ * The customers are produced and controlled by a CustomerFactory.
+ * This implementation is close to the Savina Akka implementation
+ * in terms of the work that is done in each actor/reactor and
+ * that for each message sent in Savina there is a message send
```
lhstrh, line 8:
```suggestion
 * that for each message sent in Savina there is a message sent
```

benchmark/Cpp/Savina/src/concurrency/Banking.lf
cmnrd: Due to #433 we cannot set the number to 1000. Otherwise, Epoch would crash when opening the file. I will add this to the FIXME.

benchmark/Cpp/Savina/src/concurrency/BoundedBuffer.lf
cmnrd: No, I wanted to say that the benchmark could also be designed differently. I rephrased the sentence to make this clearer (I hope).

benchmark/Cpp/Savina/src/concurrency/CigaretteSmoker.lf
cmnrd: I didn't describe the problem very well (I actually confused myself while reading it again...). I adjusted the text and hope my point is clearer now.

The cigarette smoker problem is a little more complex. The smokers implement two operations: making a cigarette and smoking. They first need to make a cigarette before they can smoke, and for making a cigarette they need access to the ingredients (resources). Since access to the resources is limited, only a single smoker can make a cigarette at any point in time. However, all the smokers that already have a cigarette can smoke whenever they like and smoking can take arbitrarily long. It is this parallelism between the cigarette smoking process and the arbitration and cigarette making that is hard to achieve in LF. In order to make a new cigarette, we need to advance time (because simultaneous access to the resources is not allowed, and also to avoid cycles), but due to the barrier synchronization we need to wait for all smoking reactions to finish before we can advance time.

benchmark/Cpp/Savina/src/concurrency/Dictionary.lf
cmnrd: Done

benchmark/Cpp/Savina/src/concurrency/SleepingBarber.lf
cmnrd: In fact the whole comment is outdated. I think this is (in part) the old comment that Hannes wrote and I forgot to update it. It's fixed now.

benchmark/Cpp/Savina/src/concurrency/LogisticMap.lf
cmnrd: Right, this one slipped through. I honestly also don't understand well what exactly it is doing. I'll give it another look

benchmark/Cpp/Savina/src/concurrency/Philosopher.lf
cmnrd: Agreed, this is still the old comment and I will update it.

benchmark/Cpp/Savina/src/concurrency/CigaretteSmoker.lf
edwardalee: It seems that it would behove us to explicitly model the timing in the problem.  The performance bottleneck you cite is because smoking takes zero (logical) time.

cmnrd: I don't understand how adding a logical delay for smoking would increase the performance here. The problem arises from the barrier synchronization in our runtime. This would of course look differently in a federated execution, but in the non-federeated runtime, we are forced to wait until all reactions complete before we can advance time and give the ingredients to the next smoker. I think it would be worthwhile though to think about ways to relax the barrier synchronization here.

benchmark/Cpp/Savina/src/concurrency/SleepingBarber.lf
```diff
@@ -0,0 +1,293 @@
...
+ * rate parameters. Depending on the rates selected and the precise runtime
+ * scheduling, the waiting room will have a constantly changing state in an
+ * actor implementation.
+ *
+ * In an LF implementation, the processing time of the workload does not
+ * influence the logical arrival and processing rates. Even with a random
+ * workload, the customer factory and the barber could operate logically
+ * synchronously in LF. To mimic the behaviour of the original Akka
+ * implementation and to have a more realistic setup, this LF implementation
+ * also uses varying logical delays. The factory will produce customers with a
```
lhstrh, line 28: Why not use physical actions?

benchmark/Cpp/Savina/src/micro/ThreadRing.lf
```diff
@@ -0,0 +1,150 @@
...
+        printBenchmarkInfo("ThreadRingBenchmark");
+        printArgs("numIterations", numIterations, "numPings", numPings, "numReactors", numReactors);
+        printSystemInfo();
+    =}
+    
+    loopOpener = new ThreadRingReactorLoopOpener();
+
+    workers = new[{=numReactors-1=}] ThreadRingReactor();
+
+    loopOpener.outNextReactor, workers.outNextReactor -> workers.inPrevReactor, loopOpener.inPrevReactor;
```
lhstrh, line 146: Wouldn't it better to first create a zero-delay chain of workers and then, in a separate statement, connect the last worker to the first using an `after 0` clause? That would obviate the need for the `ThreadRingReactorLoopOpener` altogether.

benchmark/Cpp/Savina/src/micro/Throughput.lf
```diff
@@ -1,4 +1,6 @@
 /**
+ * Copyright (C) 2020 TU Dresden
+ *
```
lhstrh, line 3: Some comments here, too, would be nice :-)

benchmark/Cpp/Savina/src/parallelism/FilterBank.lf
```diff
@@ -0,0 +1,348 @@
...
+ * This benchmark is particularly interesting for LF, as it has an interesting
+ * structure and highlights a significant advantage compared to Akka. The
+ * benchmark implements a "filter bank". Each bank consists of a pipeline of
+ * several filters. The banks operate in parallel. The input data stream is
+ * split, so that each bank receives a copy of the data value, and the data
+ * streams are merged again by summing the results. In principle, this design
+ * exposes both data level parallelism due to the separate banks and pipeline
+ * parallelism within a bank. However, this LF implementation does not exploit
+ * the pipeline parallelism. This would require additional logical actions
+ * between the pipeline stages. The effect of this unknown and it might be
```
lhstrh, line 13: As discussed earlier, it may make more sense to use a period and use matching `after` clauses to enable pipeline parallelism. Just use the `--fast` flag to ignore synchronization to physical time.

benchmark/Cpp/Savina/src/parallelism/PiPrecision.lf
```diff
@@ -1,51 +1,25 @@
...
+ * Copyright (C) 2020 TU Dresden
  * 
+ * This benchmark calculates Pi to a very high precision. The workers operate
+ * independently and each worker calculates one digit of the result. Similar to
+ * to the other parallelism benchmarks, the execution is controlled by a central
+ * manager. The overall structure in this benchmark is quite simple. The manager
+ * tells the workers which digit to calculate and collects the results to build Pi.
+ *
+ * Note that the validation isn't precise. Due to rounding errors on the last digit,
+ * it might return false although the result is correct.
```
lhstrh, line 11: I assume the same flag exists in the Akka implementation?

benchmark/Cpp/Savina/src/parallelism/RadixSort.lf
```diff
@@ -0,0 +1,197 @@
...
+            prevValue = 0;
+            errorValueLong = -1;
+            errorValueInt = -1;
+            
+            outFinished.set();
+        }
+    =}
+}
+
+// FIXME We cannot adjust maxValue here because the number of sort actors below depends on it. The parameter should
```
lhstrh, line 176: Implementing this change doesn't strike me as objectionable. How we deal with parameters isn't really a material part of the algorithm/workload.

benchmark/Cpp/Savina/src/parallelism/RadixSort.lf
```diff
@@ -0,0 +1,197 @@
...
+    
+    reaction(startup) {=
+        printBenchmarkInfo("RadixSortBenchmark");
+        printArgs("numIterations", numIterations, "dataSize", dataSize, "maxValue", maxValue, "seed", seed);
+        printSystemInfo();
+    =}
+    
+    sorters = new[60] SortActor(numValues=dataSize);
+
+    source.outNextSortReactor, sorters.outValue -> sorters.inValue, validator.inValue;
```
lhstrh, line 196: This looks like another location where we fail to exploit pipeline parallelism because the connections are all direct feed-through.

benchmark/Cpp/Savina/src/parallelism/Trapezoidal.lf
```diff
@@ -0,0 +1,141 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ *
```
lhstrh, line 3: Missing description.

example/Cpp/src/Patterns/MatrixConnectedRowsAndColumns.lf
```diff
@@ -0,0 +1,75 @@
+// This pattern creates a matrix of nodes, where each of the nodes can send
```
lhstrh, line 1: Style: use multi-line comments.

example/Cpp/src/Patterns/MatrixConnectedRowsAndColumns.lf
```diff
@@ -0,0 +1,75 @@
+// This pattern creates a matrix of nodes, where each of the nodes can send 
+// messages to all other nodes in the same row or in the same column. Since
+// banks in LF are one dimensional, we use hierachy to implement the two
```
lhstrh, line 3:
```suggestion
// banks in LF are one dimensional, we use hierachy to implement the second 
```

benchmark/Cpp/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,215 @@
...
+ * It is also important to note that the workers in this benchmark operate on
+ * shared state (the grid). Each node has an attribute that indicates its
+ * parent. Initially, this attribute is uninitialized, but it is updated when it
+ * is visited by a worker. Thus, the parent attribute can also be used to
+ * determine whether a node was already visited. However, multiple reactions may
+ * access the parent attribute of the same node at the same time. Thus,
+ * similarly to the Akka implementation, the attribute is implemented as an
+ * atomic variable. However, this does not fully eliminate the race condition.
+ * Sometimes one reaction might write first and sometimes another reaction
+ * writes first, effectively introducing none-determinism into the benchmark.
```
lhstrh, line 45:
```suggestion
 * writes first, effectively introducing nondeterminism into the benchmark.
```

benchmark/Cpp/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * The peculiarities of this design also lead to another quirk. The benchmark
+ * does not scale well with the number of threads. A single threaded execution
+ * is much faster than a multi threaded one. Interestingly the same effect can
+ * be observed in the Akka implementation. This indicates that the problem lies
+ * in the benchmark itself, and not (only) in the LF runtime. In fact,
+ * multi-threaded executions have the tendency to perform much more work as
+ * compared to single threaded execution. This is likely to the race-conditions
```
lhstrh, line 53:
```suggestion
 * compared to single threaded execution. This is likely due to the race-conditions
```

benchmark/Cpp/Savina/src/parallelism/MatMul.lf
```diff
@@ -0,0 +1,221 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ * 
+ * This benchmarks implements a parallel matrix multiplication algorithm.
+ * Overall, the benchmark uses a divide and conquer approach that is very
```
lhstrh, line 5:
```suggestion
 * Overall, the benchmark uses a divide-and-conquer approach that is very
```

benchmark/Cpp/Savina/src/parallelism/MatMul.lf
```diff
@@ -0,0 +1,221 @@
...
+ * similar to the N Queens benchmark. The Manager at first produces a WorkItem
+ * that encompasses the whole problem, and the workers then split the work items
+ * into smaller chunks that they send back to the manager. Only if the problem
+ * size is below a threshold, the workers perform an actual multiplication over
+ * a subset of the matrix. When the problem is split, always 8 new work items
+ * are generated.
+ * 
+ * It is important to note that the Savina implementation makes a severe
+ * mistake. Both the operand matrices and the result matrix are stored in shared
+ * memory. Tor the operands, this is not much of an issue since they are
```
lhstrh, line 15:
```suggestion
 * memory. For the operands, this is not much of an issue since they are
```

benchmark/Cpp/Savina/src/parallelism/MatMul.lf
```diff
@@ -0,0 +1,221 @@
...
+ * nondeterministic. The problem is illustrated nicely by the Akka
+ * implementation which, on my machine and in the default configuration, prints
+ * "valid: false". Due to the nature of race conditions, this might be hard to
+ * reproduce on other machines. However, I found that when we change the problem
+ * size, the error appears or disappears sometimes.
+ *
+ * This implementation in LF reproduces the mistake described above. This is to
+ * replicate the same behaviour. If we would only fix the LF variant, this would
+ * make a comparison to the Akka version hard. This is of course less than
+ * ideal...
```
lhstrh, line 28: For the benchmarks that are implemented incorrectly by Savina, like this one, perhaps we should check in a fixed version under a different name (e.g., `MatMulFixed.lf`). Similarly, for inefficient implementations (or implementations that are inefficient for reactors), we could do the same and use a different suffix (e.g., `Optimized` or something like that).

benchmark/Cpp/Savina/src/parallelism/NQueens.lf
```diff
@@ -0,0 +1,281 @@
...
+ * The worker sends all valid solutions back to the manager which then again
+ * distributes the positions to the workers for further processing. Only for the
+ * last `threshold` iterations, the workers directly implement the search to the
+ * end and don't produce new work items.
+ *
+ * This benchmark is very similar to the A* (GuidedSearch) benchmark. An
+ * important difference of this LF implementation compared to the Akka
+ * implementation is the handling of messages send back from the workers to the
+ * manager. In the Akka implementation, each work item produced by the workers is
+ * send as an individual message. This is not easily possible as each port in LF
```
lhstrh, line 27:
```suggestion
 * sent as an individual message. This is not easily possible as each port in LF
```

benchmark/Cpp/Savina/src/parallelism/NQueens.lf
```diff
@@ -0,0 +1,281 @@
+/**
+ * Copyright (C) 2020 TU Dresden
+ * 
+ * This benchmark implements a solution to the N queens problem. In particular,
+ * the benchmark implements a recursive search algorithm that finds all possible
+ * solution. Given N (size), the board has a size of N x N fields and N queens
+ * need to placed. The algorithm starts from an empty board. Then, the first
```
lhstrh, line 7:
```suggestion
 * need to be placed such that no two queens threaten each other.
 * The algorithm starts from an empty board. Then, the first
```

lhstrh (approved):
Terrific work! I left some minor comments/suggestions. One thing that I think will be mighty useful is to put a `README.md` in the `benchmark` directory that explains how to run the benchmarks, plot results, etc.

benchmark/Cpp/Savina/src/parallelism/RadixSort.lf
cmnrd: The problem here is rather the uniformity of parameter sets that the benchmark runner assumes. We could fix it here, but then the Savina benchmark would expect a different set of parameters. Dealing with this is currently not easily doable with our benchmark runner. It might become possible with a few changes though.

benchmark/Cpp/Savina/src/concurrency/SleepingBarber.lf
cmnrd: Good point. I changed that and it even performs faster. I think the problem with physical actions here is that they force sequential operation. It's not a large problem in this case though, since the workload of the factory and the barber is barely noticeable.

benchmark/Cpp/Savina/src/parallelism/PiPrecision.lf
cmnrd: I guess you mean bug? Yes, I think it has the same problem. Actually, many of the Savina validation routines are bogus. Here, I found it is not worth the effort to fix it.

benchmark/Cpp/Savina/src/micro/ThreadRing.lf
cmnrd: Yes, I think this is a relict of the times before we had that connection syntax. I can change it.

benchmark/Cpp/Savina/src/parallelism/PiPrecision.lf
cmnrd: I guess you mean bug? Yes, I think it has the same problem. Actually, many of the Savina validation routines are bogus. Here, I found it is not worth the effort to fix it.

cmnrd:
I haven't addressed all the comments yet, but I will merge this nonetheless to avoid blocking other open PRs and future development. I will continue working on this and open a new PR soon.

lhstrh:
Fine with me, but is there a way to track unresolved conversations after merging? Asking for a friend :-)

cmnrd:
I am not sure. I mean, the conversations do not disappear after merging. So you can always refer back to them. I guess the best is to create a new issue (like I did with #503) and list open topics there to keep track of them.

PR #411, opened 2021-07-15
Discontinue recognizing single quotes as string delimiters in LF

### LINKED ISSUE #407: Target code lexer cannot parse Rust lifetime annotations
The lexer for target code blocks expects quotes to be paired:

https://github.com/icyphy/lingua-franca/blob/0fc05d4fb3e43c60554ef4803fa6df1b57a7c82d/org.lflang/src/org/lflang/LinguaFranca.xtext#L453-L455

because of the STRING terminal, but lifetime annotations in Rust are written `'a` or `'b`.

The following reactor fails to parse:
```rust
target Rust;
main reactor StructAsState {
    preamble {=
        struct Hello {
            name: &'static str,
            value: i32,
        }
    =}
}
```
with
> ERROR:mismatched character `'<EOF>'` expecting `'''` (file:/.../Source.lf line : 5 column : 20)

----
Possible fix: add an alternative for rust lifetimes (which is also the syntax of symbols in Scala so it could be useful). Or maybe, we could stop lexing target code, though I suppose there's a good reason this is done?

lhstrh:
… fixes #407.


org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -389,6 +389,11 @@ terminal SL_COMMENT: ('//' | '#') !('\n'|'\r')* ('\r'? '\n')?;
 @Override
 terminal ML_COMMENT: ('/*' -> '*/') | ('\'\'\'' -> '\'\'\'');
 
+@Override
+terminal STRING:
+            '"' ( '\\' . /* 'b'|'t'|'n'|'f'|'r'|'u'|'"'|"'"|'\\' */ | !('\\'|'"') )* '"'
```
cmnrd, line 394: Why are the escape values (like `\n`) commented out here? Do we really want to allow all possible characters after `\`? If so, then I think the commented out code should be removed. It took me quite a while to understand this rule.

cmnrd, line 394: Why are the escape values (like \n) commented out here? Do we really want to allow all possible characters after \? If so, then I think the commented out code should be removed. It took me quite a while to understand this rule.

test/C/src/concurrent/AsyncCallbackDrop.lf
```diff
@@ -35,7 +35,7 @@ main reactor {
 	state expected_time:time(100 msec);
 	state toggle:bool(false);
 
-    physical action a(100 msec, 100 msec, 'drop'):int;
+    physical action a(100 msec, 100 msec, "drop"):int;
```
cmnrd, line 38: This makes me wonder which consequence the grammar change has in other places. Would `state foo:char('f')` be legal LF code? Or would we need to write `state foo:char({='f'=})`?

cmnrd (commented):
I am wondering if this change is a bit to drastic, because it changes what a string is in the entire LF program. Would it be possible instead just to remove the `STRING` terminal in the `Token` rule and replace it with our custom string rule? Then only the parsing of target code would be affected, but we should still be able to use single quotes for Strings everywhere else.

Soroosh129:
I think the changes to the Python tests show how drastic this fix can be for Python programs.

test/C/src/concurrent/AsyncCallbackDrop.lf
oowekyala: (Since my changes, this comment is outdated, `state foo:char('f')` is supported and tested.)

test/Python/src/ScheduleValue.lf
```diff
@@ -11,7 +11,7 @@ main reactor ScheduleValue {
     reaction(a) {=
         print("Received: ", a.value)
         if a.value != "Hello":
-            sys.stderr.write('FAILURE: Should have received "Hello"\n')
+            sys.stderr.write("FAILURE: Should have received 'Hello'\n")
```
cmnrd, line 14: Here `"'...'" seems to work. Is this correct?

org.lflang.tests/src/org/lflang/tests/LFParsingTest.java
```diff
@@ -13,19 +13,108 @@
...
+                                                 "        \"a\\\n"
+                                                     + "        bcde\"\n"
+        ));
+    }
+
+
+    @Test
+    public void testLexingSquotedString() throws Exception {
+        // we can't do that anymore
+        expectParsingErrorIn(makeLfTargetCode("Python", "a = ' a string '"));
```
cmnrd, line 69: I think not supporting `'` and having to use `"` is not ideal, but something we can live with. But not supporting `"'...'"` seems bad to me. Sometime this is simply needed... . Would `"\'...\'"` work?

oowekyala: it is supported, see your other comment ^^ the `"` you're seeing are in Java, not LF

test/Python/src/ScheduleValue.lf
oowekyala: Yes this works, because the double-quoted string production also consumes `'` fine

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -37,7 +37,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * @author{Edward A. Lee <eal@berkeley.edu>}
  */
 
-grammar org.lflang.LF with org.eclipse.xtext.common.Terminals
```
lhstrh, line 40: I don't really understand why this change was necessary. Is it? Doesn't the override mechanism suffice?

lhstrh:
I agree with @Soroosh129 that the impact of this change on the Python target is significant. I think that, before merging this in, we should:
 - customize errors thrown by the lexer to inform users to use double quotes for strings instead of single ones;
 - check any other code in the repo (benchmarks, experimental code) for single-quote errors; and
 - document in the description of the Python target (as well as the target properties) that strings delimited by single quotes are not allowed.

org.lflang/src/org/lflang/LinguaFranca.xtext
oowekyala: I don't think it's necessary, but I find it clearer to have all terminals be defined in the same place. Otherwise it's not so easy to determine how the lexer behaves. I also was editing the grammar outside of eclipse so looking up the external file was a pain.

We were anyway overriding all of them except INT, ID and WS

oowekyala:
So I checked our sources and found an interesting case where we still crash:

https://github.com/icyphy/lingua-franca/blob/26d29edda93007aa1b8a1a92f5589484a20b7523/experimental/Python/src/Intersection/Intersection.lf#L78-L85

> Single quotes can only be used around single characters, not strings. Please use double quotes instead.

(Btw kudos for the nice messages.)

It looks like we'll should add support for multiline strings.

Also I updated the wiki: https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-Python#the-python-target-specification

There's a :warning: box, though it's unclear to me if it's the best place. I think anyway that people will most likely first discover this limitation through the error message.

org.lflang/src/org/lflang/LinguaFranca.xtext
lhstrh: OK, fair enough :-)

lhstrh:
Thanks, @oowekyala! I'm still working on getting "quick fixes" to work, which are concrete suggestions provided by the IDE for fixing errors. I'll also have a look at the experimental Python code you pointed out above.

oowekyala:
@lhstrh Could we merge this soon? It's blocking the Rust PR. I added support for multiline strings, so now all experimental code should be good to go.

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.xtend
```diff
@@ -816,23 +816,23 @@ class LinguaFrancaValidationTest {
...
     /**
      * Maps a type to a list of known bad values.
      */
     val primitiveTypeToKnownBad = #{
-            PrimitiveType.BOOLEAN -> #["1 sec", "foo", "\"foo\"", "\'bar\'", "[1]", "{baz: 42}"],
-            PrimitiveType.INTEGER -> #["foo", "\"bar\"", "'baz'", "1 sec", "[1, 2]", "{foo: \"bar\"}"],
-            PrimitiveType.NON_NEGATIVE_INTEGER -> #["-42", "foo", "\"bar\"", "'baz'", "1 sec", "[1, 2]", "{foo: \"bar\"}"],
-            PrimitiveType.TIME_VALUE -> #["foo", "\"bar\"", "'baz'", "\"3 sec\"", "'4 weeks'", "[1, 2]", "{foo: \"bar\"}"],
-            PrimitiveType.STRING -> #["1 msec", "[1, 2]", "{foo: \"bar\"}"]
+            PrimitiveType.BOOLEAN -> #["1 sec", "foo", "\"foo\"", "[1]", "{baz: 42}", "'c'"],
```
lhstrh, line 831: Shouldn't `'c'` be legal? (`'cc'` should be illegal)

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -823,14 +823,14 @@ public static void produceError(String name, String description,
             v.getKeyvalue() == null && v.getArray() == null
                 && v.getLiteral() == null && v.getId() == null
                 && (v.getTime() == 0 || v.getUnit() != TimeUnit.NONE)),
-        STRING("a string", v -> v.getLiteral() != null || v.getId() != null),
+        STRING("a string", v -> v.getLiteral() != null && !isCharLiteral(v.getLiteral()) || v.getId() != null),
```
lhstrh, line 826: Why not allow chars here? They can be losslessly converted to strings...

lhstrh (commented):
Looks good generally. The only thing that stands out to me is that we don't allow string target properties to be assigned a char, which I think is overly restrictive. I had progress in the pipeline towards providing automatic suggestions for replacing single-quoted strings with doubly-quoted ones, but those got buried in git weirdness; I'll try to recover those changes and push them to a different branch.

org.lflang/src/org/lflang/TargetProperty.java
oowekyala: They can also losslessly be converted to integers, which is I believe less surprising when you come from a C-like language... 

I think we should keep types clearly separated, esp. when allowing chars here doesn't net us anything. If you allow a character literal to be interpreted as a string, then you're opening the way to some more lexer errors, as someone who wants to modify the literal might forget to change the delimiters. I think it also might be easier to teach "LF doesn't support single quoted literals" if we don't muddy the distinction between char literals and strings.

lhstrh: OK, I think you're right.

oowekyala (approved):

Issue #383, opened 2021-06-29
Fresh oomph install has problems


edwardalee:
After merging pull request #345, I did a fresh oomph install with a fresh git checkout. The result failed to compile. I opened the LinguaFranca.xtext file and regenerated xtext artifacts, then did a clean, and the compile succeeded.  Now I get:
```
git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   org.lflang.ide/.classpath
	modified:   org.lflang.ide/.project
	modified:   org.lflang.targetplatform/org.lflang.targetplatform.target
	modified:   org.lflang.tests/.project
	modified:   org.lflang.web/.classpath
	modified:   org.lflang.web/.project
	modified:   org.lflang.web/.settings/org.eclipse.wst.common.component
	modified:   org.lflang/.project
```
In order to be able to do anything at all, I will need to commit these changes.  What should I do?

Some of the changes are harmless; they just reorder the items.

The org.lflang.targetplatform/org.lflang.targetplatform.target change is problematic. It adds Mac-specific stuff:
```
+      <unit id="org.eclipse.wildwebdeveloper.embedder.node.macos.x86_64" version="0.0.0"/>
```
It also removes some stuff (see below).

The complete diff is below.

```
diff --git a/org.lflang.ide/.classpath b/org.lflang.ide/.classpath
index f2f202a2..baba5b6a 100644
--- a/org.lflang.ide/.classpath
+++ b/org.lflang.ide/.classpath
@@ -12,9 +12,9 @@
 			<attribute name="gradle_used_by_scope" value="main,test"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="src" path="/org.lflang"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11/"/>
 	<classpathentry kind="con" path="org.eclipse.buildship.core.gradleclasspathcontainer"/>
+	<classpathentry kind="src" path="/org.lflang"/>
 	<classpathentry kind="con" path="org.jetbrains.kotlin.core.KOTLIN_CONTAINER"/>
 	<classpathentry kind="output" path="bin/default"/>
 </classpath>
[...]

PR #412, opened 2021-07-16
Oomph: Remove the kotlin plugin from the developer IDE

cmnrd:
This PR removes the kotlin plugin in the oomph setup. This should remove the weird errors that pop up from time to time due to the broken Kotlin plugin. While it's not tested yet (see below), this should allow to normally build the IDE from within the developer Eclipse, only the editor support for the Kotlin files in our code-base would be gone.

So far, I was not able to test this change to the oomph setup. When I do a fresh oomph install from the `no-kotlin-plugin` branch, I get the following error:
```
ERROR: org.eclipse.oomph.resources code=0 org.eclipse.core.runtime.CoreException: Error processing changed links in project description file.
org.eclipse.oomph.resources.backend.BackendException: org.eclipse.core.runtime.CoreException: Error processing changed links in project description file.
```
A bit of googling suggest that the error stems from a read-only .project file, but I am not sure why this might be the case. @a-sr do you have any idea what might be wrong here?


lhstrh (approved):

lhstrh:
I'll go ahead and merge this into `master` because the approach described in #383 gets stuck. If it doesn't work I'll just revert b06c1fa.

PR #416, opened 2021-07-20
Add support for cross connections

### LINKED ISSUE #415: Connection pattern: fully connected and adressable banks
I noticed a pattern that is quite common in the Savina benchmarks suite, but that is not supported by our current connection syntax.  I am not sure if the name in the title captures well what this is about, so let me explain.

Building a fully connected network of reactors in a bank where each instance broadcasts its messages to all other instances in the bank is quite simple. For instance, the following program
```
target Cpp {
    threads: 1
}

reactor Node(bank_index: size_t(0), num_nodes: size_t(4)) {
    input[num_nodes] in: size_t
    output out: size_t
    
    reaction (startup) -> out{=
     	std::cout << "Hello from node " << bank_index << "!\n";
     	// broadcast my ID to everyone
     	out.set(bank_index);    
    =}
    
    reaction (in) {=
        std::cout << "Node " << bank_index << " received messages from ";
     	for (auto& port : in) {
            if (port.is_present()) {
                std::cout << *port.get() << ", ";
            }
        }   
        std::cout << '\n';
    =}
}

main reactor(num_nodes: size_t(4)) {
    nodes = new[num_nodes] Node(num_nodes=num_nodes);
    (nodes.out)+ -> nodes.in;
}
```
will print
``` 
Hello from node 3!
Hello from node 2!
Hello from node 1!
Hello from node 0!
Node 3 received messages from 0, 1, 2, 3, 
Node 2 received messages from 0, 1, 2, 3, 
Node 1 received messages from 0, 1, 2, 3, 
Node 0 received messages from 0, 1, 2, 3, 
```
All the magic lies in the line `(nodes.out)+ -> nodes.in;` which broadcasts the output of each instance, to the inputs of all instances.

So far, so good. But what if we want to make this a little more sophisticated and make the instances addressable, so that the sender does not broadcast to all receivers but can send messages directly to a specific receiver? Naively, we could try the following:
```
target Cpp {
    threads: 1
}

reactor Node(bank_index: size_t(0), num_nodes: size_t(4)) {
    input[num_nodes] in: size_t
    output[num_nodes] out: size_t
    
    reaction (startup) -> out{=
     	std::cout << "Hello from node " << bank_index << "!\n";
     	// send my ID only to my right neighbour
     	out[(bank_index + 1) % num_nodes].set(bank_index);    
    =}
    
    reaction (in) {=
        std::cout << "Node " << bank_index << " received messages from ";
     	for (auto& port : in) {
            if (port.is_present()) {
                std::cout << *port.get() << ", ";
            }
        }   
        std::cout << '\n';
    =}
}

main reactor(num_nodes: size_t(4)) {
    nodes = new[num_nodes] Node(num_nodes=num_nodes);
    nodes.out -> nodes.in;
}
```
In this program, each node is supposed to send a message to its right neighbor (higher `bank_index`). However, the output produced by the program is
```
Hello from node 3!
Hello from node 2!
Hello from node 1!
Hello from node 0!
Node 3 received messages from 3, 
Node 2 received messages from 2, 
Node 1 received messages from 1, 
Node 0 received messages from 0, 
```
This indicates that the nodes actually only send messages to themselves. Considering what `nodes.out -> nodes.in;` does, this is not all too surprising. Expressed in C++ pseudocode, the connection will be created using nested for loops similar to this:
```
for (size_t i = 0; i < num_nodes; i++) {
    for (size_t j = 0; j < num_nodes; j++) {
        connect(nodes[i].out[j], nodes[i].in[j]);
    }
}
```
However, what we actually want is this (notice the swapped indices):
```
for (size_t i = 0; i < num_nodes; i++) {
    for (size_t j = 0; j < num_nodes; j++) {
        connect(nodes[i].out[j], nodes[j].in[i]);
    }
}
```
Now my question: How can we distinguish the two different connection patterns in our syntax? I believe that we want to support both patterns, which means that they should be distinguishable somehow. Maybe we could use `-x->` for "cross connection"?

cmnrd:
This PR adds support for so-called interleaved connections using the `interleaved(b.p)` qualifier that can be optionally added to a port reference `b.p`. Normally, the connection logic would iterate first over banks, then multiports, on both sides of the connection. If the interleaved qualifier is used, this pattern is changed for the specified port and ports are iterated before banks. This allows creating patterns where bank instances can send messages to a specific second instance. See #415 for more details and a usage example.

Closes #415


example/Cpp/src/Patterns/FullyConnected.lf
```diff
@@ -0,0 +1,31 @@
+// In this pattern, each node can send broadcast messages to all other nodes
```
edwardalee, line 1: Suggest looking at the examples in `example/C/src/Patterns` for naming convention and documentation style. The naming convention is intended to lead the reader through the examples in a particular order.

test/Cpp/src/multiport/FullyConnectedAdressable.lf
```diff
@@ -0,0 +1,49 @@
...
+    
+    reaction (in) {=
+        std::cout << "Node " << bank_index << " received messages from ";
+        received = true;
+        size_t count{0};
+        size_t received{0};
+     	for (auto& port : in) {
+            if (port.is_present()) {
+                count++;
+                received = *port.get();
```
edwardalee, line 27: Here, a bool is being assigned a value of type size_t.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -231,7 +231,7 @@ Instantiation:
 Connection:
     ((leftPorts += VarRef (',' leftPorts += VarRef)*)
     | ( '(' leftPorts += VarRef (',' leftPorts += VarRef)* ')' isIterated ?= '+'?))
-    ('->' | physical?='~>') 
+    ('->' | physical?='~>' | cross?='-x->')
```
edwardalee, line 234: Suggest we change the syntax as discussed to be a modifier on VarRef to optionally reference a bank of multiports as `interleaved(b.p)`.

cmnrd:
I pushed some new changes that address @edwardalee's comments. Most importantly, the syntax changed, and we can now use interleaved on port references `nodes.out -> interleaved(nodes.in)`. Note that with this new syntax, we can alternatively also write  `interleaved(nodes.out) -> nodes.in` to achieve the same result.

I extended the tests so that they include both cases, and also added a test that checks the correct operation in conjunction with after.

org.lflang/src/org/lflang/generator/ReactorInstance.xtend
```diff
@@ -147,17 +147,17 @@ class ReactorInstance extends NamedInstance<Instantiation> {
...
-     * is a bank and `m` is a multiport, this function iterates over bank members first,
+     * If a given port reference `b.m`. where `b` is a bank and `m` is a multiport,
+     * is unqualified, this function iterates over bank members first,
      * then ports. E.g., if `b` and `m` have width 2, it returns
      * `[b0.m0, b0.m1, b1.m0, b1.m1]`. 
      * 
-     * If the cross argument is true, then for any port reference `b.m` where `b`
-     * is a bank and `m` is a multiport, this function iterates over ports first,
+     * If a given port reference `b.m`. where `b` is a bank and `m` is a multiport,
+     * is qualified with 'interleaved', this function iterates over ports first,
```
edwardalee, line 153: Suggest:
If a given port reference has the form `interleaved(b.m)`, where `b` is a bank and `m` is a multiport, then this function iterates over ports first, ...

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -1292,6 +1284,23 @@ class LFValidatorImpl extends AbstractLFValidator {
             }
         }
     }
+    
+    @Check(FAST)
+    def checkVarRef(VarRef varRef) {        
+        if (varRef.isInterleaved) {
+            if (this.target != Target.CPP && this.target != Target.C && this.target != Target.Python) {
+                error("This target does not support interleaved port referenced", Literals.VAR_REF__INTERLEAVED)
```
edwardalee, line 1296: referenced -> references

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -1292,6 +1284,23 @@ class LFValidatorImpl extends AbstractLFValidator {
...
+    @Check(FAST)
+    def checkVarRef(VarRef varRef) {        
+        if (varRef.isInterleaved) {
+            if (this.target != Target.CPP && this.target != Target.C && this.target != Target.Python) {
+                error("This target does not support interleaved port referenced", Literals.VAR_REF__INTERLEAVED)    
+            }
+            if (varRef.container === null || varRef.container.widthSpec === null || 
+                (varRef.variable as Port).widthSpec === null
+            ) {
+                error("Interleaved can only be used for multiports contained within banks", Literals.VAR_REF__INTERLEAVED)
```
edwardalee, line 1301: Shouldn't this also check that the port is a multiport?  Also, since `interleaved` is a keyword, I suggest not capitalizing here in the message.

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -1292,6 +1284,23 @@ class LFValidatorImpl extends AbstractLFValidator {
...
+            if (this.target != Target.CPP && this.target != Target.C && this.target != Target.Python) {
+                error("This target does not support interleaved port referenced", Literals.VAR_REF__INTERLEAVED)    
+            }
+            if (varRef.container === null || varRef.container.widthSpec === null || 
+                (varRef.variable as Port).widthSpec === null
+            ) {
+                error("Interleaved can only be used for multiports contained within banks", Literals.VAR_REF__INTERLEAVED)            
+            }
+            if (!(varRef.eContainer instanceof Connection)) {
+                error("Interleaved can only be used in connections", Literals.VAR_REF__INTERLEAVED)
```
edwardalee, line 1304: Interleaved -> interleaved

test/Cpp/src/multiport/FullyConnectedAdressableAfter.lf
```diff
@@ -0,0 +1,15 @@
+// In this pattern, each node can send direct messages to individual other nodes
```
edwardalee, line 1: The file name has a spelling error: FullyConnectedAdressableAfter -> FullyConnectedAddressableAfter

edwardalee (approved):
Looks great!  Minor comments and suggestions added.

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
cmnrd: It does check if the port is a multiport in the second line of the if (line 1295).

Thank you for the thorough review by the way!

cmnrd:
@edwardalee I added a description of this feature to the bottom of [Multiports and Banks of Reactors](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors). Could you check this description and see if it is understandable?

edwardalee:
Nice! I added to [Multiports and Banks of Reactors](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors) an explanation of the addressing scheme that results.

Issue #320, opened 2021-04-14
Add CI tests for scripts in /bin


cmnrd:
With the updated test framework our compiler tests are run directly from java as unit tests. While this brings many advantages, there is one downside: we do not directly test our scripts in /bin anymore. We should have tests that check the correct behavior of the scripts on all supported platforms. For `lfc`, such a test could simply compile the hello world test. The tests should also include various command line options. For `lfc`, I think the rebuild mechanism would be of particular interest because I had trouble with this mechanism while porting to Windows.

PR #417, opened 2021-07-20
Redesign command handling

### LINKED ISSUE #361: IDE complains about not finding `bash` in Windows
Saw this on a clean Windows install on a machine that doesn't have the appropriate dev tools for the used target. Instead of complaining about `bash`, it should complain about the actually required dependencies that are missing.

cmnrd:
This PR factors the handling of commands out of `GeneratorBase`. At the same time, the code is significantly simplified and generalized. Basically, this PR introduces the class `LFCommand` as our own custom abstraction of commands. Note, that this abstraction is more general than the old `createCommand` method we used in `GeneratorBase`. This allows a more flexible use of commands outside the generators (for instance in the test framework). An additional factory class `GeneratorCommandFactory` can be used to create more specialized commands for use within the code generators.

Closes #361

### DIFF
--- a/org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
+++ b/org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
@@ -1,0 +1,144 @@
+/*************
+ Copyright (c) 2021 TU Dresden
+
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+
+ 1. Redistributions of source code must retain the above copyright notice,
+ this list of conditions and the following disclaimer.
+
+ 2. Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following disclaimer in the documentation
+ and/or other materials provided with the distribution.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ ***************/
+
+package org.lflang.generator;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
+import org.lflang.ErrorReporter;
+import org.lflang.FileConfig;
+import org.lflang.util.LFCommand;
+
+/**
+ * A factory class responsible for creating commands for use by the LF code generators.
+ * <p>
+ * In addition to the basic functionality of LFCommand, this class additionally ensures that error messages (or
+ * optionally warnings) are shown when a command is not found and that certain environment variables are set (see
+ * {@link #createCommand(String, List, Path, Boolean)} createCommand}).
+ */
+public class GeneratorCommandFactory {
+
+    protected final ErrorReporter errorReporter;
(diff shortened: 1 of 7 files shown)


org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
```diff
@@ -0,0 +1,144 @@
...
+     * @see LFCommand#get(String, List, Path)
+     */
+    public LFCommand createCommand(String cmd, List<String> args, Path dir, Boolean failOnError) {
+        assert cmd != null && args != null;
+        if (dir == null) {
+            dir = Paths.get("");
+        }
+        LFCommand command = LFCommand.get(cmd, args, dir);
+        if (command != null) {
+            command.setEnvironmentVariable("LF_CURRENT_WORKING_DIRECTORY", dir.toString());
```
lhstrh, line 126: Since environment variables are likely to be used in scripts, I would probably prefer shorter names. How about `LF_CWD`, `LF_SRC`, `LF_SRC_GEN`, and `LF_BIN`?

org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public GeneratorCommandFactory(ErrorReporter errorReporter, FileConfig fileConfig) {
+        assert errorReporter != null && fileConfig != null;
+        this.errorReporter = errorReporter;
+        this.fileConfig = fileConfig;
+    }
+
+
+    /**
+     * Create a LFCommand instance form a given command and an argument list.
```
edwardalee, line 59: "form" -> "from"

org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
```diff
@@ -0,0 +1,144 @@
...
+     * @return an LFCommand object or null if the command could not be found
+     * @see #createCommand(String, List, Path, Boolean)
+     */
+    public LFCommand createCommand(String cmd, List<String> args) {
+        return createCommand(cmd, args, null, true);
+    }
+
+
+    /**
+     * Create a LFCommand instance form a given command, an argument list and a directory.
```
edwardalee, line 74: "form" -> "from"
Also, this error is repeated throughout the file.

org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -0,0 +1,312 @@
...
+     * lead to a deadlock as the subprocess blocks when output or error
+     * buffers are full. This method ensures that output and error messages
+     * are continuously read and forwards them to the system output and
+     * error streams as well as to the output and error streams hold in
+     * this class.
+     *
+     * @return the process' return code
+     * @author {Christian Menard <christian.menard@tu-dresden.de}
+     */
+    public Integer run() {
```
lhstrh, line 152: We have very similar code in the `org.lflang.test` package that is used to execute tests and capture their output. We should probably update that code and leverage this class instead...

org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -0,0 +1,312 @@
...
+        } catch (InterruptedException | IOException e) {
+            e.printStackTrace();
+            return false;
+        }
+    }
+
+
+    private static Boolean checkIfCommandIsExecutableWithBash(final String command, final Path dir) {
+        // check first if bash is installed
+        if (!checkIfCommandIsOnPath("bash", dir)) {
```
lhstrh, line 297: This should fix #361.

lhstrh (approved):
This is a huge improvement! (And another step toward a smaller `GeneratorBase`...) 🚀🚀🚀

edwardalee (approved):
Looks good to me.  One minor comment: The copyright line should probably read "TU Dresden and UC Berkeley" because much of the content come from previous files with a Berkeley copyright.

org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
Soroosh129: I think the CMake convention at least is to spell out the entire thing, e.g., [CMAKE_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_SOURCE_DIR.html#variable:CMAKE_SOURCE_DIR). I like `LF_SRC` better myself personally but the fully spelled out environment variables are perhaps more familiar in a CMake context.

Soroosh129 (approved):
These changes looks great! This will considerably decouple the CCMakeCompiler class from GeneratorBase in the `c-cmake` branch. If it is okay with you, I will merge these changes into that branch, which means this should be merged before #402.

lhstrh:
Go for it, @Soroosh129.

org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
lhstrh: Ah, yes, that makes sense.

org.lflang/src/org/lflang/util/LFCommand.java
cmnrd: Absolutely! I opened #418 to keep track of this. I was planning to use the new command infrastructure for implementing #320 anyway, and we can update the other tests in the same go.

org.lflang/src/org/lflang/generator/GeneratorCommandFactory.java
cmnrd: The wonders of copy paste... Thanks for pointing it out!

cmnrd:
Good point about the copyright notice. I updated it accordingly.

PR #420, opened 2021-07-21
Allow width to be specified in target code

cmnrd:
This PR extends the grammar slightly so that target code enclosed in `{==}` can be used in width specifications. For instance:
```
reactor Foo(a: size_t{8}, b: size_t{2}) {
    input[{=a*b=}] in: size_t;
    output[{=a/b=}] out: size_t;
}
```
This feature is really handy for some of the more complex connection patterns found in the Savina suite.

This feature is currently only supported in the C++ target. The validator will display an error message if this feature is used in any other target.

### DIFF
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -344,7 +344,8 @@ WidthSpec:
 WidthTerm:
     width=INT 
     | parameter=[Parameter] 
-    | 'widthof(' port=VarRef ')';
+    | 'widthof(' port=VarRef ')'
+    | code=Code;
 
 IPV4Host:
     (user=Kebab '@')? addr=IPV4Addr (':' port=INT)?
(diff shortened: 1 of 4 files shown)


edwardalee (approved):
I don't think this will ever be supportable in the C target and it disables a few static checks, but if you find it valuable, I'm OK with including it.  This should be described in the Wiki, perhaps in the Cpp page rather than language specification (?).

Soroosh129 (approved):
This change looks good to me as well. Maybe the TypeScript target can also utilize this feature?

cmnrd:
I think it should be documented in the language specification, even if it is only supported in C++ (for now). I will add a description before merging this. BTW, this PR does not really add a new feature, since we can use parameters with target code assignments to assign target code to widths. This PR only makes it more convenient by eliminating the need for creating an additional parameter.

cmnrd:
I added a note to the [Multiports and Banks](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors) wiki page.

PR #424, opened 2021-07-23
Made bank_index a proper parameter.

edwardalee:
This PR makes bank_index a parameter in the C and Python targets. If you define a parameter named `bank_index` in a reactor and then instantiate that reactor in a bank, the parameter will be automatically set to the bank index. This allows passing the parameter to contained reactors, something that was not possible before.

I believe this also brings the C and Python target into a consistent state with the Cpp target. However, I'm not sure. The bank_index parameter usage does not seem to be documented for Cpp, so I'm not sure how it work.

When we merge this PR with master, which I hope will be soon, we will need to update the [multiports and banks](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors) documentation.

### DIFF
--- a/test/C/src/concurrent/Tracing.lf
+++ b/test/C/src/concurrent/Tracing.lf
@@ -14,7 +14,9 @@ reactor Source {
 		self->s++;
 	=}
 }
-reactor TakeTime {
+reactor TakeTime(
+    bank_index:int(0)
+) {
 	input in:int;
 	output out:int;
 	state event:string("No ID");
--- a/org.lflang/src/org/lflang/generator/CGenerator.xtend
+++ b/org.lflang/src/org/lflang/generator/CGenerator.xtend
@@ -1763,9 +1763,6 @@ class CGenerator extends GeneratorBase {
         // Extensions can add functionality to the CGenerator
         generateSelfStructExtension(body, decl, federate, constructorCode, destructorCode)
         
-        // Handle bank_index
-        pr(body, '''«targetBankIndexType» «targetBankIndex»;''');    
-        
         // Next handle parameters.
         generateParametersForReactor(body, reactor)
         
@@ -2060,12 +2057,6 @@ class CGenerator extends GeneratorBase {
      */
     def generateParametersForReactor(StringBuilder builder, Reactor reactor) {
         for (parameter : reactor.allParameters) {
-            // Check for targetBankIndex
-            // FIXME: for now throw a reserved error
-            if (parameter.name.equals(targetBankIndex)) {
-                errorReporter.reportError('''«targetBankIndex» is reserved.''')
-            }
-
             prSourceLineNumber(builder, parameter)
             pr(builder, parameter.getInferredType.targetType + ' ' + parameter.name + ';');
         }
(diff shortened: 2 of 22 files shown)


test/C/src/concurrent/Tracing.lf
```diff
@@ -14,7 +14,9 @@ reactor Source {
 		self->s++;
 	=}
 }
-reactor TakeTime {
+reactor TakeTime(
+    bank_index:int(0)
```
Soroosh129, line 18: Does having a default value (other than 0) make sense here? 
Maybe it could be used to indicate the starting index of the `bank_index` in user code.
For example, for the following instantiation:
```
    t = new[4] TakeTime(bank_index=4);
```
the bank indexes will be `{4, 5, 6, 7}`.

To a lesser extent, I think the type might be problematic. A reactor such as
```
reactor Foo(
    bank_index:char*("1")
)
```
is a problem since it cannot be used as a bank. Perhaps, the validator can check this and disallow a type for bank_index altogether?

edwardalee: My reasoning is that there is no reason to prevent a user from creating a parameter named `bank_index` if the reactor is never used in a bank.  If it is used in a bank, and the type cannot accept an integer value, then the generated code will fail to compile. I think it would be difficult to check all the cases in the validator because validity may depend on the target language.

Soroosh129: I think at least in Python, we would not get a compile error. Most likely, the user code will be written expecting that `bank_index` is of a certain type. If the reactor is then later used in a bank, the `bank_index` will become a new type (int), which might or might not result in a runtime error (or an unexpected behavior).

I think the safest solution is to disallow an initial value and type for `bank_index` for all targets. The produced validator error can then explain that `bank_index` is reserved for banks.

edwardalee: In that case, I think that a better design would be for bank_index to always automatically be a parameter. The user would never need to declare it.  @cmnrd: How does it work in Cpp?  Does bank_index need to be declared as parameter to be used?  What if it is given the wrong type?

cmnrd: Yes, in C++ `bank_index` needs to be declared by the user. I think that this has the advantage that the user can freely choose any integer compatible type. If an incompatible type is specified, the target compiler will complain. This is exactly the same behavior as for 'normal' parameters. Since Python is not type checked and things can go wrong for any parameter, I don't think this is much of a problem. We just need to document it well and tell users not to use `bank_index` in other contexts.

I think that disallowing initial values and parameters for the `bank_index` is a bad idea. Treating it differently would just cause confusion for many users. Also I am not sure how this would solve the issue in Python, as even without an initializer you can explicitly pass any object to `bank_index` when doing a normal instantiation.

cmnrd (approved):
Looks good to me! Yes, this makes Python and C consistent with the current C++ implementation.

edwardalee:
I have also updated the [wiki page](https://github.com/icyphy/lingua-franca/wiki/Multiports-and-Banks-of-Reactors) that documents `bank_index`.

PR #427, opened 2021-07-24
Cycle detection

edwardalee:
This branch fixes a bug where reactions that are created via an AST transformation and then marked unordered were not in fact being treated as unordered.  It adds a test.  Unordered reactions are used in federated execution only, currently, although in the long run we may want to add the possibility for users to define `unordered reaction`s.  The only use case I can think of is to get parallel execution within a reactor, which would presumably only be interesting in circumstances where nondeterministic access to shared state is known to be acceptable.  But this fix is more modest and is needed for nontrivial federations to work.


org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
```diff
@@ -123,12 +125,13 @@ class ReactionInstanceGraph extends DirectedGraph<ReactionInstance> {
      * Rather than establishing a total order, we establish a partial order.
      * In this order, the level of each reaction is the least upper bound of
      * the levels of the reactions it depends on.
-     * If any cycles are present in the dependency graph, an exception is
-     * thrown. This method should be called only on the top-level (main) reactor.
+     * If any cycles are present in the dependency graph, then a graph
+     * containing the nodes in the cycle is returned. Otherwise, null is
+     * returned. This method should be called only on the top-level (main) reactor.
```
lhstrh, line 130: Since this function has no arguments:
```suggestion
     * returned.
```

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
```diff
@@ -123,12 +125,13 @@ class ReactionInstanceGraph extends DirectedGraph<ReactionInstance> {
...
      * @return true if the assignment was successful, false if it was not,
```
lhstrh, line 131: This description is no longer valid.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
```diff
@@ -69,8 +69,10 @@ class ReactionInstanceGraph extends DirectedGraph<ReactionInstance> {
         addNodesAndEdges(main)
             // Assign a level to each reaction. 
             // If there are cycles present in the graph, it will be detected here.
-            if (!assignLevels()) {
+            val leftoverReactions = assignLevels()
+            if (leftoverReactions !== null) {
                 main.reporter.reportError("Reactions form a cycle!");
+                System.err.println(leftoverReactions.toString());
```
lhstrh, line 75: We have a much more sophisticated reporting of cycles in the validator, and I would expect this code not to be executed at all if cycles were found during validation. Is the concern that AST transformations would introduce cycles, and that's why we're reporting them here?

lhstrh: It would surprise me if it would be an "expected" error when an AST transformation to introduces a cycle. In other words, if we implement AST transformations correctly, my guess is that the user should never see this message.

lhstrh (changes requested):
Left a few questions/comments.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
edwardalee: Done. Good catch.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
edwardalee: Correct.  This change was prompted by the bug that this PR fixes, where cycles are introduced in AST transformations.  I've added a comment to this effect.

edwardalee:
I am merging this PR with master because I need it to make any progress and if I instead merge it with the various branches I am working on, we will end up with a very complicated precedence order for merging.

PR #430, opened 2021-07-25
Fix sudo permissions

Feliix42:
Add `sudo` to the nightly workflow. That was not required in the local build (probably a quirk of the `act` tool) but is necessary here.

Also sorry for the PR for something small like this, but `master` is protected. Which is good practice but always annoying when doing tinkering with CI. 😅


Soroosh129 (approved):

Soroosh129:
These changes look good to me.

PR #431, opened 2021-07-26
Port TypeScript code generator to Kotlin.

hokeun:


org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -100,6 +101,43 @@ private FileConfig createCppFileConfig(Resource resource,
         }
     }
 
+    /**
+     * Create a TypeScript specific FileConfig object
+     *
+     * Since the CppFileConfig class is implemented in Kotlin, the class is is
```
lhstrh, line 107: The comments still reference `CppFileConfig`

org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -169,6 +208,49 @@ private GeneratorBase createCppGenerator(FileConfig fileConfig,
...
+     * Since the Typeenerator class is implemented in Kotlin, the class is
+     * not visible from all contexts. If the RCA is run from within Eclipse via
+     * "Run as Eclipse Application", the Kotlin classes are unfortunately not
+     * available at runtime due to bugs in the Eclipse Kotlin plugin. (See
+     * https://stackoverflow.com/questions/68095816/is-ist-possible-to-build-mixed-kotlin-and-java-applications-with-a-recent-eclips)
+     * In this case, the method returns null
+     *
+     * @return A TypeScriptGenerator object if the class can be found
+     */
+    private GeneratorBase createTsGenerator(FileConfig fileConfig,
```
lhstrh, line 223: `CppGenerator` has a very similar function with duplicated logic. Can we have a generic `createGenerator` method instead?

org.lflang/src/org/lflang/generator/ts/TsFileConfig.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.emf.ecore.resource.Resource
+import org.eclipse.xtext.generator.IFileSystemAccess2
+import org.eclipse.xtext.generator.IGeneratorContext
+import org.lflang.FileConfig
+import java.io.IOException
+import java.nio.file.Path
+
+/** Generator for TypeScript target.
+ *
+ *  @author {Hokeun Kim <hokeunkim@berkeley.edu>}
```
lhstrh, line 37: While porting to Kotlin we should inherit authors from the classes we lift code from.

org.lflang/src/org/lflang/generator/ts/TsFileConfig.kt
```diff
@@ -0,0 +1,61 @@
...
+package org.lflang.generator.ts
+
+import org.eclipse.emf.ecore.resource.Resource
+import org.eclipse.xtext.generator.IFileSystemAccess2
+import org.eclipse.xtext.generator.IGeneratorContext
+import org.lflang.FileConfig
+import java.io.IOException
+import java.nio.file.Path
+
+/** Generator for TypeScript target.
```
lhstrh, line 35: Please write multi-line comments like:
```
/**
 * Some comment...
 */
```

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+import org.lflang.Target
+import org.lflang.generator.FederateInstance
+import org.lflang.generator.GeneratorBase
+import org.lflang.generator.PrependOperator
+import org.lflang.lf.*
+import org.lflang.scoping.LFGlobalScopeProvider
+import java.lang.StringBuilder
+import java.util.LinkedList
+
+/** Generator for TypeScript target.
```
lhstrh, line 42: Newline after `/**` here also

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+    fun generateVarRefW(reference: VarRef): String =generateVarRef(reference)
+
+    /** Generate TypeScript code from the Lingua Franca model contained by the
+     *  specified resource. This is the main entry point for code
+     *  generation.
+     *  @param resource The resource containing the source code.
+     *  @param fsa The file system access (used to write the result).
+     *  @param context FIXME: Undocumented argument. No idea what this is.
+     */
+    override fun doGenerate(resource: Resource, fsa: IFileSystemAccess2,
```
lhstrh, line 109: This is OK as a first port, but we should definitely break this down into smaller parts.

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+        refreshProject()
+
+        // Invoke the protocol buffers compiler on all .proto files in the project directory
+        // Assumes protoc compiler has been installed on this machine
+
+        // First test if the project directory contains any .proto files
+        if (targetConfig.protoFiles.size != 0) {
+            // For more info, see: https://www.npmjs.com/package/ts-protoc-gen
+
+            // FIXME: Check whether protoc is installed and provides hints how to install if it cannot be found.
```
lhstrh, line 227: We should start use @cmnrd's `LFCommand` class here.

org.lflang/src/org/lflang/generator/ts/TsReactorGenerator.kt
```diff
@@ -0,0 +1,710 @@
...
+
+    private fun generateArg(v: VarRef): String {
+        return if (v.container !== null) {
+            """__${v.container.name}_${v.variable.name}"""
+        } else {
+            """__${v.variable.name}"""
+        }
+    }
+
+    fun generateReactorFederated(reactor: Reactor, federate: FederateInstance) {
```
lhstrh, line 121: This method, too, is way too long. I know it was already like that, but let's do something about it soon :-)

lhstrh (approved):
Great progress!

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+                            ". Using default configuration."
+                )
+                copyFileFromClassPath("/lib/TS/$configFile", configFileDest)
+            }
+        }
+
+        for (federate in federates) {
+            var tsFileName = fileConfig.name
+            if (isFederated) {
+                tsFileName += '_' + federate.name
```
Soroosh129, line 154: There is a [FedFileConfig](https://github.com/icyphy/lingua-franca/blob/ca8602ad799de719d111586c292fc880eecbe867/org.lflang/src/org/lflang/federated/FedFileConfig.java) class that might be useful here. [Here](https://github.com/icyphy/lingua-franca/blob/ca8602ad799de719d111586c292fc880eecbe867/org.lflang/src/org/lflang/generator/c/CGenerator.xtend#L462) is an example of its use in the C target. The behavior in C target is to put each federate into its own folder.

org.lflang/src/org/lflang/generator/LFGenerator.java
hokeun: Great suggestion. I merged the functions for CPP and TS. I named it createKotlinGenerator just to differentiate it from other generators. Let me know if this makes sense. :)

org.lflang/src/org/lflang/generator/ts/TsFileConfig.kt
hokeun: Thanks, done.

org.lflang/src/org/lflang/generator/ts/TsFileConfig.kt
hokeun: Got it, done.

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
hokeun: Done.

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
hokeun: Agreed! I added a TODO here to make sure that I follow up.

org.lflang/src/org/lflang/generator/ts/TsReactorGenerator.kt
hokeun: Same as above, I added a TODO here to make sure that I follow up.

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
hokeun: I'm not sure if I understood your comment correctly. It looks like we're already using LFCommand (created by commandFactory.createCommand() below). Could you clarify a bit?

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
hokeun: Thanks for the suggestion. I believe the federated execution is not supported yet in TypeScript and I'm planning to fix that later. I added a TODO to follow up. Would this be ok with you for now?

org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -110,60 +112,73 @@ private GeneratorBase createGenerator(Target target, FileConfig fileConfig,
...
-        // appear as an import. Instead we look the class up at runtime and
-        // instantiate it if found.
+        // play a few tricks here so that Kotlin FileConfig and
+        // Kotlin Generator do not appear as an import. Instead we look the
+        // class up at runtime and instantiate it if found.
+        String packageName;
+        String classNamePrefix;
+        String targetName;
+        switch (target) {
+            case CPP: {
```
lhstrh, line 146: If I'm not mistaken this info can be read straight from the `Target` enum. If not, it probably makes sense to store it there.

org.lflang/src/org/lflang/generator/ts/TsGenerator.kt
lhstrh: Oh, you're right. In that case the `FIXME` on that line must be obsolete.

org.lflang/src/org/lflang/generator/LFGenerator.java
lhstrh: Yes, makes sense!

org.lflang/src/org/lflang/generator/LFGenerator.java
hokeun: Thanks, updated the comments.

org.lflang/src/org/lflang/generator/LFGenerator.java
hokeun: I removed targetName, which is for the error messages.

I could toLowerCase and some string operations for package name and class name prefix; however, I think explicitly assigning the values makes the code more readable, and I'm afraid that for other languages, the relationship between enum and these names can be arbitrary.

Would this make sense to you?

lhstrh: I think it's better to have fewer strings floating around and strive for uniformity. For the package name, `toLower` of the `name()` sounds appropriate. When it comes to the class name prefix, I guess the `toString()` could work? We have both `TypeScript` and `TS` floating around in the code base, and the Kotlin classes now use `Ts`; this is a perfect example of lack of uniformity that we can solve by adopting some conventions; the `Target` enum is the place to encode them.

hokeun: Got it. I did that. Thanks for the explanation.

hokeun:
Thanks, Marten and Soroush, for reviewing my pull request! I'll be merging this pull request to master as it seems that your comments are all either addressed or marked to be followed up. Let me know if you still have comments. I'll be happy to address them soon!

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+        if (port.type !== null) {
+            return getTargetType(port.type)
+        } else {
+            return "Present"
+        }
+    }
+
+    private fun generateArg(v: VarRef): String {
+        return if (v.container !== null) {
+            """__${v.container.name}_${v.variable.name}"""
```
oowekyala, line 119: Using `"""` where you don't need it should be avoided. We should confine it to multi-line strings, and strings where you would need many escaped characters

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+    private fun getPortType(port: Port): String {
+        if (port.type !== null) {
+            return getTargetType(port.type)
+        } else {
+            return "Present"
+        }
+    }
+
+    private fun generateArg(v: VarRef): String {
+        return if (v.container !== null) {
```
oowekyala, line 118: Using `!==` or `===` is not idiomatic in kotlin. This causes compiler warnings because you sometimes do it with primitive types (`code === 0`), which causes boxing before the comparison. Identity comparison of boxed primitives is deprecated because of planned future extensions to the JVM type system.

Sticking to  `!=` and `==` is better practice.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+
+        var reactorName = reactor.name
+        if (!reactor.typeParms.isEmpty()) {
+            reactorName +=
+                reactor.typeParms.joinToString(", ", "<", ">") { it.toText() }
+        }
+        // NOTE: type parameters that are referenced in ports or actions must extend
+        // Present in order for the program to type check.
+        if (reactor.isMain()) {
+            pr("class " + reactorName + " extends __App {")
```
oowekyala, line 143: This would be a good place to use a string template:
```kotlin
            pr("class $reactorName extends __App {")
```

I tend to systematically prefer templates when the templated variable is a single identifier (you don't need braces like in `${ident.something}`)

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+        } else {
+            arguments.add("parent: __Reactor")
+        }
+
+        // For TS, parameters are arguments of the class constructor.
+        for (parameter in reactor.parameters) {
+            arguments.add(initializeParameter(parameter))
+        }
+
+        var reactorConstructor = StringBuilder()
```
oowekyala, line 166: Nearly all of the `var` in this file could be `val` instead

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+
+            pr(reactorConstructor, "this." + childReactor.getName()
+                    + " = new " + childReactor.reactorClass.name +
+                    "(" + childReactorArguments + ")" )
+        }
+
+        // Next handle timers.
+        for (timer in reactor.timers) {
+            var timerPeriod: String
+            if (timer.period === null) {
```
oowekyala, line 239: This is a weird way to put it. You don't need a `var`. You could initialize the variable like so:
```kotlin
            val timerPeriod: String =
                if (timer.period == null) "0" 
                else getTargetValue(timer.period)
```
or equivalently
```kotlin
                val timerPeriod = timer.period?.let { getTargetValue(it) } ?: "0"
```

Same for the other vars.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
```diff
@@ -0,0 +1,715 @@
...
+    // Wrapper functions and their helpers
+    private fun indent(builder: StringBuilder) = tsGenerator.indentw(builder)
+    private fun unindent(builder: StringBuilder) = tsGenerator.unindentw(builder)
+    private fun indent() = indent(code)
+    private fun unindent() = unindent(code)
+
+    private fun pr(builder: StringBuilder, text: Any) = tsGenerator.prw(builder, text)
+    private fun pr(text: Any) = tsGenerator.prw(code, text)
+
+    private fun getTargetValue(v: Value): String = tsGenerator.getTargetValueW(v)
```
oowekyala, line 44: It would be more idiomatic to use extensions for those methods. You can write
```kotlin
 private fun Value.getTargetValue(): String = tsGenerator.getTargetValueW(v)
```

With this style, in my above comment,
 
```kotlin
                val timerPeriod = timer.period?.let { getTargetValue(it) } ?: "0"
```
can be rewritten to 

```kotlin
                val timerPeriod = timer.period?.getTargetValue() ?: "0"
```

oowekyala (commented):
Just a few minor comments about Kotlin code style and idioms. Most of my comments can be fixed automatically in Intellij, for instance:

Capture d’écran de 2021-07-27 14-47-08

This is why I think it's cheap to fix them to have a more idiomatic code style.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
hokeun: Thank you, Clément, for the comments! I didn't get a chance to get to this yet, but this is certainly on my radar. I will address your comments later.

hokeun: Thanks, done for all places I found (in pull request #444 )

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
hokeun: Thanks, done.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
hokeun: Done.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
hokeun: Done.

org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt
hokeun: I tried as you commented, but it doesn't compile. Maybe I'm missing something?

oowekyala: Look at the edit to my comment. If that doesn't work please post the compiler error

hokeun: Thanks, done in https://github.com/icyphy/lingua-franca/pull/444

PR #432, opened 2021-07-26
Fix Nightly CI run and fix the git tag issue

Feliix42:
This PR should finally fix the build problems and the deployment in form of a nightly release. After a good hint from @lhstrh I switched the release action and used [this one](https://github.com/marvinpinto/action-automatic-releases). We'll see if that triggers nightly mails about a new release or not.

I also fixed the maven build, another bintray reference was hidden there.

We can discuss whether we should add a README to the `lfc` archive, but apart from that, everything should be in working order. :crossed_fingers: 

I tested this PR in my own fork [here](https://github.com/Feliix42/lingua-franca/runs/3160235942), with the result [here](https://github.com/Feliix42/lingua-franca/releases/tag/nightly).


.github/scripts/package_lfc.sh
```diff
@@ -16,14 +16,15 @@ mkdir -p $outname/lib
 mv org.lflang/build/libs/org.lflang-*-SNAPSHOT-all.jar $outname/lib
 
 # Move & patch wrappers
-sed -e '/^lfbase=/d' -e 's/\${lfbase}build\/libs/\${base}lib/g' -e '/^if \[\[ ! -f "\$jarpath" \]\]; then/{n;N;N;d}' -e '/^if \[\[ ! -f "\$jarpath" \]\]; then/a \
+sed -e '/^lfbase=/d' -e 's/\${lfbase}build\/libs/\${base}lib/g' -e '/^if \[\[ ! -f "\$jarpath" \]\]; then/{n;N;N;d}' bin/lfc > $outname/bin/lfc
```
cmnrd, line 19: Why exactly is this patching needed? I am wondering if we should make the lfc script itself a bit more flexible, so that it also works in conjunction with the nightly builds. The patching seems a bit fragile and updating the lfc scripts could break our nighty builds

cmnrd (approved):
Looks good! I am a bit concerned about the patching of lfc scripts, but this we could also address in a separate PR.

.github/scripts/package_lfc.sh
Feliix42: This patching is done so that the scripts are the same as the ones delivered with the alpha release. The reasoning is that you don't want the gradle invocation in there since we ship a finished `jar`. But I agree, it's fragile. Even if you include separate scripts for the nightly things may break if we change the original scripts.

cmnrd: Ah, right. That makes sense. Maybe we should just remove gradle from the lfc script altogether? Then the patched version here would be the default. I don't think it is asked too much to run gradle first before running `bin/lfc` the first time. @lhstrh What do you think?

As a replacement of the old functionality, we also have the new `runLfc` gradle task.

Feliix42: Regardless what the outcome of this discussion is, I'm gonna merge the PR for now. If a README in the `lfc` zip file or a change to the `bin/lfc` files is required, we can do that separately, I want to see whether the build succeeds tonight. :)

PR #435, opened 2021-07-27
Various cleanups

oowekyala:
These are small cleanups that I extracted from my rust-dev feature branch because they're not directly relevant to it.
- LFCommand now performs better checking of the "working directory" parameter before it hands it off to a ProcessBuilder. 
- Handling of invalid target name is improved. For instance if you write `target crust` or some other nonsense you'll get a validator failure instead of an NPE with no message.
  -  `Target.forName` now returns an optional to make it clear that the target may be invalid (esp. in the validator itself)
- Many unnecessary usages of primitive wrappers are replaced with their primitive counterpart. This turns `Boolean` into `boolean`, `Integer` into `int`. The problem with using those capitalized types (which are reference types) is that they may be null, and the dereference is implicit in many contexts (and may NPE). So unless we actually need the null value in the value space, and check for it, this is a time bomb. (Note that sometimes instead of using `Integer` to get a nullable `int` you can also use `OptionalInt`, same for some other primitive types)


org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -260,8 +264,9 @@ public static LFCommand get(final String cmd, final List<String> args) {
      * @param dir  The directory in which the command should be executed
      * @return Returns an LFCommand if the given command could be found or null otherwise.
      */
-    public static LFCommand get(final String cmd, final List<String> args, final Path dir) {
+    public static LFCommand get(final String cmd, final List<String> args, Path dir) {
         assert cmd != null && args != null && dir != null;
+        dir = toProperDirectory(dir);
```
cmnrd, line 269: I am not sure if I like this change. I definitely agree that a non-existent directory is a corner case that should be handled somehow, and I overlooked this. However, I find it a bit strange to have this side effect of creating a directory on a function that actually is only supposed to invoke a command. I think the caller should be responsible for ensuring that the directory exists. If it does not, then this is an error, which should probably result in an exception.

cmnrd (commented):
Thanks for those cleanups! This mostly looks good to me. I only left a comment regarding the handling of directories in LFCommand.

org.lflang/src/org/lflang/util/LFCommand.java
oowekyala: Yes I agree that if the parameter doesn't exist, it should be considered invalid. However I'm not sure how the caller can guard against this any other way than just writing `Files.createDirectories(dir);` before the call... But, maybe it's clearer anyway if it's written explicitly in the caller, and it's then something desirable

cmnrd (approved):
Looks good!

oowekyala:
Thanks for the review @cmnrd !

oowekyala:
I'll merge this today after code review if no one opposes it

Issue #433, opened 2021-07-26
Epoch and lfc become unusable for programs with large (>500) banks and multiports


cmnrd:
While experimenting with the banking benchmark, which uses by default 1000 instances of a reactor, I noticed that the IDE becomes unusable for such a large number of reactor instances. The IDE simply hangs and after a while shows an error indicating that there is no more memory. lfc shows the same behavior... I assume that this originates from the fact that we unroll the complete instantiation graph and create an object for every single reactor instance, but I am still wondering why the memory footprint is so high. Maybe there is a memory leak of some sort.

The error can be reproduced using the following simple program:
```
target Cpp;

reactor Node(bank_index: size_t(0), num_nodes: size_t(4)) {
    input[num_nodes] in: size_t
    output[num_nodes] out: size_t
}

main reactor(num_nodes: size_t(1000)) {
    nodes = new[num_nodes] Node(num_nodes=num_nodes);
    nodes.out -> nodes.in;
}
``` 
Note that depending on your available memory, you might need to increase `num_nodes` to trigger the error. While experimenting, I had the impression that the memory footprint does not scale linearly with the number of nodes.

edwardalee:
In looking at the graph classes that we are using, e.g. for detecting cycles, there are quite a few inefficiencies there. For example, to obtain a list of nodes in the graph, the code currently creates a new list on the heap and copies two lists into it to make one list.   It does this fresh each time you obtain the list. Also, the graph instances duplicate information that is already in the ReactorInstance, ReactionInstance, etc. classes. I saw quite a few other potential problems. I'm not sure that is the culprit, but graphs of size 1000 are not big, so I think we have a problem if we can't handle those.

cmnrd:
Do we need to expand the graph in order to do the cycle analysis and other steps? I find it interesting that this works fine for `num_nodes=10`, but not for `num_nodes=1000` although the visual representation is exactly the same. Maybe we should treat a bank of reactors not as N instances, but as one 'virtual instance'. For cycle analysis, it should be sufficient to use one representative. There is probably no need to repeat the analysis for every instance.

edwardalee:
Interesting conjecture.  Unfortunately, below is a counterexample.  As given below, there is no cycle.  However, if you change this line:
```
    s2 = new Split(delay = 2, nodelay = 2);
```
to
```
    s2 = new Split(delay = 1, nodelay = 3);
```
then you get a cycle.

[...]

PR #442, opened 2021-07-29
Fix out of memory error

oowekyala:
See discussion #433, I'll write up this description later

### DIFF
--- a/org.lflang/src/org/lflang/util/CollectionUtil.java
+++ b/org.lflang/src/org/lflang/util/CollectionUtil.java
@@ -1,0 +1,54 @@
+package org.lflang.util;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * Utilities to manipulate collections.
+ */
+public class CollectionUtil {
+
+    /**
+     * Returns a set which contains the elements of the given
+     * set plus the given element. The returned set should be
+     * considered unmodifiable.
+     *
+     * @param set initial set, nullable
+     * @param t   new element
+     * @param <T> Type of elements
+     * @return A new set, or the same
+     */
+    public static <T> Set<T> plus(Set<T> set, T t) {
+        if (set == null || set.isEmpty()) {
+            return Set.of(t);
+        } else if (set.size() == 1) {
+            return Set.of(set.iterator().next(), t);
+        } else if (set.size() == 2) {
+            // make mutable
+            set = new LinkedHashSet<>(set);
+        } // else it's already mutable.
+
+        set.add(t);
+        return set;
+    }
+
+
+    /**
+     * Returns a copy of the set. The returned set should be
+     * considered unmodifiable.
+     *
+     * @param set initial set, nullable
+     * @param <T> Type of elements
+     * @return A new set, or the same
+     */
+    public static <T> Set<T> copy(Set<T> set) {
+        if (set == null) {
+            return null;
+        } else if (set.size() <= 2) {
+            return set; // it's unmodifiable
+        } else {
+            return new LinkedHashSet<>(set);
+        }
+    }
+
+}
--- a/org.lflang/src/org/lflang/util/CollectionUtil.java
+++ b/org.lflang/src/org/lflang/util/CollectionUtil.java
@@ -1,7 +1,11 @@
 package org.lflang.util;
 
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
+import java.util.function.BiFunction;
 
 /**
  * Utilities to manipulate collections.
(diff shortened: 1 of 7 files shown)


lhstrh:
It looks like these changes have introduced errors in the level/chainID assignments, and therefore cause test failures...

cmnrd:
Looks like merging master fixed the test issues. I am wondering a bit why..., but it seems like we could discuss merging this now.

org.lflang/src/org/lflang/generator/InvalidSourceException.java
```diff
@@ -0,0 +1,41 @@
...
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.lflang.generator;
+
+/**
+ * Thrown when the compiled sources fails a validity check
+ * invalid and the current code cannot recover. For example,
+ * this is thrown when detecting a dependency cycle. This
+ * should be preferred
```
edwardalee, line 31: The grammar in this comment has some problems. Suggest:
```
 * This exception is thrown when a program fails a validator check
 * and code cannot be generated. For example, this is thrown when
 * a causality loop has been detected.
```

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.xtend
```diff
@@ -134,7 +133,7 @@ class ReactionInstanceGraph extends DirectedGraph<ReactionInstance> {
      * @return true if the assignment was successful, false if it was not, 
      * meaning the graph has at least one cycle in it.
      */
-    protected def DirectedGraph<ReactionInstance> assignLevels() {
+    private def DirectedGraph<ReactionInstance> assignLevels() {
```
edwardalee, line 136: The return comment for this method is now incorrect. Also the sentence before the return part.

org.lflang/src/org/lflang/graph/DirectedGraph.java
```diff
@@ -0,0 +1,288 @@
+/* Instance of an action. */
```
edwardalee, line 1: Spurious comment.

org.lflang/src/org/lflang/graph/DirectedGraph.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.*;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+import org.lflang.util.CollectionUtil;
+
+/**
+ * Directed graph that maps nodes to its upstream and downstream neighbors.
+ *
+ * @author {Marten Lohstroh <marten@berkeley.edu>}
```
edwardalee, line 38: Suggest adding Clément as an author.

org.lflang/src/org/lflang/graph/DirectedGraph.java
```diff
@@ -0,0 +1,288 @@
...
+        this.downstreamAdjacentNodes.putIfAbsent(node, Set.of());
+    }
+
+
+    @Override
+    public void removeNode(T node) {
+        this.graphChanged();
+        this.upstreamAdjacentNodes.remove(node);
+        this.downstreamAdjacentNodes.remove(node);
+        CollectionUtil.removeFromValues(this.upstreamAdjacentNodes, node);
```
edwardalee, line 112: This could use a comment.
```suggestion
        // The node also needs to be removed from the sets that represent connections to the node.
        CollectionUtil.removeFromValues(this.upstreamAdjacentNodes, node);
```

org.lflang/src/org/lflang/util/CollectionUtil.java
```diff
@@ -0,0 +1,145 @@
...
+        map.put(k, v);
+        return map;
+    }
+
+
+    /**
+     * Remove the given value from all the sets that are values
+     * in the given map. Use this if the values of the map (the
+     * sets) were build with {@link #plus(Set, Object)}.
+     *
```
edwardalee, line 65: Maybe this comment could be a bit clearer.  Suggest something like:
```
 * We use maps to represent edges, where a value in a map is a set of nodes adjacent to the key for that value.
 * Hence, when a node is removed, it needs to be removed not just as a key, but it also needs to be removed
 * from the sets of any other keys that may contain it.
```

org.lflang/src/org/lflang/util/CollectionUtil.java
```diff
@@ -0,0 +1,145 @@
...
+        if (set == null || set.isEmpty()) {
+            return Collections.emptySet();
+        } else if (set.size() == 1) {
+            return set.contains(eltToRemove) ? Collections.emptySet() : set;
+        }
+        throw new AssertionError("should be unreachable");
+    }
+
+
+    public static <K, V> Map<K, V> compute(Map<K, V> map, K k, BiFunction<K, V, V> computation) {
```
edwardalee, line 112: This method needs a comment.

edwardalee (approved):
This looks fantastic! Let's merge this ASAP. Minor suggestions included.

lhstrh (approved):
I agree with @edwardalee; these improvements look excellent. We have a few other optimizations in mind that we would like to evaluate the impact of. Could you share with us your profiling tools/methodology? That would be incredible helpful to us...

cmnrd (commented):
This also looks good to me, and I think it is a great improvement. However, since there were errors shown earlier in the CI runs, I think we should hold off merging this until we fixed and merged #535. With this merged, we can have more trust in the test results.

oowekyala:
Thank you all for your comments, I'll try to resolve them and fix the conflicts asap. I agree with @cmnrd though, that it would be better to merge #535 first --but that PR is in a more ready state as this one anyway.

PR #443, opened 2021-07-29
Better error messages when running standalone LFC

oowekyala:
Currently LFC's error messages are really... bad:

Capture d’écran de 2021-07-29 01-25-07

- there are duplicate messages
- there are warnings that were caused by another error (eg the target property warning caused by the unrecognized target error)
- there is no location information
- messages have inconsistent formatting
- messages have inconsistent ordering

Compare this to this PR:

Capture d’écran de 2021-07-29 01-23-54

~This is still missing some documentation, and testing that it doesn't affect the RCA, but I still wanted to show~

You can find the source file for the error messages displayed above here:  [MovingAverage.lf.zip](https://github.com/icyphy/lingua-franca/files/6901456/MovingAverage.zip)


lhstrh:
Closing because duplicate with https://github.com/icyphy/lingua-franca/pull/456

PR #445, opened 2021-07-29
Adding ThreadRing Benchmark (C)

mattchorlian:


benchmark/C/Savina/src/concurrency/ProdCons.lf
```diff
@@ -343,30 +341,28 @@ reactor ConsumerReactor(consCost:int(25)) {
...
         printf("numProducers: %d.\n", 40);
         printf("numConsumers: %d.\n", 40);
 		
-		
-        printSystemInfo();
-        SET(runner.inStart, true);
+        SET(manager.inStart, true);
     =}
     
     producers = new[40] ProducerReactor(numItemsToProduce=numItemsPerProducer, prodCost=prodCost);
```
cmnrd, line 368: The number of producers and consumers should be parameterized

benchmark/C/Savina/src/concurrency/ProdCons.lf
```diff
@@ -343,30 +341,28 @@ reactor ConsumerReactor(consCost:int(25)) {
 //    =} 
 //
 //}
-main reactor ProdConsBenchmark(numIterations:int(12), bufferSize:int(50), prodCost:int(25), consCost:int(25), numItemsPerProducer:int(1000)) {
```
cmnrd, line 346: The number of consumers and producers are missing here as parameters

benchmark/C/Savina/src/concurrency/ProdCons.lf
```diff
@@ -8,12 +8,10 @@
  
 target C {
     flags: "-lm",
-    logging: DEBUG,
     keepalive: true,
```
cmnrd, line 11: Keepaalive is unintuitive to me. Doesn't this cause the program to run forever?

benchmark/C/Savina/src/concurrency/ProdCons.lf
```diff
@@ -8,12 +8,10 @@
...
     timeout: 2 secs
 };
 
-import BenchmarkRunner from "../BenchmarkRunner.lf";
 
 preamble {=
     #include <stdlib.h>
```
cmnrd, line 17: It seems that the whole program can be significantly simplified. See the updated version of the C++ benchmark. This version removed the various message types defined below and instead uses dedicated ports. Would be nice if we could also simplify this version.

benchmark/C/Savina/src/micro/Counting.lf
```diff
@@ -38,7 +38,13 @@
  ]]] */
  // [[[end]]]
 
-target C;
+target C{
+    /* [[[cog
```
cmnrd, line 42: Could you update the mechanism for setting the number of threads with the one I introduced in #485? This applies of course to all the benchmarks, not just this one.

benchmark/C/Savina/src/micro/ThreadRing.lf
```diff
@@ -152,14 +160,16 @@ reactor Initializer(numReactors:int(100), numPings:int(100000)) {
...
+main reactor (numPings:int(100000), numReactors:int(100)) 
+
+/// [[[end]]]
+
+{
     
+    init = new Initializer(numReactors=numReactors, numPings=numPings);    
     
     reaction(startup) -> init.inStart {=
         SET(init.inStart, true);
```
cmnrd, line 175: This comment refers to code below (Unfortunately, I can only comment on code that changed). We have better mechanisms for connecting a large number of reactors now. Please have a look at the updated C++ version for the required pattern.

benchmark/C/Savina/src/parallelism/PiPrecision.lf
```diff
@@ -14,7 +14,7 @@
  */
 
 target C {
-    flags: "-lgmp -g",
+    flags: "-lgmp",
     keepalive: true,
```
cmnrd, line 18: Also here, keepalive seems strange to me.

benchmark/C/Savina/src/parallelism/Trapezoidal.lf
```diff
@@ -157,15 +157,14 @@ reactor Worker(instance:int(0)) {
 }
 
 
-main reactor ProdConsBenchmark(numWorkers:int(100), numIterations:int(12), numPieces:int(10000000), leftEndPoint:double(1.0), rightEndPoint:double(5.0)) {
```
cmnrd, line 160: the parameters are not cogged

benchmark/C/Savina/src/parallelism/RadixSort.lf
```diff
@@ -9,9 +9,13 @@
...
-
+target C {
+    /* [[[cog
+    cog.outl(f'threads: {threads}')
+    ]]] */
+    threads: 24 
+    // [[[end]]]
+};
 
 reactor IntSourceReactor(numValues:int(100000), maxValue:{=long=}({=1l << 60=}), seed:{=long=}(2048)) {
```
cmnrd, line 21: The comment refers to the code below. The `nextLong` function below looks wrong to me, as it does not preserve state and thus always produces the same number. There is a random number generator defined in `include/PseudoRandom.h` which should be used instead.

cmnrd (changes requested):
This looks quite good already. I added the usual nitpicks and also made a few comments regarding the code that existed before. Although that is not your fault ;) it would be nice if you could address those comments and adjust the existing benchmarks to the "modern" versions that we have now for C++.

cmnrd:
@LDeng0205 it looks like something went wrong with the merge that you pushed. There was a huge number of commits and changes in this PR. I tried to revert the merge, but I couldn't get it to work. I now force-pushed an older version of the `c-benchmark` branch and cherry picked your latest commit on top of it. I strongly recommend getting a fresh clone of Lingua Franca to avoid problems when you pull from this repo.

benchmark/C/Savina/src/parallelism/RadixSort.lf
mattchorlian: If I just delete that code for nextLong and include PseudoRandom.h, will that suffice?  The nextlong function in PseudoRandom.h is identical to the nextLong function that was in this preamble, though.

benchmark/C/Savina/src/micro/ThreadRing.lf
mattchorlian: Is there  a special syntax for this in the C target?  When I tried using the C++ pattern the compiler reports "extraneous input '-1' expecting ']" in reference to this line:

workers = new[numReactors-1] ThreadRingReactor();

Is this expected?

edwardalee: AFAIK, the `new` construct does not accept arbitrary arithmetic expressions. I think it will accept sums of constants or parameters, but not arbitrary arithmetic expressions.  And even those sums are not documented in the [language specification](https://github.com/lf-lang/lingua-franca/wiki/Multiports-and-Banks-of-Reactors#banks-of-reactors).

benchmark/C/Savina/src/parallelism/RadixSort.lf
cmnrd: The difference is that the random number generator in PseudoRandom.h has internal state that is updated on each call to `nextLong`. Thus, it will produce different numbers on each call, whereas the implementation here should always produce the same number. In order to call `nextLong` from `PseudoRandom.h` you also need to instantiate the `PseudoRandom` struct and then initialize it with the seed in a startup reaction.

benchmark/C/Savina/src/micro/ThreadRing.lf
cmnrd: Our syntax doesn't allow this. The C++ version uses the target code escape `{=...=}` to do the calculation in target code:
```
workers = new[{=numReactors-1=}] ThreadRingReactor();
```
This, however, is only supported by the C++ target and does not work in C. I think your only option is to use cog here. So you would do the subtraction in python code and generate the line accordingly.

PR #448, opened 2021-07-30
Convert some test classes to Java

oowekyala:
This PR converts some of the test files that are written in XTend, including TestBase, to Java. There are also a few automated cleanups of TestRegistry. The conversion to Java allows us to run those tests from within IntelliJ.

This branch was originally meant to integrate our test infrastructure more tightly within Junit, so that IDE integrations would work better. This is not as useful as I thought, because of limitations in Gradle's test task, so I reverted the commit. I kept it in the branch history to continue work on this later if things change.


lhstrh (approved):
Thanks for porting these!

Issue #447, opened 2021-07-30
Cycle highlighting in the diagram view no longer works


lhstrh:
While an error message is displayed, the "Filter cycle" or "Display cycle" features ceased to work (all nodes disappear and only the collapsed top-level reactor is shown).

edwardalee:
I have pushed to branch cycle-visualization-bugfix a partial fix for this. However, Connections are still not highlighted.  The offending code is in CycleVisualization.xtend. Specifically, this code is wrong somewhere:
```
                        // Reactor edges
                        for (cycleEgde : reactorNode.outgoingEdges.filter [
                            affectedElements.contains(sourcePort.sourceElement()) && (
            				(
            					!target.sourceIsReactor() && allAffectedElements.values.contains(target.sourceElement())
            				) || (
            					target.sourceIsReactor() &&
                                allAffectedElements.keySet.contains(
                                    target.getProperty(LinguaFrancaSynthesis.REACTOR_INSTANCE)) &&
                                allAffectedElements.get(target.getProperty(LinguaFrancaSynthesis.REACTOR_INSTANCE)).
                                    contains(targetPort.sourceElement())
            				)
            			)
                        ]) {
                            // FIXME: Still hard-coded semantics
                            if (!(cycleEgde.sourceElement() instanceof Connection) ||
                                (cycleEgde.sourceElement() as Connection).delay === null) {
                                cycleEgde.setProperty(DEPENDENCY_CYCLE, true)
                                highlighter.accept(cycleEgde)
                            }
                        }
```
Unfortunately, with the way this code is written, it seems to not be possible to set breakpoints see why the filter is erroneously filtering out Connection objects.  I could rewrite this code in straightforward Java and then debug it, but that would significantly change the coding style. So perhaps it's better for Alex to fix this?

a-sr:
Yes, I will take care of that.

a-sr:
I fixed the remaining problems and opened a PR (#455). In my tests, all cycles are again completely highlighted and the buttons are working. It seems I forgot to adjust the cycle visualization to the new reaction graph.
The code should be more debug-friendly now as well.

I also had a look at the FixMe (see above) that I added when initially writing the highlighting. The problem was (and still is) that a reported cycle only consists of reaction and port instances. Hence, if two ports in a cycle are connected with multiple direct connections but not all of them contribute to the cycle, then I need to filter some out based on semantics.
Here is an example:
[...]

PR #455, opened 2021-08-09
Cycle Visualization Bugfix

a-sr:
See #447


lhstrh (approved):
Thanks for these fixes! I'll look into the issue with physical connections and delays not being properly recognized. That seems to be a problem on our end.

PR #456, opened 2021-08-17
Error reporting

lhstrh:
Copy of https://github.com/icyphy/lingua-franca/pull/443


lhstrh:
@oowekyala: our reliance on Kotlin classes in this new error reporting mechanism breaks the ability to execute our JUnit tests from within Eclipse. This does pose an inconvenience...

lhstrh:
An update on this: it appears we only use `Main` in order to obtain a resource set (TestRegistry.java:173). There must be a better way to do this that doesn't depend on `ReportingUtil`.

lhstrh (commented):
Thanks for taking the initiative for implementing these improvements, @oowekyala! There are a few comments missing in `ReportUtil`, but this looks very good overall.

PR #463, opened 2021-08-18
Fixes #459. Fixes #460.

### LINKED ISSUE #459: Warnings are not repored by lfc
If an LF program has warnings, lfc does not show these warnings in its output.

### LINKED ISSUE #460: Invalid program triggers NullPointerException
The following program triggers a NullPointerException:
```
target C;

main reactor {
  test = new Test();
}
```
While the program is invalid and compilation is expected to fail, the compiler should produce an actual error message instead of throwing a NullPointerException.

lhstrh:


edwardalee (approved):
Looks good to me...

PR #464, opened 2021-08-19
Branding

lhstrh:
- [x] Update splash screen
- [x] Update icons
- [x] Adjust very small icons for scale (simplify)
- [x] Rename `eclipse` executable to `epoch`
- [x] Adjust icon associated with `.lf` file type (see `ui` package)
- [x] Update "about" text


lhstrh:
image

cmnrd:
While we are at it: I noticed that in the .tar.gz produced by the nightly build, the root directory is called `lflang`. Thus, when extracting, this creates the directory `lflang` and places epoch in it. Maybe we should rename this repository to `epoch` or `epoch-ide`?

lhstrh:
I would have to dig a little bit to figure out how this is adjusted. Maybe @a-sr knows?

lhstrh:
I keep getting warnings about the 16x16, 32x32, and 48x48 icons not being the right depth (32 instead of 8), but even after changing this (`convert lf-logo-16px.png -colors 256 lf-logo-16px.png`) the warnings don't go away. Any ideas, @a-sr?

a-sr:
> I would have to dig a little bit to figure out how this is adjusted. Maybe @a-sr knows?

I think it is the rootFolder entry in the pom of org.lflang.rca: https://github.com/icyphy/lingua-franca/blob/51161b20a86d707ebc0c1726974559e10ee145ad/org.lflang.rca/pom.xml#L57

a-sr:
> I keep getting warnings about the 16x16, 32x32, and 48x48 icons not being the right depth (32 instead of 8), but even after changing this (`convert lf-logo-16px.png -colors 256 lf-logo-16px.png`) the warnings don't go away. Any ideas, @a-sr?

To be honest, I always ignored this warning in our Kieler product because I wondered who would have a modern Windows system with a color depth lower than 32-bit.
I found this guide that uses Gimp and creates an ICO file instead of separate BMPs, maybe this works.

cmnrd:
Looks fantastic!

PR #467, opened 2021-08-19
Cpp: Support generic parameters

cmnrd:
Type parameters could be used as types for reactor reactor parameters as is shown in the following example:
```
reactor<T> (bar:T(0)) {
}
```
However, this is currently not properly supported by the C++ generator and it produces invalid code. This PR slightly modifies the generator, so that it produces correct code for generic parameters. This PR also adds a new test case.


HugoForrat (approved):

lhstrh (commented):
🚀🚀🚀

PR #468, opened 2021-08-19
file-config: place binaries in bin, not src-gen/bin

cmnrd:
It seems that #435 introduced a bug that caused all binaries to be located in `src-gen/bin`, while they are expected to be placed in `bin`. It seems a bit surprising that no one noticed this subtle change. I am not sure why exactly the behavior changes, as @oowekyala's commits in #435 look good in a first glance. I suspect that `resolve()` works different for URIs and Paths, which could have caused this bug. In any case, this PR should fix the problem and revert to the old behavior.


lhstrh:
@mattchorlian noticed this odd behavior but hadn't gotten around to file a bug report yet. Thanks for pushing a fix, @cmnrd!

edwardalee (approved):
Looks good to me. Thanks for fixing it.

PR #469, opened 2021-08-23
FileConfig: make resolution of project root more robust

cmnrd:
This makes the resolution of the project root more robust. Before, the resulting
path would include the '..'. Also, for some reason that I don't quite
understand, this broke the check for existence of a directory and, in consequence,
lfc -c did not work as expected. This is fixed now.

### DIFF
--- a/org.lflang/src/org/lflang/FileConfig.java
+++ b/org.lflang/src/org/lflang/FileConfig.java
@@ -201,7 +201,7 @@ public class FileConfig {
         this.srcGenPath = getSrcGenPath(this.srcGenBasePath, this.srcPkgPath,
                 this.srcPath, name);
         this.srcGenPkgPath = this.srcGenPath;
-        this.outPath = srcGenBasePath.resolve("..");
+        this.outPath = srcGenBasePath.getParent();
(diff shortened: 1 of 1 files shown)


org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -175,7 +175,7 @@ public FileConfig(Resource resource, IFileSystemAccess2 fsa, IGeneratorContext c
         this.srcGenPath = getSrcGenPath(this.srcGenBasePath, this.srcPkgPath,
                 this.srcPath, name);
         this.srcGenPkgPath = this.srcGenPath;
-        this.outPath = srcGenBasePath.resolve("..");
+        this.outPath = srcGenBasePath.getParent();
```
oowekyala, line 178: In #468 you also changed line 204, maybe it should be updated too?

cmnrd: Good catch. Thanks!

lhstrh (approved):
Yes, this looks right. I recall consistently using `getParent`; I'm not sure where/how the `resolve('..')` came in..

cmnrd:
That was my bad. I changed it in #469 ;)

PR #478, opened 2021-08-30
Contributor resources and code style guide

oowekyala:
This PR creates a `CONTRIBUTING.md` document in our repo to serve as a reference for our contributors. This PR was initially created by @oowekyala a long time ago and has been updated by @lhstrh based on discussions had during our last retreat in Sonoma.


CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+The overarching guideline we follow is to use each language's most widely used idioms and conventions, which are fortunately very similar.
+
+A couple of more specific code hygiene guidelines are described in [code-style.md](code-style.md) (mostly for Java right now).
+
+#### Code formatting
+
+The following code formatters can be imported into Eclipse:
+* [Java](https://github.com/lf-lang/lingua-franca/blob/master/JavaFormatterEclipse.xml)
+* [Xtend](https://github.com/lf-lang/lingua-franca/blob/master/XtendFormatterEclipse.xml)
+
```
lhstrh, line 30: - [ ] 4 spaces for indentation
- [ ] no tabs

code-style.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Visibility
+
+- avoid public non-final fields
+- avoid public collection fields (even if final, they may be mutable)
+
+### Design for inheritance or prohibit it
+
+Make new classes final by default, and don't use protected members.
+The class should only be unsealed if a use case for inheritance shows up, which is in many cases never.
```
petervdonovan, line 39: There was a comment from Alexander some months ago that went something like, "If you make a class final, you patronize the programmer... Sometimes you just need to change one method, and then it would work perfectly... One time [someone] became so frustrated that he copied an entire class, pasted it into a new file, and called it [name of the original class]2." I think that this issue does not really apply to us since we use a different project structure where we do not have multiple separate projects that only depend on specific releases. As a result, the use of the `final` keyword for us would in practice become a fairly mild constraint -- basically just a suggestion and a signal of the original programmer's intent. Therefore it seems unnecessary for this issue to further delay this PR.

lhstrh:
@oowekyala: I'm a little unsure what the change to [.idea/codeStyles/Project.xml](https://github.com/lf-lang/lingua-franca/pull/478/files#diff-11745b49c51a1ecde7d4d4cffda53d76594e6711ec81bf1148a05942b0438141) is for. Should it be preserved?

petervdonovan:
> I'm a little unsure what the change to [.idea/codeStyles/Project.xml](https://github.com/lf-lang/lingua-franca/pull/478/files#diff-11745b49c51a1ecde7d4d4cffda53d76594e6711ec81bf1148a05942b0438141) is for. Should it be preserved?

This was a change that I pushed to prevent imports from automatically being replaced with star imports in IntelliJ. I think it is harmless.

CONTRIBUTING.md
```diff
@@ -1,7 +1,186 @@
...
+- _Keep classes concise._ Classes should be limited in scope and not become too large. Anything above 1000 lines probably needs refactoring.
+- _Do not leave FIXMEs_. Unaddressed `FIXME`s should not be allowed to pass code review unless they are accompanied with a link to a GitHub issue.
+
+_Comments_
+
+Please adhere to the following principles when writing documentation for your code:
+- Write descriptions in English.
+- Do not use contractions like "aren't" or "isn't".
+- It is OK to use phrases instead of complete sentences, in the interests of brevity. This holds especially in `@param` JavaDoc tag descriptions.
+- Use 3rd person (descriptive) not 2nd person (prescriptive), i.e., write "Compute the shortest path," not "Computes the shortest path."
```
cmnrd, line 100: I am not very fit in grammar, but isn't "Computes ... " 3rd person?

CONTRIBUTING.md
```diff
@@ -1,7 +1,186 @@
...
+#### Merging
+Perform merges to bring your feature branch up-to-date with master locally (do not issue a PR). This is very easy to do.
+1. Ensure you are on your feature branch (run `git branch` to find out; `git switch my-feature-branch` to switch to your feature branch).
+2. Make sure you have the latest changes (run `git fetch --all`)
+3. Perform the merge (run `git merge origin/master`, assuming that you want to merge the branch `master` from remote `origin` into `your-feature-branch`).
+
+#### Addressing reviews
+To address feedback from code review, implement changes and push to your feature branch. If you are certain that a reviewer's concern has been addressed by your new changes, hit the `Resolve conversation` button. If you are unsure, ask for follow up in the conversation and let the reviewer determine whether the feedback was addressed accordingly.
+
+### Code style and formatting
```
cmnrd, line 76: We should mention and/or link our own guidelines here.

cmnrd (approved):
Thanks for the effort! I really like how this came out. I still found a couple of nitpicks, but it should be easy to sort this out.

CONTRIBUTING.md
lhstrh: Jep, that's why the word *not* is written in before the quoted text :-)

lhstrh: I see what you mean now. The description (which was a remainder of a previous draft) was copied without scrutiny :-) I fixed it. Thanks!

CONTRIBUTING.md
lhstrh: They are all mentioned in this section. There is no separate `code-style.md` anymore.

CONTRIBUTING.md
```diff
@@ -1,7 +1,186 @@
...
+```
+./gradlew test --tests org.lflang.tests.compiler.*
+```
+**Integration tests** consist of complete Lingua Franca programs that are located in the `test` directory in the root of the repository and are organized first by target, then by category. These tests get indexed automatically and are also invoked using Gradle. For instance, to run all the tests for the C target, use the following command:
+```
+ ./gradlew test --tests org.lflang.tests.runtime.CTest.*
+ ```
+ Tests in most categories are expected to compile without errors and return exit code `0` when executed. Tests in some categories are not attempted to run and are only expected to compile as they might require the presence of particular hardware or exotic software configurations that are not manageable in GitHub Actions, our current platform for Continuous Integration (CI). Only pushes to [feature branches](#feature-branches) associated with an active [pull request](#pull-requests) trigger CI.
```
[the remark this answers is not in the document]
lhstrh: This is a good point. I'll add some clarifying text about this...

PR #479, opened 2021-08-30
Change `cmake-include` to files list

Soroosh129:
This branch changes the `cmake-include` target property to take a list of file paths rather than just a single string.

### DIFF
--- a/org.lflang/src/org/lflang/TargetConfig.xtend
+++ b/org.lflang/src/org/lflang/TargetConfig.xtend
@@ -58,11 +58,6 @@ class TargetConfig {
      * Parameter passed to cmake. The default is 'Release'.
      */
     public BuildType cmakeBuildType = BuildType.RELEASE
-
-    /**
-     * An optional additional .cmake file to include.
-     */
-    public String cmakeInclude = ""
     
     /**
      * Enable or disable the use of CMake to build.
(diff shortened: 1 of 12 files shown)


cmnrd (commented):
I think this is a good idea! I pushed a small change to the kotlin code to simplify it ;) 

I have one remark about the tests: They use two cmake files, but, since both include files add the same library, they do not check if both files are actually included. Would it make sense to have one file include another library, or maybe define a macro using `add_compile_definitions`?

Soroosh129:
> I think this is a good idea! I pushed a small change to the kotlin code to simplify it ;)

Thanks for the simplification! It looks a lot better. I'm still adjusting to Kotlin.

> 
> I have one remark about the tests: They use two cmake files, but, since both include files add the same library, they do not check if both files are actually included. Would it make sense to have one file include another library, or maybe define a macro using `add_compile_definitions`?

I was looking for a universal library like `m` to use for the second included file, but I couldn't find any (even the `m` library is apparently built into the standard library that comes with Visual Studio). I really like the `add_compile_definitions` suggestion, which is more platform-independent.

Soroosh129:
I went ahead and changed the second included file to contain a `add_compile_definitions`.

lhstrh (approved):

PR #480, opened 2021-08-31
Fix error message display when line has tabs

oowekyala:
The new error messages are off when the error line has tabs:

Capture d’écran de 2021-08-31 17-22-23

This is now fixed:
Capture d’écran de 2021-08-31 17-25-33


lhstrh (approved):
Nice touch! Looks good!

Issue #481, opened 2021-09-01
Sleeping Barber benchmark slowing on large inputs


ith8:
I am working on benchmarks for the C target. [Barber.lf ](https://github.com/icyphy/lingua-franca/blob/benchmarks/benchmark/C/Savina/src/concurrency/Barber.lf) has trouble running on my local machine when I increase the problem size to over 1000 customers. For instance, I was able to run a smaller input size of 500 customers and finish within a few seconds:

```c
191 | main reactor Barber(cut_time:time(1 nsec), num_chairs:int(100), num_customers:int(500)){
 ```

But when I increase the parameters `num_customers` or `num_chairs`, the compiled program just hang and does not terminate even after 10-20 minutes.

```c
191 | main reactor Barber(cut_time:time(1 nsec), num_chairs:int(100), num_customers:int(1000)){ 
```

cmnrd:
Interesting! At first, I thought this is related to #433, but it seems that not the slow compilation process is the issue here, but that there is a problem in the generated C code or the C runtime.

lhstrh:
It was moved [here](https://github.com/lf-lang/lingua-franca/blob/478291f38a659d2bbb72a496526b5a025811bfc1/example/C/src/SleepingBarber.lf) in https://github.com/lf-lang/lingua-franca/commit/478291f38a659d2bbb72a496526b5a025811bfc1

edwardalee:
OK, I guess it's not in master. 478291f seems to imply it's in benchmarks. I get this:
```
EALMAC:src eal$ git checkout origin/benchmarks
Note: checking out 'origin/benchmarks'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 478291f38 move sleeping barber to examples and remove cog
EALMAC:src eal$ git branch
* (HEAD detached at origin/benchmarks)
  master
```
Being in a "detached head" state seems kind of disconcerting, but what the hey, this is git (not Eclipse, the favorite whipping boy).  If I follow the instructions in the above message:
```
EALMAC:src eal$ git checkout -b origin/benchmarks
Switched to a new branch 'origin/benchmarks'
EALMAC:src eal$ 
EALMAC:src eal$ 
EALMAC:src eal$ git status
[...]

Issue #290, opened 2021-03-19
Align the meaning of the 'threads' target property across targets.


cmnrd:
Currently we have two targets supporting multi-threaded execution: C and C++. The number of worker threads utilized by the runtime can be configured with the `threads=N` target property. However, `threads=0` has a different meaning in the two targets. In C++, `threads=0` tells the runtime to choose the optimal number of worker threads (currently this is the number of hardware threads in the executing system). In C, `threads=0` is the default and tells the compiler to use the unthreaded runtime. If `threads` is set to one or more, the compiler will use the threaded runtime and utilize the given amount of worker threads.

Since `threads` has a double meaning in C (specifying the number of worker threads and switching between threaded and unthreaded runtime), I think we should introduce a new C-specific target property `multi-threaded` (or `threaded`) to switch between runtimes,  and give `threads` (in particular `threads=0`) the same meaning as in C++. The default would be `multi-threaded=false` and explicitly setting `multi-threaded=true` would enable multi-threaded execution. In case of multi-threaded execution, `threads=N` can further set the number of worker threads to `N`. There should be an additional sanity check ensuring that `multi-threaded=false` and `threads=N` cannot be used in conjunction.

oowekyala:
Is there a use case for using the multi-threaded runtime with a single thread? It looks like you could avoid some complexity if you didn't support this, and used threads=0 for auto-selection, threads=1 for single-threaded, threads=n > 1 for multi-threaded, the selected runtime being an implementation detail.

cmnrd:
I guess that could also be an option. For C, the default would then be `threads=1`, meaning unthreaded runtime. However, I think it is desirable to make the difference between the two C runtimes more explicit. The runtime selection also has some implications that are not so obvious and not just an implementation detail. For instance, the unthreaded C runtime does not support physical actions.

oowekyala:
Ok thanks for explaining!

edwardalee:
A cleaner solution would be that the target property could be:
```
   threads: auto
   threads: 0       // No threads. Unthreaded runtime.
   threads: 1 or more   // Number of threads.
```

lhstrh:
I like @edwardalee's suggestion. That said, perhaps we should use `workers` instead of `threads`?

cmnrd:
[...]

PR #485, opened 2021-09-03
Update benchmark runner to support hydra 1.1 and distinguish threaded and unthreaded C runtime.

cmnrd:
This PR does 2 things:
- Updating the hydra dependency. A new version of hydra (1.1) has been released. Hydra 1.1  brings a few important changes and also deprecates several functions. This PR adjusts both our configs and the runner script to the new requirements of hydra 1.1. This also means, that older hydra versions (<1.1) are not supported anymore.
- Adding a new target: the unthreaded C runtime. With this change, we can use `target=lf-c` for threaded C programs, and `target=lf-c-unthreaded` for unthreaded ones. Note, that this requires appropriate cog code in the individual C benchmarks to work correctly.

This PR is based on #406 and should only be merged after merging #406.

### DIFF
--- a/benchmark/C/Savina/src/micro/Counting.lf
+++ b/benchmark/C/Savina/src/micro/Counting.lf
@@ -38,7 +38,16 @@
  ]]] */
  // [[[end]]]
 
-target C;
+target C {
+    /* [[[cog
+      if (threaded_runtime=="True"):
+          cog.outl(f"threads: {threads},")
+      else:
+          cog.outl("threads: 0,")
+    ]]] */
+    threads: 0,
+    /// [[[end]]]
+}
 
 reactor ProducerReactor(countTo:int(1000000)) {
 
--- a/benchmark/C/Savina/src/micro/PingPong.lf
+++ b/benchmark/C/Savina/src/micro/PingPong.lf
@@ -23,9 +23,17 @@
  * @author Edward A. Lee
  */
 target C {
+    /* [[[cog
+      if (threaded_runtime=="True"):
+          cog.outl(f"threads: {threads},")
+      else:
+          cog.outl("threads: 0,")
+    ]]] */
+    threads: 0,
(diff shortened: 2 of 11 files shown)


Feliix42 (approved):
LGTM

benchmark/C/Savina/src/micro/Counting.lf
```diff
@@ -38,7 +38,16 @@
  ]]] */
  // [[[end]]]
 
-target C;
+target C {
+    /* [[[cog
+      if (threaded_runtime=="True"):
+          cog.outl(f"threads: {threads},")
```
lhstrh, line 44: Should we take this PR as an opportunity to address https://github.com/icyphy/lingua-franca/issues/290?

lhstrh (approved):
Looks good, but I'm thinking we might want to address https://github.com/icyphy/lingua-franca/issues/290 in this PR as well, to avoid breaking this functionality by the time we get around to implementing the change proposed in that issue.

benchmark/C/Savina/src/micro/Counting.lf
cmnrd: I think #290 is a bigger change and I would rather address this separately

lhstrh: OK, fair enough!

mattchorlian (approved):
Looks good to me as well! Will all the config files need an additional lf-c-unthreaded configuration? And if so, should we address that here or in another PR?

cmnrd:
No, the benchmark configs don't need to change. All the magic happens in the target configurations (and partly in the runner script). The benchmarks only need to contain the appropriate cog code. This we should also adjust in #445 and #481 once this is merged. BTW, I will have a closer look at those PRs next week, so that we can ready them up for merging.

PR #486, opened 2021-09-04
Refactor TypeScript code generation in Kotlin to make code modular and more readable.

hokeun:
Significant parts of TypeScript code generation are now similar to that of Cpp code generator.
A lot of pr(), indent(), unindent() functions got removed.
TSReactorGenerator is more than 320 lines shorter than before.

### DIFF
--- a/org.lflang/src/org/lflang/generator/ts/TSActionGenerator.kt
+++ b/org.lflang/src/org/lflang/generator/ts/TSActionGenerator.kt
@@ -1,0 +1,78 @@
+package org.lflang.generator.ts
+
+import org.lflang.generator.PrependOperator
+import org.lflang.lf.Action
+import org.lflang.lf.Reactor
+import org.lflang.lf.Type
+import org.lflang.lf.Value
+import java.util.*
+
+class TSActionGenerator (
+    private val tsGenerator: TSGenerator,
+    private val reactor: Reactor
+) {
+    private fun Value.getTargetValue(): String = tsGenerator.getTargetValueW(this)
+    private fun Type.getTargetType(): String = tsGenerator.getTargetTypeW(this)
+
+    /**
+     * Return a TS type for the specified action.
+     * If the type has not been specified, return
+     * "Present" which is the base type for Actions.
+     * @param action The action
+     * @return The TS type.
+     */
+    private fun getActionType(action: Action): String {
+        if (action.type != null) {
+            return action.type.getTargetType()
+        } else {
+            return "Present"
+        }
+    }
+
+    fun generateClassProperties(): String {
+        val stateClassProperties = LinkedList<String>()
+        for (action in reactor.actions) {
+            // Shutdown actions are handled internally by the
+            // TypeScript reactor framework. There would be a
+            // duplicate action if we included the one generated
+            // by LF.
+            if (action.name != "shutdown") {
+                stateClassProperties.add("${action.name}: __Action<${getActionType(action)}>;")
+            }
+        }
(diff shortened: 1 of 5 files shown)


hokeun:
Hi Marten, I kind of got stuck with the approach we discussed last week (copying each generator class from Cpp and make them work with NewTSGenerator). Thus, I switched to step-by-step refactoring while making sure the TS code generator works. I found this works better for me and I made quite some progress. Let me know if this also works for you!

Thanks,
Hokeun

org.lflang/src/org/lflang/generator/ts/TSActionGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import org.lflang.generator.PrependOperator
+import org.lflang.lf.Action
+import org.lflang.lf.Reactor
+import org.lflang.lf.Type
+import org.lflang.lf.Value
+import java.util.*
+
+class TSActionGenerator (
+    private val tsGenerator: TSGenerator,
```
lhstrh, line 11: It shouldn't be necessary to pass in a reference to the generator. However, looking at `CppActionGenerator`, I'm not sure where it gets `$targetType` from. @cmnrd could you explain this?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -0,0 +1,345 @@
...
+ * Reaction generator for TypeScript target.
+ *
+ *  @author{Matt Weber <matt.weber@berkeley.edu>}
+ *  @author{Edward A. Lee <eal@berkeley.edu>}
+ *  @author{Marten Lohstroh <marten@berkeley.edu>}
+ *  @author {Christian Menard <christian.menard@tu-dresden.de>}
+ *  @author {Hokeun Kim <hokeunkim@berkeley.edu>}
+ */
+class TSReactionGenerator(
+    private val tsGenerator: TSGenerator,
```
lhstrh, line 21: Same comment as above; we should try to decouple from the generator.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -0,0 +1,345 @@
...
+        ${" |            "..reactEpilogue}            
+            |            // =============== END react epilogue
+            |        }
+        ${" |    "..if (reaction.deadline != null) generateDeadlineHandler(reaction, reactPrologue, reactEpilogue, reactSignature) else "}"}
+            |);
+        """.trimMargin()
+        }
+    }
+
+    fun generateReaction(
```
lhstrh, line 144: This can definitely be decomposed further, but it's a good start to have factored this stuff out of `TSGenerator`.

lhstrh (approved):
Looks like good progress to me!

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
hokeun: Agreed, I added a TODO to make sure to break this function into multiple functions.

org.lflang/src/org/lflang/generator/ts/TSActionGenerator.kt
hokeun: I took a note as a TODO in the code. I'll try to figure out how to remove this dependency.

hokeun:
Thanks, Marten. I added a few TODOs so that I can address your comments in the next pull requests.

PR #488, opened 2021-09-08
Rust dev

oowekyala:
What this initial version of the rust target supports is described in [this file](https://github.com/icyphy/lingua-franca/blob/rust-dev/org.lflang/src/org/lflang/generator/rust/readme.md). Most things work for simple examples, but there are still some problems that should be fixed in the runtime. That checklist is also WIP and I'll expand it to make each bullet point covered by a test case

Note: when checking out the branch for the first time, you might have to run `git submodule update --init` to clone the rust runtime submodule.

This is a draft until
- [ ] the failing tests are fixed
- [ ] the CI workflow for rust-test is setup


oowekyala:
Note that you can now find docs for the runtime's API here: https://lf-lang.github.io/reactor-rust/reactor_rt/index.html

PR #498, opened 2021-09-11
`cmake-include` improvements

Soroosh129:
This branch updates the search procedure of cmake-include in the C target to match the files target property.

This relies on the changes made in the `ROS-serialization` branch.


cmnrd:
It seems that this branch somehow includes several old commits that have already been merged to master. Did you do a rebase?

Is this new search path policy documented somewhere? And should we adjust the C++ target as well?

Soroosh129:
> It seems that this branch somehow includes several old commits that have already been merged to master. Did you do a rebase?

A lot of the commits also belong to the `ROS-serialization` branch. This PR should really have 4 commits. I will ask for reviews once that branch is merged in.

> 
> Is this new search path policy documented somewhere? And should we adjust the C++ target as well?

We should definitely have a conversation about this.

The search policy is not new and is used in the files target property. I just factored it out here to reuse it for cmake-include. The search procedure is currently only documented [here](https://github.com/icyphy/lingua-franca/blob/defeead30ff304d7bb2edb28e4467fc91660d278/org.lflang/src/org/lflang/FileConfig.java#L738). I will add it to the wiki when we are close to merging this branch.

cmnrd:
A lot of the commits also belong to the ROS-serialization branch. This PR should really have 4 commits. I will ask for reviews once that branch is merged in.

> Ah, ok. I somehow thought this is merged already. You can change the target branch of this PR from `master` to the ROS serialization branch. Then it would only show the diff between the two branches, and once the ROS branch is merged, it automatically updates to master. I have used this a lot for PR that depend on each other. See for instance  #485 which depends on #406.

Soroosh129:
> > A lot of the commits also belong to the ROS-serialization branch. This PR should really have 4 commits. I will ask for reviews once that branch is merged in.
> 
> Ah, ok. I somehow thought this is merged already. You can change the target branch of this PR from `master` to the ROS serialization branch. Then it would only show the diff between the two branches, and once the ROS branch is merged, it automatically updates to master. I have used this a lot for PR that depend on each other. See for instance #485 which depends on #406.

Interesting! Thanks for pointing this feature out! I have switched the base branch.

edwardalee (approved):
This change looks good to me.  Other targets should also support the same scheme for finding files, if they don't already.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -873,29 +837,44 @@ class CGenerator extends GeneratorBase {
...
+            mainDef.reactorClass.toDefinition.eResource
         ) {
             // Extract the contents of the imported file
-            val contents = federate.instantiation.reactorClass.toDefinition.eResource.contents;
+            val contents = reactor.toDefinition.eResource.contents;
             val model = contents.get(0) as Model
-            for (p : model.preambles?: emptyList) {
-                pr(p.code.toText)
+            for (c : model.eContents) {
+                if (c instanceof TargetDecl) {
```
lhstrh, line 861: There will likely be other target properties for which we want to do something similar. Rather than replicate logic for finding target properties and manually updating a `TargetConfig` instance, I think we should factor out and reuse the `setTargetConfig` method that's currently in `GeneratorBase`. We can then just create a separate `TargetConfig` for the imported file and merge its contents with the main instance as deemed fit. I think it makes sense to either turn `setTargetConfig` into a constructor for `TargetConfig` or make it a factory method.

lhstrh (commented):
I commented on an opportunity to promote code reuse and some stuff out of our bloated `GeneratorBase`.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: I agree that this is a good idea. However, there will be issues regarding the merging policy of target properties and issues related to federates (e.g., `cmake-include` might need to be federate-based). This looks to be out of scope of this PR. Would you like to create an issue for this?

lhstrh: Not sure I agree that it is out-of-scope, because this PR already introduces code replication through the way this is handled. Do you want me to take a stab at this and push a fix to this branch?

Soroosh129: I'm not sure what code replication you are referring to. This piece of code is just moved  from `doGenerate` to a separate `inspectReactorEResource` function.

lhstrh: I'm referring to the fact that the code below replicates what the `TargetProperty`/`TargetConfig` classes were already designed to do.

Soroosh129: Well, this is a piece of code that was already there. I think if we want to address this issue, a branch named `target-config-import` or something to that effect would be more appropriate than just a `cmake-include-improvements`. The former has a much broader implication on all targets and needs quite a bit of thought and discussions. The latter is much more limited in scope and only applies mostly to the `cmake-include` property. The changes are even further only limited to the C target.

lhstrh: The change I suggested here is also strictly limited to the `cmake-include` property and doesn't require much though or discussion IMO, but if you prefer I address it in a separate branch so you go ahead and merge this, I'm also fine with that.

Soroosh129: I can take a stab at fixing it for `cmake-include`.

Soroosh129: This issue I believe has been fixed 5ee652f.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -434,34 +435,132 @@ public static void copyDirectory(Path src, Path dest) throws IOException {
...
+   public String copyFileOrResource(String fileName, Path srcDir, Path dstDir) {
+       // Try to copy the file from the file system.
+       Path file = findFile(fileName, srcDir);
+       if (file != null) {
+           Path target = dstDir.resolve(file.getFileName());
+           try {
+               Files.copy(file, target, StandardCopyOption.REPLACE_EXISTING);
+               return file.getFileName().toString();
+           } catch (IOException e) {
+               // Will try to find the file as a resource.
```
cmnrd, line 543: The exception should be handled, or there should be an explanation of why this is not needed

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -434,34 +435,132 @@ public static void copyDirectory(Path src, Path dest) throws IOException {
...
+     *  in the filesystem.
+     * 
+     *  This also creates new directories for any directories on the destination
+     *  path that do not yet exist.
+     * 
+     *  @param source The source file as a path relative to the classpath.
+     *  @param destination The file system path that the source file is copied to.
+     * @throws IOException 
+     */
+    public void copyFileFromClassPath(String source, String destination) throws IOException {
```
cmnrd, line 478: Should we make the source a `URI` (or `URL`?) and the destination a `Path`?

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -434,34 +435,132 @@ public static void copyDirectory(Path src, Path dest) throws IOException {
...
+    * 2- Search in CLASSPATH. @see findFile()
+    * 3- Search for 'fileName' as a resource.
+    *  That means the `fileName` can be '/path/to/class/resource'. @see java.lang.Class.getResourceAsStream()
+    * 
+    * @param fileName Name of the file
+    * @param srcDir Where the file is currently located
+    * @param dstDir Where the file should be placed
+    * @return The name of the file in destinationDirectory
+    */
+   public String copyFileOrResource(String fileName, Path srcDir, Path dstDir) {
```
cmnrd, line 534: Why the three arguments and not just two as in `copyFile()` and `copyFileFromClasspath()`?

org.lflang/src/org/lflang/FileConfig.java
cmnrd: And shouldn't this be a static method?

org.lflang/src/org/lflang/FileConfig.java
cmnrd: And also this method should probably be static

cmnrd (commented):
I added a few comments to the code.

To be honest, I am not quite convinced that we need the copying functionality and personally prefer to refer to source files directly (without copying). Also the capability to refer to files in our classpath worries me. I understand that this PR is needed to get some examples to work for the C target, so I am not blocking this. But we should have a discussion at some point when we think about aligning the behavior in the cmake based targets.

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: I don't think it can be static since it relies on `this.getClass()` to actually look for the file.

Soroosh129: > Should we make the source a `URI` (or `URL`?) and the destination a `Path`?

I think that is potentially a really nice feature.

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: > Why the three arguments and not just two as in `copyFile()` and `copyFileFromClasspath()`?

Just because of `findFile` since it takes in the `fileName` and the `srcDir` separately. We can attempt to extract those from a source path like the functions that you have mentioned but that can be error-prone. This keeps the logic of this function simple.

> And also this method should probably be static

This can't be static because it uses `copyFileFromClassPath()` which needs a class object to exist I think.

Soroosh129:
I realized that there was an issue with the way `cmake-include`s and `file`s from imported resources were being used.
The problem was that I was adding the path from the included `.lf` files verbatim to the `cmake-include`/`file` of the main .lf file, which can be problematic if an imported `.lf` file is not in the same folder as the main `.lf` file. The generator was previously attempting to resolve these paths relative to the src file of the main .lf file.

To solve this, I have added a new class called `LFResoruce` which keeps track of all the EResources of `.lf` files that we've seen as well as a separate FileConfig and TargetConfig for that resource. I also updated the function `copyUserFiles()` to take a TargetConfig and a FileConfig and perform the lookup relative to each LFResource. 

You can see the changes in e9ff67c.

org.lflang/src/org/lflang/FileConfig.java
cmnrd: If the source path is a `Path` object, we can simply use `getFileName()` to get the name of the file and `getParent()` to get its directory. Or is there something more complicated that I am missing here?

Soroosh129: Ah, okay. But I think this signature is still more convenient for my use case.
The `files` and `cmake-include` target properties take a file name, and the `srcDir` is the src directory of the relevant `.lf` file. If we switch to a source path, then we need to join the file name and `srcDir` before calling this function.

Should we prioritize the signature?

Soroosh129:
I have made a stab at documenting the files search procedure here: https://github.com/icyphy/lingua-franca/wiki/Target-Specification#files

I have also updated the `cmake-include` documentation for the C target to reflect the recent updates: https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#cmake-include

The links to examples don't work for now pending the merge to `master`. I would appreciate any comments. Please feel free to change them as you see fit.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
lhstrh: Doesn't look like this commit exists anymore (the link is broken).

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -434,34 +435,134 @@ public static void copyDirectory(Path src, Path dest) throws IOException {
...
+    public void copyFilesFromClassPath(String srcDir, String dstDir, List<String> files) throws IOException {
+        for (String file : files) {
+            copyFileFromClassPath(srcDir + '/' + file, dstDir + File.separator + file);
+        }
+    }
+    
+   /**
+    * Copy the 'fileName' from the 'srcDirectory' to the 'destinationDirectory'.
+    * This function has a fallback search mechanism, where if `fileName` is not
+    * found in the `srcDirectory`, it will try to find `fileName` via the following procedure:
```
lhstrh, line 523: IMO, it should not look in the src directory, but the package root, meaning the parent of the src directory. This is the way it used to work, and I don't think we should change this.

lhstrh: The reason being that not all files to copy are necessarily source files. They could be images. They could be database blobs.

org.lflang/src/org/lflang/TargetConfig.xtend
```diff
@@ -70,6 +70,13 @@ class TargetConfig {
      * Optional additional extensions to include in the generated CMakeLists.txt.
      */
     public List<String> cmakeIncludes = newArrayList
+    
+    /**
+     * List of cmake-includes from the cmake-include target property with no path info.
+     * Useful for copying them to remote machines. This is needed because
+     * target cmake-includes can be resources with resource paths.
+     */
+    public List<String> cmakeIncludesWithoutPath = newLinkedList;
```
lhstrh, line 79:
```suggestion
    public List<String> cmakeIncludesWithoutPath = newArrayList;
```

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -171,6 +174,9 @@
     FILES("files", UnionType.FILE_OR_FILE_ARRAY, Arrays.asList(Target.ALL),
             (config, value) -> {
                 config.fileNames = ASTUtils.toListOfStrings(value);
+            },
+            (config, value) -> {
+                config.fileNames.addAll(ASTUtils.toListOfStrings(value));
```
lhstrh, line 179: Not sure this will work as expected. Are file names resolved with respect to the main file or with respect to the imported file?

lhstrh: I think the files need to be resolved while constructing the target config for the imported file, meaning the file names should be a list of files or paths, not strings.

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -349,6 +362,45 @@ private TargetProperty(String description, TargetPropertyType type,
...
+     *                    should conform to.
+     * @param supportedBy List of targets that support this property.
+     * @param setter      Function for setting configuration values.
+     * @param updater     Function for updating configuration values.
+     */
+    private TargetProperty(String description, TargetPropertyType type,
+            List<Target> supportedBy,
+            BiConsumer<TargetConfig, Element> setter,
+            BiConsumer<TargetConfig, Element> updater) {
+        this.description = description;
```
lhstrh, line 383: Why not just call the other constructor here?

lhstrh (commented):
Left some more comments.

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: I didn't change how this works though, I think. The files directive has always [been](https://github.com/lf-lang/lingua-franca/blob/9e635d8d6405983c184550e7a6bd5d9269624a9a/org.lflang/src/org/lflang/generator/GeneratorBase.xtend#L482) relative to the [source file](https://github.com/lf-lang/lingua-franca/blob/9e635d8d6405983c184550e7a6bd5d9269624a9a/org.lflang/src/org/lflang/generator/GeneratorBase.xtend#L488) as far as I can remember. I agree that switching to the parent might be better. This is a relatively easy fix here, but I think we should have a discussion about it.

org.lflang/src/org/lflang/FileConfig.java
Soroosh129: I think this is better left as to do since if we make this change, we need to add tests for it as well.
This change can go in conjunction with changing the files search algorithm as part of a `files` improvement effort.

org.lflang/src/org/lflang/TargetProperty.java
Soroosh129: Yes, I agree that this could potentially be dangerous. The CGenerator does not merge `files` target properties like that.
We currently keep a list of [LFResources](https://github.com/lf-lang/lingua-franca/blob/21a68aa44bf6f0f97746a92d8ee51d37788d7f9a/org.lflang/src/org/lflang/generator/LFResource.java) and copy user files for each resource as we come across reactors that belong to that resource at the time of code generation.

org.lflang/src/org/lflang/TargetProperty.java
Soroosh129: Ah nice suggestion.

Soroosh129: Unfortunately, `updater` being final means that we can't call the other constructor, which sets the `updater` to an empty lambda function.

lhstrh: Ah, right.

PR #499, opened 2021-09-12
Migrate TypeScript code generation from pr(), indent(), unindent() style, using similar style as C++ code generators. Also modularize TypeScript code generators.

hokeun:
TSReactorGenerator got much shorter: 378 lines -> 154 lines

### DIFF
--- a/org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
+++ b/org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
@@ -8,6 +8,14 @@ import org.lflang.lf.Parameter
 import org.lflang.lf.Reactor
 import java.util.*
 
+/**
+ * Generator for code in the constructor of reactor class in TypeScript target.
+ * Specifically, this generator generates the code for constructor argument,
+ * call to the super class constructor. This generator uses other code generators
+ * for child reactors, timers, parameters, state variables, actions, ports, connections,
+ * and code to register reactions. This generator also generates federate port action
+ * registrations.
+ */
 class TSConstructorGenerator (
     private val tsGenerator: TSGenerator,
     private val errorReporter: ErrorReporter,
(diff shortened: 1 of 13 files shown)


org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
```diff
@@ -0,0 +1,131 @@
...
+import org.lflang.generator.PrependOperator
+import org.lflang.lf.Action
+import org.lflang.lf.Parameter
+import org.lflang.lf.Reactor
+import java.util.*
+
+class TSConstructorGenerator (
+    private val tsGenerator: TSGenerator,
+    private val errorReporter: ErrorReporter,
+    private val reactor : Reactor,
```
lhstrh, line 14: Isn't the `Reactor` obtainable from the `FederateInstance`? It should be something like `getInstantiation().getReactorClass()`

lhstrh (approved):
Looks good! Left a minor comment.

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
hokeun: I think the reactor from federate.instantiation.reactor is the top-level reactor, and it could be different from the reactor being generated here (the reactor here could be any reactor inside the federate).

https://github.com/icyphy/lingua-franca/blob/master/org.lflang/src/org/lflang/generator/FederateInstance.xtend#L115

hokeun:
> Looks good! Left a minor comment.

Thanks, Marten! I replied to your comment. Please let me know if it makes sense to pass federate as well!

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
lhstrh: Ah, I see, That makes sense.

lhstrh (approved):
This looks like good progress, so I'm fine with merging this in, but some of these new classes completely lack documentation, which needs to be addressed soon.

hokeun:
> This looks like good progress, so I'm fine with merging this in, but some of these new classes completely lack documentation, which needs to be addressed soon.

Thanks, Marten, for merging the pull request! As you probably noticed, I added comments to the new classes.

lhstrh:
I did! Thanks for putting those in!

PR #501, opened 2021-09-13
Modal Models

a-sr:
Prototype for modal reactors. See also [reactor-c PR #6](https://github.com/lf-lang/reactor-c/pull/6)
Currently only supports C code generation + non-federated.

ToDo:
- [x] Fix test models that use new keywords (e.g. initial in [SendersAndReceivers.lf](https://github.com/icyphy/lingua-franca/blob/master/example/C/src/Patterns/lib/SendersAndReceivers.lf))
- [x] Fix current implementation of suspending events that breaks next references that model super dense time
- [x] Fix current implementation of suspending events that does not take microsteps into account when re-enqueuing (history)
- [x] Fix handling of after delay on connections
- [x] Fix build
- [x] Reset state variables
- [x] Allow mixing writers on output ports if they are in mutually exclusive modes
- [x] Adjust cycle analysis to allow mutually exclusive modes to separate ordered reactions
- [x] Turn examples into tests

ToDiscuss:
- [ ] Compatibility to federated?
- [ ] Assess experimental multi-threaded implementation?


lhstrh:
@a-sr: since the C runtime moved, we'll have to replicate some of your changes there... See https://github.com/lf-lang/reactor-c

I could help with that, if you'd like.

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -1004,8 +1004,8 @@ public String toString() {
      * @author{Edward A. Lee <eal@berkeley.edu>}
      */
     public enum ClockSyncMode {
-        OFF, INITIAL, ON;
-        
+        OFF, INIT, ON; // TODO Discuss initial in now a mode keyword (same as startup) and cannot be used as target property value, thus changed it to init
+        // FIXME I could not test if this change breaks anything
```
lhstrh, line 1008: @Soroosh129: could you comment on this?

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -192,8 +199,12 @@ protected void addNodesAndEdges(ReactorInstance reactor) {
                     List<Runtime> previousRuntimes = previousReaction.getRuntimeInstances();
                     int count = 0;
                     for (Runtime runtime : runtimes) {
-                        this.addEdge(runtime, previousRuntimes.get(count));
-                        count++;
+                        // Only add the reaction order edge if previous reaction is outside of a mode or both are in the same mode
+                        // This allows modes to break cycles since modes are always mutually exclusive.
+                        if (runtime.getReaction().getMode(true) == null || runtime.getReaction().getMode(true) == reaction.getMode(true)) {
+                            this.addEdge(runtime, previousRuntimes.get(count));
```
lhstrh, line 205: I'm aware that this terminology was inherited, but I find `Runtime` and "runtimes" a very strange choice... @edwardalee: how about "replicas"?

lhstrh (approved):
This looks ready to me! A _very_ cool feature that I think will find lots of use.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
edwardalee: No, "replicas" doesn't work because the vast majority of the time, there is only one of these. It's only when nested within banks that you get more than one, and even then, they aren't really replicas. They can have different parameters and serve different roles. "Replicas" makes me think of triple modular redundancy...

experimental/C/src/ModalModels/BehaviorTrees/robohub_example_advanced.lf
```diff
@@ -0,0 +1,141 @@
+/*
+ * Implements a modal LF version of the hierarchical statemachine 
+ * for the behavior tree in presented in this article:
+ * https://robohub.org/introduction-to-behavior-trees/
+ * 
+ * Compared to the simple variant is uses modes more extensively, which
```
edwardalee, line 6:
```suggestion
 * Compared to the simple variant this uses modes more extensively, which 
```

experimental/C/src/ModalModels/BehaviorTrees/robohub_example_advanced.lf
```diff
@@ -0,0 +1,141 @@
+/*
+ * Implements a modal LF version of the hierarchical statemachine 
+ * for the behavior tree in presented in this article:
+ * https://robohub.org/introduction-to-behavior-trees/
+ * 
+ * Compared to the simple variant is uses modes more extensively, which 
+ * results in the correct behavior.
+ * Moreover, modeling the sequence in Nominal as model enables the potential
```
edwardalee, line 8: Should "as model" be "as modal"?

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
lhstrh: OK, a better fitting term -- one that's more specific than "runtime" -- that you can think of?

edwardalee (approved):
This is a real tour de force! I have not reviewed much, but I think we should merge because almost anything we do will lead to more conflicts.

experimental/C/src/ModalModels/BehaviorTrees/robohub_example_advanced.lf
lhstrh, line 8:
```suggestion
 * Moreover, modeling the sequence in Nominal as modal enables the potential 
```

PR #504, opened 2021-09-15
Update benchmark runner to support hydra 1.1 and distinguish threaded and unthreaded C runtime

cmnrd:
This is a duplicate of #485. Apparently I hit the merge button too quickly after merging #406, so that the target branch wasn't updated. Now #485 got merged into the `cpp-benchmarks` branch, but is still not in `master`. Since there is no way of reopening a merge request, I created this new one to merge the changes to master.


lhstrh (approved):
Looks good!

PR #505, opened 2021-09-15
Test all the benchmarks with GitHubActions

### LINKED ISSUE #439: Test the Benchmarks
We currently don't run test on our benchmarks. We should design a test that checks both that all benchmarks are compilable and that the benchmark runner works as expected.

cmnrd:
This extends the benchmark runner by a 'test mode'. In this mode, it builds and runs all the benchmarks as normal, but uses a timeout for the actual benchmark run. This way, we can test if everything works as expected, without waiting for a potential lengthy benchmark run to complete.

I think that this way of testing via the benchmark runner should be preferred over an integration of the benchmarks with our Java based test setup. This way of testing via the runner not only tests the LF benchmarks, but more importantly also the benchmark runner and the provided configurations.

This PR depends on #504.

Closes #439

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -229,7 +229,7 @@ jobs:
         run: pip3 install -r benchmark/runner/requirements.txt      
       - name: Build lfc;
         run: |
-          ./gradlew generateStandaloneComiler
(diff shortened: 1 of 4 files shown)


lhstrh (approved):
This is a very good idea!

PR #506, opened 2021-09-15
Spaces for Tabs

### LINKED ISSUE #484: Epoch should use spaces by default, not tabs
Epoch currently uses tabs by default, which I personally find pretty annoying. There seems to be an agreement in this dev group that spaces should be preferred over tabs. We should also make this the default in Epoch.

I would be grateful for any pointers on how the default configuration of Epoch can be adjusted.

a-sr:
Resolves #484

Also removes ineffective/duplicate plugin_customization.ini file.


cmnrd (approved):
Very cool :) Thanks!

lhstrh (approved):

PR #513, opened 2021-09-19
CCpp target

Soroosh129:
This (re)adds the CCpp target as a way to mix C/C++ code in the C target.
This feature has been documented [here](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#the-ccpp-target).


edwardalee (approved):
Looks good to me!

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -1113,7 +1113,7 @@ abstract class GeneratorBase extends AbstractLFValidator {
         // In case errors occur within an imported file, record the original path.
         val originalPath = path;
         
-        var severity = IMarker.SEVERITY_ERROR
+        var severity = IMarker.SEVERITY_WARNING
```
lhstrh, line 1116: What is the reason for this change?

Soroosh129: To treat any output from the compiler as a warning until we find an actual instance of an error. Sometimes a warning seem to have been interpreted as an error.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -364,6 +370,22 @@ class CGenerator extends GeneratorBase {
...
     }
 
+    /**
+     * 
+     */
+    override setTargetConfig(IGeneratorContext context) {
+        super.setTargetConfig(context);
+        // Set defaults for the compiler after parsing the target properties
+        // of the main .lf file.
+        if(this.CCppMode && targetConfig.useCmake == false && targetConfig.compiler.isNullOrEmpty) {
```
lhstrh, line 380: Better not to repeat the same conditional twice. Cleaner to have something like:
```
if (targetConfig.useCmake == false && targetConfig.compiler.isNullOrEmpty) {
  if (this.CCppMode) {
    // ..
  } else {
    // .. 
  }
}
```

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -340,13 +340,13 @@ class LFValidatorImpl extends AbstractLFValidator {
 
     @Check(FAST)
     def checkWidthSpec(WidthSpec widthSpec) {
-        if (this.target != Target.C && this.target != Target.CPP && this.target != Target.Python) {
+        if (this.target != Target.C && this.target != Target.CCPP && this.target != Target.CPP && this.target != Target.Python) {
             error("Multiports and banks are currently only supported by the C and Cpp targets.",
```
lhstrh, line 344:
```suggestion
            error("Multiports and banks are currently not supported by the given target.",
```

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -340,13 +340,13 @@ class LFValidatorImpl extends AbstractLFValidator {
...
                 Literals.WIDTH_SPEC__TERMS)
         } else {
             for (term : widthSpec.terms) {
                 if (term.parameter !== null) {
-                    if (this.target != Target.C && this.target != Target.Python && this.target != Target.CPP) {
+                    if (this.target != Target.C && this.target != Target.CCPP && this.target != Target.Python && this.target != Target.CPP) {
```
lhstrh, line 349: Sidenote: I thought they were supported for `CPP`?

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
```diff
@@ -549,7 +549,7 @@ class LFValidatorImpl extends AbstractLFValidator {
 
     @Check(FAST)
     def checkDeadline(Deadline deadline) {
-        if (this.target == Target.C &&
+        if ((this.target == Target.C || this.target == Target.CCPP) &&
```
lhstrh, line 552: The fact that this has to be repeated everywhere is a bit inelegant, but it is an unavoidable consequence of the CCpp target approach. With the `cmake-force-cpp` proposal this would not have been the case...

lhstrh (approved):

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
lhstrh: Got it. Thanks!

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
cmnrd: They are supported in C, Python and C++, but not all the other targets. Notice the `!=` check. I think we could really do a better job in keeping track of the features supported by the targets. The code would be easier to read if each target would have a `supportsParameterixedWidths()` method.

org.lflang/src/org/lflang/validation/LFValidatorImpl.xtend
cmnrd: The conditions became really difficult to parse with the nested or operation. Maybe we could introduce a small method that does the check? For instance, `isCTarget()` or `isCBasedTarget()`

cmnrd (commented):
This looks pretty good. I only added a few minor comments

example/C/src/ROS/ROSBuiltInSerialization.lf
```diff
@@ -21,11 +21,10 @@
  * 
  * @author Soroush Bateni <soroush@utdallas.edu>
  */
- target C {
+ target CCpp {
     cmake: true,  // Only CMake is supported
```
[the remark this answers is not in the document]
Soroosh129: I think the target `CCpp` works without CMake. We have [tests](https://github.com/icyphy/lingua-franca/blob/ccpp-target/test/C/src/target/HelloWorldCCPPNoCMake.lf) for [that](https://github.com/icyphy/lingua-franca/blob/ccpp-target/test/C/src/target/HelloWorldFederatedCCPPNoCMake.lf) in this branch.

ROS Serialization on the other hand only works with CMake. That's a great point. I will add a check.

org.lflang/src/org/lflang/ASTUtils.xtend
```diff
@@ -162,6 +162,16 @@ class ASTUtils {
         return true
     }
     
+    /**
+     * Change the target name to 'newTargetName'.
+     * For example, change C to CCpp.
+     */
+    static def boolean changeTargetName(Resource resource, String newTargetName) {
```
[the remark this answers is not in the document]
Soroosh129: > I assume it is needed by the test that compile C tests as CCpp?

Correct.

> Maybe it would be better placed in the test package.

I think semantically, it belongs to ASTUtils because it is an AST transformation after all.
I think in terms of being dangerous, there will be validation and compiler errors if a target name is changed to an incompatible language, so it might be relatively safe perhaps.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: Thanks, that's a great suggestion.

PR #514, opened 2021-09-19
Turn off cmake so that this example compiles

edwardalee:
One line change so that example file compiles. Are we compiling examples in the tests? If so, why didn't this show up as a test failure?  Is there a better way to fix this example?


Soroosh129:
>Are we compiling examples in the tests? If so, why didn't this show up as a test failure? 

I think @lhstrh can answer this question better than me, but it looks like the test step that compiles examples always returns true (meaning that they always pass).

>Is there a better way to fix this example?

We have tests that appear to pass on GitHub actions that combine Protobufs and MacOS: 
[PersonProtocolBuffers.lf](https://github.com/icyphy/lingua-franca/blob/master/test/C/src/PersonProtocolBuffers.lf)
[ProtoNoPacking.lf](https://github.com/icyphy/lingua-franca/blob/master/test/C/src/ProtoNoPacking.lf)

That is why I think this error has gone unnoticed. 
Upon running these locally myself, I get a compile error using CMake.
This is mostly due to the fact that finding the `protoc-c` compiler using CMake seems to be rather difficult because there is no find module for it (there is [one](https://cmake.org/cmake/help/latest/module/FindProtobuf.html) for protoc itself though).
What we do currently is that we search for it in the CMakeLists.txt:
```
include(FindPackageHandleStandardArgs)
FIND_PATH( PROTOBUF_INCLUDE_DIR protobuf-c/protobuf-c.h)
find_library(PROTOBUF_LIBRARY 
NAMES libprotobuf-c.a libprotobuf-c.so libprotobuf-c.dylib protobuf-c.lib protobuf-c.dll)
find_package_handle_standard_args(libprotobuf-c DEFAULT_MSG PROTOBUF_INCLUDE_DIR PROTOBUF_LIBRARY)
target_include_directories( ${LF_MAIN_TARGET} PUBLIC ${PROTOBUF_INCLUDE_DIR} )
target_link_libraries( ${LF_MAIN_TARGET} ${PROTOBUF_LIBRARY})
```
 
The issue on MacOS seems to arise when a user, not root, uses Homebrew or MacPorts to install protobuf-c, which puts the relevant files in non-standard paths in /opt which are not searched by CMake by default it looks like, rather than /usr/local which is searched.

The solution seems to be to use the [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) variable to add `/opt/homebrew` and `/usr/local/opt` to the search path.

Maybe @cmnrd has a better suggestion?

cmnrd:
The problem with the tests could be related to #456. I think this is pretty serious and there might be more problems that we don't know about.

Regarding the cmake problem: Yes, extending `CMAKE_PREFIX_VARIABLE` would be a good solution. We could either ask the user to do that, or manipulate the variable ourselves before calling `find` with the paths that we know.

Soroosh129:
> Do we know that the example also does not compile in the GitHub runner? How do we install protobuf there? Maybe the compilation just works fine in CI.

My guess is that compilation works fine on GitHub actions because it installs the dependencies as root, which causes Homebrew to install things to [`/usr/local/Cellar`](https://github.com/lf-lang/lingua-franca/runs/3646764582#step:5:27) (you can't use `brew` with sudo anymore, so I think there is a non-standard or old setup going on here).

Soroosh129:
Actually, this is hard to nail down. The [man](https://docs.brew.sh/Installation) page of Homebrew says that 

> [The] preferred prefix [is] /usr/local for macOS Intel, /opt/homebrew for Apple Silicon and /home/linuxbrew/.linuxbrew for Linux

So the setup on GitHub actions might not be weird, but due to the fact that it is using an Intel architecture. 
If so, the aforementioned PATH search issue could only be due to my Apple Silicon laptop using a path that is not searched by CMake.

@edwardalee would you please kindly verify that you have installed all the following three dependencies:
```
brew install protobuf
brew install protobuf-c
brew install coreutils
```

edwardalee:
OK, for some reason, redoing the brew install for protobufs made this work with cmake, so I've reversed the disabling of cmake and added to the documentation for the protobuf example.

However, now I see that the MQTT examples are also not working. First, they were pointing to the wrong location for net_utils.h/c. I've fixed that. Now, MQTTDistributed works with cmake disabled, but not with cmake enabled.  MQTTPhysical does not work with either.

MQTTPhysical exposes another potential bug, which is that a linker failure is reported as a warning only, not an error. Is there a reason for this?

Soroosh129:
I see. I personally only tested the ROS and Rhythm examples.

The examples in general are hard to keep up-to-date as the code evolves, because some like MQTT examples need special supporting libraries to even compile. Even if they do compile, there might be differences between machines. I guess the burden is on the author of a PR to manually test all the examples.

Soroosh129:
I also have it installed in the same place. I have tried to push a more elaborate finding mechanism for the Eclipse Paho MQTT client. Does this resolve the issue on your machine?

I realized that this example still doesn't compile on Linux.

Soroosh129:
I think I found a solution that fixes the compile error on Linux. Apparently, I was linking against `paho-mqtt3as`, but we had to link against `paho-mqtt3cs`!

edwardalee:
Unfortunately, this still doesn't work for me.  The CMakeInclude.txt file that is generated for MQTTPhysical.lf has an erroneous line in it:
```
include()
```
This yield the following error:
```
Cleaning /Users/eal/lingua-franca-master/git/lingua-franca/example/C/src-gen/MQTT/MQTTPhysical/build
--- Current working directory: /Users/eal/lingua-franca-master/git/lingua-franca/example/C/src-gen/MQTT/MQTTPhysical/build
--- Executing command: bash --login -c cmake -DCMAKE_INSTALL_PREFIX=/Users/eal/lingua-franca-master/git/lingua-franca/example/C -DCMAKE_INSTALL_BINDIR=bin /Users/eal/lingua-franca-master/git/lingua-franca/example/C/src-gen/MQTT/MQTTPhysical
-- The C compiler identification is AppleClang 11.0.0.11000033
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found OpenSSL: /usr/local/lib/libcrypto.dylib (found version "3.0.0")  
-- Found PahoMqttC: /usr/local/lib/libpaho-mqtt3cs.so  
-- Found MOSQUITTO: /usr/local/lib/libmosquitto.dylib  
-- libmosquitto include dir: /usr/local/include
-- libmosquitto: /usr/local/lib/libmosquitto.dylib
CMake Error at CMakeLists.txt:47 (include):
  include called with wrong number of arguments.  include() only takes one
  file.


See also "/Users/eal/lingua-franca-master/git/lingua-franca/example/C/src-gen/MQTT/MQTTPhysical/build/CMakeFiles/CMakeOutput.log".
ERROR: CMake Error at CMakeLists.txt:47 (include):
  include called with wrong number of arguments.  include() only takes one
  file.
```

Soroosh129:
> Unfortunately, this still doesn't work for me. The CMakeInclude.txt file that is generated for MQTTPhysical.lf has an erroneous line in it:

Whoops, I had forgotten to push `include/net_utils.cmake`. I believe the handling of missing `cmake-include`s should be more robust than this.

edwardalee:
Cool, that got it working. I'm going to merge this with master so these examples work again.

PR #517, opened 2021-09-20
SVG Export for Epoch RCA

a-sr:
Adds support for SVG export of diagrams to the Epoch RCA.

Also includes the Eclipse Marketplace in the product.


edwardalee (approved):
Looks pretty straightforward. Let's go ahead and merge into master.

PR #523, opened 2021-09-21
Fix the error "Could not find or load main class org.lflang.generator.Main".

hokeun:
Update the main class path in build.gradle to match the current location of Main class.

### DIFF
--- a/org.lflang.lfc/build.gradle
+++ b/org.lflang.lfc/build.gradle
@@ -34,5 +34,5 @@ task runLfc(type: JavaExec) {
     description = "Build and run LFC, use --args to pass arguments"
     group = "application"
     classpath = sourceSets.main.runtimeClasspath
-    mainClass = 'org.lflang.generator.Main'
+    mainClass = 'org.lflang.lfc.Main'
 }


cmnrd (commented):
This change shouldn't be needed, as I already made it in #437. I just double-checked and on master it says `mainClassName = 'org.lflang.lfc.Main'`. Are you maybe on an outdated branch?

hokeun:
> This change shouldn't be needed, as I already made it in #437. I just double-checked and on master it says `mainClassName = 'org.lflang.lfc.Main'`. Are you maybe on an outdated branch?

I'm pretty sure I was on master when I made the change, otherwise, I think this pull request shouldn't have been created.

I think also what's in #437 is mainClass = 'org.lflang.generator.Main' actually.
I also checked the file: https://github.com/lf-lang/lingua-franca/blob/master/org.lflang.lfc/build.gradle and it looks like this was updated 5 days ago.

Please let me know if I missed anything!

Screen Shot 2021-09-21 at 4 42 02 AM
?

cmnrd (approved):
Sorry, I got confused... GitHub only shows a small excerpt of the file, and I didn't realize that this change is in a different section of the gradle file.

Thanks for fixing this!

Issue #434, opened 2021-07-27
Make the lfc scripts more robust such that they don't need patching in our releases


cmnrd:
Our current release process requires patching of the lfc scripts (see #432). We should aim at making the scripts more robust, such that they work as expected both in the context of a release and in a developer's setup.

lhstrh:
@cmnrd: can we close this, or are there still any rough edges?

cmnrd:
AFAIK this is fixed

Issue #521, opened 2021-09-20
`lfc -r` no longer works


Soroosh129:
It appears that with some of the recent path changes, `lfc -r` no longer works as expected:
```bash
$ lfc -r Rhythm.lf
lfc: info: Jar file is missing or out-of-date; running Gradle.
lfc: fatal error: Rebuild failed. Reason: Cannot run program "./gradlew" (in directory "/home/soroush/lingua-franca-master/git/lingua-franca/org.lflang.lfc/build/libs/org.lflang.lfc-0.1.0-SNAPSHOT-all.jar!"): error=2, No such file or directory
```

lhstrh:
Yeah, I know why. Will push a fix soon!

PR #530, opened 2021-09-23
Drop the rebuild feature of lfc

cmnrd:
The automatic rebuild feature of lfc (`lfc -r`) is convenient for development, but can easily cause trouble (see #521 and #524). I argue that we should drop this feature and let gradle handle the rebuilding for us (see #528). There is already a `runLfc` task which provides the same functionality as `lfc -r`. This PR completely removes the rebuild feature of lfc and slightly improves usability of the `runLfc` task, by setting the default working directory to the root of the LF project (instead of `org.lflang.lfc` as before).

### DIFF
--- a/org.lflang.lfc/build.gradle
+++ b/org.lflang.lfc/build.gradle
@@ -28,11 +28,14 @@ task buildLfc() {
 buildLfc.finalizedBy shadowJar
 
 task runLfc(type: JavaExec) {
-    // Note: when you use --args, you need to escape cli flags which start with --
-    // For instance --args ' --help'
-    // Otherwise they're parsed as arguments to the Gradle CLI, not LFC.
-    description = "Build and run LFC, use --args to pass arguments"
+    // builds and runs lfc
+    // The working directory will be the root directory of the lingua franca project
+    // CLI arguments can be passed to lfc by using --args. Note that you need
+    // to escape cli flags which start with --.For instance --args ' --help'.
+    // Otherwise they're parsed as arguments to the Gradle CLI, not lfc.
+    description = "Build and run lfc, use --args to pass arguments"
     group = "application"
     classpath = sourceSets.main.runtimeClasspath
     mainClass = 'org.lflang.lfc.Main'
+    workingDir = '..'
 }
--- a/org.lflang.lfc/src/org/lflang/lfc/Main.java
+++ b/org.lflang.lfc/src/org/lflang/lfc/Main.java
@@ -1,6 +1,7 @@
 /**
  * Stand-alone version of the Lingua Franca compiler (lfc).
  */
+
 package org.lflang.lfc;
 
 import java.io.File;
@@ -35,6 +36,7 @@ import org.eclipse.xtext.util.CancelIndicator;
 import org.eclipse.xtext.validation.CheckMode;
 import org.eclipse.xtext.validation.IResourceValidator;
 import org.eclipse.xtext.validation.Issue;
+
 import org.lflang.ASTUtils;
 import org.lflang.ErrorReporter;
 import org.lflang.FileConfig;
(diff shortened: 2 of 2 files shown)


org.lflang.lfc/build.gradle
```diff
@@ -28,11 +28,14 @@ task buildLfc() {
...
+    // builds and runs lfc
+    // The working directory will be the root directory of the lingua franca project
+    // CLI arguments can be passed to lfc by using --args. Note that you need
+    // to escape cli flags which start with --.For instance --args ' --help'.
+    // Otherwise they're parsed as arguments to the Gradle CLI, not lfc.
+    description = "Build and run lfc, use --args to pass arguments"
     group = "application"
     classpath = sourceSets.main.runtimeClasspath
     mainClass = 'org.lflang.lfc.Main'
+    workingDir = '..'
```
hokeun, line 40: Just a clarifying question, does this mean we don't have to prepend "../" in the path argument anymore?

hokeun (approved):

org.lflang.lfc/build.gradle
lhstrh: I think this is because it adopts the root of the LF project as the default working directory (instead of `org.lflang.lfc` as before).

cmnrd: Yes, this would eliminate the "../" from the paths we provide to runLfc

Soroosh129 (approved):
This looks good to me. This change cleans up `Main.java` considerably.

lhstrh:
This isn't so much of a clean up as it is a feature removal, and it happens to concern a feature that I use very frequently. One of the motivations behind this code was to simplify the `lfc` script, in part to ease porting it to power shell, to avoid code replication, and to take advantage of nice library code for printing menus.

I'm OK with these changes, but I would like to keep the `-r` feature as an undocumented flag that only works in a development setup, because I have it in muscle memory now, and it will be a nuisance to switch to using gradle commands to achieve the same thing. I think this is best done in the `lfc` script. I can take a stab at implementing this (in the bash script, not the ps1 script) and I'll push to the same branch so we can evaluate it as part of the this PR.

lhstrh (commented):
I agree that the Java-based rebuild functionality was:
1. too complicated;
2. too brittle; and
3. not suitable for non-development use.

I've instead worked rebuild functionality into the Bash script, which is tiny compared to the amount of Java code that was slashed. This, however, reminded me of another issue, which is that we do some work on the scripts as part of the release script, which probably has to be looked at (I expect this might break the release). Ideally, we just work some conditionals into our scripts so that no adaptations are necessary for releases.

cmnrd:
Currently, the benchmark tests fail because the bash script detects `--runtime-version` as `-r`. The benchmark runner uses this flag to tell lfc which version of the C++ runtime it should use.

Honestly, I think that lfc should really just be a launcher. It shouldn't be able to build or rebuild the jars. This dual use of the lfc script for launching and building is the main reason why we have to do the brittle patching in our nightly builds. If it would just be a launcher, there would also be no need for patching.

I understand that `lfc -r` is in your muscle memory, but it also quite complicates things in our scripts. I think a good alternative could be to make the build-lfc script more powerful. Maybe `build-lfc -r` could build and **r**un lfc?

lhstrh:
@cmnrd: You might be right. I like your suggestion adding the capability to also run stuff using `build-lfc`. In any case, I agree that the patching that's currently being done for the nightly build has to go. I'll push a fix.

org.lflang.lfc/src/org/lflang/lfc/Main.java
```diff
@@ -109,11 +109,9 @@
      * @author Marten Lohstroh <marten@berkeley.edu>
      */
     enum CLIOption {
-        COMPILER(null, "target-compiler", true, false, "Target compiler to invoke.", true),
-        CLEAN("c", "clean", false, false, "Clean before building.", true),
```
cmnrd, line 133: Why did the clean option disappear? It has been used for cleaning bin, src-gen, etc, and isn't related to the rebuild feature. It is an option that I quite frequently use...

lhstrh: It didn't look like it was used anywhere, so my suspicion is that the functionality (except for the flag) was removed prior to my commit.

cmnrd:
I worked on the lfc script, and I think I was able to make it much more robust. With the new design, the script checks first if it is in a developer setup (i.e. are the sources available) or in a pre-built package. Depending on this, it looks up the appropriate jar. The version of the jar is not fixed anymore. In case the script finds multiple options, it uses the latest version. If we are in a developer setup and there is no jar (or a rebuild is requested with `-r`), then the jar is build with gradle.

These changes also remove the patching from our packaging script. Thus, this closes #434.

With these changes, I also convinced myself that it is possible to have an lfc script that supports the rebuild flag and that works in a developer and a packaged context without modification. I am unsure what to do now. We can go for the rebuild feature in lfc, but then the build-lfc script is redundant and I would delete it. Or we use lfc purely for running the standalone compiler and extend the build-lfc script such that it can optionally also run lfc. What do you think?

Soroosh129:
I have a question: Why can't `lfc -r` just use `build-lfc` to avoid duplicate logic? They are after all in the same bin directory in a developer setup, and non-existent in a non-developer setup, so perhaps only the existence of `build-lfc` can be checked.

PR #533, opened 2021-09-23
Replaced files with submodules

lhstrh:
Opening this as a draft PR to track progress. To do:

 - [ ] adjust paths


lhstrh:
I've gone above and beyond to save all history for all of the files in the factored-out repos. It is simply not true that `pythontarget.c` has a one-commit history. Try `git log --follow` to account for files having been moved.

Issue #451, opened 2021-08-02
Federated Execution in TypeScript


lhstrh:
Currently, the `TypeScript` target is behind in its capability to support federated execution. The code generator needs to be updated. Aspects that needs work:
 - [ ] control reactions
 - [ ] data serialization (go beyond `Buffer`)
 - [ ] federated tests (none are present at the moment)
 - [ ] ...

hokeun:
To enable federated tests for the TypeScript target, I think we need to generate **a shell script to run the RTI and federates** as part of the TypeScript code generation process and use the shell script for testing.

This will require significant changes to:

1. the TypeScript code generation (to generate the shell script as well as .ts and .js files), and
2. testing (to run the shell script instead of running by changing the switch case in the "private void execute(LFTest test)" method in [org.lflang.tests/src/org/lflang/tests/runtime/TestBase.java](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang.tests/src/org/lflang/tests/runtime/TestBase.java)).

Here is an example shell script that runs the RTI and federates for the C target.

```
#!/bin/bash
# Launcher for federated HelloDistributed.lf Lingua Franca program.
# Uncomment to specify to behave as close as possible to the POSIX standard.
# set -o posix

# Enable job control
set -m
shopt -s huponexit

# Set a trap to kill all background jobs on error or control-C
# Use two distinct traps so we can see which signal causes this.
cleanup() {
    printf "Killing federate %s.\n" ${pids[*]}
    # The || true clause means this is not an error if kill fails.
    kill ${pids[@]} || true
    printf "#### Killing RTI %s.\n" ${RTI}
    kill ${RTI} || true
    exit 1
}
cleanup_err() {
    echo "#### Received ERR signal on line $1. Invoking cleanup()."
    cleanup
}
cleanup_sigint() {
    echo "#### Received SIGINT signal on line $1. Invoking cleanup()."
    cleanup
}

trap 'cleanup_err $LINENO' ERR
trap 'cleanup_sigint $LINENO' SIGINT

# Create a random 48-byte text ID for this federation.
# The likelihood of two federations having the same ID is 1/16,777,216 (1/2^24).
FEDERATION_ID=`openssl rand -hex 24`
echo "Federate HelloDistributed in Federation ID '$FEDERATION_ID'"
# Launch the federates:
echo "#### Launching the runtime infrastructure (RTI)."
# First, check if the RTI is on the PATH
[...]

PR #538, opened 2021-09-25
Revive basic federated execution for TypeScript target.

hokeun:
Now we can code-generate Distributed HelloWorld example and run the source and print in TypeScript.


org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
```diff
@@ -106,6 +106,20 @@ class TSConstructorGenerator (
         return connectionInstantiations.joinToString("\n")
     }
 
+    fun generateFederateConfigurations(): String {
```
Soroosh129, line 109: Would you please add documentation for this function? It is a bit unclear what it does.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -349,6 +355,13 @@ class TSReactionGenerator(
         if (reactor.isFederated) {
             generatedReactions = LinkedList<Reaction>()
             for (reaction in reactor.reactions) {
+                // TODO(hokeun): Find a better way to gracefully handle this skipping.
+                // Do not add reactions created by generateNetworkOutputControlReactionBody
```
Soroosh129, line 359: Can't the body of the control reactions just be empty for now, instead of skipping them altogether?

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -383,15 +382,92 @@ class TSGenerator(
...
+        |// generateNetworkReceiverBody
         |if (${action.name} !== undefined) {
-        |    ${receivingPort.container.name}.${receivingPort.variable.name} = 
-        |    ${action.name}; // defaults to utf8 encoding
+        |    ${receivingPort.container.name}.${receivingPort.variable.name} = ${action.name}.toString(); // defaults to utf8 encoding
         |}
         """.trimMargin()}
     }
 
+    /**
```
Soroosh129, line 394: Perhaps these functions are better put in a `TSFedGenerator.kt`.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -383,15 +382,92 @@ class TSGenerator(
...
+        sendingChannelIndex: Int,
+        receivingFed: FederateInstance,
+        type: InferredType,
+        isPhysical: Boolean,
+        delay: Delay?,
+        serializer: SupportedSerializers
+    ): String {
+        return with(PrependOperator) {"""
+        |// FIXME: For now assume the data is a Buffer, but this is not checked.
+        |// Replace with ProtoBufs or MessagePack.
```
Soroosh129, line 424: Just regarding the comment, the `SupportedSerializers` should be what dictates the serialization. There are both `native` and `proto` serializer options.

example/TS/src/DistributedHelloWorld/HelloWorld.lf
```diff
@@ -0,0 +1,72 @@
...
+    // Output type char* instead of string is used for dynamically
+    // allocated character arrays (as opposed to static constant strings).
+    output message:string;
+    state count:number(1);
+    // Send first message after 1 sec so that the startup reactions
+    // do not factor into the transport time measurement on the first message.
+    timer t(1 sec, 1 sec);
+    reaction(t) -> message {=
+        message = root + " " + count++;
+        console.log(`At time (elapsed) logical tag ${util.getElapsedLogicalTime()}, source sends message: ${message}`);
```
Soroosh129, line 52: This is interesting. I don't know how the TS target works, but shouldn't the `message` be set somewhere using an API? Is `=` doing the job of a `.set`?

hokeun: I think the way TypeScript works for now is to handle .set in the generated code. Here is an example: https://github.com/lf-lang/reactor-ts/blob/master/src/example/generated/DistributedLogical/Distributed_dsp.ts#L81

I don't think this is ideal, but I'm just using it as is since this is what's available for now.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
hokeun: These are functions that need to be overridden, so I put them in TSGenerator which inherits GeneratorBase. Do you have any specific suggestion?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
hokeun: Actually, I think it's cleaner to skip these reactions until we figure out what is the right thing to do, I mean in terms of implementation. We have to handle the special case of generating an empty body for these reactions and I'm concerned that will make the code much more complicated. Please let me know if my concern makes sense.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
Soroosh129: I see. You're right. It is unfortunate that these functions should end up in here. Perhaps we should rethink this design altogether and move these functions out of `GeneratorBase`

example/TS/src/DistributedHelloWorld/HelloWorld.lf
```diff
@@ -0,0 +1,72 @@
+/**
```
Soroosh129, line 1: I think if this example was turned into a test, it might be less of a headache to keep whatever is already working intact as we make changes to federated execution.

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
hokeun: Sure, added comments in https://github.com/lf-lang/lingua-franca/pull/538/commits/a64a9e9b9b9c1d56ed13a2ef01b42dce08c8e9c8.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Soroosh129: I think this string matching that is happening here is more prone to breakage than just generating an empty reaction. What if the name of the reaction changed? Are you worried about cluttering the generated code?

hokeun: I'm not sure if I'm following "what if the name of the reaction changed". This code doesn't depend on the name of the reaction.

It just uses the string created inside the overridden function, generateNetworkOutputControlReactionBody(). So as long as the TypeScript-specific code generation from generateNetworkOutputControlReactionBody() doesn't change, we're fine. 

What I'm worried about is that I may have to have temporary codes in multiple places (i.e., in multiple .kt) and it can become harder to maintain later.

May I ask what benefits we can get by having an empty body rather than skipping the reaction entirely?

hokeun: Ah, by the way, I'm not sure what generateNetworkOutputControlReactionBody() does. Would this be necessary for all targets? Or it may not be used at all by some targets?

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
hokeun: Ah, I just noticed SupportedSerializers was added here after I started working on the code and got added here when I pulled the recent master to this branch.

I updated this FIXME to use serializer in https://github.com/lf-lang/lingua-franca/pull/538/commits/c58677c8c28bb8b9b785856b0ee61d681d99a8e9.

By the way, this FIXME was here since I just copied and pasted the code used in the legacy TypeScript federated execution example (which doesn't work for now) here without thinking about it too much: https://github.com/lf-lang/reactor-ts/blob/master/src/example/generated/DistributedLogical/Distributed_dsp.ts

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Soroosh129: These reactions are essential to enable support for distributed cycles. You can sometimes receive PTAGs from the RTI and need these reactions in place to block certain reactions from execution using the input control reactions until you either receive a follow-up TAG or a port absent (via an output control reaction), which in both cases will unblock the input control reactions.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
Soroosh129: I think now that the TS target is getting attention, we should keep these TS federated functionalities up-to-date with any changes related to federated execution. I will make a mental note to look here whenever I make a change related to federated execution. Adding tests will also help with this matter. We should also make changes more collaboratively between all of us going forward.

Soroosh129: Just as a heads-up, I am going to make changes to how and where output control reactions are generated. These changes are needed for a correct functionality. I don't expect these changes to affect the TS target that much, but to err on the safe side, we can have a meeting about it if you are interested.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
hokeun: Got it. Thank you for your explanation! I think we will need to implement that functionality in reactor-ts first before supporting these reactions here.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
hokeun: Sure, having a meeting sounds good to me! I'd like to learn the overall communication and control flow between the RTI and federates as well. What would be a good time for a meeting for you? Maybe we could invite others who might be interested in as well? For the upcoming three weeks, I'll be mostly available before 6 PM in Pacific time.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
cmnrd: Yes, I think this is a symptom of our monolithic generator base. I had similar issues in the C++ target, for instance with the methods used for generating after delays.

example/TS/src/DistributedHelloWorld/HelloWorld.lf
hokeun: Thank you for the great suggestion. I tried to turn this into a test, but realized that we need to make significant changes to the TypeScript code generation and the testing process, to support the federated test for the TypeScript target. I planned out how to enable federated tests in TypeScript in my comment [here](https://github.com/lf-lang/lingua-franca/issues/451#issuecomment-928521464).

Do you think you could approve this pull request for now? I'm afraid that if the pull request is left for too long, it might be harder to merge later.

PR #541, opened 2021-09-27
Support after on connections composed of multiple single ports

cmnrd:
Currently, if we connect multiple individual ports (not multiports) like so `o1, o2 -> i1, i2`, then after delays do not work as expected. If only single ports are involved, the connection is still treated as a single port connection. This PR modifies the `isWide` method to also return true, if multiple ports are listed on either side. This small change is sufficient to get after delays working as expected on such connections. I also added a C and a C++ test.

### DIFF
--- a/org.lflang/src/org/lflang/ASTUtils.xtend
+++ b/org.lflang/src/org/lflang/ASTUtils.xtend
@@ -173,24 +173,19 @@ class ASTUtils {
     }
     
     /**
-     * Return true if any port on the left or right of the connection involves
-     * a bank of reactors or a multiport.
+     * Return true if the connection involves multiple ports on the left or right side of the connection, or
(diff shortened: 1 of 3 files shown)


edwardalee (approved):
Looks great! Thanks for fixing this.

Soroosh129:
I guess the Python target was overlooked?

Going forward, how should we handle these types of new tests? I'd be more than happy to add the Python-equivalent test in any given branch if the author pings me.

cmnrd:
That is a good question. I wasn't even aware that Python supports multiports, and I wouldn't have known to ping you. So I guess a good first step would be the feature matrix that we discussed a while ago.

Maybe we should just be a bit slower with merging and give everyone the opportunity to have a look. I think the code review is the perfect place to comment on missing tests.

Issue #336, opened 2021-05-16
Code Coverage with Kotlin


lhstrh:
It appears that Jacoco should be able to compute code coverage for our Kotlin code as well, see:http://vgaidarji.me/blog/2017/12/20/how-to-configure-jacoco-for-kotlin-and-java-project/
and
https://kevcodez.de/posts/2018-08-19-test-coverage-in-kotlin-with-jacoco/

lhstrh:
Closing this issue because CodeCov now lists the Kotlin files.

PR #543, opened 2021-09-27
Fix #537 - support warnings that span multiple lines in standalone compiler

### LINKED ISSUE #537: Formatting Multiline Errors/Warnings
Are per https://github.com/lf-lang/lingua-franca/pull/536, multi line issues are reported on the first line of the involved code segment, like so:

```
lfc: error: Reactor must be named.
--> ../sandbox/Foo.lf:3:1
  |
2 | 
3 | reactor {
  | ^^^^^^^^^ Reactor must be named.
  |
4 | 
```

It would be nice if we could indicate a range. Perhaps something like this:

```
lfc: error: Reactor must be named.
--> ../sandbox/Foo.lf:3-8
  |
2 | >>>>>>>>>
3 | reactor {
  | ...
8 | } 
  | < Reactor must be named.
```

Or this?

```
lfc: error: Reactor must be named.
--> ../sandbox/Foo.lf:3-8
  |
2 | vvvvvvvvv
3 | reactor {
  | ...
8 | } 
  | ^ Reactor must be named.
```

oowekyala:
This also adds tests for the way the standalone compiler formats messages.

This is a sample multiline warning:
```

lfc: error: Reactor must be named.
--> test/resources/org/lflang/lfc/tests/multilineWarning.lf:3:1
  |
2 |
  | >>>>>>>>>>>>>>
3 | reactor {
4 |
5 | }
  | <<<<<<<<<<<<<< Reactor must be named.
```

If the block is too long it's shortened with `...` 

Fix #537


.github/workflows/build.yml
```diff
@@ -48,6 +48,36 @@ jobs:
...
+        uses: actions/setup-java@v1.4.3
+        with:
+          # The Java version to make available on the path. Takes a whole or semver Java version, or 1.x syntax (e.g. 1.8 => Java 8.x). Early access versions can be specified in the form of e.g. 14-ea, 14.0.0-ea, or 14.0.0-ea.28
+          java-version: 11
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+      - name: Run standalone LFC tests
+        run: |
+          ./gradlew :org.lflang.lfc:test
+      - name: Report to CodeCov
```
lhstrh, line 68: I don't think this should be here. Only one job is supposed to report to CodeCov, and it has to be one that is driven by JUnit.

org.lflang.lfc/src/org/lflang/lfc/LFStandaloneModule.java
```diff
@@ -60,5 +61,13 @@ public void configure(Binder binder) {
         binder.bind(ErrorReporter.class).to(StandaloneErrorReporter.class);
         binder.bind(ReportingBackend.class).toInstance(helper);
         binder.bind(ValidationMessageAcceptor.class).to(StandaloneIssueAcceptor.class);
+        // This is required to force the ResourceValidator to
+        // use a new registry instance (which is reused by the injector as a singleton).
+        // Otherwise, it uses the static EValidator.Registry.INSTANCE which is bad
+        // as the first validator to be created would be persisted in that static instance.
```
lhstrh, line 67:
```suggestion
        // as the first validator to be created would persist in that static instance.
```

lhstrh (approved):
Very nice feature! It's also a great step forward to have tests for this. The only thing I worry about is the reporting to CodeCov. What I expect to happen is that whichever job finishes last will report it's coverage as overall coverage. I might be wrong, because I'm not completely sure how this works, but we should be on the look out for this...

lhstrh:
There's also this issue: https://github.com/lf-lang/lingua-franca/issues/336

org.lflang.lfc/test/resources/org/lflang/lfc/tests/tabs.stderr
```diff
@@ -0,0 +1,16 @@
...
+4 |     state foo(1);
+  |     ^^^^^^^^^^^^ State must have a type.
+  |
+5 |     state    bar(1);
+lfc: error: State must have a type.
+--> test/resources/org/lflang/lfc/tests/tabs.lf:5:2
+  |
+4 |     state foo(1);
+5 |     state    bar(1);
+  |     ^^^^^^^^^^^^ State must have a type.
```
oowekyala, line 14: actually there's a problem here which I should fix. The first warning also spans over the `(1)`, which I think is the correct behavior. This one doesn't.

lhstrh:
Why don't we make the tests you added part of the compiler tests? Is it because of Kotlin? We should really find a way to calculate code coverage over the entire code base...

lhstrh:
Maybe this helps? https://docs.codecov.com/docs/merging-reports

lhstrh:
Judging by what is described there, reporting is incremental by default, so it might actually just work the way you were doing it initially... Let's just give it a shot and see what happens?

oowekyala:
@lhstrh Our codecov integration looks broken anyway, this should probably be the focus of another fix PR.
- lf-lang/lingua-franca [has no coverage data](https://app.codecov.io/gh/lf-lang/lingua-franca/)
- the most recent data in icyphy/lingua-franca [is six months old](https://app.codecov.io/gh/icyphy/lingua-franca/)

Maybe this data is private and I need to log in?

Here's [an error in the latest successful build on master](https://github.com/lf-lang/lingua-franca/runs/3722262978#step:5:3702).

lhstrh:
RE: CodeCov: The move from the `icyphy` to `lf-lang` org required some configuration changes, but those changes didn't solve the underlying issue that apparently the reporting had been broken for quite a while already; still troubleshooting.

cmnrd (approved):
This looks good! Thanks for taking care of this and also adding tests.

> Why don't we make the tests you added part of the compiler tests? Is it because of Kotlin? We should really find a way to calculate code coverage over the entire code base...

I think it is a good idea to keep tests close to the source code. I am not sure how this is commonly handled in Java projetcs, but in Python projects it's quite common to have tests in every package. The tests here test lfc's ability to produce correctly formatted error messages. This shouldn't be a concern relevant to any other package.

Also, we need to be careful about dependencies here. Adding dependencies to Kotlin code from projects that we can compile from Eclipse will cause trouble. The `org.lflang.lfc` project is currently not even visible to our developer Eclipse.

lhstrh:
Looks like the stack trace option kills code coverage?

oowekyala:
> Looks like the stack trace option kills code coverage?

I'm not sure what you mean, but I think the reason we got a coverage report is that that commit is the first to this branch since your changes to code coverage reporting on master. I don't think it's related to the stack trace option itself

lhstrh:
Yes, you're right. What about the test failures in the standalone tests though?

lhstrh:
Judging by recent activity, it seems that this PR is subsumed by https://github.com/lf-lang/lingua-franca/pull/488? If this is the case, then I guess we should close this one.

lhstrh (approved):

lhstrh (changes requested):
Apart from the test failure that are still there, this looks good!

oowekyala:
> Yes, you're right. What about the test failures in the standalone tests though?

It seems to be failing only on windows, I'll need some time to reproduce it :'(

PR #544, opened 2021-09-27
Support list initializer syntax in Python

oowekyala:
Implement list literals for Python. Now:

- using `(1, 2, 3)` as an initializer creates a tuple (previously created a list)
- using `([1, 2, 3])` as an initializer creates a list

This makes LF syntax in line with python syntax itself and allows writing
- initializers with an empty list `([])`
- initializers with a one element list `([1])`

Previously these scenarios required fat braces, eg `{=[]=}`.

Using list literals in other targets than Python produces a validator error. When this is merged I'll follow up with the necessary changes for Rust to support this. TS could also support it.

The python code generator is simplified as there is no more special treatment for lists in parameters. Previously, initializing a parameter with `(1, 2)` created a tuple, while the same initializer for a state variable creates a list. I argue that this was hard to learn and not ergonomic, because
- the difference is not obvious to a user reading the code
- it breaks a useful refactoring flow, that abstracts a state variable initializer to a parameter. Namely, when you have the following:
```python
reactor R {
   state var(...)
}
```
you can easily abstract it by turning it into
```python
reactor R(p(...)) {
   state var(p)
}
```
and the code means the same, regardless of what "`...`" is. Previously, this was not true for python if `...` is a comma separated list, because the refactoring effectively makes the state variable be initialized to an immutable tuple instead of a mutable list, which may cause unexpected errors at runtime.

I understand that this was supposed to honor the principle that parameters should be immutable, whereas state variables are not. However I think as shown above, this corner case has potent downsides, and doesn't buy us safety anyway: you could always initialize your parameter to a mutable thing, eg `{=[1,2]=}`, put mutable objects inside the tuple fields, or even set the parameter variable itself (I think), as Python doesn't support final bindings.

### DIFF
--- a/test/Python/src/ListLiterals.lf
+++ b/test/Python/src/ListLiterals.lf
@@ -2,10 +2,19 @@
 // For this test, success is just compiling and running.
 target Python;
 
+
+reactor Sub(seq([1,2])) {
+
+}
+
 main reactor {
     state l12([1,2]);
     state empty([]);
     state empty2([,]);
+
+    sub  = new Sub(seq = [1, 2]);
+    sub2 = new Sub(seq = (1, 2));
+    sub3 = new Sub(seq = ([1, 2]));
 
     reaction(startup) {=
         assert self.empty == []
--- a/test/Python/src/ListLiterals.lf
+++ b/test/Python/src/ListLiterals.lf
@@ -1,0 +1,16 @@
+// This example illustrates state variables and parameters on the wiki.
+// For this test, success is just compiling and running.
+target Python;
+
+main reactor {
+    state l12([1,2]);
+    state empty([]);
+    state empty2([,]);
+
+    reaction(startup) {=
+        assert self.empty == []
+        assert self.empty2 == []
+        assert self.l12 == [1, 2]
+        print("Success")
+    =}
+}
--- a/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.xtend
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.xtend
@@ -611,7 +611,29 @@ class LinguaFrancaValidationTest {
             }
         }
     }
-    
+
+    @Test
+    def void forbidListLiterals() {
+        parseWithoutError('''
+            target C;
+            reactor Contained {
+                state x: int[]([1, 2]);
+            }
+        ''').assertError(LfPackage::eINSTANCE.listLiteral,
+            null, 'Target C does not support LF list literals')
+    }
+
+
+    @Test
+    def void allowListLiteralsInPython() {
+        parseWithoutError('''
+            target Python;
+            reactor Contained {
+                state x([1, 2]);
+            }
+        ''').assertNoErrors()
+    }
+
     
     /**
      * Tests for state and parameter declarations, including native lists.
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -341,10 +341,10 @@ Type:
 ;
 
 ArraySpec:
-    ofVariableLength?='[]' | '[' length=INT ']';
+    '[' ( ofVariableLength?=']' | length=INT ']' );
     
 WidthSpec:
-    ofVariableLength?='[]' | '[' (terms+=WidthTerm) ('+' terms+=WidthTerm)* ']';
+    '[' ( ofVariableLength?=']' | (terms+=WidthTerm) ('+' terms+=WidthTerm)* ']' );
     
 WidthTerm:
     width=INT 
@@ -527,7 +527,7 @@ Token:
     // Braces
     '(' | ')' | '{' | '}' |
     // Brackets
-    '[' | ']' | '<' | '>' | '[]' |
+    '[' | ']' | '<' | '>' |
     // Punctuation
     ':' | ';' | ',' | '.' | '::' |
     // Slashes
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -323,7 +323,7 @@ Expr :
 // If it is a constant, the validator should check that if the value
 // is non-zero, it is accompanied by a unit.
 Value:
-    (parameter=[Parameter] | time=Time | literal=Literal | code=Code);
+    (parameter=[Parameter] | time=Time | literal=Literal | list=ListLiteral | code=Code);
 
 
 Time:
@@ -383,6 +383,11 @@ Generic:
 
 SignedInt:
     INT | NEGINT
+;
+
+// list literals support even empty lists, and a trailing comma.
+ListLiteral:
+    '[' ( items+=Value (',' items+=Value)* )? ','? ']'
 ;
 
 Literal:
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -385,9 +385,14 @@ SignedInt:
     INT | NEGINT
 ;
 
-// list literals support even empty lists, and a trailing comma.
+// List literals support even empty lists, and a trailing comma.
+
+// Note: this expansion is weirdly written, but without the
+// emptyList property, Xtext fails to push a ListLiteral node
+// when the list is empty... Another bug of Xtext.
 ListLiteral:
-    '[' ( items+=Value (',' items+=Value)* )? ','? ']'
+    '[' (items+=Value (',' items+=Value)* ','? ']'
+        | ','? emptyList?=']')
 ;
 
 Literal:
--- a/org.lflang/src/org/lflang/Target.java
+++ b/org.lflang/src/org/lflang/Target.java
@@ -397,6 +397,14 @@ public enum Target {
         this(description, requiresTypes, "N/A", "N/A", keywords);
     }
 
+    /**
+     * Returns true if the target supports list literals in LF syntax.
+     * If not, list literals produce validator errors.
+     */
+    public boolean supportsLfListLiterals() {
+        return this == Python;
+    }
+
 
     /**
      * Return the target that matches the given string.
--- a/org.lflang/src/org/lflang/generator/PythonGenerator.xtend
+++ b/org.lflang/src/org/lflang/generator/PythonGenerator.xtend
@@ -29,7 +29,6 @@ package org.lflang.generator
 import java.io.File
 import java.util.ArrayList
 import java.util.LinkedHashSet
-import java.util.LinkedList
 import java.util.List
 import java.util.regex.Pattern
 import org.eclipse.emf.ecore.resource.Resource
@@ -47,12 +46,10 @@ import org.lflang.lf.Input
 import org.lflang.lf.Instantiation
 import org.lflang.lf.Model
 import org.lflang.lf.Output
-import org.lflang.lf.Parameter
 import org.lflang.lf.Port
 import org.lflang.lf.Reaction
 import org.lflang.lf.Reactor
 import org.lflang.lf.ReactorDecl
-import org.lflang.lf.StateVar
 import org.lflang.lf.TriggerRef
 import org.lflang.lf.Value
 import org.lflang.lf.VarRef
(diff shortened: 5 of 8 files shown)


Soroosh129:
Thanks for taking the time to simplify the Python generator. I like the simplified logic of the generator itself. However, I have the following concerns:
- As you correctly pointed out, the `()` initializer was interpreted as an immutable tuple for parameters and a mutable list for state variables to prevent serious error conditions that can occur when a parameter is changed in the body of the reaction, or when a reaction attempts to change a state variable, but fails (with a syntax error). I agree that the corner case of assigning parameters to a state variable as an initial value can be problematic. If this corner case is a concern, we can use the [`list()`](https://www.w3schools.com/python/ref_func_list.asp) constructor to convert a tuple parameter to a list when it is initially assigned to a state variable. However, I can see why this would be considered unergonomic to a Python programmer, which brings up my next point.
- I think that `state a([1,2,3])` is very unintuitive. With the updated syntax, defining a `state s(1,2,3)` will create an immutable tuple, which will produce a syntax error if an attempt is made to modify it. I would say the next natural thing to be tried by the user (assuming that they know Python) is `state s[1,2,3]`, which will fail with a cryptic error.  I don't think that it will ever occur to someone, unless they read the instructions carefully somewhere, that state variable array-types should almost always be defined as `state a([1,2,3])`, especially since the syntax for tuples is not `state a((1,2,3))`. I would say to be consistent, the syntax for defining array-type state variables should be one of the following:
    1. `state a((1,2,3))` for tuple and `state a([1,2,3])` for list
    2. `state a(1,2,3)` for tuple and `state a[1,2,3]` for list
    3. `state a({=(1,2,3)=})` for tuple and `state a(1,2,3)` for lists (this one is not so much consistent but is less error-prone than the proposal IMO).
My inclination would be more toward 2 or 3. With option 2, `state a(1,2,3)` should produce a warning. Similarly, using `a[1,2,3]` for a parameter should also produce a warning.

Finally, I suggest that the following two tasks must also be dealt with:
- [ ] Update [Writing-Reactors-in-Python](https://github.com/lf-lang/lingua-franca/wiki/Writing-Reactors-in-Python) to reflect the new syntax.
- [ ] Make sure that examples, experimental examples, and benchmarks all use the new syntax.

oowekyala:
Thanks for your review! I'll get busy with the tasks you outlined.

I don't find `([a,b,c])` unintuitive, as I understand it as "parens around an expression", where the expression is here the `[a,b,c]` list. In case of `(a,b,c)`, the expression is an unparenthesized tuple (which is perfectly valid python). In case of `(a)`, it's "just a". It's regular and easy to explain to a user.

The `state a[1,2,3]` syntax (proposition ii) would cause parsing ambiguity in languages that use types... Eg in `state a:int[2]`, do the brackets denote the array type or the initializer?

I would be fine with your proposition (i), but IMO a more satisfactory candidate would be to stop using parentheses to denote assignment, which to me is the unintuitive thing here. If we replace them with an `=` sign, we get the following pythong:
- `state a = [a, b]` for a list
- `state a = (a, b)` for a tuple
- `state a = 1` for an integer
- `state a = {= ... =}` for something weird 
- ...

The general grammar is just `state a = expr` for any `expr`. Then, what exactly an "expr" is can be restricted to expression forms that are close to the target language. Eg as per #492, it's not very useful to support list or tuple syntax in C. The equal sign syntax also plays well with typed languages, as indeed it's the syntax of many typed languages like MLs, Kotlin, Rust, (and also Python's type annotations). So I think it's the most intuitive syntax around. Lastly this syntax already exists in LF, as it's the one of constructor parameter assignments. The syntax `state x: t(a, b)` can mean "constructor call on type t" for C++ and other languages that have constructors.

If that is not an option, then I'd prefer your proposal (i) or my proposal.

cmnrd:
I agree with @oowekyala that our current syntax can be described as `state name: type(expr)`, or `state name: type(expr, expr, expr, ...)` for C++ constructors. In this regard, I find it intuitive (even as a Python programmer), to write `state foo([1, 2, 3])`. The parenthesis simply enclose the expression that is to be assigned. I think that this provides the best clarity. Using `state foo[3]` would be very confusing, as we use the `[N]` syntax to denote fixed size lists.

@oowekyala might be right about our `(expr)` syntax being unintuitive (at least for some languages). I would be fine with extending the grammar so that we can either write `state name(expr)` or `state name = expr`. Which option we use could be target specific. This would probably help to reduce the syntax clutter in some of our targets. However, I am not sure how the assignment syntax would play along with our type annotations. `state name: type = expr` looks a bit weird to me.

Soroosh129:
I also agree with @oowekyala that our current assignment syntax could use a bit of a boost. I think the end result of having `state a = [a, b]` for a list and `state a = (a, b)` for a tuple will be intuitive and easy to understand. One downside is that people might automatically assume that the right-hand side can contain any arbitrary target language expression without knowing about `{=...=}`. We might be able to mitigate this by a clear syntax error that points to using `{=...=}` if possible, I'm not sure.

> `state name: type = expr` looks a bit weird to me

How so? I think Swift, Scala, and Rust seem to be using this `var name: type = expr` type of assignment.

After reading your comments and trying out a few examples and patterns in Python, I am convinced that `()` and `([])` are intuitive enough, although just to be safe, can we allow both `()` and `(())` (both will be equivalent)?

cmnrd:
> How so? I think Swift, Scala, and Rust seem to be using this var name: type = expr type of assignment.

I guess it's just me then :). This syntax would definitely be more natural for Rust and Python programmers

edwardalee (approved):
This looks good to me.  Please be sure to update the wiki pages for the affected target languages so that this new syntax is documented.

Soroosh129:
@oowekyala Is this ready to merge?

cmnrd (approved):
This looks really good!

Should we open an issue for the assignment operator discussion? I think it's important to keep track of it.

Soroosh129:
Please don't forget to update the wiki and the examples and benchmarks before merging in.

oowekyala:
I've just updated the wiki. I reorganised the language specification page a bit (added an appendix about types and one about expressions, to clean up the section about parameter declarations, which had the burden of explaining everything since it's the first section on the page). Feel free to change things.

I just realised i forgot to update the python page though...

oowekyala:
Ok I think I'm done with the wiki, and AFAICS no more code needs to be updated

I'll merge this shortly and open the issue @cmnrd referred to

Soroosh129:
> Ok I think I'm done with the wiki, and AFAICS no more code needs to be updated
> 
> I'll merge this shortly and open the issue @cmnrd referred to

Thank you for updating the wiki. There is a sentence in the Python entry that is left unfinished:

> The initial value is given by the expression enclosed in parentheses. Many different

oowekyala:
thanks for spotting this, it's fixed now!

test/Python/src/TuplePrint.lf
```diff
@@ -1,5 +1,5 @@
-// Source produces a statically allocated array, which it passes
-// to Print. The destination references the array directly.
+// Source produces a tuple, which it passes to Print.
```
Soroosh129, line 1: `ArrayAsType` is a common test. Renaming it to `TuplePrint` causes the test framework to complain about the absence of this test. If you decide to keep the name, I think it should be moved to `test/Python/src/target`.

oowekyala: I think I'll just rename it back and keep a tab on it

test/Python/src/ListLiterals.lf
```diff
@@ -0,0 +1,26 @@
...
+
+main reactor {
+    state l12([1,2]);
+    state l12_trailing([1,2,]);
+    state empty([]);
+    // doesn't parse state empty2([,]);
+
+    sub  = new Sub(seq = [1, 2]);
+    sub2 = new Sub(seq = (1, 2));
+    sub3 = new Sub(seq = ([1, 2]));
```
lhstrh, line 18: This looks weird to me. What is the difference between this and line 16?

lhstrh: Is `sub3` a tuple of size one that has a list in it?

lhstrh (commented):
I'm a little confused about parameter assignments like `foo = ([1, 2])`.

lhstrh:
> I agree with @oowekyala that our current syntax can be described as `state name: type(expr)`, or `state name: type(expr, expr, expr, ...)` for C++ constructors. In this regard, I find it intuitive (even as a Python programmer), to write `state foo([1, 2, 3])`. The parenthesis simply enclose the expression that is to be assigned. I think that this provides the best clarity. Using `state foo[3]` would be very confusing, as we use the `[N]` syntax to denote fixed size lists.
> 

I agree with this, but this reasoning should not extend to parameter assignments which use a `=` instead of parentheses. In other words, I would like to make sure that `new Foo(bar=([1, 2]))` means that `bar` is assigned a tuple of size 1 that has a list of numbers in it. If this is not true (and we're effectively ignoring the outer parentheses, saying that the assignment is just a list of numbers) then I think we have a problem.

test/Python/src/ListLiterals.lf
oowekyala: There is no difference. Parameter assignment syntax just allows that currently. You can also write `= 1` or `= (1)`.

oowekyala:
> I agree with this, but this reasoning should not extend to parameter assignments which use a = instead of parentheses. In other words, I would like to make sure that new Foo(bar=([1, 2])) means that bar is assigned a tuple of size 1 that has a list of numbers in it. If this is not true (and we're effectively ignoring the outer parentheses, saying that the assignment is just a list of numbers) then I think we have a problem.

This was always a problem in the parameter syntax. You could always add useless parentheses.

I think it's counter intuitive that `(expr)` would not be the same thing as `expr`. Parentheses don't change the meaning of an expression in any language I know, they just change operator precedence. In Python, to create a tuple of size 1, you need a trailing comma: `(1,)`. In Rust, tuples of size 1 don't exist at all, `(T) = T`.

So I'm fine with having duplicate syntax in parameter assignment, especially if we're phasing out using parentheses for assignment anyway. Also, parenthesized expressions could become a useful expression form if at some point we introduce syntax for infix operators.

But, we should possibly make sure that `(1,)` creates a tuple properly in Python and not an integer, which I think is currently not the case.

lhstrh:
After some discussion with @Soroosh129, I'm convinced that this (and the current implementation of lists in LF) needs more discussion. I think it's better to hold off from merging this into master; I propose to table this until after the tutorial.

oowekyala:
Ok so this has grown quite a bit since last time.

One large refactoring that affects the diff is that `Value` is no longer a concrete AST node type, instead it's an interface implemented by all expression types we support. Implementors are:
- `CodeExpr` and `ParamRef`. These are new nodes that wrap a `Code` and `Parameter` node, respectively. These intermediary nodes are there to avoid having to make Parameter and Code implemented Value directly, when in other contexts they're actually not expressions.
 - `Literal`. Previously, `Value::getLiteral` directly returned a string. The `Literal` node has the same method now. It might be useful to split this later into different node kinds for different literals.
 - `ListExpr`: list expressions, eg `[]`, `[a, b]`
 - `TupleExpr`: tuple expressions, eg `()`, `(1,)`, `(1,2)`, but NOT `(1)`. Parenthesized expression, which match `(expr)` without trailing comma, are actually parsed as just the enclosed expression, without intermediary node to represent the parentheses.
 - `MulExpr`: replaces the previous Product and Quotient
 - `AddExpr`: replaces the previous Sum and Difference

The latter two are now Values, which means we can now use simple arithmetic everywhere. For instance:
```
state foo = 2 * 5
```
A validator check was added to make sure we can't write `(1,2) * [1]` or other nonsensical expressions. Arithmetic expressions don't support parameter references properly yet. I've done some nice changes in #488 that will make implementing this much easier, so I don't want to do this now. In the meantime, this feature is not super useful.

Btw I fixed the precedence of those arithmetic expressions. Previously, if you wrote `a - b - c`, instead of parsing as `(a - b) - c` it parsed as `a - (b - c)` which of course means `a - b + c`. It doesn't matter much because I think those productions were unused anyway.

Note that there is also a TODO that says `rename Value to Expr` in the grammar. Maybe this is a good time to do this...


Summary:
- Value is now an interface
- Arithmetic is theoretically allowed in initializers
- List expressions and tuple expressions are new expression forms. They're currently only allowed in Python.

Future work includes
  - purging GeneratorBase using #488
  - replacing WidthSpec contents with regular Value

----

The other large change is the new `=` initializer syntax. Since that syntax already existed for constructor arguments, I went ahead and factored it into an `Initializer` node. Initializer has three forms:
- `name(e_1, ..., e_n)`: the current form used everywhere in state and parameter declarations, seldom used in ctor arguments (but allowed).
- `name{e_1, ..., e_n}`: the braced form, only allowed in C++
- `name = e`: the new form, exactly one expression is required. If you want to write a tuple, then that expression is... a tuple expression: eg `name = (1,2)`.


Those three initializer forms are now allowed in the following three places:
- parameter declaration: `reactor R(a = 2)`, `reactor R(a(2))`, `reactor R(a{2})`
- state initializer: `state s = 2`, `state s(2)`, `state s{2}`
- ctor parameter: `new R(a = 2)`, `new R(a(2))`, `new R(a{2})`

Within a given target, *the three initializer forms are not necessarily equivalent*. (Though currently `name(e)` is equivalent to `name = e` in all targets (this is only true if there is a single expression, read on)).

However, within a given target, a given initializer form means the same thing regardless of *where* it is used. So in the following three fragments, whatever `...` is, it means the same thing in the three positions:
```python
state var(...)
reactor R(param(...))
r = new R(param(...))
```

Currently, since `name(e)` === `name = e`, we have an equivalence between the meaning of `reactor R(param(...))` and `new R(param = ...)`, which makes the new syntax backwards compatible with the old. But there is one exception: C arrays. It's still allowed to initialize a C array with `name(1,2)`, but not with `name = (1,2)` (including in ctor *calls*), because that's a tuple expression. There are different ways to provide an alternative here (eg `param: int[] = {1, 2}`, which looks more like C than `param: int[](1,2)`), but I don't think this PR should necessarily provide it yet.

Summary:
- there's a new initializer syntax with `=`
- there is nearly perfect backwards compatibility, so we should use it liberally
- there are consistency rules that are now explicitly laid out for future development of our expression syntax

opinion note: I think it's useful to not make the `=` syntax an alias for the `(...)` syntax, because it makes implementation and documentation easier. But this also means that C arrays should be initialized like `new R(param(1,2))` which looks like a method call. I think this is an argument for deprecating the `(...)` syntax in the future.

petervdonovan:
> NOTE: Adopting these changes will give @petervdonovan some extra work due to the changes it will require in the VS Code extension. The same probably holds for the IntelliJ plugin? Is there a way to create tests for these plugins so we'd detect it when they break due to changes like these?

It might break some semantic highlighting, but maybe nothing else. I could write tests in TypeScript specifically for that purpose and maintain the tests as changes like these come up, although admittedly this would be a non-issue if semantic highlighting came from the language server, as others have suggested it should. And yes, there are [recommendations](https://code.visualstudio.com/api/working-with-extensions/testing-extension) for how to create VS Code integration tests.

oowekyala:
Things that are diverging from the spec (#986)
* ctor calls allow the three assignment syntaxes and not just `=`
* unclear if c++ codegen for ctor calls matches the spec
* Python dictionaries are unsupported, though braced init lists are.

lhstrh:
Now that you're back in action, @oowekyala, should we try to revive this effort? I feel like there were a lot of good contributions in this PR. @cmnrd addressed some of them, I believe, but there's still work in this branch that hasn't made it to `master`...

cmnrd:
Right. I already briefly spoke with @oowekyala about this. I also still have it on my TODO List to create an issue detailing several action items on the way forward. I'll try to take care of this soon.

lhstrh:
Note that I silently moved this to release `0.5.0`. Do you think, @oowekyala and @cmnrd, that you would have the bandwidth to address these lingering issues this fall? I realize this PR is very stale at this point, but the ideas that were implemented here I think are important and should find their way into `master`...

PR #546, opened 2021-09-27
Set keepalive to true when physical actions exist.

Soroosh129:
Set `keepalive` to `true` if physical actions exist and the user has not explicitly set it to `true` or `false`. See #547.

Also produces a warning whenever both `keepalive` and `threads` are changed because of the presence of a physical action.


edwardalee (approved):
This looks right to me, especially since the user can override this and set keepalive to false.  In particular, a (peculiar) use of physical actions would schedule them within a reaction rather than externally, in which case one might want keepalive set to false.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -284,21 +284,47 @@ abstract class GeneratorBase extends AbstractLFValidator {
...
+                if (targetConfig.threads < 1) {
+                    targetConfig.threads = 1
+                    errorReporter.reportWarning(
+                        target, 
+                        '''Using the threaded C runtime to allow for asynchronous handling of«
+                        » physical action «action.name».'''
+                    );
+                }
+                // Check if the user has explicitly set keepalive to false or true
+                if (target.config?.pairs?.filter[
```
lhstrh, line 308: I don't think this is the right place to do this. I would just add a member `keepAliveSpecified` or something like that to `TargetConfig` that indicates as to whether the property was set explicitly.

lhstrh (changes requested):
Looks like a sensible policy, but we should not be iterating through the AST to figure out whether a target property was set. If we want to generalize the solution that I suggested in my comment, we could have a `Map<TargetProperty, boolean> isSpecified` in `TargetConfig` that tracks for each property whether it has been set.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
lhstrh: See below.

Soroosh129:
> If we want to generalize the solution that I suggested in my comment, we could have a `Map<TargetProperty, boolean> isSpecified` in `TargetConfig` that tracks for each property whether it has been set.

I like this suggestion a lot. I added a list called `setByUser` to keep track of this.

lhstrh:
I suppose a list works too, but why not use a map? This is mostly about finding out whether a _particular_ property is set, which requires iterating through the list. I know the time complexity doesn't really matter because it's a short list, but still...

org.lflang/src/org/lflang/TargetConfig.xtend
```diff
@@ -35,6 +35,11 @@ import org.lflang.lf.TimeUnit
  * @author{Marten Lohstroh <marten@berkeley.edu>}
  */
 class TargetConfig {
+    
+    /**
+     * Keep track of every target property that is explicitly set by the user.
+     */
+    public List<TargetProperty> setByUser = newLinkedList;
```
lhstrh, line 42: Good name!

lhstrh:
Actually, you want to use a set, not a map.

Soroosh129:
> I suppose a list works too, but why not use a map? This is mostly about finding out whether a _particular_ property is set, which requires iterating through the list. I know the time complexity doesn't really matter because it's a short list, but still...

Perhaps a set is more optimized than a list, but adding it will add a whole new dependency to `TargetConfig`.  Most likely for such a small list, the difference in performance is marginal.

lhstrh:
This is Java, not C. We should use data structures that are optimal for their purpose. Dependencies on standard Java collections implementations are a nonissue.

Soroosh129:
It looks like the consensus might be that this is a sensible design. I would still encourage anyone to take a look at the related discussion in #547.

lhstrh (approved):
Looks good to me! I'll go ahead and merge this.

PR #549, opened 2021-09-27
Respond to cancel

### LINKED ISSUE #368: The build process should be interruptable in the LF IDE
Currently, when opening an existing project in the LF IDE, it will compile all the `*.lf` files within the project. This can be quite annoying and unfortunately there is no way for interrupting the build process. While the IDE shows a little stop button in the Progress view, pressing it has no effect.

petervdonovan:
Fixes #368.

- Since user cancellations already did have an effect -- they prevent additional LF files from being compiled -- the problem must be that compilation finishes for the current file, which can take time (especially for C++ and TypeScript). This pull request is to abort slow processes involved in compilation.
- Regrettably, some form of polling/busy-waiting seems necessary in order to accomplish this because
   1. A blocking `read` operation from an `InputStream` cannot be interrupted without risking undefined behavior
   2. No notification is sent when the current operation is marked as canceled.
- It is not obvious which operations are slow enough to require changes. For example, based on the `test` directory, I found no examples where C compilation was slow enough to justify interruption of CMake.


org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -42,6 +49,9 @@
  */
 public class LFCommand {
 
+    private static final int PERIOD = 200;
```
cmnrd, line 52: It would be great to have a brief explanation here as to what exactly these constants control and which unit they are in.

org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -73,74 +83,59 @@
...
+            collectOutput(process.getErrorStream(), errors, System.err);
         }
     }
 
 
     /**
      * Execute the command while forwarding output and error streams.
      * <p>
-     * Executing a process directly with `processBuiler.start()` could
+     * Executing a process directly with `processBuilder.start()` could
```
cmnrd, line 138: Could you also add an explanation here on how the cancellation works?

cmnrd (commented):
This is great, thanks for fixing this!

> It is not obvious which operations are slow enough to require changes. For example, based on the test directory, I found no examples where C compilation was slow enough to justify interruption of CMake.

Is there a downside for using it on all commands? I think we should also use it on the compile commands in the C generator. Even if our tests compile fast, users could write larger programs, maybe with additional library files that compile longer.

petervdonovan:
Thank you for the comments, @cmnrd. I just pushed some commits to try to address them.

cmnrd (approved):
Looks good!

PR #551, opened 2021-09-28
Allow empty target properties block

oowekyala:
This is especially handy when you comment out a target property. For now if the block happens to be empty you get a really unfriendly parsing error:
```
lfc: error: no viable alternative at input '}'
--> ../sandbox/R.lf:3:1
  |
2 |  // timeout: 3 sec
3 | };
  | ^^ no viable alternative at input '}'
  |
4 | main reactor R(p: int(2)) {

lfc: error: missing '}' at ';'
--> ../sandbox/R.lf:3:2
  |
2 |  // timeout: 3 sec
3 | };
  |  ^^ missing '}' at ';'
  |
4 | main reactor R(p: int(2)) {

lfc: fatal error: Aborting due to 2 previous errors
```

With this change, this produces no error anymore. If there is a need to produce an error (which here, I don't think there is) it should be produced by the validator instead.

Note that this also allows empty dictionary values, eg
```
target C {
   clock-sync-options: { /* everything defaulted */ }
}
```
I think this is fine for the same reasons (commenting out bindings).

I think it would be nice to make arrays in target properties accept empty arrays too. But this would require that the value of target properties be checked more thoroughly, so I'm leaving this for later.

---

Note: this branch is also merged in the PR #488


cmnrd (approved):
This was bugging me as well, but I never got to opening an issue or fixing it. Thanks!

I think we should be aware that this grammar rule also allows `target Foo { , }`, but this is probably fine.

Soroosh129:
Thanks for fixing this rather annoying issue.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -251,12 +251,12 @@ Serializer:
 /////////// For target parameters
 
 KeyValuePairs:
-    '{' pairs+=KeyValuePair (',' (pairs+=KeyValuePair))* ','? '}';
+    '{' ( pairs+=KeyValuePair (',' (pairs+=KeyValuePair))* )? ','? '}';
```
lhstrh, line 254:
```suggestion
    {KeyValuePairs} '{' (pairs+=KeyValuePair (',' (pairs+=KeyValuePair))* ','?)? '}';
```
I'm OK with a trailing comma for convenience, but `{ , }` just looks sloppy. This grammar rule accepts `{ }` and `{ timeout: 3 sec, }` but not `{ , }`.

org.lflang.tests/src/org/lflang/tests/LFParsingTest.java
```diff
@@ -26,6 +26,17 @@
     private ParseHelper<Model> parseHelper;
 
 
+    @Test
+    public void testLexingEmptyTargetProperties() throws Exception {
+        assertNoParsingErrorsIn("target C { };      \nreactor Foo {}");
+        assertNoParsingErrorsIn("target C {,};      \nreactor Foo {}");
```
lhstrh, line 32: With my suggested change, this would yield something like:
```
lfc: error: extraneous input ',' expecting '}'
--> foo.lf:1:12
  |
1 | target C { , }
  |            ^ extraneous input ',' expecting '}'
  |
2 | 
```

lhstrh (changes requested):
This is a good idea! I left a suggestion for improvement.

org.lflang/src/org/lflang/LinguaFranca.xtext
oowekyala: Ok I agree `{,}` should be avoided, but what does this `{KeyValuePairs}` mean?

lhstrh: It's an explicit "action" that forces the creation of an empty pairs object if no pairs were found.

lhstrh: Try it out :-)

oowekyala: Very cool!! This will actually help with #544

lhstrh (approved):
Awesome! Let's merge this.

PR #552, opened 2021-09-28
Support cmake include files with spaces in path

cmnrd:
This is a quick fix to support paths with spaces in cmake includes. Currently, spaces in paths will trigger syntax errors in cmake


Soroosh129 (approved):
Thanks for fixing this bug! It looks good to me.

PR #553, opened 2021-09-28
Support after delays on broadcast connections in C++

cmnrd:
Currently, after delays do not work on broadcast connections (using the `(...)+` syntax). In part, this is because the `iterated` property of the connection is not preserved in our AST transformation. I fixed this, so that the newly generated upstream connection is also iterated if the original connection was iterated. This solves the issue for C++. However, a similar test written in C still fails, and I haven't found out why. I will push the test that I have to `broadcast-after-c` based on this branch, and will open a separate issue.

This PR also does a few cleanups and adds C++ tests.

### DIFF
--- a/org.lflang/src/org/lflang/ASTUtils.xtend
+++ b/org.lflang/src/org/lflang/ASTUtils.xtend
@@ -216,7 +216,7 @@ class ASTUtils {
         output.variable = delayClass.outputs.get(0)
         upstream.leftPorts.addAll(connection.leftPorts)
         upstream.rightPorts.add(input)
-        upstream.isIterated = connection.isIterated
+        upstream.iterated = connection.iterated
         downstream.leftPorts.add(output)
         downstream.rightPorts.addAll(connection.rightPorts)
 
--- a/org.lflang/src/org/lflang/ASTUtils.xtend
+++ b/org.lflang/src/org/lflang/ASTUtils.xtend
@@ -216,6 +216,7 @@ class ASTUtils {
         output.variable = delayClass.outputs.get(0)
         upstream.leftPorts.addAll(connection.leftPorts)
         upstream.rightPorts.add(input)
+        upstream.isIterated = connection.isIterated
         downstream.leftPorts.add(output)
         downstream.rightPorts.addAll(connection.rightPorts)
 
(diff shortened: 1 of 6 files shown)


org.lflang/src/org/lflang/ASTUtils.xtend
```diff
@@ -177,7 +177,7 @@ class ASTUtils {
      * if the port on the left or right of the connection involves a bank of reactors or a multiport.
      * @param connection The connection.
      */
-    private static def boolean isWide(Connection connection) {
+    static def boolean isMultiportConnection(Connection connection) {
```
edwardalee, line 180: The name isMultiportConnection for this method seems misleading because the connection may not involve any multiports and the method could still return true.

cmnrd: I found `isWide` also not very clear, because it could be multiple individual connections in case of `o1, o2 -> i1, i2`. Do you have a suggestion?

test/Cpp/src/multiport/Broadcast.lf
```diff
@@ -0,0 +1,27 @@
+target Cpp;
+
+reactor Source {
```
Soroosh129, line 3: I think at least the `Source` reactor (but also a parametrized version of `Sink`) can be imported for a bit of a more minimalistic design. This is by no means a critical issue though.

Soroosh129 (approved):

org.lflang/src/org/lflang/ASTUtils.xtend
edwardalee: How about isMultipleConnection?

test/Cpp/src/multiport/Broadcast.lf
cmnrd: Yep, having better guidelines for writing tests and establishing a set of basic library reactors would be nice.

org.lflang/src/org/lflang/ASTUtils.xtend
cmnrd: I renamed it to `hasMultipleConnections`, which seems better from a grammatical point of view.

PR #555, opened 2021-09-28
Add gradle task to run a single test from command line

oowekyala:
One can use it using `gradle runSingleTest --args test/Python/src/Minimal.lf`

TODO
- [x] document it somewhere in the wiki (or test/README.md)

I'm probably going to merge this into #488

### DIFF
--- a/org.lflang.tests/build.gradle
+++ b/org.lflang.tests/build.gradle
@@ -50,4 +50,13 @@ test {
     maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
 	useJUnitPlatform()
     finalizedBy jacocoTestReport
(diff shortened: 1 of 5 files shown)


cmnrd (approved):
Cool! I think we should add a description to both the wiki and the README.

We should probably also implement a test for this functionality as part of #320.

oowekyala:
@cmnrd I added doc in [this page](https://github.com/lf-lang/lingua-franca/wiki/Regression-Tests) and copied part of it to the README.md. This duplicates info, but, fixing the organisation of this doc is part of #478 I think

lhstrh (approved):
This looks great! Thanks for also jotting down instructions in `README.md`.

PR #556, opened 2021-09-29
Attempt at fixing concurrency bug in C runtime.

lhstrh:


Soroosh129:
I think this resolved the nondeterminism. Thanks Marten!
I will mark this PR as ready to review since the C tests pass. The Windows Python tests currently fail, but since this branch is merging into `windows-c-target`, I think we should defer fixing these these errors to that branch.

org.lflang/src/lib/core/reactor_threaded.c
```diff
@@ -721,6 +721,17 @@ void request_stop() {
     lf_mutex_unlock(&mutex);
 }
 
+/**
+ * Return true of r1 has precedence over r2, false otherwise.
```
lhstrh, line 725:
```suggestion
 * Return true if the first reaction has precedence over the second, false otherwise.
 * @param r1 The first reaction.
 * @param r2 The second reaction.
```

lhstrh (commented):
@Soroosh129: Thanks for catching the missing return statement, and for discovering this problem in the first place!

edwardalee (commented):
While I think this strategy may work, I no longer understand exactly how we are defining chain IDs. The comment documenting _lf_is_blocked_by_executing_reaction() is no longer correct. It gives my definition of chain ID, not the one that is apparently implemented.  Also, the comment to the assignChainIDs method of ReactionInstanceGraph says, "The assigned IDs are such that the bitwise conjunction between two chain IDs is always nonzero if there exists a dependency between them."  This comment also needs to be changed to give the new definition, but I'm not sure what that new definition should be.

PR #564, opened 2021-10-02
Python target: Allow the execution of async Python threads concurrently with the C runtime

Soroosh129:
Fixes #563.

Issue #563, opened 2021-10-01
python target does not schedule physical action when timer reaction is absent. 


housengw:
Here's a minimal example that reproduces the problem:

With the timer reaction commented out:
```
target Python {
    timeout: 2 sec,
    keepalive: true 
};

main reactor {
    preamble {=
        import time
        import threading

        def take_time(self, a):
            self.time.sleep(0.1)
            a.schedule(0)
    =}
    timer t(0, 200 msec);
    physical action a(100 msec);
  
    reaction(startup) -> a {=
        x = self.threading.Thread(target=self.take_time, args=(a,))
        x.start()
    =}
    
//    reaction(t) {=
//        # do nothing
//        pass
//    =}
    
    reaction(a) {=
        print("async call back!")
    =}
}
```
Output:
```
---- Start execution at time Fri Oct  1 16:45:04 2021
---- plus 444451000 nanoseconds.
---- Elapsed logical time (in nsec): 2,000,000,000
---- Elapsed physical time (in nsec): 2,011,236,000
```

After uncommenting the timer reaction, the program outputs:
```
---- Start execution at time Fri Oct  1 16:49:03 2021
---- plus 865908000 nanoseconds.
async call back!
---- Elapsed logical time (in nsec): 2,000,000,000
---- Elapsed physical time (in nsec): 2,013,625,000
```

Soroosh129:
Thanks for reporting this issue @housengw. I have created a fix in #564.

PR #565, opened 2021-10-02
add piano demo for python target

housengw:


edwardalee:
I get a segfault when I try to run this demo:
```
EALMAC:Piano eal$ python3 /Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/Piano.py
---- Start execution at time Sat Oct  2 07:33:03 2021
---- plus 774026000 nanoseconds.
starting PianoGui...
Fatal Python error: (pygame parachute) Segmentation Fault

Thread 0x00000001127ec5c0 (most recent call first):
  File "/Users/eal/anaconda3/lib/python3.7/site-packages/mingus/midi/pyfluidsynth.py", line 258 in start
  File "/Users/eal/anaconda3/lib/python3.7/site-packages/mingus/midi/fluidsynth.py", line 64 in start_audio_output
  File "/Users/eal/anaconda3/lib/python3.7/site-packages/mingus/midi/fluidsynth.py", line 130 in init
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 143 in __init__
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 102 in start_gui
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/process.py", line 99 in run
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/process.py", line 297 in _bootstrap
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/popen_fork.py", line 74 in _launch
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/popen_fork.py", line 20 in __init__
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/context.py", line 277 in _Popen
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/context.py", line 223 in _Popen
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/process.py", line 112 in start
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 57 in start_sensor_simulator
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/Piano.py", line 56 in reaction_function_0
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/Piano.py", line 90 in main
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/Piano.py", line 95 in <module>
Exception in thread Thread-1:
  File "/Users/eal/anaconda3/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/Users/eal/anaconda3/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 78, in sensor_receiver
    key_packet = pout.recv()
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 250, in recv
    buf = self._recv_bytes()
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 407, in _recv_bytes
    buf = self._recv(4)
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 383, in _recv
    raise EOFError
EOFError
```

Soroosh129:
@edwardalee Here is what I had to do to get the demo to work (admittedly it's quite a lot):
```bash
$ brew install sdl2 sdl2_gfx sdl2_image sdl2_mixer sdl2_net sdl2_ttf
$ brew install Caskroom/cask/xquartz
$ brew install fluid-synth
$ python3 -m pip install mingus
$ python3 -m pip install pygame
```
And I had to go to **System Preferences -> Security & Privacy -> Privacy -> Input Monitoring** and check **Terminal** so that the Terminal application can read my keystrokes.

example/Python/src/Piano/sensor_simulator.py
```diff
@@ -0,0 +1,333 @@
+from multiprocessing import Process, connection
+from threading import Thread
+from os import sys
+import os
+os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = "hide"
+try: 
+    from mingus.containers.note import Note
+    from mingus.midi import fluidsynth
```
Soroosh129, line 8: It seems like `fluidsynth` has to be installed separately (via a package manager perhaps). Therefore, even after installing `mingus`, I would still get `"Import Error: Failed to import 'mingus'. Try 'pip3 install mingus'"`, but I think the message should be to install `fluidsynth` (e.g., on macOS, it's `brew install fluid-synth` and most likely `sudo apt-get install -y fluidsynth` on Linux) if the `fluidsynth` import fails..

With that in mind, I'm not sure how it might work on Windows since it doesn't have a robust package manager to install `fluidsynth`.

example/Python/src/Piano/Piano.lf
```diff
@@ -0,0 +1,81 @@
...
+    =}
+    
+    reaction(shutdown) {=
+        print("PianoGui: reactor shutting down")
+    =}
+    
+    reaction (t) {=
+        # do nothing. 
+        # I need to put this reaction here for logical time to progress.
+        pass
```
Soroosh129, line 62: Fixed in #564 I think.

Soroosh129 (approved):
This is a really cool demo! Thanks for providing extensive documentation in `sensor_simulator.py`.

My suggestion is to put the MacOS-specific requirements in a README.md. I will also try this demo on Linux and Windows shortly and post a requirements list here as well.

example/Python/src/Piano/sensor_simulator.py
housengw: I added messages in the import statement according to your suggestions:

```python3
try:
    from mingus.midi import fluidsynth
except:
    if sys.platform == "darwin":
        print("Import Error: fluidsynth is missing. Try 'brew install fluidsynth'")
    elif sys.platform == "linux" or sys.platform == "linux2":
        print("Import Error: fluidsynth is missing. Try 'sudo apt-get install -y fluidsynth'")
    else:
        print("Import Error: fluidsynth is missing. ")

```

edwardalee: Wow, sdl2 is pretty cool. I wonder why I didn't find that when I was looking for a way to do the Rhythm demo.  It's not clear to me why xquartz is needed if you have sdl2, since sdl2 apparently does cross-platform graphics using native libraries.

Soroosh129: I think installing `xquartz` might be optional: https://github.com/pygame/pygame/issues/2452#issuecomment-864364494

lhstrh:
> @edwardalee Here is what I had to do to get the demo to work (admittedly it's quite a lot):
> 
> ```shell
> $ brew install sdl2 sdl2_gfx sdl2_image sdl2_mixer sdl2_net sdl2_ttf
> $ brew install Caskroom/cask/xquartz
> $ brew install fluid-synth
> $ python3 -m pip install mingus
> $ python3 -m pip install pygame
> ```
> 
> And I had to go to **System Preferences -> Security & Privacy -> Privacy -> Input Monitoring** and check **Terminal** so that the Terminal application can read my keystrokes.

This is where a package manager would be really helpful. I may have mentioned this before, but we could call a package a `cask` and we could call our tool `barrel`.

example/Python/src/Piano/README.md
```diff
@@ -0,0 +1,21 @@
...
+$ brew install sdl2 sdl2_gfx sdl2_image sdl2_mixer sdl2_net sdl2_ttf
+$ brew install Caskroom/cask/xquartz
+$ brew install fluid-synth
+$ python3 -m pip install mingus
+$ python3 -m pip install pygame
+```
+Then, go to ```System Preferences -> Security & Privacy -> Privacy -> Input Monitoring``` and check Terminal so that the Terminal application can read keystrokes.
+
+### Linux:
+```
```
lhstrh, line 17:
```suggestion
```
 $ pip3 install mingus
 $ pip3 install pygame
```

**Comment by Soroosh129** [comment]
I get the following error on Linux, even after installing `jack`
```
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
jackd 0.125.0rc1
Copyright 2001-2009 Paul Davis, Stephane Letz, Jack O'Quinn, Torben Hohn and others.
jackd comes with ABSOLUTELY NO WARRANTY
This is free software, and you are welcome to redistribute it
under certain conditions; see the file COPYING for details

JACK compiled with System V SHM support.
loading driver ..
creating alsa driver ... hw:0|hw:0|1024|2|48000|0|0|nomon|swmeter|-|32bit
ALSA: Cannot open PCM device alsa_pcm for playback. Falling back to capture-only mode
cannot load driver module alsa
no message buffer overruns
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
fluidsynth: error: Failed to connect to Jack server.
```

The game then proceeds to open but doesn't produce any sound. I did finally manage to make this error go away by:

```bash
sudo apt install jackd2
pulseaudio --kill
jack_control start
```
but I still don't get any sound. The only useful output that I get (that is not printed by the game itself) is at the beginning: `fluidsynth: Jack sample rate mismatch, adjusting. (synth.sample-rate=44100, jackd=48000)`. 

I tried to fiddle with the sound output device that is being used using `qjackctl` with no luck. Is there anyway we have control over what audio server/driver is being used? `puleseaudio` would be nice :). I can see why Linux gaming hasn't caught on yet...

Soroosh129:
**An update about Linux:**
I was able to get the demo to produce sound by making a simple change in `sensor_simulator.py:153`:
```python
- if not fluidsynth.init(SF2):
+ if not fluidsynth.init(SF2, "alsa"):
```
I think the second argument can be OS-specific.

**An update about Windows:**

While installing `pygame` and `mingus` was easy, installing FluidSynth is not. 
It looks like it either has to be compiled with [CMake ](https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake) or binaries need to be downloaded from [here](https://github.com/FluidSynth/fluidsynth/releases/latest) and then added to PATH.  Neither has worked for me so far. I will try to debug this further tomorrow.

But I think the takeaway might be that `mingus` (or `fluidsynth` rather) is not very portable maybe.

example/Python/src/Piano/README.md
Soroosh129, line 17:
```suggestion
```bash
 $ sudo apt install libsdl2-2* libsdl2-image-2*
 $ sudo apt install fluidsynth
 $ pip3 install mingus
 $ pip3 install pygame
```

Soroosh129: I've added the `libsdl` and `fluidsynth` requirements to @lhstrh's suggestion. This resulted in a working demo on my Ubuntu Linux machine with [this](https://github.com/lf-lang/lingua-franca/pull/565#issuecomment-933153502) fix.

housengw: the ```if not fluidsynth.init(SF2, "alsa"):``` line from ```sensor_simulator.py``` basically calls functions from [this file](https://github.com/bspaans/python-mingus/blob/master/mingus/midi/fluidsynth.py).

```Python
 def start_audio_output(self, driver=None):
        """Start the audio output.
        The optional driver argument can be any of 'alsa', 'oss', 'jack',
        'portaudio', 'sndmgr', 'coreaudio', 'Direct Sound', 'dsound',
        'pulseaudio'. Not all drivers will be available for every platform.
        """
        self.fs.start(driver)
```
from the comments in the file, it seems the "alsa" fix is indeed platform dependent.

edwardalee:
This example still segfaults for me with the same stack trace as above after following all the MacOS instructions in the README file except the one about permissions.  My MacOS has no `Input Monitoring` entry in `System Preferences -> Security & Privacy -> Privacy` (running MacOS Mojave). Google search for this problem reveals nothing.

Soroosh129:
> This example still segfaults for me with the same stack trace as above after following all the MacOS instructions in the README file except the one about permissions. My MacOS has no `Input Monitoring` entry in `System Preferences -> Security & Privacy -> Privacy` (running MacOS Mojave). Google search for this problem reveals nothing.

This could potentially be related to #563 according to this: https://stackoverflow.com/a/64309867/783868
I've merged master into this branch which contains a fix to #563.

I've also noticed that you are using a Python installed using Anaconda. Have you tried using brew to install a more recent version: https://formulae.brew.sh/formula/python@3.9

housengw:
I downloaded anaconda and got the same segfault message as @edwardalee's, but I was able to resolve the segfault by setting the "start method" of the multiprocessing module according to [here](https://stackoverflow.com/questions/30669659/multiproccesing-and-error-the-process-has-forked-and-you-cannot-use-this-corefou)

edwardalee:
Cool. This now starts up for me and I can play sounds. It isn't clear how to stop the program. If I control-C it, I get the following errors. Looking at the code, it looks like a key `None` will request a stop. But what is `None`?
```
---- Elapsed logical time (in nsec): 52,285,000,000
---- Elapsed physical time (in nsec): 52,285,463,000
Process Process-1:
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 116, in start_gui
    gui.start()
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 349, in start
    pygame.display.update()
KeyboardInterrupt
Exception in thread Thread-1:
  File "/Users/eal/anaconda3/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/Users/eal/anaconda3/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/eal/lingua-franca-master/git/lingua-franca/example/Python/src-gen/Piano/Piano/sensor_simulator.py", line 286, in actuator_receiver
    key_packet = self.actuator_pout.recv()
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 250, in recv
    buf = self._recv_bytes()
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 407, in _recv_bytes
    buf = self._recv(4)
  File "/Users/eal/anaconda3/lib/python3.7/multiprocessing/connection.py", line 383, in _recv
    raise EOFError
EOFError
```

housengw:
closing the GUI window will let pygame detect a pygame.QUIT event, which makes the sensor thread send a "None" to the LF program and shutting down everything else. Currently, that's the only way to gracefully shutdown the program.

edwardalee:
Ah, OK. Maybe it would be better to also intercept Control-C and Control-D to result in an orderly shutdown.

I had a look at the code, and it seems that there is a busy-wait loop that probably should be event triggered rather than busy waiting. My CPU usage remains at 140% even when nothing is happening. Is this because of the 1 msec timer you added? It seems that if I remove that timer, the CPU usage is still high, however, so it seems not. Is `pygame.event.get()` a blocking call? I'm suspecting not, in which case the `start()` function is performing a busy wait. 

There are four distinct things happening in this application:
1. Reading key presses
2. Translating key characters to musical notes
3. Updating the keyboard graphic
4. Playing the sound.
I think it would be interesting to separate these into four distinct reactors. That would enable a more modular design that would make it easier to elaborate. For example, one could create an automated accompanist that turns each key press into a riff, or, for each keypress, play a rhythmic pattern and align all the rhythmic patterns so that they are on beat. These would show off the timing capabilities in LF.

1. 

1.

example/Python/src/Piano/Piano.lf
```diff
@@ -0,0 +1,78 @@
+// Preambles are put inside the Python reactor class
+// Therefore, accessing them requires the use of self.
+target Python {
+    files: [sensor_simulator.py, keys.png, soundfont.sf2],
```
lhstrh, line 4: The soundfont file is rather large and is really slowing down checkouts and fetches. We should probably use something like LFS (see: https://git-lfs.github.com/) in situations like these.

lhstrh (changes requested):
This looks good, but the large binary file `soundfont.sf2` should not be in the repo. I've provided a suggestion as to how we can deal with it, but I haven't tried that method myself. Alternatively, we can let the user retrieve it themselves from where ever you downloaded it from.

housengw:
I removed the SoundFont file and added a link to the SoundFont used for testing the demo as a fix for now.

PR #566, opened 2021-10-04
Convert some classes to Java

oowekyala:
This converts a bunch of Xtend classes to Java, normally, without semantic changes.

One significant difference is that TimeValue now doesn't use BigInteger at all. Previously we used a BigInteger for the number of nanoseconds, but we were still capping the maximum time value to `1 << 64` nanoseconds. This makes sense, since we can't use infinite precision numbers like BigInteger in all target languages, so we have to set a limit somewhere. Also this maximum representable time value is roughly 292 years so it's probably enough for most use cases :) I used this to drastically clean up the code.

Another difference is that some of the classes that were declared side by side in TargetConfig.xtend are now nested within the class TargetConfig, so some imports need to change.


org.lflang/src/org/lflang/TimeValue.java
```diff
@@ -0,0 +1,168 @@
...
+***************/
+
+package org.lflang;
+
+import org.jetbrains.annotations.NotNull;
+
+import org.lflang.lf.TimeUnit;
+
+/**
+ * Represents an amount of time (a duration).
```
oowekyala, line 33: Currently nothing in the code prevents negative time values to be constructed. If this is intentional (which I'm not sure it is, since LF syntax does not support writing negative time values), we should document it. Otherwise we should ensure it's impossible.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -4294,8 +4293,9 @@ class CGenerator extends GeneratorBase {
                 reactionInstance.definition
             )) {
                 val reactionStructName = '''«selfStructName(reactionInstance.parent)»->___reaction_«reactionCount»'''
-                val reactionIndex = "0x" + (reactionInstance.deadline.toNanoSeconds.shiftLeft(16)).or(
-                    new BigInteger(reactionInstance.level.toString)).toString(16) + "LL"
+                // xtend doesn't support bitwise operators...
+                val indexValue = CGeneratorExtension.longOr(reactionInstance.deadline.toNanoSeconds << 16, reactionInstance.level)
```
Soroosh129, line 4297: `longOr` is unrelated to federated execution.

org.lflang/src/org/lflang/TimeValue.java
Soroosh129: Based on the comment alone, time durations can be negative.

Soroosh129 (approved):
Wow this is an impressive improvement! This looks good to me (minus the minor comment).

org.lflang/src/org/lflang/TimeValue.java
oowekyala: I wrote this comment based on the current definition of the class though

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
oowekyala: Yeah that's true... but we currently don't have a utility class for this kind of things... Maybe this should go into a   `MissingLanguageFeaturesUtil`... jk I'll find something

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaDependencyAnalysisTest.xtend
```diff
@@ -135,7 +135,7 @@ class LinguaFrancaDependencyAnalysisTest {
         Assertions.assertNotNull(model)
 		
 		var info = new ModelInfo()
-		info.update(model, DefaultErrorReporter.DEFAULT)
+		info.update(model, new DefaultErrorReporter())
```
lhstrh, line 138: Why create a new instance and not use the static one?

org.lflang/src/org/lflang/DefaultErrorReporter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.nio.file.Path;
+
+/**
+ * Simple implementation of the ErrorReport interface that simply prints to
+ * standard out.
+ */
+public class DefaultErrorReporter implements ErrorReporter {
+
+    private boolean errorsOccurred = false;
+
```
lhstrh, line 14: The static member was removed. I'm sure you didn't do this by accident, but could you explain why?

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -129,159 +131,161 @@
...
      */
-    public boolean collectStats = true
+    public static class ClockSyncOptions {
 
-    /**
-     * FIXME
-     */
-    public boolean localFederatesOn
+        /**
+         * FIXME
```
lhstrh, line 210: @Soroosh129: maybe you could quickly jump in and address these FIXMEs?

Soroosh129: What is this FIXME referring to?

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -129,159 +131,161 @@
...
     /**
-     * The name to use as the root of the trace file produced.
-     * This defaults to the name of the .lf file.
+     * Settings related to tracing options.
      */
-    public String traceFileName = null
+    public static class TracingOptions {
+        /**
+         * The name to use as the root of the trace file produced.
+         * This defaults to the name of the .lf file.
```
lhstrh, line 286: This comment does not describe the code below, so the default is probably set elsewhere, but that makes it a little confusing...

org.lflang/src/org/lflang/TimeValue.java
```diff
@@ -0,0 +1,168 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON 
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT 
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+***************/
+
+package org.lflang;
+
+import org.jetbrains.annotations.NotNull;
```
lhstrh, line 28: What is this for?

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaDependencyAnalysisTest.xtend
oowekyala: Because that object is stateful. Keeping a static instance is basically keeping a global mutable variable which may be accessed concurrently and behave incorrectly, as we don't know who's using it.

org.lflang/src/org/lflang/DefaultErrorReporter.java
oowekyala: Yes it's because of the field we can see here

org.lflang/src/org/lflang/TimeValue.xtend
```diff
@@ -1,244 +0,0 @@
...
-DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
-ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES 
-(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
-LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON 
-ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT 
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
-SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-***************/
-
-package org.lflang
```
lhstrh, line 28: Just noticing that a `git mv <class>.xtend <class>.java` yields a more useful diff. So for future class ports, we should try to follow that recipe as much as possible.

org.lflang/src/org/lflang/graph/Graph.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Remove the given node from the graph. This also eliminates any
+     * edges from upstream and to downstream neighbors of this node.
+     *
+     * @param node The node to remove.
+     */
+    void removeNode(T node);
+
+
+    // todo order of parameters here is unintuitive... from -> to is more usual
```
lhstrh, line 60: I agree. This is a remnant of an earlier methods with different names that referenced the concept of dependencies, so the idea was "first arg _depends on_ second arg," but this reverses the polarity of the relations between nodes when compared to the flow of data. So we went with "from" and "to", but the complete refactoring should have also changed the order of the arguments, which we didn't (yet) do.

org.lflang/src/org/lflang/TimeValue.java
oowekyala: Ah this is Intellij importing stuff automatically, I usually remove it. Basically this is an annotation you can use in Java code to make it clear that some variable is never null. There's also `@Nullable` which means that a variable may be null, and so you must always check for null before dereferencing the variable. Using those annotations improves compatibility with the kotlin type system, gives more info to static analysis tools, and helps document API, the end goal being to avoid NullPointerExceptions. But if we are to use them, we need to decide which annotations to use, as those from `org.jetbrains.annotations` are only one option.

Maybe I can add an iterm to #478?

org.lflang/src/org/lflang/TargetConfig.java
lhstrh: Missing comments.

org.lflang/src/org/lflang/TimeValue.java
lhstrh: It may be better to use something more neutral. Doesn't Java have its own?

lhstrh (approved):
Thanks for doing this, @oowekyala! If we keep working at this pace we can soon be Xtend-free!
This all looks good to me, with the caveat that the classes for this changes weren't tracked in the diff are a bit difficult to review, meaning that any mistake in those translations is easy to miss. But the tests pass, so I say, let's go 🚀

org.lflang/src/org/lflang/TimeValue.java
oowekyala: No, there are `javax` ones but we should really not use those. We have [those other choices](https://kotlinlang.org/docs/java-interop.html#nullability-annotations). Those from jetbrains are available as they're part of the kotlin dependency. Maybe we can use those from the JDT instead?

org.lflang/src/org/lflang/TimeValue.xtend
oowekyala: I didn't know about that, I'll use this in the future

org.lflang/src/org/lflang/TimeValue.java
lhstrh: JSpecify maybe?

lhstrh: Since this just works, let's keep it like this for now

PR #573, opened 2021-10-05
Fixes #572

### LINKED ISSUE #572: The import org.jetbrains cannot be resolved
After merging in #566, I get the following error in `org.lflang.TimeValue.java`:

image

```
The import org.jetbrains cannot be resolved
```

Is a dependency missing?

Soroosh129:
This gets rid of the `org.jetbrains.annotations` dependency in `org.lflang.TimeValue`.

### DIFF
--- a/org.lflang/src/org/lflang/TimeValue.java
+++ b/org.lflang/src/org/lflang/TimeValue.java
@@ -25,8 +25,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package org.lflang;
 
-import org.jetbrains.annotations.NotNull;
-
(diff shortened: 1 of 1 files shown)


org.lflang/src/org/lflang/TimeValue.java
```diff
@@ -115,7 +113,10 @@ public boolean isEarlierThan(TimeValue other) {
     }
 
     @Override
-    public int compareTo(@NotNull TimeValue o) {
+    public int compareTo(TimeValue o) {
+        if (o == null) { 
+            throw new NullPointerException("Can't compare against null timevalue");
```
lhstrh, line 118: Adding a note here for @oowekyala before merging in this quick fix.

lhstrh (approved):

PR #574, opened 2021-10-06
add reflex game demo for python target

housengw:


example/Python/src/ReflexGame/ReflexGame.lf
```diff
@@ -0,0 +1,192 @@
+target Python {
+    keepalive: true,
+    files: [gui.py]
+}
+
+preamble {=
```
Soroosh129, line 6: I think it's better to move this preamble to `RandomSource`. This makes the reactor a lot more self-contained.

example/Python/src/ReflexGame/ReflexGame.lf
```diff
@@ -0,0 +1,192 @@
...
+/*
+ * Receives key presses from the pygame process. 
+ */
+reactor GetUserInput {
+    preamble {=
+        import threading
+        def listen_for_input(self, user_response):
+            while 1:
+                try:
+                    c = self.user_input.recv()
```
Soroosh129, line 43: Just FYI, apparently, `recv()` blocks until something is received. There is an alternative `poll()` method that busy waits.
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Connection

So this `while 1:` might be okay here.

example/Python/src/ReflexGame/ReflexGame.lf
housengw: sounds good, I'll update that momentarily

example/Python/src/ReflexGame/ReflexGame.lf
housengw: I believe I made sure that all ```while 1``` loops in this program are blocking, including the pygame loop.

example/Python/src/ReflexGame/ReflexGame.lf
```diff
@@ -0,0 +1,192 @@
...
+    reaction(startup) -> user_input_pipe, update_graphics_pipe {=
+        # Starts the gui pygame process
+        user_input_pout, update_graphics_pin = self.gui.start_gui()
+        
+        # Sets the outputs to trigger the initialization of GetUserInput and UpdateGrpahics 
+        user_input_pipe.set(user_input_pout)
+        update_graphics_pipe.set(update_graphics_pin)
+    =}
+    
+    reaction(t) {=
```
Soroosh129, line 176: I don't think this is needed anymore right?

Soroosh129: It does look like this branch is behind `master` by quite a bit. You would have to merge in master first and then rebuild lfc and/or Eclipse based on the tools you use.

example/Python/src/ReflexGame/gui.py
```diff
@@ -0,0 +1,140 @@
...
+os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = "hide"
+
+try:
+    import pygame
+except:
+     print("Import Error: Failed to import 'pygame'. Try 'pip3 install pygame'")
+     sys.exit(1)
+
+
+def start_gui():
```
Soroosh129, line 15: This should be something we could do in a reactor with federated Python once that's ready.

Soroosh129:
It seems like some commits are duplicated like `remove redundant logical action` and `fix error in comment`.

This has resulted in a strange history. Any idea why this might have happened?

Soroosh129:
> Added failing test that replicates #574

Sorry about this. I used the wrong reference number for an unrelated issue (it was meant to be 515).

housengw:
I think the duplicate commit messages are due to the fact that I misused ```git rebase``` when merging changes from the master branch. I'll fix that shortly, sorry about that.

PR #577, opened 2021-10-06
Filter out connections that break dependencies

lhstrh:
Fixes https://github.com/lf-lang/lingua-franca/issues/447, but problems remain in the code generator when it comes to the handling of physical connections. I think we should merge this in and address the remaining problem separately.


lhstrh:
> Looks good to me, except that I don't see where the ports get added as nodes. Is this a side effect of addEdge()?

Indeed, it is.

PR #581, opened 2021-10-07
Pytorch example

Soroosh129:
This adds two examples that pass images from the user webcam (if available) through the YOLOv5 object detector deep neural network (DNN) and overlay the detected object on top of the image and show it back to the user. The DNN framework used is PyTorch.

One of the examples is driven by a physical action and one is driven by a timer.


lhstrh:
Looks like there are still conflicts. Could you resolve these?

PR #600, opened 2021-10-12
Fixes #426

### LINKED ISSUE #426: `printf` Statements in C Tests/Examples/Benchmarks that print `instant_t/interval_t` Produce Compiler Warnings
Many of our C tests/examples/benchmarks have assumed that `instant_t` (which is used to represent [time](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#timed-behavior) in the C target) is a `long long int`, which was the old data type for `instant_t`. 

However, to ensure a more consistent representation for `instant_t` across platforms, the current type was switched to `int64_t` as part of the [platform abstraction effort](https://github.com/icyphy/lingua-franca/blob/f5c904713a2ef268aa2d19e3ac36fed2f7bca438/org.lflang/src/lib/core/platform/lf_linux_support.h#L51). 

The issue is that `int64_t` happens to correspond to a `long int` on  64-bit machines, which is a problem for legacy code such as
```
printf("Current logical time: %lld", get_elapsed_logical_time());
``` 
This code results in a warning produced due to the type mismatch between `%lld` and `instant_t` (a.k.a. `long int`).

There are three suggested solutions so far:

1- Revert `instant_t` back to `long long int` which has a _minimum_ width of 64 bits.

2- Convert all legacy `printf` code to use `%ld`, which will cause a problem in machines that are less than 64-bit.

3- Convert all legacy `printf` code to use the [proper formatting](https://en.cppreference.com/w/cpp/types/integer), which is `PRId64` for `int64_t`:
```
printf("%" PRIu64, get_elapsed_logical_time());
```

Soroosh129:
### **Summary:**

Many of our C tests/examples/benchmarks have assumed that `instant_t` (which is used to represent [time](https://github.com/icyphy/lingua-franca/wiki/Writing-Reactors-in-C#timed-behavior) in the C target) is a `long long int`, which was the old data type for `instant_t`. 

However, to ensure a more consistent representation for `instant_t` across platforms, the current type was switched to `int64_t` as part of the [platform abstraction effort](https://github.com/icyphy/lingua-franca/blob/f5c904713a2ef268aa2d19e3ac36fed2f7bca438/org.lflang/src/lib/core/platform/lf_linux_support.h#L51). 

The issue is that `int64_t` happens to correspond to a `long int` on  64-bit machines, which is a problem for legacy code such as
```
printf("Current logical time: %lld", get_elapsed_logical_time());
``` 
This code results in a warning produced due to the type mismatch between `%lld` and `instant_t` (a.k.a. `long int`).

### **Solution:**

Revert `instant_t` back to `long long int` which has a _minimum_ width of 64 bits.

### DIFF
--- a/org.lflang/src/lib/core/platform/lf_linux_support.h
+++ b/org.lflang/src/lib/core/platform/lf_linux_support.h
@@ -60,7 +60,7 @@ typedef long long int _interval_t;
 /**
  * Microstep instant.
  */
-typedef unsigned long int _microstep_t;
+typedef unsigned int _microstep_t;
 
 
 // The underlying physical clock for Linux
--- a/org.lflang/src/lib/core/platform/lf_macos_support.h
+++ b/org.lflang/src/lib/core/platform/lf_macos_support.h
@@ -60,7 +60,7 @@ typedef long long int _interval_t;
 /**
  * Microstep instant.
  */
-typedef unsigned long int _microstep_t;
+typedef unsigned int _microstep_t;
 
 // The underlying physical clock for MacOS
 #define _LF_CLOCK CLOCK_MONOTONIC
--- a/org.lflang/src/lib/core/platform/lf_windows_support.h
+++ b/org.lflang/src/lib/core/platform/lf_windows_support.h
@@ -85,7 +85,7 @@ typedef long long int _interval_t;
 /**
  * Microstep instant.
  */
-typedef unsigned long int _microstep_t;
+typedef unsigned int _microstep_t;
 
 #define _LF_TIMEOUT ETIMEDOUT
 


lhstrh:
@edwardalee: Yeah I recall that discussion... But do you think that having `_ll` is preferable over what's in the current PR, or not?

Soroosh129:
> I started down this path at some point but then realized that this would not be thread safe because multiple threads could call the function, and even if a mutex is used, the second invocation of the function would overwrite the char[] that was written by the first, possibly before the printf has successfully used the string.

I think the C developers have encountered the same exact issue with `ctime()`, which as you know is an ANSI C function that returns a human-readable string. `ctime()` is not thread-safe, but there is a [`ctime_s()`](https://en.cppreference.com/w/c/chrono/ctime) variant that is. Both of these functions are still part of C17.

For our purposes, perhaps a `lftime()` and a `lftime_s()` would suffice? These would follow the same convention as `ctime` and should be graspable by a C developer. They can take in a `tag_t` and do the appropriate conversion to `long long int`.

PR #606, opened 2021-10-14
Introduce a target property to trigger export of the runtime dependency graph

cmnrd:
This change introduces the `export-dependency-graph` target property. This is a useful debugging tool for C++ programs. If set to true, the resulting program will produce a *.dot file after its normal execution. This .dot file visualizes the internal dependency graph produced by the reactor-cpp runtime.

### DIFF
--- a/org.lflang/src/org/lflang/TargetConfig.java
+++ b/org.lflang/src/org/lflang/TargetConfig.java
@@ -200,6 +200,14 @@ public class TargetConfig {
     public TracingOptions tracing = null;
 
 
+    /**
+     * If true, the resulting binary will output a graph visualizing all reaction dependencies.
+     *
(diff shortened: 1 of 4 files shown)


lhstrh:
> I slightly modified reactor-cpp, so that it now also prints the location where the dot file was created. I also documented this feature in the C++ part of the wiki.
> 
> > Seems useful to also add this as a CLI option for reactor-ccp?
> 
> I don't really understand this comment. Do you mean we should add a CLI option to the compiled binaries that allows us to export the graph? I was considering this, but I decided against it because I don't want to clutter the help text of binaries too much and the feature is only really useful for us developers.

Yes, that's what I meant, but I think I agree that it's not necessary to have it as a CLI option

PR #609, opened 2021-10-14
Disable chain ID

Soroosh129:
The current chain ID assignment and usage is incorrect and will lead to nondeterminism in the exectuion of the program. Until that is fixed, we need to disable the usage of chain IDs. This is what was causing the PingPongDistributed example/benchmark to fail. See #608. However, the problem is not limited to federated execution.


Soroosh129:
@edwardalee My conjecture has been that `BankToBankMultiportAfter.lf` is failing due to the chain ID bug (there is an issue open for this: https://github.com/lf-lang/lingua-franca/issues/688). I have now pulled in `master` into this (`disable-chain-id`) branch and I will re-run the CI tests a few times to see if this issue still persists.

Soroosh129:
Looks like I was wrong. Pushed a fix here: https://github.com/lf-lang/lingua-franca/pull/702

edwardalee (approved):
Although Soroush has pointed out that this change increases the burden that target developers have to support federated execution, the change is important because the implementation in master is incorrect.  It has two errors that happen to complement each other so that, luckily, tests pass.  The first error is that levels are assigned incorrectly to network control reactions because the AST transformation breaks the precedences by removing connections between federates. The second error is that chain IDs are computed incorrectly, which coincidentally happens to sometimes compensate for the first error.

This PR should be merged before the scalability work is completed because merging after will be quite difficult.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -1505,6 +1508,42 @@ abstract class GeneratorBase extends AbstractLFValidator implements TargetTypes
             return time.time.toString()
         }
     }
+    
+    
+    
+    /**
+     * Remove disconnected network ports in federates, if any.
```
lhstrh, line 1515: I think already mentioned this elsewhere in the PR, but I find the term "disconnected network port" confusing. I think it would be better to call it something like `removeFederateConnectionPorts` (because the interfederate connections are being substituted by network connections).

lhstrh (approved):
Looks good. Left a minor comment regarding method naming/terminology.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
Soroosh129: Thanks. That's a good point. I changed the name and documentation.

PR #620, opened 2021-10-16
Allow after 0, refactor time units

lhstrh:
// Creating this draft PR because I forgot I had been working on this...

This PR
- [x] allows writing a zero delay (which conflicts with #544 but whatever)
- [x] refactors time units to disallow writing `0 NONE`
- [x] adds support for SI time units #742 (ns, us, ms, s, h, d)
- [x] makes time units soft keywords instead of keywords (they're lexed as identifiers), because they would otherwise be reserving too many useful identifiers to my liking
- [ ] disallows negative time values
 
This conflicts with #544 but also extracts parts of it. I'll merge this pr into that one.
This conflicts with @edwardalee's changes (edits some converted xtend files) but I can take care of merging master into your branch myself


oowekyala:
I'm picking this up, I'll use it to fix #742 as well

org.lflang/src/org/lflang/JavaAstUtils.java
```diff
@@ -150,15 +151,86 @@ public static InferredType getInferredType(Port p) {
     }
 
     /**
-     * If the given string can be recognized as a floating-point number that has a leading decimal point, 
+     * Returns the time value represented by the given AST node.
+     */
+    public static TimeValue toTimeValue(Time e) {
+        if (!isValidTime(e)) {
+            // invalid unit, will have been reported by validator
+            return new TimeValue(e.getInterval(), TimeUnit.SECOND);
```
lhstrh, line 159: I think it's safer to throw an exception here...

lhstrh (commented):
This looks great! Thanks for picking up this work and bringing it to the finish line :-) I only left a minor comment.

org.lflang/src/org/lflang/JavaAstUtils.java
```diff
@@ -32,6 +32,7 @@
 import org.lflang.lf.Parameter;
 import org.lflang.lf.Port;
 import org.lflang.lf.StateVar;
+import org.lflang.lf.Time;
```
lhstrh, line 35:
```suggestion
import org.lflang.lf.Time;
import java.lang.IllegalArgumentException;
```

org.lflang/src/org/lflang/validation/LFValidator.xtend
```diff
@@ -43,6 +43,7 @@ import org.lflang.FileConfig
 import org.lflang.ModelInfo
 import org.lflang.Target
 import org.lflang.TargetProperty
+import org.lflang.TimeUnit
```
lhstrh, line 46:
```suggestion
```

PR #629, opened 2021-10-18
Added symlink from CCpp to C in test directory

### LINKED ISSUE #622: Allow running CCpp tests with `gradle runSingleTest`
In our gradle runSingleTest task (#555) we can't run CCpp tests right now. The problem is that the task determines the target by parsing the file name argument. Eg if you call `gradle runSingleTest test/Rust/...`, it will know this is a rust test by literal pattern matching on the `test/Rust/...` argument. However, CCpp doesn't have a separate test directory in our test sources (it uses C test cases), so if you write `test/CCpp/...` the task will fail with an invalid file name

We have several options
- add a separate cli argument to force CCpp to be used instead of C when passing `test/C/...` 
- special-case `test/CCpp/...` (even though it's an invalid file name)
- check in a symlink `test/CCpp -> test/C` 
- ???

See https://github.com/lf-lang/lingua-franca/pull/488#discussion_r730327058

lhstrh:
Fix #622


lhstrh:
We organize the tests by target. That's not conflating anything; doing so was a deliberate design choice. But now we have a target with no tests, even though we want to run tests for it. This is precisely what has caused this awkwardness.

Soroosh129:
> But now we have a target with no tests.

We have a target with no test folder.

I'd be happy to create tests for the CCpp target but it would be a lot of duplicate code. To me, this does look like a special case that we need to account for just to avoid duplicating things unnecessarily.

PR #631, opened 2021-10-19
Fix Rust build

oowekyala:
Fixes the build issues found after merging #488 into master.


oowekyala:
Everything looks green. For reference, the bug was caused by two threads trying to access the same physical action concurrently. This is a data race, but the smart pointer we use detected that and panicked. I replaced it with a mutex. I'm not sure how to reproduce this in a test case, as the faulty behavior is only observable in a certain interleaving of threads. That's also why the test failed only unpredictably.

lhstrh:
I think this PR already _fixes_ the issue @oowekyala describes here.

PR #643, opened 2021-10-20
Update lfc.ps1 to work in both dev and release setups

petervdonovan:
This allows `lfc` to work in Windows in a way that is consistent with recent changes to the nightly build workflow and the corresponding Bash scripts. Related to #530.

### DIFF
--- a/bin/lfc.ps1
+++ b/bin/lfc.ps1
@@ -1,23 +1,36 @@
 #==========================================================
 # Description: 	    Run the lfc compiler.
-# Authors:          Christian Menardn
+# Authors:          Christian Menard
 # Usage:            Usage: lfc [options] files...
 #==========================================================
 
 $base="$PSScriptRoot\.."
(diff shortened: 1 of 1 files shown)


petervdonovan:
> In addition to these fixes, I think we need to add a check similar to the one performed in the Bash script for ensuring that a JRE >= 11 is installed.

The Java version is checked on lines 43-47.

I tested it on my machine for all possible cases (dev setup with JAR available/unavailable, release setup with JAR available/unavailable).

lhstrh:
> > In addition to these fixes, I think we need to add a check similar to the one performed in the Bash script for ensuring that a JRE >= 11 is installed.
> 
> The Java version is checked on lines 43-47.
> 
> I tested it on my machine for all possible cases (dev setup with JAR available/unavailable, release setup with JAR available/unavailable).

Oh, sorry, I missed that. GitHub collapsed the view because those checks weren't part of the diff.

PR #646, opened 2021-10-20
Add a federated test for TypeScript

hokeun:
Add a federated test for TypeScript after fixing issues with federated TypeScript code generation.


Soroosh129:
Yes it makes perfect sense! Thanks for adding the TODO.

PR #658, opened 2021-10-22
Fix validation of interleaved in case the variable is not a port

cmnrd:
This fixes the Exception we saw yesterday in the Big benchmark


lhstrh (approved):
:+1:

PR #672, opened 2021-10-25
DOT Generation in the Graph package

Soroosh129:
Added functionality to generate DOT representation of the reaction instance graph.

TO-DO:
- [ ] Allow the user to request the DOT representation of the reaction instance graph (e.g., by passing a flag to lfc)


cmnrd:
Looks like this is a stale PR. What is the status and what should we do with it?

Soroosh129:
> Looks like this is a stale PR. What is the status and what should we do with it?

I think the solution proposed by @a-sr sounds better. I'll close this PR and we can revisit this issue when we have a more modular solution to reporting intermediate data structures.

PR #675, opened 2021-10-26
Python federated

Soroosh129:
This adds support for local federated programs written using the Python target. The data exchanged between federates is transparently serialized and deserialized using Python's pickle module.

It also adds a simple HelloDistributed.lf test.


housengw:
#703

Soroosh129:
I think this is ready to be merged. There are a few failing federated tests but fixing them requires augmenting the Python API to add a few missing functionalities. These changes can be put in their own PR to avoid making this PR larger than it already is. I suggest we could ignore the tests in the `failing` folder for now during the review process.

Thanks @housengw for meticulously porting all the federated C tests to Python. I am optimistic we can get them all passing.

org.lflang/src/org/lflang/validation/LFValidator.xtend
```diff
@@ -83,7 +83,7 @@ import org.lflang.lf.WidthSpec
 
 import static extension org.lflang.ASTUtils.*
 import static extension org.lflang.JavaAstUtils.*
-import org.lflang.federated.SupportedSerializers
+import org.lflang.federated.serialization.SupportedSerializers
```
lhstrh, line 86: Should serialization be part of the `federated` package? As you mentioned this morning, we might want to employ serialization to enable communication between reactors of different targets without relying to inter-process communication.

Soroosh129: That's an excellent point.

I think for the time-being, the serialization package is mainly focused on federated execution. Once we are able to generalize its functionality, we should definitely consider moving it out of the federated package.

lhstrh (approved):
Looks good! Should we first merge https://github.com/lf-lang/lingua-franca/pull/713?

Soroosh129:
> Looks good! Should we first merge #713?

I think the order doesn't matter (hopefully). If we merge this in first, we can change the base of #713 to master.

PR #690, opened 2021-10-27
Rust benchmarks

oowekyala:
Adds support for banks of reactors, and a couple of benchmarks.

Before merging I'd like to add more tests for some corner cases where outputs and inputs of the same reactor are connected.

-----

Here are results from the Trapezoidal benchmark, which is the first parallelism benchmark for Rust:

trapezoidal-smalltalk-threaded-nice

> Trapezoidal.lf - Intel(R) Core(TM) i7-9700T CPU @ 2.00GHz / 8 cores

--- 

And results for Dictionary.lf

concdict


benchmark/Rust/Savina/src/concurrency/CigaretteSmoker.lf
```diff
@@ -0,0 +1,93 @@
...
+    logical action chooseSmoker;
+    
+    reaction(startup) -> chooseSmoker {=
+        ctx.schedule(chooseSmoker, Asap);
+    =}
+    
+    reaction(chooseSmoker) -> startSmoking, chooseSmoker {=
+        // assume resources grabbed instantaneously and choose a random smoker to start smoking
+        let new_smoker_index = self.random.gen_range(0..self.num_smokers);
+        let busy_wait_period = self.random.gen_range(10usize..1010);
```
cmnrd, line 54: This does not use the pseudo random number generator as provided in Savina (and implemented in C++), does it? I think using the same generator is really important for comparability of the results.

benchmark/Rust/Savina/src/concurrency/Dictionary.lf
```diff
@@ -0,0 +1,167 @@
...
+                        self.data_map.insert(*key, *value);
+                        self.answers_to_send[i] = (*value).try_into().unwrap_or(0);
+                        info!("Worker {} writes {} to key {}", i, value, key);
+                    },
+                    Message::Read { key } => {
+                        // read the value. If the key is not yet present, this will add it and initialize the value to 0.
+                        // This does not match 100% the Savina implementation which returns null if the key is not present,
+                        // but should do a similar job.
+
+                        // todo wat? does the savina impl insert a new (key->null) mapping or just return null?
```
cmnrd, line 90: Reading a value that is not in the map, should just return null and don't insert a value. This was actually a bug in the C++ implementation.

benchmark/Rust/Savina/src/concurrency/dictionary_support.rs
```diff
@@ -0,0 +1,29 @@
...
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/// Message sent by Dictionary benchmark
+pub enum Message {
```
cmnrd, line 26: Would this maybe be better placed in a preamble?

benchmark/Rust/Savina/src/parallelism/RadixSort.lf
```diff
@@ -0,0 +1,155 @@
...
+    state num_values(num_values);
+    state max_value(max_value);
+
+    output outValue: u64;
+
+    logical action sendRandomNum;
+
+    // @label gen_random
+    reaction(startup, sendRandomNum) -> outValue, sendRandomNum {=
+        let value = self.random.gen_range(0..self.max_value);
```
cmnrd, line 49: Also her we should use the pseudo random number generator.

benchmark/Rust/Savina/src/parallelism/support.rs
```diff
@@ -0,0 +1,31 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/// For Trapezoidal
+#[derive(Copy, Clone, Debug)]
+pub struct WorkMessage {
```
cmnrd, line 27: This also seems like it would be better placed in a preamble

cmnrd (approved):
Looks great! I only made a few remarks on the benchmarks. Feel free to ignore them for now.

benchmark/Rust/Savina/src/concurrency/CigaretteSmoker.lf
jhaye: This is pretty much the standard random library for Rust. When I ported benchmarks I implemented the PRNG from C++, which can be found [here](https://github.com/lf-lang/lingua-franca/blob/rust-more-benchmarks/benchmark/Rust/Savina/src/pseudo_random.rs).
When porting existing benchmarks to the new benchmark runner I will switch the RNGs out accordingly. For example, [here](https://github.com/lf-lang/lingua-franca/blob/rust-more-benchmarks/benchmark/Rust/Savina/src/parallelism/RadixSort.lf) is RadixSort, which I already ported. Performance almost doubled after switching.

benchmark/Rust/Savina/src/concurrency/Dictionary.lf
oowekyala: Ok, well the rust implementation does insert a mapping if one is missing. I added this behavior because otherwise we're always asking for random numbers which aren't in the map. Maybe it's because we use SmallRng and not the custom prng. I would assume then, that this implementation of Dictionary is not comparable to the c++ or Akka one. Probably it's the same thing with all benchmarks that use a prng (RadixSort and CigaretteSmokers). Should I still include them in my report?

oowekyala: this would also explain why the curves for dictionary are different in the screenshots above

PR #692, opened 2021-10-27
Federate refactoring

Soroosh129:
This refactors code in FedASTUtils and FederateInstance to make functions shorter and hopefully more readable.

### DIFF
--- a/org.lflang/src/org/lflang/federated/FedASTUtils.java
+++ b/org.lflang/src/org/lflang/federated/FedASTUtils.java
@@ -483,14 +483,10 @@ public class FedASTUtils {
     
     /**
      * Add a network sender reaction for a given input port 'source' to
-     * destination's parent reactor. This reaction will react to a generated
-     * network action (triggered asynchronously, e.g., by federate.c). This
-     * network action will contain the actual message that is sent by the sender
-     * in 'action->value'. This value is forwarded to 'destination' in the network
-     * receiver reaction.
+     * source's parent reactor. This reaction will react to the 'source'
+     * and then send a message on the network destined for the rightFederate.
      * 
      * @note Used in federated execution
-     * 
      * 
      * @param source The source port instance.
      * @param destination The destination port instance.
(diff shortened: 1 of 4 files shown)


org.lflang/src/org/lflang/federated/FedASTUtils.java
```diff
@@ -74,25 +74,203 @@
...
+     * @param sourceFederate The source federate.
+     * @param destinationFederate The destination federate.
+     * @param rightBankIndex The right bank index or -1 if the right reactor is not in a bank.
+     * @param rightChannelIndex The right channel index or -1 if the right port is not a multiport.
+     * @param generator The GeneratorBase instance used to perform some target-specific actions
+     * @param coordination One of CoordinationType.DECENTRALIZED or CoordinationType.CENTRALIZED.
+     * @param serializer The serializer used on the connection
+     */
+    private static void addNetworkReceiverReaction(
+            Action networkAction,
```
lhstrh, line 162: The number of arguments of this function (as well as it's length) suggests that it's doing too many things at once. Without having carefully looked at what this function does exactly, I bet this can be factored out in two or three smaller functions, each of which has only a few arguments.

org.lflang/src/org/lflang/federated/FedASTUtils.java
```diff
@@ -292,23 +464,117 @@ private static TimeValue findMaxSTP(Variable port,
...
+     * @param connection The network connection.
+     * @param sourceFederate The source federate.
+     * @param leftBankIndex The left bank index or -1 if the left reactor is not in a bank.
+     * @param leftChannelIndex The left channel index or -1 if the left port is not a multiport.
+     * @param destinationFederate The destination federate.
+     * @param generator The GeneratorBase instance used to perform some target-specific actions
+     * @param coordination One of CoordinationType.DECENTRALIZED or CoordinationType.CENTRALIZED.
+     * @param serializer The serializer used on the connection
+     */
+    private static void addNetworkSenderReaction(
```
lhstrh, line 486: I think my previous comment applies here too.

lhstrh (commented):
These look like good improvements to me! Thanks for taking the time to do some refactoring. I've provided some comments that should be relatively easy to address.

org.lflang/src/org/lflang/federated/FedASTUtils.java
Soroosh129: That's a great a suggestion. I will take a stab at this refactoring in a day or two.

PR #698, opened 2021-10-29
Update C benchmark Pi Precise

cmnrd:
This simplifies the Pi Precise benchmark for C and aligns it with the updated C++ benchmark. I also fixed a few issues 
 in the existing C code. However, I didn't get this to work properly. The benchmark should compute 5000 digits of Pi, but for some reason it stops at 4160. Also it prints "Result is incorrect", but this doesn't necessarily mean much as the check implemented in Savina does not account for rounding errors on the last digit. I don't know why the program ends too early. Maybe @Soroosh129 has an idea?


Soroosh129 (approved):
Looks good to me!

I fixed the reporting stuff in the shutdown reaction. 
The C++ version of this benchmark directly sets the precision of the result stringstream to the requested scale. The C version now directly uses scale to limit the number of character comparisons.

lhstrh (approved):

cmnrd:
Thanks for fixing this @Soroosh129. I guess I got confused about the two versions at some point :laughing:. I slightly adjusted the C++ version, so that it produces a similar output.

PR #700, opened 2021-10-29
fixed docker file generation for C/Ccpp target and python target

housengw:
C/Ccpp target:
fixed both standalone and federated docker file generation.

Python target:
added standalone docker file generation.

Also added an example at example/C/ContainerizedFederatedHelloWorld


Soroosh129:
Hi Steven
Why did you close this PR?
You can mark it as a draft PR if you think something needs to be added.

housengw:
oh i was trying to resolve the merge conflict

example/C/src/ContainerizedFederatedHelloWorld/HelloWorld.lf
```diff
@@ -0,0 +1,67 @@
+/**
+ * Containerized distributed LF program where a MessageGenerator creates a string
```
Soroosh129, line 2: I suggest it might be better to put this in `example/C/src/DistributedHelloWorld/` but in a separate folder, called `docker` for example. This way, you can utilize the [import](https://github.com/lf-lang/lingua-franca/wiki/Language-Specification#import-statement) system in LF to import from `HelloWorld.lf` and avoid duplicate code.

example/C/src/ContainerizedFederatedHelloWorld/README.md
```diff
@@ -0,0 +1,51 @@
+# ContainerizedFederatedHelloWorld
+This is a basic example showing containerized federated execution. 
+
+For more details, see: https://github.com/icyphy/lingua-franca/wiki/
```
Soroosh129, line 4: This looks great!

I suggest that if you want to link to a specific page, you could link to [Distributed Execution](https://github.com/lf-lang/lingua-franca/wiki/Distributed-Execution). We should probably add a docker section there.

housengw: wait... that was a mistake. I meant to link to https://github.com/lf-lang/lingua-franca/wiki/Containerized-Execution

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -1239,50 +1244,133 @@ class CGenerator extends GeneratorBase {
      * The file will go into src-gen/filename.Dockerfile.
      * If there is no main reactor, then no Dockerfile will be generated
      * (it wouldn't be very useful).
-     * @param The root filename (without any extension).
+     * @param the name given to the docker file (without any extension).
      */
-    def writeDockerFile(String filename) {
+    def writeDockerFile(String dockerFileName) {
```
Soroosh129, line 1249: If you want, you could move this to GeneratorBase and override it here.

example/C/src/ContainerizedFederatedHelloWorld/README.md
Soroosh129: Ah okay, that one is even better.

housengw:
converting this to draft because I am getting a 
```
lfc: error: Platform mac os x is not supported
 --> (unknown file):null:1 - Platform mac os x is not supported
```
error after resolving conflicts

org.lflang/src/org/lflang/generator/PythonGenerator.xtend
```diff
@@ -1053,13 +1053,21 @@ class PythonGenerator extends CGenerator {
         targetConfig.noCompile = true;
         targetConfig.useCmake = false; // Force disable the CMake because 
                                        // it interferes with the Python target functionality
+        var dockerOptions = targetConfig.dockerOptions;
+        targetConfig.dockerOptions = null;
```
Soroosh129, line 1057: Perhaps a comment here can help.

If you are overriding `writeDockerFile`, the CGenerator will call your overridden function.

Soroosh129:
> converting this to draft because I am getting a
> 
> ```
> lfc: error: Platform mac os x is not supported
>  --> (unknown file):null:1 - Platform mac os x is not supported
> ```
> 
> error after resolving conflicts

Hmm, the merge created a weird `addPlatformFiles`. My guess is that it should look like [this](https://github.com/lf-lang/lingua-franca/blob/39ff0b946b9c15790fae238d5d6676d75e880608/org.lflang/src/org/lflang/generator/c/CGenerator.xtend#L1153). Not sure what happened there.

Soroosh129:
My guess is that the merge conflict happened because of this line in `master`:
```
// All platforms use this one.
coreFiles.add("platform/lf_tag_64_32.h");
```

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -1239,50 +1244,133 @@ class CGenerator extends GeneratorBase {
...
+            To build the docker image, use:
+               
+                docker build -t rti -f «dockerFile» «rtiDir»
+            
+            #####################################
+        ''')
+    }
+
+    /**
+     * Write a Dockerfile for the RTI at rtiDir.
```
Soroosh129, line 1358: I think this comment needs to be updated.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -1158,11 +1158,8 @@ class CGenerator extends GeneratorBase {
     }
     
     /**
-     * Add the platform files to 'coreFiles'.
-     * 
-     * Also if useCmake is set to false, detect the host OS
-     * and add the appropriate platform support file(s) to 
-     * the compile command (compileAdditionalSources).
+     * Add the appropriate platform files to 'coreFiles'. These platform files
```
Soroosh129: Thanks for fixing this daunting merge conflict. I think the old comment reflects the intent of this function more accurately.

org.lflang/src/org/lflang/generator/PythonGenerator.xtend
```diff
@@ -1699,6 +1698,51 @@ class PythonGenerator extends CGenerator {
...
+        if (this.mainDef === null) {
+            return
+        }
+        
+        var srcGenPath = fileConfig.getSrcGenPath
+        val dockerFile = srcGenPath + File.separator + filename + '.Dockerfile'
+        val contents = new StringBuilder()
+        
+        // If a dockerfile exists, remove it.
+        var file = new File(dockerFile)
```
lhstrh, line 1718: If the programmer removes the `main` keyword after a successful build, then the produced docker image will not be removed. We probably want to remove the old image in this case as well (i.e., move 1718:1721 to 1709:1712).

lhstrh (changes requested):
This looks good! I left a minor comment (and I saw there was also an unaddressed comment from @Soroosh129).

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -843,6 +844,15 @@ class CGenerator extends GeneratorBase {
             
             // Create docker file.
             if (targetConfig.dockerOptions !== null) {
+                if (isFederated) {
+                    var rtiPath = fileConfig.getSrcGenBasePath().resolve("RTI")
+                    var rtiDir = rtiPath.toFile()
+                    if (!rtiDir.exists()) {
```
Soroosh129, line 850: I didn't realize this before, but with this design, if we change something in the RTI's source code in `c/reactor-c/core/federated/RTI`, it will not show up in the RTI's source code that is in `src-gen/RTI` if this folder already exists, because `copyRtiFiles` is skipped. Perhaps `copyRtiFiles(rtiDir, coreFiles)` should be out of the if statement? Apologies for delaying the merge, but I think this is rather important.

housengw: that's a good point. I think a better design would be to put the RTI docker file generation outside of the for (federate : federates) loop. I also moved the copyRtiFiles outside the if statement. Thanks for your input!

lhstrh (approved):

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: That's a good design! Thanks for addressing this so quickly

PR #705, opened 2021-11-01
Implement matrix ADT.

petervdonovan:
This is just a small header file for use in a few benchmarks.


benchmark/C/Savina/src/Matrix.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+void mat_destroy(Matrix matrix) {
+    free(matrix.content);
+}
+
+/*
+ * Return a pointer to entry (i, j) of the given Matrix.
+ */
+double* mat_at(Matrix matrix, size_t i, size_t j) {
+    return (double*) matrix.content + i * matrix.n_cols + j;
```
Soroosh129, line 35: I think `matrix.content[i * matrix.n_cols + j]` is slightly more readable. Moreover, I would strongly suggest that some form of memory bound check be performed here (e.g., `(i * matrix.size_y + j) < (matrix.size_y * matrix.size_x)`).

Soroosh129 (commented):
I left comments for what, in my opinion, are considered best practices that could be useful here.

petervdonovan:
Thank you, @Soroosh129, for your help. I think that I have followed your directions.

Soroosh129 (approved):
Looks good to me.

One comment:
Since there are no `set` APIs, it is a bit awkward to set an element. For example:
```
double* value = mat_at_d(mat, 3, 4);
if (value == NULL) error_print_and_exit("Null value");
*value = 3;
```
If you want, you can abstract these steps here.

I also moved the .c file to the `lib` directory and the .h file to the `include` directory (please feel free to revert this if you don't like the new structure) and added a `.cmake` file that can be included in the `cmake-include` target property.

Soroosh129:
This turned out excellent! Thanks, @petervdonovan.

PR #706, opened 2021-11-01
Only generate a dockerfile for the RTI if docker is set to true

housengw:


housengw:
closing this pull request since the change was merged into master via #675

PR #709, opened 2021-11-02
Remove warning about keepalive from Rust target

oowekyala:
Remove warning added by b3b225b (#594) from the Rust target: the Rust target doesn't need the keepalive option. The warning looks like so:

```
Compiled binary is in /home/clem/.eclipse/lingua-franca-master/git/lingua-franca/example/Rust/bin
lfc: warning: Setting keepalive to true because of key_press. This can be overridden by setting the keepalive target property manually.
 --> Snake/KeyboardEvents.lf:20:5
   |
19 | 
20 |     physical action key_press: Key;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Setting keepalive to true because of key_press. This can be overridden by setting the keepalive target property manually.
   |
21 | 
```

### DIFF
--- a/org.lflang/src/org/lflang/Target.java
+++ b/org.lflang/src/org/lflang/Target.java
@@ -462,6 +462,10 @@ public enum Target {
         return this.keywords.contains(ident);
     }
 
+    /**
+     * Return true if the target supports multiports and banks
+     * of reactors.
+     */
     public boolean supportsMultiports() {
         switch (this) {
         case C:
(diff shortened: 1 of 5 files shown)


org.lflang/src/org/lflang/Target.java
```diff
@@ -485,6 +495,15 @@ public boolean supportsParameterizedWidths() {
         return false;
     }
 
+    /**
+     * Return true if the keepalive option is set automatically
+     * for this target if physical actions are detected in the
+     * program (and keepalive was not explicitly unset by the user).
+     */
+    public boolean setsKeepAliveOptionAutomatically() {
```
Soroosh129, line 503: I think there is an extra 's' after `set` in the name of the function.

Soroosh129: NVM, I see the intent of the function.

lhstrh (approved):

PR #710, opened 2021-11-02
Scalability

lhstrh:
Creating this draft PR to track progress...


cmnrd:
I did a quick test trying to compile the Banking C benchmark with 1000 accounts. I used time to determine execution time and memory footprint. On master `/bin/time -v  ./bin/lfc --no-compile benchmark/C/Savina/src/concurrency/Banking.lf` reports the following numbers (I pruned the output to the relavant lines):
```
        User time (seconds): 180.39
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:18.69
        Maximum resident set size (kbytes): 11040696
```
On this branch, it reports:
```
        User time (seconds): 74.16
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:50.59
        Maximum resident set size (kbytes): 7624160
```
There is a more than 2x speedup in user time, which unfortunately does not translate to similarly significant speedup in elapsed wall clock time. The memory footprint was reduced from about 11GB to about 7.6GB (about 30% improvement).

I am currently seeing if I get the generated C code to compile...

edwardalee:
I do not understand the test failures in the benchmarks, but they seem unrelated to this PR. I would like to get this massive PR merged, even though it is still a work in progress. I have a long list of further improvements to make, including handling banks, which should improve scalability by quite a lot, but the changes are so extensive that not merging will prevent other work from being done without conflicts.

petervdonovan:
(In case anyone is wondering about the benchmark tests -- Apsp/MatMul failed simply because of a syntax error in the `matrix_t` implementation that has since been fixed in `master`. However, it looks like SleepingBarber failed because the code generator ran out of heap space, which might be slightly more concerning.)

PR #712, opened 2021-11-02
C-Benchmarks: Concurrent Dictionary and Concurrent Linked List

mattchorlian:
I have made these benchmarks compatible with the python script and cleaned them up a bit.  I'm making this a draft PR because I'm expecting to make some more changes to make sure they align more closely with the C++ versions, pending review from some of you


benchmark/C/Savina/src/concurrency/Dictionary.lf
```diff
@@ -218,24 +154,24 @@ reactor Worker(instance:int(0), num_requests:int(10), write_percentage:int(10)){
...
+{
+
+    dictionary = new DictionaryImpl(num_workers=numWorkers);
+	workers = new[numWorkers] Worker(num_requests=numMessagesPerWorker,write_percentage=writePercentage);
 	
 	workers.read->dictionary.read;
 	workers.write->dictionary.write;
 	dictionary.result->workers.result;
 	
 	reaction(startup)->dictionary.start, workers.start{=
```
cmnrd, line 171: Similar to other benchmarks, we don't really need the start ports for C and both the dictionary and the workers could react directly to startup.

benchmark/C/Savina/src/concurrency/Dictionary.lf
```diff
@@ -7,90 +9,25 @@
...
+    ]]] */
+    threads: 0
+    /// [[[end]]]
 };
 
-reactor Dictionary(num_workers:int(20)){
+
+reactor DictionaryImpl(num_workers:int(20)){
     preamble {=
 		struct node{
```
cmnrd, line 32: The main problem with this benchmark is the implementation of the dictionary structure. I don't think it is complete and realistic, but I don't know if we want to invest in implementing a complete dictionary in C.... Maybe there is an implementation with a permissive license somewhere in the internet?

I cannot comment on the code below unfortunately, so I will summarize some of my thoughts here:

- In the preamble (definition of the dictionary): 
  - `table` does not seem to be initialized. There is no malloc for it
  - Elements can be inserted, but there is no function for cleaning up the dictionary at shutdown
  - The hash function seems to be overly simplistic
- In the Dictionary reactor:
  - The initialization in startup is strange, the dictionary is supposed to be empty
  - The C++ version does not distinguish read and write ports. It uses a single request port. This should be a bit faster as there is only a single reaction in the dictionary that needs to be processed.
  - The C++ version introduces a logical action in the Dictionary to break up the dependency cycle. In this implementaton, the logical action is contained in the workers. While both options should work fine, the one implemented for C++ is more efficient, as per iteration it has only a single call to schedule. In this version, each worker schedules its logical action resulting in multiple calls and multiple events being created.
- In the Worker reactor:
  - The Savina benchmark uses the Scala random number generator instead of the PseudoRandom generator. This is a bad design in Savina. However, we need to think how to best translate this to C. Simply using `rand()` does not work, since each worker needs its own generator initiailized with its own seed. In C++, we use [`std::minstd_rand`](http://www.cplusplus.com/reference/random/minstd_rand/) to create an instance of a random number generator. Is there something similar for C?
  - The workers limit the produced Dictionary keys to be smaller than 1000. I think this is due to the limited size of the dictionary implementation. However, we should be able to deal with arbitrary keys.

benchmark/C/Savina/src/concurrency/Dictionary.lf
```diff
@@ -218,24 +154,24 @@ reactor Worker(instance:int(0), num_requests:int(10), write_percentage:int(10)){
     =}
 }
 
-main reactor(num_workers:int(20),num_requests:int(8000),write_percentance:int(10)){
```
cmnrd, line 221: the default number of requests should be 10000

benchmark/C/Savina/src/concurrency/SortedLinkList.lf
```diff
@@ -7,51 +10,64 @@
...
+        if (self->requests_sent>=self->num_requests){
+            SET(finish,true);
+        }else{
+            schedule(request,0);
+        }
+    =}
+}
+
 reactor LinkedList(num_workers:int(20)){
     preamble {=
```
cmnrd, line 72: Also the implementation of linked list looks incomplete. At least there is no way to delete the list on shutdown

benchmark/C/Savina/src/concurrency/SortedLinkList.lf
```diff
@@ -7,51 +10,64 @@
...
+    output read:int;
+    output write:int;
+    output finish:bool;
+    
+    reaction(start)->request{=
+        self->requests_sent = 0;
+        schedule(request,0);
+    =}
+    reaction(request)->read, write{=
+        int coin = rand()%100;
```
cmnrd, line 50: Here we should use the PseudoRandom number generator

benchmark/C/Savina/src/concurrency/SortedLinkList.lf
```diff
@@ -7,51 +10,64 @@
...
+    
+    output read:int;
+    output write:int;
+    output finish:bool;
+    
+    reaction(start)->request{=
+        self->requests_sent = 0;
+        schedule(request,0);
+    =}
+    reaction(request)->read, write{=
```
cmnrd, line 49: In this benchmark, we can also produce a "contains" request to check if a value is contained in the list. Similar to the dict, I think there should be a single request port here.

benchmark/C/Savina/src/concurrency/SortedLinkList.lf
```diff
@@ -7,51 +10,64 @@
...
+preamble {=
+    #include "PseudoRandom.h"
+=}
+
+reactor Worker(instance:int(0), num_requests:int(10), write_percentage:int(10)){
+    input start:bool;
+    input result:bool;
+    
+    state requests_sent:int;
+    logical action request;
```
cmnrd, line 39: Similar to the Dictionary, the action should probably be moved to the List reactor, in order to reduce overhead for scheduling.

cmnrd (commented):
Your changes to cog and the yaml files look good! However, it seems that there are many problems remaining in Thee's code. I am not sure how to address them best. I think we should discuss this in our meeting on Thursday. The biggest challenge is getting a working implementation of a Dictionary and a linked list. Maybe we should do as @Soroosh129 suggested and use the CCpp target here.

lhstrh:
This looks like a good resource: https://www.clibs.org/packages/
Here's a BSD-licensed dictionary implementation, for example: https://www.clibs.org/packages/willemt/quadratic-probing-hashmap

mattchorlian:
I just pushed a starter BoundedBuffer implementation.  There are 3 things to address:

1) There is segmentation fault in the preamble, likely because I am mallocing a new PseudoRandom generator every time the function is called.  Any tips on how to do this without mallocing every time the function is called would be greatly appreciated

2)  Line 75 needs to be cogged.  In Cpp, you are allowed to use the difference of two parameters to instantiate a state, but this is not allowed in C.  So, I likely need to use some cog trickery to be able to get the appropriate adjustedBufferSize

3) Cogging the main reactor

benchmark/C/Savina/src/concurrency/BoundedBuffer.lf
```diff
@@ -0,0 +1,230 @@
...
+
+preamble {=
+    #include "PseudoRandom.h"
+    #include <math.h>
+    
+    double processItem(double curTerm, size_t cost) {
+        double res = curTerm;
+        
+        //FIXME: segfaulting here
+        PseudoRandom random = malloc(sizeof(struct PseudoRandom));
```
cmnrd, line 44: This should be `PseudoRandom*`

benchmark/C/Savina/src/concurrency/BoundedBuffer.lf
```diff
@@ -0,0 +1,230 @@
...
+        random->mValue = malloc(sizeof(long));
```
cmnrd, line 45: Why is the value allocated? It is already part of the struct

benchmark/C/Savina/src/concurrency/BoundedBuffer.lf
```diff
@@ -0,0 +1,230 @@
...
+
+
+
+reactor ManagerReactor(bufferSize: size_t(50), numProducers: size_t(40), numConsumers: size_t(40)) {
+
+    preamble {=
+        #include "deque.c"
+    =}
+    
+    //FIXME: need to cog this
```
cmnrd, line 74: Right... Instead of using cog, you could also initialize `adjustedBuffersize` in a reaction to startup

benchmark/C/Savina/src/concurrency/BoundedBuffer.lf
```diff
@@ -0,0 +1,230 @@
...
+    input[numProducers] producerFinished: bool;
+    output[numProducers] producerCommand: bool;
+    
+    input[numConsumers] consumerAvailable: bool;
+    output[numConsumers] consumerData: double;
+    
+    reaction(startup) -> producerCommand {=
+        // reset local state
+        self->numTerminatedProducers = 0;
+        self->producerTerminated[self->numProducers];
```
cmnrd, line 94: It doesn't work like that in C, unfortunately. The array needs to be allocated here

benchmark/C/Savina/src/concurrency/Dictionary.lf
```diff
@@ -0,0 +1,203 @@
...
+                    // to the map was successful.
+                    self->answersToSend[i] = msg->value;
+                } else if(msg->type == AccessType::Read) {
+                    // read the value. If the key is not yet present, this will add it and initialize the value to 0.
+                    // This does not match 100% the Savina implementation which returns null if the key is not present,
+                    // but should do a similar job.
+                    int value = self->dataMap[msg->key];
+                    self->answersToSend[i] = value;
+                }
+                //FIXME: free the message here?
```
cmnrd, line 131: I think the C runtime interprets the type `Message *` as a token and implements its own mechanism to free it. Could @Soroosh129 or @edwardalee help here? The program currently segfaults. I am not sure if this is the issue, though.

cmnrd (commented):
This looks good already. I think the Sorted List is finished. Dictionary currently segfaults, but this is likely related to the FIXME in the code regarding a `free`. But I don't know the C runtime well enough to tell for sure...

benchmark/C/Savina/src/concurrency/Dictionary.lf
mattchorlian: I looked into this and still can't figure out what would be causing the segfault.  I tried putting the Messages on the stack and allocating memory for some of the local reactor states but that doesn't work (plus, putting Messages on the stack will likely cause memory problems if we ever want to test the performance on larger problem sizes).  I also tried letting the C runtime free the Messages, per your comment, but that also segfaults.  The only thing I can think of now would be the std::map<int, int> used in the Dictionary reactor causing this, but that map is one of the major reasons we are using the CCpp target- so, I'm not sure what the most elegant way around this is...

cmnrd: Indeed the problem was with the map, but I don't think it is a problem with the map itself, but rather with how it is initialized (or not) in the CCpp target. See #741.

PR #713, opened 2021-11-02
Python iterable multiports

Soroosh129:
This makes multiports in the Python target iterable, which makes accessing them more Pythonic.

For example, previously, a `multiport` could only be iterated on as:
```Python
for i in range(len(multiport)):
    print(multiport[i].value)
```
Now, you could iterate over each port directly:
```Python
for port in multiport:
    print(port.value)
```


lhstrh (approved):
This is a nice improvement!

cmnrd (approved):
Great idea! Should we merge this?

PR #714, opened 2021-11-03
Initial port of Chameneos from C++ to C

lhstrh:
It runs, but I have not verified that it is correct.

### DIFF
--- a/benchmark/C/Savina/src/micro/Chameneos.lf
+++ b/benchmark/C/Savina/src/micro/Chameneos.lf
@@ -29,7 +29,9 @@
  */
 
 
-target C;
+target C {
+    logging: warn
+}
 
 preamble {=
     char const* color_names[] = {"Red", "Yellow", "Blue", "Faded"};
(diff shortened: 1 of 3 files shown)


cmnrd (approved):
This looks good. I pushed a few small fixes and integrated it with the benchmark runner. Both C and C++ produce the same output, so it seems to be correct. 

While looking at the output, I noticed another issue: our mall always pairs the same chameneos! For 5 chamenos, it always pairs  0 and 1 as well as 2 and 3. 4 is never paired. This is of course because in LF, at each iteration all the chameneos are present and we always iterate over them in the same order. I wonder: Should we add some variability here? In actors, there is automatically some variability, as there is no defined order in which messages arrive.

Also the C compiler takes a long time to compile for the 100 chameneos used in the Savina suite.

lhstrh:
@cmnrd: with respect to the pairing, I guess this depends on what the goal of the program is; is randomness required or merely an emergent property? I would assume that an actor implementation also is biased to some extent, there's just no obvious way to control that bias. I'd say let's leave it as is...

cmnrd:
I think you are right. We could see this as one possible behavior of an actor implementation.

PR #716, opened 2021-11-03
Port MatMul benchmark from C++ to C

petervdonovan:
This is a port of the matrix multiplication benchmark from C++ to C.

### DIFF
--- a/benchmark/C/Savina/src/include/matrix.h
+++ b/benchmark/C/Savina/src/include/matrix.h
@@ -89,4 +89,30 @@ static inline int* mat_at_i(matrix_t matrix, size_t i, size_t j) {
     return &(matrix.data_i[i * matrix.size_y + j]);
 }
 
+/*
+ * Set the (i, j) entry of the given double matrix.
+ * @param matrix the matrix to be modified.
+ * @param i The row to be accessed.
+ * @param j The column to be accessed.
+ * @param value The value to be placed in the matrix.
+ */
+static inline void mat_set_d(matrix_t matrix, size_t i, size_t j, double value) {
+    // This is safe in debug mode because of the assertions in mat_at_d. It is not safe
+    // when not in debug mode because in that case the assertions are disabled. However,
+    // these functions are called a billion times in the MatMul benchmark. Similar
+    // comments may apply to mat_set_i.
+    *mat_at_d(matrix, i, j) = value;
+}
+
+/*
+ * Set the (i, j) entry of the given integer matrix.
+ * @param matrix the matrix to be modified.
+ * @param i The row to be accessed.
+ * @param j The column to be accessed.
+ * @param value The value to be placed in the matrix.
+ */
+static inline void mat_set_i(matrix_t matrix, size_t i, size_t j, int value) {
+    *mat_at_i(matrix, i, j) = value;
+}
+
 #endif
--- a/benchmark/C/Savina/src/include/matrix.h
+++ b/benchmark/C/Savina/src/include/matrix.h
@@ -6,6 +6,7 @@
 #define MATRIX_H
 
 #include <stddef.h>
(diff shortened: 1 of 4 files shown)


benchmark/C/Savina/src/include/matrix.h
```diff
@@ -76,24 +82,11 @@ double* mat_at_d(matrix_t matrix, size_t i, size_t j);
...
-void mat_set_d(matrix_t matrix, size_t i, size_t j, double value);
-
-/*
- * Set the (i, j) entry of the given integer matrix.
- * @param matrix the matrix to be modified.
- * @param i The row to be accessed.
- * @param j The column to be accessed.
- * @param value The value to be placed in the matrix.
- */
-void mat_set_i(matrix_t matrix, size_t i, size_t j, int value);
```
Soroosh129, line 97: Looks like set definitions were also removed.

Soroosh129:
The changes in the matrix implementation are promising! Should the set functions also be inline?

petervdonovan:
Unless I am mistaken:
* Most CPU time is spent accessing the matrices. This is why I put the matrix access functions in a header file in commit 2d451ba -- apparently it helps for gcc to inline them.
* The speed increases almost linearly with the number of threads, which means that the overhead introduced by the C runtime might be relatively unimportant (at least for the default parameters used in the benchmark)

 

However, this is a simplification. I do not know why the C version of this benchmark is still slower than the C++ version, even though the matrix implementation is now essentially the same.

benchmark/C/Savina/src/include/matrix.h
petervdonovan: Oh! Thanks for the catch. I will amend that and inline the functions, and we can see if anything changes.

Soroosh129:
I'm a bit worried that all this copying the `matrix_t` structure around is causing a performance hit. Not sure. 
Perhaps, passing by reference is the way to go here.

benchmark/C/Savina/src/include/matrix.h
Soroosh129: Thanks! I did the same in #717.

petervdonovan:
My reasoning was that it only happens once, on startup.

Soroosh129:
Even though the `at` function and the `set` function are static inline, having a `matrix_t` and not a `matrix_t*` as an argument will cause a shallow copy of the matrix to be created each time the function is called. This copy could later be optimized away by the compiler, but that's not guaranteed.

petervdonovan:
Oh, right! Okay, I'll do what you said.

Soroosh129:
> Oh, right! Okay, I'll do what you said.
> 
> (**Update:** I observed no difference on my machine, but of course I have no idea about Eiffel.)

Okay. Thanks for trying this though!

cmnrd (approved):
This looks fantastic! I did a few quick tests and found that if we compare C and the first iteration of the C++ benchmarks, they are very close (+- 2ms). However, the second and subsequent iterations are much faster in C++. This is very likely due to cache effects (the first iteration warmed the cache up). I think this shows a significant disadvantage of the choice not to use the benchmark runner for the C benchmarks.

benchmark/C/Savina/src/parallelism/MatMul.lf
```diff
@@ -0,0 +1,290 @@
...
+
+    state A: matrix_t;
+    state B: matrix_t;
+    state C: matrix_t;
+
+    state work_stack: work_stack_t;
+
+    reaction (startup) -> data, next {=
+        // Fill both input arrays with data
+        self->A = mat_new_d(self->data_length, self->data_length);
```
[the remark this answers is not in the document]
petervdonovan: Thanks for pointing that out!

PR #717, opened 2021-11-04
C APSP and A* port

Soroosh129:
This ports the APSP and A* benchmarks from the Cpp target to C.

One awkwardness that I had to deal with was replacing `getElementAt` which took multiports as arguments. I tried to rewrite the code as compactly as possible without using this function.

Another awkwardness was that a lot of functionality for the A* benchmark was in the GuidedSearch folder with code written in C++. Rather than converting it to C, I used the CCpp target and reused most of that code.

Someone please verify that I haven't made a mistake in porting the logic.


benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,246 @@
...
+    output finished: bool;
+    
+    output[numWorkers] search: WorkMessage;
+    input[numWorkers] pathFound: bool;
+    input[numWorkers] moreNodesToVisit: std::deque<GridNode*>;
+    
+    logical action next;
+    
+    reaction (start) -> next {=
+        self->grid = new Grid(self->gridSize, self->priorities);
```
cmnrd, line 102: this should be freed somewhere. Actually, why is this a pointer? Couldn't this be a simple state variable? Since this is the CCpp target, you could also consider using a smart pointer to free the data automagically.

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,246 @@
...
+    state grid: Grid*;
+    state nodesToVisit: std::deque<GridNode*>;
+    state pathWasFound: bool(false);
+    
+    input start: bool;
+    output finished: bool;
+    
+    output[numWorkers] search: WorkMessage;
+    input[numWorkers] pathFound: bool;
+    input[numWorkers] moreNodesToVisit: std::deque<GridNode*>;
```
cmnrd, line 97: Won't this copy the entire deque when you set the port?

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,246 @@
...
+        info_print("Found a path through the grid");
+        info_print("Result valid: %d", valid);
+        SET(finished, true);
+    =}
+    
+    reaction (moreNodesToVisit) {=
+        // collect more nodes to visit from all workers and store them in the internal queue
+        for (int i=0; i< moreNodesToVisit_width; i++) {
+            if (moreNodesToVisit[i]->is_present) {
+                auto nodes = moreNodesToVisit[i]->value;
```
cmnrd, line 145: This copies the entire deque from `moreNodesToVisit[i]->value` to `node`. Is there a reason why you removed the const reference from the original code?

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,246 @@
...
+
+        void busyWait() {
+            for (int i = 0; i < 100; i++) {
+                rand();
+            }
+        }
+    =}
+    
+    reaction(search) -> pathFound, moreNodesToVisit {=
+        auto workMessage = search->value;
```
cmnrd, line 170: Also this copies (although it's only two pointers and will likely not matter)

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -0,0 +1,246 @@
...
+                        return;
+                    } else {
+                        workQueue.push_back(loopNeighbor);
+                    }
+                }
+            }
+        }
+        
+        if(!workQueue.empty()) {
+            SET(moreNodesToVisit, std::move(workQueue));
```
cmnrd, line 207: I am not sure how the `std::move` plays along here with the `SET` macro. I think you are good here and this will only copy the deque object to the port value, but not copy its data.

benchmark/C/Savina/src/parallelism/Apsp.lf
```diff
@@ -0,0 +1,337 @@
...
+    =}
+    
+    reaction(from_row, from_col) -> notify_neighbors, finished {=
+        // do nothing if complete
+        if(self->k == self->graph_size) {
+            return;
+        }
+        
+        // perform computation
+        matrix_t local_matrix = mat_new_i(self->block_size, self->block_size);
```
cmnrd: It doesn't seem like the matrices created during the execution are freed up anywhere.

cmnrd (commented):
This looks great! I found a few issues, mostly related to memory management. Could you have another look?

cmnrd:
Oh, seems like the issue of freeing matrices has been addressed in the meantime :smile:

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
Soroosh129: I changed this back to be a const reference.

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
Soroosh129: I think it did have a slight performance benefit to use smart pointers.

Soroosh129:
Thanks @cmnrd for your great comments. I have addressed them as best as I could.

cmnrd (approved):

PR #719, opened 2021-11-04
Revised LS Compile Action Class

a-sr:
Converted compile action handler from Xtend to Java.

Also added LF compile button to toolbar.


lhstrh (approved):
Thanks, @a-sr! This looks great!

PR #720, opened 2021-11-04
Fix off by 1 error in C++ benchmark runner

jhaye:
The `nextIteration` reaction, where `count` is incremented, is always executed before `finished`. Thus this would skip `measuredTimes[0]` and leads to inaccurate reporting of the timings.


cmnrd (commented):
Good catch, thanks! It might be cleaner to increment `count` in the reaction to `finished` and not in `nextIteration`. What do you think?

I think we never really relied on the printed summary. Our benchmark runner script does its own calculations, so this error had no impact on the results that we already obtained .

jhaye:
> Good catch, thanks! It might be cleaner to increment `count` in the reaction to `finished` and not in `nextIteration`. What do you think?

That's probably more clear, yes.

cmnrd (approved):

PR #721, opened 2021-11-04
Big and NQueens C Benchmarks

LDeng0205:
There is an issue in the NQueens benchmark of getting the warnings of freeing already freed tokens every time `reaction(next)` is triggered in Manager. Every time "next" is triggered, 20 (current number of Workers) warnings appear. It shouldn't be due to any `free()` calls as I tried commenting out all the `free()` and the issue persisted.

### DIFF
--- a/benchmark/C/Savina/src/include/deque.c
+++ b/benchmark/C/Savina/src/include/deque.c
@@ -1,0 +1,213 @@
+/**
+@file
+@author Arthur Deng
+@author Edward A. Lee
+@section LICENSE
+Copyright (c) 2021, The University of California at Berkeley.
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+1. Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+@section DESCRIPTION
+This provides an implementation of a double-ended queue.
+Each node contains a void* pointer.
+To use this, include the following in your target properties:
+To use this, include the following in your target properties:
+<pre>
+target C {
+    cmake-include: "/lib/c/reactor-c/util/deque.cmake"
+    files: ["/lib/c/reactor-c/util/deque.c", "/lib/c/reactor-c/util/deque.h"]
+};
+</pre>
+In addition, you need this in your Lingua Franca file:
+<pre>
+preamble {=
+    #include "deque.h"
+=}
+</pre>
+To create a deque, use calloc to ensure that it gets initialized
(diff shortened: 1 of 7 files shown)


cmnrd (approved):
This looks great! I will merge this despite the warnings so that we can get some first measurements. 

Regarding the warnings about multiple frees, I couldn't find a smoking gun in the code. What strikes me as odd is that it says "WARNING: Token being freed that has already been freed". So it seems this comes from the C runtime. However, as far as I can see, no token type is used in the program. Or is `{=foo*=}` interpreted as a token despite the `{=...=}`? @Soroosh129 @edwardalee could you take a look?

edwardalee:
I have no bandwidth, but yes, I think that `{=foo*=}` is interpreted as a token type.  To suppress that interpretation, you can do a `typedef` in the the preamble so that the declared type does not have  a trailing `*`. But why is a pointer type being used?

LDeng0205:
Thanks! I changed `more_work` from `{=work_queue_t *=}` to `work_queue_t` and it fixed the warnings! However, I am still curious as to why using a pointer would trigger the warnings.. If I wanted to send a larger `struct`, wouldn't it be more efficient to use a pointer instead of the value?

lhstrh:
The way it currently works is that if you use a pointer type, then the runtime system assumes responsibility of managing the allocated memory. You can circumvent this by introducing a typedef because we're just string matching and looking for a `*`...

There's more written about this [here](https://github.com/lf-lang/lingua-franca/wiki/Writing-Reactors-in-C).

edwardalee:
First, you are not passing a large data structure. You have this:
```
    typedef deque_t work_queue_t;
```
and your port types are `work_queue_t`.  This type is defined in `deque.h` and is:
```
typedef struct deque_t {
    struct deque_node_t* front;
    struct deque_node_t* back;
    size_t size;
} deque_t;
```
The total size of what you are passing through the ports is two pointers and one int.  If you want to pass just one pointer, you can do it by declaring this:
```
    typedef deque_t* work_queue_t;
```
The code generator will see the type `work_queue_t`, which does not have a `*` suffix, and hence it will not assume responsibility for memory management.

Note that passing pointers between reactors (which you are doing even without the above suggested change) is quite dangerous.  In general, reactors can execute in parallel, which means that more than one reactor could end up manipulating the same data structure in two separate threads. This can result in your deque becoming corrupted. If the application has no parallelism between reactors accessing the data structure, then you can get away with it. But such reactors are not reusable.  Perhaps a deque is not really the right thing to be passing between reactors? I haven't looked at the application carefully, so I have no suggestions here.

LDeng0205:
Yep, I saw that in this case passing in the struct wouldn't actually make a difference as the size is small so I changed it accordingly [#729 ](https://github.com/lf-lang/lingua-franca/pull/729). In this benchmark no two Workers should be modifying the same queue (so there should be no parallel access), which is more like a buffer for transmitting and processing solutions (now it shouldn't matter as I'm passing the struct item). But I do see how passing in pointers can be dangerous, and will definitely take this into account when I write programs in the future. Thanks!

edwardalee:
You are still passing pointers, so I'm not sure what you mean by "now it shouldn't matter as I'm passing the struct item." The struct item contains pointers, so you are passing pointers, which is generally dangerous in LF.

LDeng0205:
I see.. One way to improve it might be using arrays and C Target's macros (instead of deques) for setting output variables, which would not affect efficiency but would make the code cleaner and safer. Should I make this change?

edwardalee:
What does the Cpp version do?  The Akka version?  Message sizes could have an impact on performance if they are copied. The token mechanism avoids copying when possible...

cmnrd:
I think the solution as it is now is fine. Copying the struct shouldn't cost too much. Akka moves around references, and LF-Cpp uses smart pointers (as for all values).

LDeng0205:
Under the current parameters (`threshold = 4` and `size = 12`), a total of ~5,000 arrays of size <= 12 will be copied, but the numbers go up exponentially if `threshold` or `size` is increased. Is there a way to circumvent this copying without using a pointer? How would I send an array of arrays using the current macros?

edwardalee:
You are asking the right questions. These questions are the most difficult part of programming in C, especially when your programs execute concurrently. There is no general answer to your question. How to handle this depends on many things. If you can guarantee that no two concurrent processes will simultaneous access or modify the data structure being pointed to, then you can use pointers.  However, even that is too simple.  If you can guarantee that there is only one writer that modifies the data structure, and that readers do not start reading until the writer is done, and that the writer does not update until all readers are done, then you can again use pointers. The point of automated memory management (including our token types with reference counting) is to relieve you of that burden. But if you are willing to take on the burden, you can make things much more efficient.

cmnrd:
I don't understand... Where are the arrays copied? In the N Queens benchmark, the manager sends a `work_item_t` to the workers. This struct will be copied, but not the underlying data. The `work_item_t` has `size_t* data` and this pointer will simply be copied. So copying a `work_item_t` only requires copying two 64-bit values. Or is there something else that I am missing?

cmnrd:
Or are you referring to these copies?
```
// copy current configuration into new array b and set next row to contain a queen
// at position "cnt"
for (idx = 0; idx < depth + 1; idx++) {
    b[idx] = a[idx];
}
```

```
  // copy depth items from a to b
  int j = 0;
  for (j = 0; j < depth; j++) {
      b[j] = a[j];
  }
```

These copies are essential for the algorithm, since new lists are created. There might be ways to avoid a few of those copies, but I am not sure how to do it safely. However, one obvious optimization would be to use `memcopy` instead of copying all array elements individually in a for loop.

LDeng0205:
I was just thinking of whether it's possible to remove all memory management in C by using current macros. I agree that for this benchmark the implementation should remain as is, with potential optimizations. I don't think race conditions will happen in spite of the use of pointers.

PR #723, opened 2021-11-05
Faster PiPrecision

Soroosh129:
This speeds up the PiPrecision considerably (by about 9x on my machine for the unthreaded C runtime).

The main reason is the usage of `mpf_div_2exp` instead of `mpf_div`.

Also, I left a FIXME about a weird problem. GMP doesn't let me clear the global `mpf_t` variables (which frees the memory).
I think making these global `mpf_t` variables local again to the `calculateBbpTerm` function is not going to be a huge performance hit if you think we should do that.

### DIFF
--- a/benchmark/C/Savina/src/parallelism/PiPrecision.lf
+++ b/benchmark/C/Savina/src/parallelism/PiPrecision.lf
@@ -22,7 +22,8 @@ target C {
     ]]] */
     threads: 0,
     /// [[[end]]]
-    cmake-include: "PiPrecision.cmake"
+    cmake-include: "PiPrecision.cmake",
+    build-type : RelWithDebInfo
 };
 
 preamble {=
@@ -156,23 +157,21 @@ reactor Worker {
     
     preamble {=
         #include <gmp.h>
+        mpf_t sixteen;
+        mpf_t four;
+        mpf_t two;
+        mpf_t one;
         void calculateBbpTerm(mpf_t result, uint32_t k) {
             
+            instant_t time1, time2;
+            time1 = get_physical_time();
             uint32_t eightK = 8ul * k;
-            mpf_t sixteen;
-            mpf_t four;
-            mpf_t two;
-            mpf_t one;
             mpf_t term;
             mpf_t intermediate;
-            
-            mpf_init_set_ui(sixteen, 16ul);
-            mpf_init_set_ui(four, 4ul);
-            mpf_init_set_ui(two, 2ul);
-            mpf_init_set_ui(one, 1ul);
             mpf_init_set_ui(term, 0ul);
             mpf_init_set_ui(intermediate, 0ul);
             
+
             mpf_div_ui(term, four, (eightK + 1ul));
             mpf_div_ui(intermediate, two, (eightK + 4ul));
             mpf_sub(term, term, intermediate);
@@ -181,13 +180,9 @@ reactor Worker {
             mpf_div_ui(intermediate, one, (eightK + 6ul)); 
             mpf_sub(term, term, intermediate);
             
-            mpf_pow_ui(sixteen, sixteen, k);
-            mpf_div(result, term, sixteen);
+
+            mpf_div_2exp(result, term, 4 * k);
             
-            mpf_clear(sixteen);
-            mpf_clear(four);
-            mpf_clear(two);
-            mpf_clear(one);
             mpf_clear(intermediate);
             mpf_clear(term);
         }
@@ -196,12 +191,23 @@ reactor Worker {
     input doWork:uint32_t;
     output response:mpf_t;
     
+    reaction(startup) {=
+        mpf_init_set_ui(sixteen, 16ul);
+        mpf_init_set_ui(four, 4ul);
+        mpf_init_set_ui(two, 2ul);
+        mpf_init_set_ui(one, 1ul);
+    =}
+    
     reaction(doWork) -> response {=
         mpf_t result;
         mpf_init(result);
         calculateBbpTerm(result, doWork->value);
         mpf_swap(response->value, result);
         SET_PRESENT(response);
+    =}
+    
+    reaction(shutdown) {=
+        // mpf_clears(one, two, four, sixteen, NULL); FIXME: This causes a "double free or corruption" error
     =}
 }
 


Soroosh129:
By the way, a very useful tool was a whole lot of calls to `get_physical_time()` and printing the difference. Since we use high-precision clock, these measurements ended up being very effective.

benchmark/C/Savina/src/parallelism/PiPrecision.lf
```diff
@@ -196,13 +189,24 @@ reactor Worker {
...
     reaction(doWork) -> response {=
         mpf_t result;
         mpf_init(result);
         calculateBbpTerm(result, doWork->value);
         mpf_swap(response->value, result);
         SET_PRESENT(response);
     =}
+    
+    reaction(shutdown) {=
+        // mpf_clears(one, two, four, sixteen, NULL); FIXME: This causes a "double free or corruption" error
```
cmnrd, line 208: That is not surprising, as this shutdown reaction will be invoked multiple times for each worker. If we use global variables, then we should probably allocate and free them in the main reactor.

benchmark/C/Savina/src/parallelism/PiPrecision.lf
```diff
@@ -196,13 +189,24 @@ reactor Worker {
     input doWork:uint32_t;
     output response:mpf_t;
     
+    reaction(startup) {=
```
cmnrd, line 192: Each worker will now initialize the global variables (potentially in parallel). This doesn't look safe...

cmnrd (commented):
I think it is fine to make the variables global, but the current way of initializing them is error prone (see my comments).

I think your points about increasing the performance are valid. However, we should make the same changes in C++ as well, and then we will have a 9x improvement in both versions. This still doesn't explain the difference between C and C++. I would even expect the effect to be bigger, since we reduce the workload and hence any overhead will have more weight.

benchmark/C/Savina/src/parallelism/PiPrecision.lf
Soroosh129: Ah! You're right. Completely missed that.

Soroosh129:
> This still doesn't explain the difference between C and C++. 

I did an analysis on this using the tracing tool and realized that about 98% of the execution time in the unthreaded version of this benchmark was seemingly spent in the worker's reaction. I used `get_physical_time()` to measure the overhead of each part of the worker, and realized that the aforementioned line is to be blamed. We can do a similar analysis in C++. One guess is that perhaps some of the GMP we perform in C are slower than the C++ implementation.

cmnrd:
The GMP code is actually pretty much the same (in the end the C++ version only uses a wrapper). Anyhow, I updated the C++ versiona and made quick run on Eiffel
piprecise

Both the C and C++ version got significantly faster. For 1 thread, C seems to be a bit faster now than C++. However, when 
adding more threads, C becomes slower again. Since the workload is now lower, the thread synchronization overhead in C now plays a more significant role.

Soroosh129:
> The GMP code is actually pretty much the same (in the end the C++ version only uses a wrapper).

The C unthreaded version went from being significantly slower to being faster compared to 1 thread in Cpp with a 1 line change. I think there might be some discrepancies here. Just pausing the [gmpxx.h](https://github.com/alisw/GMP/blob/master/gmpxx.h) code, it does look like the Cpp layer is doing quite a lot of things that look like optimizations to me before calling the C API.

cmnrd:
Actually, I just looked back at the Akka implementaion:
``` 
   protected static BigDecimal calculateBbpTerm(final int scale, final int k) {
        final RoundingMode roundMode = RoundingMode.HALF_EVEN;

        final int eightK = 8 * k;
        BigDecimal term = four.divide(new BigDecimal(eightK + 1), scale, roundMode);
        term = term.subtract(two.divide(new BigDecimal(eightK + 4), scale, roundMode));
        term = term.subtract(one.divide(new BigDecimal(eightK + 5), scale, roundMode));
        term = term.subtract(one.divide(new BigDecimal(eightK + 6), scale, roundMode));
        term = term.divide(sixteen.pow(k), scale, roundMode);
        return term;
    }
```
I wasn't thinking about this earlier, but now it seems to me that the optimization we implemented is rather unfair in comparison to the Akka version. It creates new Values for the constants every time and also uses a general `pow` function instead of a specialized base-2 one. Surely this could also be optimized in Scala, but since it isn't, we should probably not make this optimization.

Soroosh129:
> I wasn't thinking about this earlier, but now it seems to me that the optimization we implemented is rather unfair in comparison to the Akka version. It creates new Values for the constants every time and also uses a general `pow` function instead of a specialized base-2 one. Surely this could also be optimized in Scala, but since it isn't, we should probably not make this optimization.
> 
> What do you think?

The problem I think is that, in this benchmark, we're seemingly pretty much comparing the performance of BigDecimal in Java with GMP in C and C++. Each of these three have their own implementation (I'm separating the C and C++ GMP because the C++ version does seem to have a significant translation layer on top of the C version). Looking around, I couldn't find anything about how BigDecimal performs its division. However, it doesn't seem to have a function specific to power of two denominators. I think as long as we're using the best APIs in BigDecimal vs. the best tools in GMP, we might be okay.

cmnrd:
Yes, we are comparing apples to pears here. There is not much we can do about this, because there is no implementation of BigDecimal outside of the Java sphere. My point was more, that we at least should try to replicate the original workload. And this workload includes the creation (and deletion) of several new values each time the function is called, as well as a general power and division computation. This way, also the resulting plot is somewhat comparable to Akka. We are still a bit faster, but not 10x.

At least the creation of the constant values would be avoidable in Java/Scala, and I am sure there is a way to implement an optimized division for powers of 2 (even though this might not be part of the standard API).

Soroosh129:
Okay, fair enough. I guess up to you how you want to proceed. The GMP general operations in C will be a lot slower than C++.

cmnrd:
> Okay, fair enough. I guess up to you how you want to proceed. The GMP general operations in C will be a lot slower than C++.

Do you know which operations are used by the C++ code? In the end, it is just a wrapper around the C GMP library. So we should be able to do the same in both versions.

Soroosh129:
> Do you know which operations are used by the C++ code? 

I don't know, but we can certainly look at `gmpxx.h` a lot closer.

> In the end, it is just a wrapper around the C GMP library. So we should be able to do the same in both versions.

We could replicate some of the optimization-related stuff in `gmpxx.h` but I would rather use the right API from the beginning. If you are arguing for no optimizations, then we should probably not use `gmpxx.h` in both benchmarks since it appears to contain significant optimizations on top of `gmp.h`. Or, we could just use the CCpp target and use `gmpxx.h` in the C version as well.

cmnrd:
PiPrecise
This is from a new run on Eiffel using our master branch. Looks like C and C++ have about the same performance. C only gets a bit worse when adding more threads. So it seems the two implementation use the same gmp calls and there is nothing to worry about. The data we looked at last time was obtained on our other machine Shakespeare, and I am not sure why there is a difference. I could imagine that this was also a caching effect, since most memory accesses have a higher latency on Shakespeare.

Should we just close this?

Soroosh129:
> Should we just close this?

I think we shouldn't dismiss the previous results entirely, but I'm okay with tabling this for now.

PR #724, opened 2021-11-05
Simplified throughput aka fork join

lhstrh:
This draft is a WIP towards finding performance bottlenecks in the `micro/Throughput.lf` benchmark. @edwardalee  first worked on simplifying the implementation quite drastically and then started looking at the schedules by inspecting traces. We noticed that after one of the workers advances time, it goes on to pick up work, but other workers remain waiting. The issue appears to be a lack of notification after populating the reaction queue at the beginning of a new logical time.


lhstrh:
This is with 60 workers and 20 threads:
image

cmnrd:
Why is it a problem that the worker threads are waiting most of the time? This should be similar in C++. Unfortunately, the C++ tracing currently does not include the waiting status of the workers, but I could try to produce a similar graph for C++ if that helps.

In this particular benchmark there is not much parallelism. So one worker will easily process multiple reactions before other threads wake up. Quite quickly all reactions will be processed and some worker threads will still be waiting. There is simply not enough work for all threads. But that problem is inherent to the benchmark. To me, the question is: Why is there such a high-performance penalty when worker threads wait most of the time?

cmnrd (approved):
I like this simplification. I also simplified the C++ version, but I cannot use all the tricks there due to the benchmark runner. I also made a few adjustments. The computation function is modified to avoid dead code elimination, and I reintroduced a state variable to count the number of sent messages, as the timeout is tricky to handle from cog.

Issue #725, opened 2021-11-05
Timeout with value zero has different behaviour in C and Cpp


Rodario:
I have ported missing tests from the Cpp to C target and happened on a "bug" in one of them.
The test in question is the  [TimoutZero test.](https://github.com/lf-lang/lingua-franca/blob/ac2e091b5b52d50e67720db8b48bdd3b590b5921/test/C/src/TimeoutZero.lf).
My implementation looks like [this](https://github.com/lf-lang/lingua-franca/blob/62afd9652a12b67a972317f2fda47b7d52a0a787/test/Cpp/src/TimeoutZero.lf).

The "bug" is more of a implementation question it seems. A timeout of 0 in Cpp is handled as if there is no timeout where in C there is a termination after the 0th step.
I talked with @cmnrd about this and we wanted to ask, whats the correct behavior here?

Soroosh129:
Others might want to discuss the requirements of the language.

However, from a practical standpoint, is there a way to stop the program at tag `(0,0)` in the Cpp target?

cmnrd:
Yes, absolutely. Actually, the C++ runtime does not implement a timeout. Instead, the code generator simply injects a timeout reactor that sets a timer and it's reaction requests to stop the execution. In principle, the delay could be zero.

However, the issue at hand is the interpretation of zero. In C++, a timeout of zero is interpreted as invalid and hence there is no timeout. This is very convenient when using the CLI arguments. We can simply write `--timout 0` to disable a timeout.

How does the C runtime handle this? Which value has timeout if there is no timout configured?

Soroosh129:
> Instead, the code generator simply injects a timeout reactor that sets a timer and it's reaction requests to stop the execution.

I don't see how with this implementation, a stop tag of `(0,0)` could be achieved. A `request_stop()` call should stop the program at least one microstep later.

> Which value has timeout if there is no timout configured?

I don't understand this question. Do you mean what the default timeout is? It is infinity by default (or FOREVER to be specific).

PR #727, opened 2021-11-05
Port more C tests to Cpp

Rodario:
The TimeoutZero test is referenced in #725 and currently I would remove it from the PR if the discussion does not resolve the issue.
The Timeout test (without the Zero) had to be renamed to Timeout_Test because of a naming collision.
I hope I retained the original meaning of the tests. Maybe people who worked on those could comment if they spot any glaring issues, thought maybe @Soroosh129  for example?


Rodario:
#683db3a seems to have fixed the last bug on windows

test/Cpp/src/DoublePort.lf
```diff
@@ -0,0 +1,62 @@
...
+    =}
+}
+
+reactor Print {
+    input in:int;
+    input in2:int;
+    reaction(in, in2) {=
+        auto elapsed_time = get_elapsed_logical_time();
+        if(in.is_present()){
+            reactor::log::Info() << "At tag (" << elapsed_time << ", " << get_logical_time()
```
cmnrd, line 37: `get_logical_time()` will return a time point on the logical time axis. This is not the microstep. Unfortunately, there is currently no convenient API for getting a tag in C++ (see lf-lang/reactor-cpp#3). Currently, you would need to use `environment()->logical_time().micro_step()` to get the microstep of the current tag.

test/Cpp/src/ImportComposition.lf
```diff
@@ -0,0 +1,42 @@
...
+    imp_comp = new ImportedComposition();
+    state received:bool(false);
+    reaction(startup) -> imp_comp.x {=
+        imp_comp.x.set(42);
+    =}
+    reaction(imp_comp.y) {=
+        auto receive_time = get_elapsed_logical_time();
+        reactor::log::Info() << "Received " << *imp_comp.y.get() << " at time " << receive_time;
+        received = true;
+        if(receive_time.count() != 55000000LL) {
```
cmnrd, line 27: Thanks to chrono, this can be done simpler in C++: `if(receive_time != 55ms)`

test/Cpp/src/Alignment.lf
```diff
@@ -0,0 +1,100 @@
...
+        =}
+}
+
+reactor Destination {
+    input ok:bool;
+    input in:int;
+    state last_invoked:{=reactor::TimePoint=};
+    reaction(ok, in) {=
+        if (ok.is_present() && in.is_present()) {
+            reactor::log::Info() << "Destination: Input " << *in.get() << " is a prime at tag ( "
```
cmnrd, line 82: I think here it is fine to only print a logical time (as opposed to the complete tag including the microstep). However, the text should then better say "logical time" instead of tag. Also the output becomes easier to read when using `get_ellapsed_logical_time`, as it then starts counting from 0, instead of the physical start time.

test/Cpp/src/PhysicalConnection.lf
```diff
@@ -0,0 +1,31 @@
...
+    output out:int;
+    reaction(startup) -> out {=
+        out.set(42);
+    =}
+}
+reactor Destination {
+    input in:int;
+    reaction(in) {=
+        auto time{get_elapsed_physical_time()};
+        reactor::log::Info() << "Received " << *in.get() << " at logical time " << time.count() << ".";
```
cmnrd, line 20: You can print time directly, it will be formatted for you ;)

test/Cpp/src/PhysicalConnection.lf
```diff
@@ -0,0 +1,31 @@
...
+reactor Source {
+    output out:int;
+    reaction(startup) -> out {=
+        out.set(42);
+    =}
+}
+reactor Destination {
+    input in:int;
+    reaction(in) {=
+        auto time{get_elapsed_physical_time()};
```
cmnrd, line 19: Shouldn't this be *logical* time?

test/Cpp/src/PhysicalConnection.lf
```diff
@@ -0,0 +1,31 @@
...
+        reactor::log::Info() << "Received " << *in.get() << " at logical time " << time.count() << ".";
+        if(time.count() <= 0LL) {
```
cmnrd, line 21: It is better to compare with `reactor::TimePoint::zero()`

test/Cpp/src/ScheduleLogicalAction.lf
```diff
@@ -0,0 +1,51 @@
...
+    =}
+}
+
+reactor print {
+    state expected_time:time(0);
+    input x:int;
+    reaction(x) {=
+        auto elapsed_time = get_elapsed_logical_time();
+        reactor::log::Info() << "Result is " << *x.get();
+        reactor::log::Info() << "Current logical time is: " << elapsed_time.count();
```
cmnrd, line 34: Also here, no `count` is needed.

test/Cpp/src/SlowingClock.lf
```diff
@@ -0,0 +1,42 @@
...
+main reactor SlowingClock {
+    logical action a(100 msec);
+    state interval:time(100 msec);
+    state expected_time:time(100 msec);
+    reaction(startup) -> a {=
+        a.schedule(0ns);
+    =}
+    reaction(a) -> a {=
+        auto elapsed_logical_time = get_elapsed_logical_time();
+        reactor::log::Info() << "Logical time since start: " << elapsed_logical_time.count() << " nsec.";
```
cmnrd, line 25: Here you can also remove the count() and "nsec"

test/Cpp/src/Timeout_Test.lf
```diff
@@ -0,0 +1,44 @@
...
+
+import Sender from "lib/LoopedActionSender.lf"
+
+reactor Consumer {
+    input in:int;
+    state success:bool(false);
+    reaction(in) {=
+        auto current{get_elapsed_logical_time()};
+        if(current > 11ms ){
+            reactor::log::Error() << "Received at: " << current.count() << ". Failed to enforce timeout.";
```
cmnrd, line 20: No `count` needed.

cmnrd (commented):
Great! I added a few minor comments, mostly regarding the use of time in C++ code bodies.

test/Cpp/src/PhysicalConnection.lf
Rodario: I think I copied the the wrong error text here. This is the PhysicalConnection Test so the physical time is correct I think.

test/Cpp/src/PhysicalConnection.lf
Rodario: I could not find zero() in TimePoint but reactor::Duration::zero() seems to do the job?

cmnrd: Yep, right, that is what I meant ;)

cmnrd (approved):
Great! Thanks!

PR #729, opened 2021-11-05
Fixed Warnings in NQueens Benchmark

LDeng0205:
Fixed use of * to remove warnings of freeing already freed token.


cmnrd (approved):
Great!

PR #730, opened 2021-11-06
Python port c builtins

housengw:
Implement a python wrapper class for tag_t in the c target. 
Add support for get_microstep() and compare_tags() in the python target.
Add rich comparator for the Tag object (ex. get_current_tag() < some_other_tag)
Add test for get_microstep() and compare_tags() for the python target.


Soroosh129:
This looks great! Thanks for implementing this.

One note about `compare_tags()`: This is a necessity in C when it comes to comparing tags because there is no operator overloading in C. 

For Python, I think we can provide a better mechanism for comparison. I looked around and it looks like we can provide a [`.tp_richcompare`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_richcompare) function on the new `py_tag_t` type (which you named `Tag` in Python, which is a great name). This rich compare function would essentially do what `compare_tags` does but in a neater way so that the user can just directly compare two tags (e.g., using `<` and `>`).

Soroosh129 (approved):
Looks good to me!

PR #731, opened 2021-11-06
Update C and C++ Philosophers benchmarks

cmnrd:
This PR simplifies the C++ Philosophers benchmark and adjusts the C version accordingly. Also this PR integrates the C benchmark with the runner script.

My simplification had a significant impact on performance:
philosophers
For C++ it went down from 400-500 ms to a bit more than 100 ms. This improvement was achieved by throwing out unnecessary actions in the philosophers. There is now only a single action in the arbitrator to break the loop. While this is probably not a good example of LF usage, it implements the original benchmark very efficiently.

There is obviously a significant scaling problem in C. Oddly, also the single threaded C versions are significantly slower than C++. I am pretty sure that both versions implement the same workload.


edwardalee (approved):
Looks good to me. I stated that `info_print` has negligible overhead, but it may be worth verifying that using this benchmark.

lhstrh (approved):
Looks good!

lhstrh:
> Looks good to me. I stated that `info_print` has negligible overhead, but it may be worth verifying that using this benchmark.

Commenting out the `info_print` statements didn't appear to make a significant impact.

lhstrh (approved):

PR #732, opened 2021-11-06
Formatted C++ examples for better readability

revol-xut:
# Formatting C++ Examples

**What did I change ?**
- the C++ examples had very inconsistent and irritating indentation. 
- removed trailing white spaces
- moved some code around
- added some extra lines where necessary

**What did I test ?**
- [x] successfully compiled and run every changed example except ROS where for some reason only the code generator was triggered without compiling the results.

### DIFF
--- a/example/Cpp/src/CarBrake/CarBrake.lf
+++ b/example/Cpp/src/CarBrake/CarBrake.lf
@@ -13,76 +13,75 @@ reactor Camera {
 }
 
 reactor BrakingAssistant {
-    input frame: void;
+  input frame: void;
 	output trigger_brake: void;
-
 	
 	state counter: int(0);
 
 	reaction(frame) -> trigger_brake {=
-        // processing takes some time
+    // processing takes some time
  		std::this_thread::sleep_for(10ms);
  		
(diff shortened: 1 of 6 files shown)


lhstrh (changes requested):
While I see that there were some indentation problems in the code, this PR seems to mostly reduce the indentation from four spaces to two. The rest of the code base (mostly) uses two. If you want to propose to change this convention, you could start a [discussion](https://github.com/lf-lang/lingua-franca/discussions) thread about this. Could you change it to 4-space indentation for the time being? We should also document this in our "contributing" document that is still being drafted. I'll make a note. Thanks!

revol-xut:
Thanks for the advice @lhstrh. I changed it to use 4 spaces. 

It's just a little thing (indentation) but it bothered me while learning lingua franca. :D

lhstrh (approved):
Looks good! Thanks for the quick turnaround.

cmnrd:
Thanks @revol-xut for cleaning that mess ;) In the past, Epoch used tabs for indentation which created this mess. We have it now configured to insert 4 spaces.

PR #734, opened 2021-11-06
Updated script to work with the current LF program outputs

Soroosh129:


lhstrh (approved):
🚀

cmnrd:
Do we actually need that script given that we have the benchmark runner in `benchmark/runner`? Also, we should probably make clear that this script only works for the C target.

Soroosh129:
> Do we actually need that script given that we have the benchmark runner in `benchmark/runner`? 

The runner script doesn't report stats such as min, max, and average. For these stats, you would also need to integrate the BenchmarkRunner reactor into your program for example. We can add this functionality to `benchmark/runner` and then we don't need this script.

cmnrd:
It does already ;), just not directly. The runner script is intended to work together with the `collect_results.csv` script. This will give you min, max, average, and median of a single benchmark or from multiple if you did a multirun.

Soroosh129:
Okay. It does involve a bit more steps but I'm okay with removing the `measure-lf-time` script.

PR #738, opened 2021-11-09
C++ Throughput benchmark: make computed value volatile to defeat dead code elimination

jhaye:
The previous measures did not actually defeat dead code elimination. This does however and can be easily checked by putting a for loop around the function call. A difference should be easily visible with 1000+ iterations.


cmnrd (approved):
Thanks!

PR #743, opened 2021-11-11
C new scheduler

Soroosh129:
Creating a draft PR to test the new scheduler.

The code is still in very rough shape, but the scheduler is functional.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
@@ -818,7 +818,7 @@ class CGenerator extends GeneratorBase {
                     bool _lf_trigger_shutdown_reactions() {                          
                         for (int i = 0; i < _lf_shutdown_reactions_size; i++) {
                             if (_lf_shutdown_reactions[i] != NULL) {
-                                _lf_enqueue_reaction(_lf_shutdown_reactions[i]);
+                                _lf_enqueue_reaction(_lf_shutdown_reactions[i], -1);
                             }
                         }
                         // Return true if there are shutdown reactions.
@@ -929,7 +929,7 @@ class CGenerator extends GeneratorBase {
             pr('''
                 for (int i = 0; i < _lf_startup_reactions_size; i++) {
                     if (_lf_startup_reactions[i] != NULL) {
-                        _lf_enqueue_reaction(_lf_startup_reactions[i]);
+                        _lf_enqueue_reaction(_lf_startup_reactions[i], -1);
                     }
                 }
             ''')
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
@@ -532,7 +532,8 @@ class CGenerator extends GeneratorBase {
         if (targetConfig.threads === 0) {
             coreFiles.add("reactor.c")
         } else {
-            coreFiles.add("reactor_threaded.c")
+            coreFiles.add("threaded/scheduler.c")
+            coreFiles.add("threaded/reactor_threaded.c")
         }
         
         addPlatformFiles(coreFiles);
@@ -4958,7 +4959,7 @@ class CGenerator extends GeneratorBase {
         }
     }
 
-    /** Generate #include of pqueue.c and either reactor.c or reactor_threaded.c
+    /** Generate #include of pqueue.c and either reactor.c or threaded/reactor_threaded.c
      *  depending on whether threads are specified in target directive.
      *  As a side effect, this populates the runCommand and compileCommand
      *  private variables if such commands are specified in the target directive.
@@ -5084,7 +5085,7 @@ class CGenerator extends GeneratorBase {
     /** Add necessary source files specific to the target language.  */
     protected def includeTargetLanguageSourceFiles() {
         if (targetConfig.threads > 0) {
-            pr("#include \"core/reactor_threaded.c\"")
+            pr("#include \"core/threaded/reactor_threaded.c\"")
         } else {
             pr("#include \"core/reactor.c\"")
         }
--- a/test/C/src/concurrent/ThreadedThreaded.lf
+++ b/test/C/src/concurrent/ThreadedThreaded.lf
@@ -10,7 +10,9 @@ target C {
     timeout: 2 sec,
     tracing: true,
     flags: "", // Disable compiler optimization so that TakeTime actually takes time.
-    threads: 4
+    threads: 4,
+    build-type: RelWithDebInfo,
+    logging: DEBUG
 };
 reactor Source {
 	timer t(0, 200 msec);


Soroosh129:
Question for the weekly meeting next week: Is the method of testing all implemented runtime schedulers on all relevant LF tests a good approach?

If not, what other alternatives sound reasonable?

Soroosh129:
With considerable effort, I was able to resolve the merge conflict that this PR had with master. While resolving the conflict, I took the time to clean up the code (mostly in reactor-c) and also implement an optimization for the default NP scheduler, where the maximum number of reactions is statically determined at compile time and used to allocate the data structures in that scheduler (once we have mutations, we can think about resizing these data structures at runtime or use GEDF_NP).

Here are the latest results on the Savina benchmark suite, comparing the three implemented schedulers in this PR with the C runtime in master and also with the C++ target (thanks @petervdonovan for the script to run the benchmarks and generate this figure):
e76957c34-benchmarks

Looks like the default NP scheduler performs very well across the board. I believe this PR (and it's accompanying lf-lang/reactor-c#16) should be ready to merge. Let's try to merge this as soon as possible to avoid further conflicts down the road.

org.lflang.tests/src/org/lflang/tests/AbstractTest.java
```diff
@@ -192,4 +200,43 @@ public void runWithFourThreads() {
...
+                Message.DESC_SCHED_SWAPPING + scheduler.toString() +".",
+                categories::contains,
+                test -> {
+                    test.getContext()
+                        .getArgs()
+                        .setProperty(
+                                "scheduler", 
+                                scheduler.toString()
+                                ); 
+                    return true;
```
petervdonovan, line 235: Should the scheduler tests be run with four threads like [this](https://github.com/lf-lang/lingua-franca/blob/a1fe31566cad6f9afc834fbdd0b89abedc1f5ef0/org.lflang.tests/src/org/lflang/tests/AbstractTest.java#L235)? I only just remembered to mention it since testing seems to be so high in priority.

Soroosh129: I think it should be mostly okay. 

The reason that we force four threads is that multiport tests are not threaded (at least not in C, e.g., they don't specify `threads` in their target property). Also, some concurrent tests also specify one thread, which isn't particularly challenging for a scheduler. I thought setting the number of worker threads to four sounded like a good compromise.

We could also only set threads to four for the multiport test category.

petervdonovan: Oh, okay -- that makes sense. Thanks for explaining!

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -94,6 +94,14 @@ public void rebuild() {
         }
     }
     
+    /*
+     * Get an array of non-negative integers representing the maximum number of reactions 
+     * per each level, where levels are indices of the array.
+     */
+    public Integer[] getMaxNumOfReactionPerLevel() {
+        return maxNumOfReactionPerLevel.toArray(new Integer[0]);
```
lhstrh, line 102: I find the terminology used here a bit confusing; this is not really about the maximum, but rather the total number of reactions per level.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -206,6 +214,16 @@ protected void addNodesAndEdges(ReactorInstance reactor) {
         }
     }
     
+    ///////////////////////////////////////////////////////////
+    //// Private fields
+    
+    /**
+     * Maximum number of reactions per level, represented as a list of
```
lhstrh, line 221: Again, total; not maximum.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -254,4 +277,43 @@ private void assignLevels() {
...
+        }
+    }
+    
+    /**
+     * Adjust {@link #maxNumOfReactionPerLevel} at index <code>level<code> by
+     * <code>valueToAdd<code>.
+     */
+    private void adjustMaxNumOfReactionPerLevel(int level, int valueToAdd) {
+        if (maxNumOfReactionPerLevel.size() > level) {
+            maxNumOfReactionPerLevel.set(level, maxNumOfReactionPerLevel.get(0) + valueToAdd);
```
lhstrh, line 313: This I don't understand. Why is this happening?

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -254,4 +277,43 @@ private void assignLevels() {
...
+     * @param runtime The reaction runtime instance.
+     * @param level The level to assign.
+     */
+    private void assignLevel(Runtime runtime, Integer level) {
+        runtime.level = level;
+        adjustMaxNumOfReactionPerLevel(level, 1);
+    }
+    
+    /**
+     * Update the level of the reaction <code>runtime<code> instance
```
lhstrh, line 293: What is the rationale behind this logic?

lhstrh (commented):
This all looks good to me, but I got really stuck while reviewing the changes in `ReactionInstanceGraph`...

benchmark/runner/conf/default.yaml
```diff
@@ -1,8 +1,9 @@
 iterations: 12
 threads: null
+timeout: 1200
 savina_path: "${oc.env:SAVINA_PATH}"
 lf_path: "${oc.env:LF_PATH}"
-continue_on_error: False
+continue_on_error: True
```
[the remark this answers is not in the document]
Soroosh129: I think @petervdonovan might have a better idea about this.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -483,14 +483,19 @@ class CGenerator extends GeneratorBase {
...
             "reactor.h",
-            "pqueue.c",
-            "pqueue.h",
             "tag.h",
             "tag.c",
             "trace.h",
             "trace.c",
-            "util.h",
-            "util.c",
+            "utils/pqueue.c",
```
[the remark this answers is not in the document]
Soroosh129: It does, but reactor-c is really three different libraries: unthreaded, threaded, and federated. Copying the entirety of reactor-c files will cause a lot of unnecessary files to be copied over and bloating the `src-gen` directory.

Also, I remember that I tried to make copying of folders from classpath work a while ago, but couldn't get it to work the last time I tried. The issue was that the copy mechanism had to also work for JAR files, which was a headache to get to work right for folders. If we had such a mechanism, it would be really useful at least for our `utils`, `threaded`, and `federated` folders when they are needed. What do the other target code generators do?

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -916,6 +922,50 @@ class CGenerator extends GeneratorBase {
...
+                if (targetConfig.setByUser.contains(TargetProperty.SCHEDULER)) { 
+                    // FIXME: This potentially belongs in the validator
+                    // TODO: Move after LFValidator is ported
+                    errorReporter.reportWarning("This program contains deadlines, but the chosen "
+                        +targetConfig.schedulerType.name
+                        +" scheduler does not prioritize reaction execution "
+                        +"based on deadlines. This might result in a sub-optimal "
+                        +"scheduling.");
+                } else {
+                    targetConfig.schedulerType = TargetProperty.SchedulerOption.GEDF_NP;
```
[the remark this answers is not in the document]
Soroosh129: That's certainly doable. Should we create an issue for this or try to address it first before the merge?

benchmark/runner/conf/default.yaml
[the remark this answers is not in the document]
petervdonovan: I did this a while ago because it was not possible to run the benchmarks without it. Since the `PEDF_NP` scheduler is not currently in use, I think we can revert this change.

org.lflang/src/org/lflang/generator/python/PythonGenerator.xtend
```diff
@@ -856,13 +856,13 @@ class PythonGenerator extends CGenerator {
         from setuptools import setup, Extension
         
         linguafranca«topLevelName»module = Extension("LinguaFranca«topLevelName»",
-                                                   sources = ["«topLevelName».c", «FOR src : targetConfig.compileAdditionalSources SEPARATOR ", "» "«src»"«ENDFOR»],
-                                                   define_macros=[('MODULE_NAME', 'LinguaFranca«topLevelName»')«IF (targetConfig.threads !== 0 || (targetConfig.tracing !== null))», 
-                                                       ('NUMBER_OF_WORKERS', '«targetConfig.threads»')«ENDIF»])
+                                               sources = ["«topLevelName».c", «FOR src : targetConfig.compileAdditionalSources SEPARATOR ", "» "«FileConfig.toUnixString(src)»"«ENDFOR»],
```
Soroosh129, line 859: @lhstrh the string version of `FileConfig.toUnixString` is called here. Does it make sense to convert `src` to a path first? If so, would you please show me how I could do that efficiently?

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -718,6 +718,15 @@ public static IPath toIPath(URI uri) throws IOException {
     public static String toUnixString(Path path) {
         return path.toString().replace('\\', '/');
     }
+
+    /**
+     * Convert a given string path to a unix-style string.
+     * 
+     * This ensures that '/' is used instead of '\' as file separator.
+     */
+    public static String toUnixString(String path) {
```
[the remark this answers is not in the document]
Soroosh129: I added a comment on GitHub referencing the line at which this is used. I would appreciate your help on what to do about that piece of code.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
Soroosh129: The level assignment algorithm could reassign a new level to a reaction that was previously assigned a smaller level (based on a newly found `origin`). If that happens, we need to adjust the values in `adjustMaxNumOfReactionPerLevel` accordingly.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
Soroosh129: This is adjusting the number of reactions for `level` by the requested `valueToAdd`.

But, `maxNumOfReactionPerLevel` is a list and it might not have an element at index `level`, hence the `else` statement.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
Soroosh129: You're right.

org.lflang/src/org/lflang/FileConfig.java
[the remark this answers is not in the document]
lhstrh: You could use `Paths.get(src)` and then just invoke the existing `toUnixString` method.

test/C/src/Deadline.lf
```diff
@@ -44,6 +45,16 @@ reactor Destination(timeout:time(1 sec)) {
         }
         (self->count)++;
     =}
+
+    reaction(x) {=
```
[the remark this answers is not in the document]
lhstrh: @edwardalee pointed out that the test does not achieve what it aims to test because the two reactions execute in a predefined order since they are in the same reactor. Splitting them into separate reactors, however, would mean they wouldn't share state, making the test impossible. The only way I see this working is using a global variable and a mutex, which is an anti-pattern but could be OK for a test.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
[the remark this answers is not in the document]
lhstrh: Perhaps we should defer this until https://github.com/lf-lang/lingua-franca/pull/971 has been merged, since it provides functionality for copying entire directories. I think "bloat" in `src-gen` isn't really really a concern; it won't affect the size of the produced binary...

test/C/src/Deadline.lf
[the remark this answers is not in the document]
lhstrh: OK, then the solution should be to just revert the changes to this test.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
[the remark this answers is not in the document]
lhstrh: See #983.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
Soroosh129: I think I see at least one bug in `updateLevel`. Let's wait for @edwardalee.

edwardalee (approved):
@lhstrh and I have made a pass over this and I believe it is ready to merge. The c-new-scheduler branch of reactor-c should be merged first and then the reactor-c pointer updated to point to reactor-c/main.

lhstrh (approved):

PR #750, opened 2021-11-14
refactor test framework to support docker tests

housengw:
major changes:
- getExecCommand() returns a list of ProcessBuilders instead of one ProcessBuilder.
- execute() loops through the list of ProcessBuilders and marks the test as failed if any of the ProcessBuilders returns a non-zero exit code or times out. 
- add a special case in getExecCommand() before the target switch statement to make a list of docker related test commands 

Making execute() loop through a list of ProcessBuilders allows more flexible testing. 
Currently, in order to run a test that spans multiple commands (preprocessing, processing, cleanup), a script needs to be created and ran.


housengw:
Seems like docker is not natively supported on MacOS for Github Action. 
https://github.community/t/why-is-docker-not-installed-on-macos/17017

org.lflang.tests/src/org/lflang/tests/runtime/CTest.java
```diff
@@ -59,6 +59,11 @@ protected boolean supportsFederatedExecution() {
         return true;
     }
 
+    @Override
+    protected boolean supportsDockerOption() {
```
Soroosh129, line 63: We could use this opportunity to create an issue for what remains to be done for Python. I have created a new project called [Docker Support](https://github.com/lf-lang/lingua-franca/projects/11) to track the progress.

org.lflang/src/org/lflang/generator/python/PythonGenerator.xtend
```diff
@@ -1911,10 +1911,9 @@ class PythonGenerator extends CGenerator {
             # For instructions, see: https://github.com/icyphy/lingua-franca/wiki/Containerized-Execution
             FROM python:alpine
             WORKDIR /lingua-franca/«topLevelName»
-            COPY . src-gen
             RUN set -ex && apk add --no-cache gcc musl-dev \
-             && cd src-gen && python3 setup.py install && cd .. \
-             && apk del gcc musl-dev
```
Soroosh129, line 1917: I think the purpose of deleting things like `gcc` is to minimize the final image size. Are these dependencies still needed after the `python3 setup.py install`?

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -426,48 +456,163 @@ private void generateCode(LFTest test) {
...
+     * @param fedNameToDockerFile A mapping of federateName -> absolutePathToDockerFile
+     * @param testNetworkName The name of the network used for testing the docker option. 
+     *                        See https://github.com/lf-lang/lingua-franca/wiki/Containerized-Execution#federated-execution for more details.
+     */
+    private String getDockerRunScript(Map<String, Path> fedNameToDockerFile, String testNetworkName) {
+        StringBuilder shCode = new StringBuilder();
+        shCode.append("#!/bin/bash\n");
+        int n = fedNameToDockerFile.size();
+        shCode.append("pids=\"\"\n");
+        shCode.append(String.format("docker run --rm --network=%s --name=rti rti:test -i 1 -n %d &\n", testNetworkName, n));
```
Soroosh129, line 522: Would `ProcessBuilder` not be able to execute these commands and read their return value?

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -426,48 +456,163 @@ private void generateCode(LFTest test) {
...
+                             new ProcessBuilder("docker", "run", "--rm", "lingua_franca:test"),
+                             new ProcessBuilder("docker", "image", "rm", "lingua_franca:test"));
+    }
+
+    /**
+     * Return a list of ProcessBuilders used to test the docker option under federated execution.
+     * @param test The test to get the execution command for.
+     */
+    private List<ProcessBuilder> getFederatedDockerExecCommand(LFTest test) {
+        var rtiPath = test.fileConfig.getSrcGenBasePath().resolve("RTI");
```
Soroosh129, line 556: At some point, there should be a test that `docker` actually exists, and print an informative message to the user if not. We have similar checks for things like `gcc` somewhere. Remember that a user can run these tests locally.

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -426,48 +456,163 @@ private void generateCode(LFTest test) {
...
+        var rtiDockerPath = rtiPath.resolve("rti.Dockerfile");
+        Map<String, Path> fedNameToDockerFile = getFederatedDockerFiles(test);
+        try {
+            File testScript = File.createTempFile("dockertest", null);
```
Soroosh129, line 560: I think that the contents of this function can be made more compact by creating a list that contains the RTI as well as all the federates. Also, perhaps instead of creating a bash script, the execution commands could be appended to the list of commands? If the need for `sleep 3` is a concern, we should probably address the underlying problem, where the RTI must be up and running before federates can execute.

Soroosh129: Actually, I wonder if this type of testing would become redundant when we have `docker compose` tests (and figure out how to read the return value)? Maybe we should leave a FIXME here to come back to it later? That would make the workflow of this testing a lot cleaner perhaps.

Soroosh129 (changes requested):
Looks great! I left some suggestions. I wonder if we could just go for testing the federated docker support using Docker Compose. If that is complicated, I would advocate for merging this in and revisiting this issue later.

It would be great if @lhstrh could also comment on the changes made to the test framework.

org.lflang.tests/src/org/lflang/tests/runtime/CCppTest.java
```diff
@@ -33,18 +33,23 @@ public CCppTest() {
...
+     * Exclusion function for runAsCCpp test
+     */
+    static private boolean runAsCCppExclusion(TestCategory category) {
+        boolean excluded = false;
+        // Don't need to test examples.
+        // If any of them uses CCpp, it will
+        // be tested when compileExamples is
+        // run.
+        excluded |= (category == TestCategory.EXAMPLE);
+        excluded |= (!isLinux() && (category == TestCategory.DOCKER_FEDERATED || category == TestCategory.DOCKER_NONFEDERATED));
```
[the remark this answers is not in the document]
housengw: The docker tests are excluded when the OS is non-linux. The reason for that is docker does not natively support Mac on Github action, and workaround solutions are unstable (see: https://github.community/t/why-is-docker-not-installed-on-macos/17017). Docker on Mac runs on a Linux virtual machine. I decided that it would be the simplest to just test on Linux.

Soroosh129:
I have not tested this myself, but apparently, [this](https://github.com/saltzmanjoelh/RunDockerInMacOSAction/blob/a91b09d613f8d0aec2ff50afe9b25a0016618b47/.github/workflows/swift.yml) workaround exists to get MacOS docker support working on GitHub Actions. What do you think?

Also, have you by any chance tried enabling Windows for non-federated tests? Apparently, Docker is fully supported on Windows on GitHub Actions. Maybe it's worth giving it a try. It could help us be more confident to deploy/publish a non-federated LF application as a Docker image publicly.

org.lflang.tests/src/org/lflang/tests/runtime/CCppTest.java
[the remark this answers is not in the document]
Soroosh129: Ah I see. Yes I know about that limitation (see my comment below).
I think I misread this line of code.
Thanks for clarifying!

test/Python/src/HelloWorld.lf
```diff
@@ -2,8 +2,19 @@ target Python {
     tracing: true
 };
 reactor HelloWorld2 {
+    preamble {=
+        import sys
```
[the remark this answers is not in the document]
housengw: That's right

test/C/src/docker/docker_nonfederated/HelloWorldContainerized.lf
```diff
@@ -0,0 +1,12 @@
+target C {
```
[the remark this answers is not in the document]
housengw: I think keeping docker_nonfederated is more clear. Alternatively, I could remove the parent folder "docker" and put "docker_federated" and "docker_nonfederated" under "src".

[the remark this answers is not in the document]
Soroosh129: Well, so far we haven't distinguished LF applications as "non-federated" anywhere else. Federated is a specialized category of tests for now. I do think that these underscores are a bit awkward in the folder names (I've commented about `docker_federated` here as well).

There should be a more elegant way of combining two test categories, and I personally think that joining the category names with an underscore is not the prettiest.

org.lflang/src/org/lflang/generator/python/PythonGenerator.xtend
housengw: There is a tradeoff between build time and image size. Docker treats each RUN instruction as a "layer" and caches them. If I delete the dependencies, every build of a new lf program needs to re-add those dependencies whereas if I don't delete them, docker will use the cached layer.

test/C/src/docker/docker_nonfederated/HelloWorldContainerized.lf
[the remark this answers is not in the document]
housengw: that makes sense.

org.lflang.tests/src/org/lflang/tests/TestBase.java
housengw: To test federated execution, several processes need to run concurrently. The current implementation of the test framework does not allow that.

org.lflang.tests/src/org/lflang/tests/TestBase.java
housengw: `sleep 3` is an arbitrary number. I agree that testing with docker compose would be simpler. I'll look into that.

org.lflang/src/org/lflang/generator/python/PythonGenerator.xtend
Soroosh129: Interesting. Perhaps speeding up subsequent build times makes more sense for the Python target then.

org.lflang.tests/src/org/lflang/tests/TestBase.java
Soroosh129: I see. I guess the problem is that some commands do need to run sequentially (e.g., `docker build` and `docker run` for each federate) and some can (and should) run concurrently.

I think if we design this carefully enough, we could use something like Java thread pools to achieve concurrency. Not using bash means better compatibility with Windows, and Docker-enabled federated programs can really be a shortcut to getting federated execution to work on Windows. It's your choice though of course.

org.lflang.tests/src/org/lflang/tests/runtime/CTest.java
housengw: I added docker tests for non-federated python programs

org.lflang.tests/src/org/lflang/tests/TestBase.java
housengw: one way to do it is to run `which docker` and a non-zero exit code indicates that docker is missing.

Soroosh129: We have an [`LFCommand`](https://github.com/lf-lang/lingua-franca/blob/ce52695f2a07296cd63ae1a71cc05b45ad377b57/org.lflang/src/org/lflang/util/LFCommand.java#L50) class that already does this. There are some nuances, such as using `where` instead of `which` on Windows.

housengw:
Ok. Running Linux containers on Windows seems to require running them inside a Linux VM, as per: https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers

So I assume it suffices that we test the docker option in the Linux VM? What do you think @Soroosh129 ?

lhstrh:
> Ok. Running Linux containers on Windows seems to require running them inside a Linux VM, as per: https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers
> 
> So I assume it suffices that we test the docker option in the Linux VM? What do you think @Soroosh129 ?

I would say that suffices. If it runs in a VM anyway, we'd essentially just be testing the VM stack.

lhstrh:
I'll have a closer look at this PR tomorrow.

Soroosh129:
> Ok. Running Linux containers on Windows seems to require running them inside a Linux VM, as per: https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers
> 
> So I assume it suffices that we test the docker option in the Linux VM? What do you think @Soroosh129 ?

Interesting find. Thanks for digging into this. Based on this article and reading about how Linux-based Docker images work on MacOS (i.e., using linuxkit as a Linux VM), I agree with @lhstrh that testing on Linux seems to suffice. I think it was essential to find the limitations of what Docker can and cannot do for us. Thanks anyway :)

housengw:
according to [this](https://megamorf.gitlab.io/2020/05/06/why-it-s-better-not-to-use-alpine-linux-for-python-projects/) and [this](https://pythonspeed.com/articles/alpine-docker-python/), seems like python wheels do not work well with musl, which is the libc version alpine linux uses. Wheels of common packages (ex. numpy) are usually compiled glibc. I suspect this was the reason the CI was failing for the python docker tests, although I was able to build and run the python images locally.

As a temporary fix, I have changed the python docker base image from `python:alpine` to `python:latest`, which is based on Debian. This drastically increase the size of the built image from 177MB to around 900MB.

At first glance, the disk usage looks pretty bad. However, docker containers that use the same base image do not make a copy of that image, so building multiple containerized python LF programs only results in an overhead of 900MB. I was able to verify the disk usage using `docker system df` according to [here](https://stackoverflow.com/questions/26753087/how-to-analyze-disk-usage-of-a-docker-container)

Soroosh129:
> As a temporary fix, I have changed the python docker base image from `python:alpine` to `python:latest`, which is based on Debian. This drastically increase the size of the built image from 177MB to around 900MB.

To me, using `python:latest` is a perfectly reasonable choice since space-savings is not as a big deal for the Python target as it is for the C target. 

With that said, I was able to produce an image of size 446 MB using `python:slim` and adding `RUN apt-get update && apt-get install -y python3-pip`. In theory, `pip` is all you need here since setup tools is able to automatically install any required packages using `pip` as long as they are listed in the [`install_requires`](https://github.com/lf-lang/lingua-franca/blob/cd9b4fb3aae9e210cc2b90f21bddce5b29446f02/org.lflang/src/org/lflang/generator/python/PythonGenerator.xtend#L788) field in the `setup.py`. We don't currently have a way for the user to add to this list, but that can be straightforward to add as a target property. Do you foresee issues if we use `python:slim`?

housengw:
sounds good. I think using `python:slim` is fine.

org.lflang.tests/src/org/lflang/tests/TestBase.java
housengw: I think it is best to add docker-compose testing in a PR branched from #754. I'll mark this as resolved for now

Soroosh129 (approved):
Looks good to me!

I still think that @lhstrh should perhaps have a look at the changes before merging in, since this PR makes modifications to the test framework.

test/Python/src/docker/docker_nonfederated/PingPongContainerized.lf
```diff
@@ -0,0 +1,38 @@
...
+ * These are about an order of magnitude faster than anything reported in the paper.
+ * 
+ * @author Edward A. Lee
+ */
+target Python {
+    fast: true,
+    docker: true
+};
+
+import Ping from "../../PingPong.lf"
```
lhstrh, line 29: This file should also be in `docker`, not `docker/docker_nonfederated`, I think.

lhstrh (approved):
Looks good! I've made a couple suggestions here and there. It would be great if you could address them before merging this in.

org.lflang.tests/src/org/lflang/tests/Configurators.java
```diff
@@ -82,17 +82,17 @@ static boolean noChanges(LFTest test) {
      * categories.
      */
     public static boolean defaultCategoryExclusion(TestCategory category) {
```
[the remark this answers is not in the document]
housengw: will do!

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -426,48 +460,182 @@ private void generateCode(LFTest test) {
...
      */
-    private ProcessBuilder getExecCommand(LFTest test) {
+    private List<ProcessBuilder> getExecCommand(LFTest test) {
         final var nameWithExtension = test.srcFile.getFileName().toString();
         final var nameOnly = nameWithExtension.substring(0, nameWithExtension.lastIndexOf('.'));
+        
+        var srcGenPath = test.fileConfig.getSrcGenPath();
+        var srcBasePath = test.fileConfig.srcPkgPath.resolve("src");
+        var relativePathName = srcBasePath.relativize(test.fileConfig.srcPath).toString();
+        System.out.println(relativePathName);
```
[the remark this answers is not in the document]
housengw: good catch, my bad.

test/Python/src/docker/docker_nonfederated/HelloWorldContainerized.lf
```diff
@@ -0,0 +1,9 @@
+target Python {
+    docker: true
+};
+
+import HelloWorld2 from "../../HelloWorld.lf"
```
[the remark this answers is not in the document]
housengw: I forgot to change it for the Python target

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -426,48 +460,182 @@ private void generateCode(LFTest test) {
...
+     * @param test The test to get the execution command for.
+     */
+    private List<ProcessBuilder> getNonfederatedDockerExecCommand(LFTest test) {
+        if (!checkDockerExists()) {
+            System.out.println(Message.MISSING_DOCKER);
+            return Arrays.asList(new ProcessBuilder("exit", "1"));
+        }
+        var srcGenPath = test.fileConfig.getSrcGenPath();
+        var dockerPath = srcGenPath.resolve(test.fileConfig.name + ".Dockerfile");
+        return Arrays.asList(new ProcessBuilder("docker", "build", "-t", "lingua_franca:test", "-f", dockerPath.toString(), srcGenPath.toString()),
```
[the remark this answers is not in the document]
housengw: The instructions are documented here:
docker build: https://docs.docker.com/engine/reference/commandline/build/
docker run: https://docs.docker.com/engine/reference/run/
docker image: https://docs.docker.com/engine/reference/commandline/image/

lhstrh (approved):
🚀🚀🚀

PR #754, opened 2021-11-16
generate docker compose files for federated execution

housengw:
make sure tests pass after adding code generation for docker compose.
This branch is branched from #750 

Also, I will try to refactor CGenerator.doGenerate() since it currently has about 500 lines.


Soroosh129 (approved):
I think this looks great!

In the interest of a more modular design, it would be great if the docker-related functions were in a separate `CDockerGenerator.java` class (following the naming convention of the Cpp and TS generators).

lhstrh:
@housengw could you please provide a bit more explanation on what this PR is meant to achieve?

housengw:
This PR is meant to autogenerate `docker-compose.yml` files that help ease the deployment of docker containers. The user could do `docker compose up` in the directory where `docker-compose.yml` is located and the instruction in the yaml tells docker how to build and run the containers. Without the yaml, the user will have to build and run the an image for each federate manually.

PR #759, opened 2021-11-18
Scalability part I: Multiports

edwardalee:
This branch does a major refactoring of the instance classes, ReactorInstance, ReactionInstance, PortInstance, etc., converting them all to Java, eliminating the MultiportInstance class, and consolidating so that a multiport results in exactly one PortInstance being created rather than one PortInstance per channel.  This is a large PR with a lot of changes, but it is still a work in progress. Quite a bit more work needs to be done, most notably to handle banks more intelligently and create only one ReactorInstance per bank rather than one ReactorInstance per bank member.  That additional work will occur in a new branch derived from this one.


org.lflang/src/org/lflang/federated/FederateInstance.xtend
```diff
@@ -389,6 +353,84 @@ class FederateInstance {
         return !excludeReactions.contains(reaction)
     }
     
+    /** 
+     * FIXME: This redundant function is here only because it is used in
+     * another repo (for the TS target) and it is a royal pain to refactor
```
lhstrh, line 358: I'm confused about this comment. What code/repo is this comment referring to? The `reactor-ts` submodule only has TypeScript code in it...

org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.xtend
```diff
@@ -291,6 +291,20 @@ class LinguaFrancaShapeExtensions extends AbstractSynthesisExtensions {
 			]
 		}
 
+        // optional reaction level
+        val hasLevel = SHOW_REACTION_LEVEL.booleanValue
+        if (hasLevel) {
```
lhstrh, line 296:
```suggestion
        if (SHOW_REACTION_LEVEL.booleanValue) {
```

org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.xtend
```diff
@@ -291,6 +291,20 @@ class LinguaFrancaShapeExtensions extends AbstractSynthesisExtensions {
 			]
 		}
 
+        // optional reaction level
+        val hasLevel = SHOW_REACTION_LEVEL.booleanValue
```
lhstrh, line 295:
```suggestion
```

org.lflang/src/org/lflang/generator/ActionInstance.java
```diff
@@ -0,0 +1,123 @@
...
+    //// Protected fields.
+
+    /** The minimum delay for this action. */
+    TimeValue minDelay = DEFAULT_MIN_DELAY;
+    
+    /** The minimum spacing between action events. */
+    TimeValue minSpacing = null;
+    
+    /** The replacement policy for when minimum spacing is violated. */
+    String policy = null;
```
Soroosh129, line 119: Should the default be null?

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -106,7 +106,7 @@ abstract class GeneratorBase extends AbstractLFValidator implements TargetTypes
     /**
      * Constant that specifies how to name generated delay reactors.
      */
-    public static val GEN_DELAY_CLASS_NAME = "__GenDelay"
+    public static val GEN_DELAY_CLASS_NAME = "_GenDelay"
```
Soroosh129, line 109: Maybe this should be `_lf_GenDelay` to be consistent with other LF-related reserved keywords in C.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -1759,19 +1748,24 @@ abstract class GeneratorBase extends AbstractLFValidator implements TargetTypes
...
-                        federate.bankIndex,
-                        source.index,
-                        rightFederate,
-                        destination.parent.bankIndex,
-                        destination.index,
-                        this,
-                        targetConfig.coordination
-                    )
+                    // Make one communication for each channel.
+                    // FIXME: There is an opportunity for optimization here by aggregating channels.
```
Soroosh129, line 1752: Should we create an issue for this? I can take this up as an action item.

org.lflang/src/org/lflang/generator/ParameterInstance.java
```diff
@@ -0,0 +1,121 @@
...
+ * assignment in the instantiation that this parameter instance is a result of.
+ * If it is overridden, the parameter gets initialized using the value looked up
+ * in the instantiation hierarchy.
+ * @author{Marten Lohstroh <marten@berkeley.edu>}
+ * @author{Edward A. Lee <eal@berkeley.edu>}
+ */
+public class ParameterInstance extends NamedInstance<Parameter> {
+        
+    /** 
+     * Create a runtime instance from the specified definition
```
Soroosh129, line 53:
```suggestion
     * Create a parameter instance from the specified definition
```

org.lflang/src/org/lflang/generator/ParameterInstance.java
```diff
@@ -0,0 +1,121 @@
...
+        if (parent == null) {
+            throw new InvalidSourceException("Cannot create a ParameterInstance with no parent.");
+        }
+        
+        this.type = JavaAstUtils.getInferredType(definition);
+        this.init = parent.initialParameterValue(definition);
+        
+        // If the parent is in a bank and the parameter name is "bank_index", then
+        // override the default value provided to make it equal to the bank index.
+        if (parent.bankIndex >= 0 && getName().equals("bank_index")) {
```
Soroosh129, line 69: Would type cause an issue here (e.g., if the user defines `bank_index:string`)?

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+ * Representation of a runtime instance of a port.
+ * This may be a single port or a multiport.
+ *  
+ * @author{Marten Lohstroh <marten@berkeley.edu>}
+ * @author{Edward A. Lee <eal@berkeley.edu>}
+ */
+public class PortInstance extends TriggerInstance<Port> {
+
+    /**
+     * Create a runtime instance from the specified definition
```
Soroosh129, line 53:
```suggestion
     * Create a port instance from the specified definition
```

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+     * and with the specified parent that instantiated it.
+     * @param instance The Instance statement in the AST.
+     * @param parent The parent.
+     */
+    public PortInstance(Port definition, ReactorInstance parent) {
+        this(definition, parent, null);
+    }
+
+    /**
+     * Create a runtime instance from the specified definition
```
Soroosh129, line 63:
```suggestion
     * Create a port instance from the specified definition
```

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+     */
+    public PortInstance(Port definition, ReactorInstance parent, ErrorReporter errorReporter) {
+        super(definition, parent);
+        
+        if (parent == null) {
+            throw new NullPointerException("Cannot create a PortInstance with no parent.");
+        }
+        
+        // If this is a multiport, determine the width.
+        WidthSpec widthSpec = definition.getWidthSpec();
```
lhstrh, line 77: This could be factored out into a private method.

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+                    Parameter parameter = term.getParameter();
+                    if (parameter != null) {
+                        Integer parameterValue = parent.initialIntParameterValue(parameter);
+                        // Only a Literal is supported.
+                        if (parameterValue != null) {
+                            // This could throw NumberFormatException
+                            width += parameterValue;
+                        } else {
+                            errorReporter.reportWarning(definition,
+                                "Width of a multiport cannot be determined. Assuming 1."
```
Soroosh129, line 99:
```suggestion
                                "Width of multiport "+ getFullName() +" cannot be determined at compile time. Assuming 1."
```

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+     * * destinations A list of port ranges for destination ports, each
+     *   of which has the same width as rangeWidth.
+     *   
+     * Each item also has a method, getNumberOfDestinationReactors(),
+     * that returns the total number of unique destination reactors for
+     * its range. This is not necessarily the same as the number
+     * of ports in its destinations field because some of the ports may 
+     * share the same container reactor.
+     */
+    public List<SendRange> eventualDestinations() {
```
lhstrh, line 168: This method is too long.

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+                // The width may be given by a parameter or even sum of parameters.
+                width = 0;
+                for (WidthTerm term : widthSpec.getTerms()) {
+                    Parameter parameter = term.getParameter();
+                    if (parameter != null) {
+                        Integer parameterValue = parent.initialIntParameterValue(parameter);
+                        // Only a Literal is supported.
+                        if (parameterValue != null) {
+                            // This could throw NumberFormatException
+                            width += parameterValue;
```
Soroosh129, line 96: I think for the following comment to be accurate, this statement should perhaps be in a `try...catch`, with catch using the provided error reporter, if any:
> An error reporter, or null to throw exceptions.

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+                }
+            }
+        }
+    }
+
+    //////////////////////////////////////////////////////
+    //// Public methods
+
+    /**
+     * Cleared cached information about the connectivity of this port.
```
Soroosh129, line 115:
```suggestion
     * Clear cached information about the connectivity of this port.
```

org.lflang/src/org/lflang/generator/PortInstance.java
```diff
@@ -0,0 +1,616 @@
...
+     * In particular, eventualDestinations() and eventualSources()
```
Soroosh129, line 116:
```suggestion
     * In particular, {@link #eventualDestinations()} and {@link #eventualSources()}
```

lhstrh (approved):
These changes improve the handling of multiports and port a number of classes to Java. These are both great improvements! That said, I believe that if we want to address the fundamental memory consumption problems we've been seeing, the approach capitalized on in this design -- to create a full instantiation tree in memory -- is not tenable. While working on the operational semantics, I've sketched out an alternative design that would use significantly less memory. Such an overhaul, however, would not be easy and obviously outside of the scope of this effort. Having a clean implementation of multiports and banks will certainly also be helpful as a reference for such alternative implementation, so I nonetheless think it's a good idea to merge this into master. I've only left a few minor comments.

edwardalee:
I agree that creating a "full instantiation tree" is not tenable, and in fact, this PR is precisely addressing this issue by not instantiating all the channels of a multiport.  The scalability-banks branch (a work in progress) is further extending this by not instantiating all bank members.  Once these two are done, I do not believe further reduction will yield any benefit because the resulting graph structure is the minimum size required for diagramming, so the graph will need to be constructed anyway if we are to have diagramming.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -3869,194 +3896,136 @@ class CGenerator extends GeneratorBase {
...
-        generateReactorInstanceExtension(initializeTriggerObjects, instance, federate)
 
         // Generate code to initialize the "self" struct in the
         // _lf_initialize_trigger_objects function.
         pr(initializeTriggerObjects, "//***** Start initializing " + fullName)
-
-        // Start with parameters.
-        generateParameterInitialization(initializeTriggerObjects, instance)
+        
+        generateTraceTableEntries(instance, instance.actions, instance.timers)
```
Soroosh129, line 3904: There seems to be duplicate function calls here that are also in `generateFederate`. I wonder if these two could be consolidated, or the latter could be moved to `federated.CGeneratorExtension.java`

Soroosh129 (approved):
This is indeed an impressive improvement over the existing design. Looking forward to seeing part II of this effort. I left a few comments and some suggestions. I hope they can be helpful.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
edwardalee: Yes, this is on the to-do list. I've added a FIXME.

org.lflang/src/org/lflang/federated/FederateInstance.xtend
edwardalee: This function has been renamed "contains" from "containsReaction", so this is here just for backward compability in another repo.  The rename has to be done in the other repo, and then everything aligned. I don't have the bandwidth to do this.  I've added a more detailed explanation.

org.lflang/src/org/lflang/generator/PortInstance.java
edwardalee: Noted.

org.lflang/src/org/lflang/generator/ActionInstance.java
```diff
@@ -0,0 +1,123 @@
+/* Instance of an action. */
```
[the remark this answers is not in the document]
edwardalee: Good idea. Noted but deferred.

org.lflang/src/org/lflang/generator/ActionInstance.java
edwardalee: This is set unconditionally in the constructor, so null is not the default.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
edwardalee: Good idea. Done.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
edwardalee: It would be great if you would take this as an action item, but I suggest waiting until after scalability-banks is merged.

org.lflang/src/org/lflang/generator/PortInstance.java
edwardalee: Thanks. Comment is obsolete. Removed.

org.lflang/src/org/lflang/federated/FederateInstance.xtend
lhstrh: I still don't understand. What repo are we talking about? Can you provide a link?

cmnrd: If the method is here only to preserve compatibility, it should be labeled with `@Deprecated`. However, I also don't see how the TS runtime uses this method.

org.lflang/src/org/lflang/generator/ReactorInstance.java
```diff
@@ -0,0 +1,1297 @@
...
+ * Representation of a runtime instance of a reactor.
+ * For the main reactor, which has no parent, once constructed,
+ * this object represents the entire Lingua Franca program.
+ * The constructor analyzes the graph of dependencies between
+ * reactions and throws exception if this graph is cyclic.
+ *
+ * @author{Marten Lohstroh <marten@berkeley.edu>}
+ * @author{Edward A. Lee <eal@berkeley.edu>}
+ */
+public class ReactorInstance extends NamedInstance<Instantiation> {
```
cmnrd, line 69: This class is huge. A good rule of thumb seems to be to keep classes below 500 lines. Is there maybe a way to split this class up. Also several methods in this class are too long (The method on line 911 for instance). Generally, method bodies should fit on a single screen.

cmnrd (approved):
Looks good! This is quite difficult to review carefully due to the many changes made, but I didn't notice any read flags.

cmnrd:
BTW: There is something fishy going with the history in this branch. The commits between 
2cbe84e and 631e261 look like they are duplicated from master. It looks as if they were rebased. Was this on purpose?

As for the effectiveness of the changes made, I did a quick performance test compiling the bank transaction benchmark with 1000 accounts. It looks like this branch improves execution time, but increases the memory footprint even further.
`time -v` reports 123.68s user time on `master` and 10.2GB memory usage. On `scalability`, it reports 62.45s user time and 12.2GB memory usage. I am not sure if the changes made in this branch are supposed to influence the size of generated C code, but it is 1.2GB large on both branches.

edwardalee:
As for the fishy history, I have no idea... All I did was merge in master. Was this the wrong thing to do?

edwardalee:
The interleaved connection in Banking foils the methods used in this PR, I think. It will require a bit more work to handle those efficiently.

cmnrd:
> As for the fishy history, I have no idea... All I did was merge in master. Was this the wrong thing to do?

Merging master shouldn't be a problem. But it looks like what actually happened is not a merge, but a rebase (there is also no merge commit). I observed this happening a few time now (we also had this in one of Marten's branches), but I don't know what causes this. Consider for instance the commit ce5269 on master. It is the same as commit 631e261 on this branch, but they got different hashes. The two commits are duplicates. I don't know what is causing this though... How did you perform the merge? Directly with `git merge master`, or did you use some other tool?

edwardalee:
I used `git merge master`, then resolved conflicts and committed them, then did either `git rebase --continue` or `git rebase --abort` (the latter apparently needed if the conflicts are resolved in favor of what's in master).

cmnrd:
Oh, that explains it. When you are merging, you should use `git merge --continue` instead. I don't know what precisely happens if you start a merge and then continue with a rebase, but apparently it aborts the merge and rebases all the changes instead (i.e. duplicating the master commits and putting them on top of the branch).

edwardalee:
Ugh, ok.  Does this need to be undone? If so, how?

cmnrd:
No, I don't think that we need to fix it. Nothing actually broke, we will only have a few duplicated commits when we merge this into master. For fixing it we would need to rewrite the history of this branch, but his would come at the high risk of breaking something else.

lhstrh:
I want to be extra clear here to avoid further misunderstandings. When you do a `git pull --rebase` (`--rebase` is implicit if you have `pull.rebase` set to `true` in your configuration) and conflicts arise, them after resolving conflicts, you do a `git rebase --continue`. If you do a `git merge`, then after resolving conflicts you do a `git merge --continue`. Git tells also you this, but I the `merge` vs. `rebase` difference in the suggested commands is easily missed.

I'm glad we were able to pinpoint this, because it likely also explains some other issues that you have been experiencing with Git that really puzzled me.

PR #771, opened 2021-11-24
Support cmake versions older than 3.12 in the C++ target

cmnrd:


cmnrd:
I will merge this in without review as this issue is blocking someone. Please feel free to add comments even though this is merged.

PR #773, opened 2021-11-24
Platform

soyerefsane:
Added spike support for cross code generation.


lhstrh:
This is a rather old PR, but it never got merged, I think, because the tests never passed. The changes are now too stale because they affect a previous version of `CGenerator` that was written in Xtend rather than Java. In other words, this PR needs work :-) It would be great if you could make a start at making these changes ready to merge, @arengarajan99. 

Also tagging @soyerefsane who might be able to provide more context.

arengarajan99:
I can fix the PR specific to the platform specific information.

lhstrh:
OK, let's close this then.

Issue #419, opened 2021-07-21
Initializing pointer parameters in C++


cmnrd:
A drawback of #399 is that the initialization of parameters which have a pointer type does not work anymore. For instance
```
target Cpp;

reactor Foo(ptr: int*({=nullptr=})) {
    reaction (startup) {=
            std::cout << "Got " << *ptr << "\n";
    =}
}

main reactor {
    private preamble {=
        int bar{42};
    =}
    // don't do this at home
    foo = new Foo(ptr={=&bar=});
}
```
does not compile.

While generally speaking pointers should be avoided, sometimes they are needed and there is no way around it. So we should support the above program and have it compile correctly. However, this does not seem to be easy to solve because of C++'s peculiarities.

For now, this simple workaround can be used:
```
target Cpp;

reactor Foo(ptr: int_ptr({=nullptr=})) {
    public preamble {=
        using int_ptr = int*;
    =}
    reaction (startup) {=
        std::cout << "Got " << *ptr << "\n";
    =}
}

main reactor {
    private preamble {=
        int bar{42};
    =}
    // don't do this at home
    foo = new Foo(ptr={=&bar=});
}
```

PR #774, opened 2021-11-26
Init pointer parameters cpp

### LINKED ISSUE #605: Parameter types with more than one word in their name cause error in CPP target
Given a main reactor with a parameter of a type with more than one word:
`main reactor FooReactor( value:{=unsigned long=}(42) ){}`.

the build fails with the error message:
`error: expected primary-expression before ‘unsigned’
   63 |     typename std::add_lvalue_reference<typename std::add_const<unsigned long>::type>::type unsigned_long_value = unsigned long(42));`

This only occurs for parameters of a main reactor. Normal reactors can use the `{= unsigned long=}` syntax.

Rodario:
Fixes #605 and #419 by adding brackets to the righthandside types.


cmnrd:
This solution doesn't seem to work for general types and we need to find something better... closing this for now.

Issue #765, opened 2021-11-22
Use the same implementation of deque accross all C benchmarks


cmnrd:
Currently there are multiple implementations of a deque floating around. There is at least the one implemented in `benchmark/test/C/Savina/src/include/` and the one provided by our compiler via `files: ["/lib/c/reactor-c/util/deque.h", "/lib/c/reactor-c/util/deque.c"]`. There might be more versions hidden in some preamble code. We should decide for one version and use it consistently.

PR #624, opened 2021-10-18
banking and pseudorandom

LDeng0205:
Banking.lf and pseudorandom change

PR #777, opened 2021-12-01
Benchmark Runner for C Benchmarks

mattchorlian:
Opening this draft PR to track progress of converting all C benchmarks to use the benchmark runner:

- [x] Port current C++ benchmark runner to C
- [x] Big.lf
- [x] Chameneos.lf
- [x] Counting.lf
- [x] PingPong.lf
- [x] ThreadRing.lf
- [x] Throughput.lf
- [x] BoundedBuffer.lf
- [x] CigaretteSmoker.lf
- [x] Dictionary.lf
- [x] LogisticMap.lf
- [x] Philosophers.lf
- [x] SleepingBarber.lf
- [x] SortedLinkList.lf
- [x] Apsp.lf
- [x] FilterBank.lf
- [x] GuidedSearch.lf
- [x] MatMul.lf
- [x] NQueens.lf
- [x] PiPrecision.lf
- [x] RadixSort.lf
- [x] Trapezoidal.lf


mattchorlian:
I'm noticing that we have BoundedBuffer.lf and ProdCons.lf both here even though they are the same benchmark.  I suppose this is (in part) the Savina naming convention causing confusion again.  In any case, we should decide which one to keep around before merging this in!

cmnrd (changes requested):
See the comments I made in #852

lhstrh:
Does this PR also address https://github.com/lf-lang/lingua-franca/issues/765? If not, I think it probably should.

benchmark/C/Savina/src/parallelism/GuidedSearch.lf
```diff
@@ -84,6 +86,8 @@ preamble {=
     #include <memory>
 =}
```
lhstrh, line 88: This file, along with the included files, should be in a separate directory, as is done with the C++ benchmarks.

lhstrh (changes requested):
This looks great, @mattchorlian and @LDeng0205! I've only requested minor changes. Let's merge this into master ASAP.

housengw:
Seems like running `MatMul.lf` with more than 1 thread produces a random validation error. I'm guessing it's a memory issue..

benchmark/runner/conf/benchmark/savina_concurrency_concsll.yaml
```diff
@@ -38,10 +38,11 @@ targets:
...
       workers: ["--numWorkers", "<value>"]
   lf-c:
-    copy_sources:
+    copy_sources: 
+      - "${lf_path}/benchmark/C/Savina/src/BenchmarkRunner.lf"
       - "${lf_path}/benchmark/C/Savina/src/include"
       - "${lf_path}/benchmark/C/Savina/src/concurrency"
-    lf_file: "concurrency/SortedLinkList.lf"
+    lf_file: "concurrency/SortedList.lf"
     binary: "SortedLinkList"
```
cmnrd, line 46:
```suggestion
    binary: "SortedList"
```

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
   # Run the C benchmark tests. 
   c-benchmark-tests:
-    uses: lf-lang/lingua-franca/.github/workflows/benchmark-tests.yml@master
+    uses: lf-lang/lingua-franca/.github/workflows/benchmark-tests.yml@c-benchmark-runner
```
cmnrd, line 29: This should be reverted to master. There should be no need to change this unless the `benchmark-tests.yaml` workflow itself is updated in this branch.

cmnrd (commented):
This looks great! Looks like we are close to completing the C benchmarks.

Should we also include the Banking Benchmark #624 here and fix it up?

> Seems like running MatMul.lf with more than 1 thread produces a random validation error. I'm guessing it's a memory issue.

Could you describe what the issue seems to be? Is it a segfault that occurs sometimes? Or is the result invalid (since you mention a validation error)? The MatMul benchmark has a built-in race condition and hence the result is expected to be invalid for parallel execution. This is a bug in the original Savina benchmark...

housengw:
the `validation error` in multithreaded run of `MatMul.lf` is likely to be a race condition.

lhstrh:
> the `validation error` in multithreaded run of `MatMul.lf` is likely a race condition.

Could you elaborate?

housengw:
The MatMul benchmark has random validation errors when running with more than 1 thread. Since there is no segfault and the validation error happens at random indices in the matrix, it is most likely a race condition. 
As @cmnrd mentioned, the MatMul benchmark has a built-in race condition, so running it with multiple threads would cause such an error.

lhstrh (approved):
Looks ready to me! This is a big feat and will be very useful for the evaluation of the new schedulers in `reactor-c` 🚀🚀🚀 Thanks!

mattchorlian:
> Should we also include the Banking Benchmark #624 here and fix it up?

We decided to get this merged first and deal with Banking separately @cmnrd

lhstrh:
> > Should we also include the Banking Benchmark #624 here and fix it up?
> 
> We decided to get this merged first and deal with Banking separately @cmnrd

Comt to thing of it, there is one distinct advantage to adding the other stuff to this PR: we'll have a cleaner history and release notes (if we automatically generate them). Doesn't seem like it would be much work?

cmnrd (approved):
Looks great! I am happy with both options, merging as is or including banking as well.

LDeng0205:
I just added benchmark runner to Banking as well so we could include it with this PR. However, I still have a   java.lang.OutOfMemoryError when I compile it with NumAccounts = 1000.

edwardalee:
Is this branch merged with master? It looks like it is still generating in-line code, which suggests it is using a version of master that predates scalability-banks.

PR #784, opened 2021-12-05
add tests for python federated docker file generation

housengw:
Add tests for python federated docker file generation.


Soroosh129 (approved):
Looks great! Thanks for adding these tests.

PR #785, opened 2021-12-06
fix the lfc zip artefact in nightly builds

cmnrd:
zip does not preserve symbolic links and replaces bin/lfc with the actual
script. This breaks our mechanism for identifying the correct path and leads
to errors when running lfc. This patch fixes the problem by adding the
--symlinks flag to the zip command, so that symbolic links are preserved.


lhstrh (approved):
Nice catch! Thanks for the fix...

PR #789, opened 2021-12-08
Cpp: Fix handling of fast and keepalive; add C++ tests for fast, keepalive and timeout properties

### LINKED ISSUE #786: Cpp Reflex Game Early Termination
Cpp Reflex Game Example terminates before input is given.

image

cmnrd:
#561 introduced a bug to the handling of `fast` and `keepalive` in C++. If either of those properties was set to `true`, the command line parser would reset the runtime value to `false`, if the properties were not given explicitly as command line arguments. This bug caused #786.

This PR fixes this bug and, more importantly, adds C++ tests for the fast, timeout, and keepalive target properties. Since no test category seemed to fit these tests, I added a new category 'properties'.

Closes #786.


lhstrh (approved):
This looks great!

PR #790, opened 2021-12-08
Updated Language Server Implementation

lhstrh:
This PR provides the language server implementation developed by @petervdonovan and @a-sr. The VS Code Extension is removed because it is now hosted in its own repo.


org.lflang.diagram/src/org/lflang/diagram/lsp/LFLanguageServerExtension.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.xtext.ide.server.ILanguageServerExtension;
+import org.eclipse.xtext.ide.server.ILanguageServerAccess;
+
+/**
+ * Provides Lingua-Franca-specific extensions to the
+ * language server's behavior.
+ */
```
lhstrh, line 15:
```suggestion
 * @author Peter Donovan <peterdonovan@berkeley.edu>
 */
```

build.gradle
```diff
@@ -1,83 +1,84 @@
...
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+    }
 }
 
 plugins {
-	id 'com.github.johnrengelman.shadow' version '6.0.0'
-	id 'java'
-	id 'jacoco'
+    id "com.github.johnrengelman.shadow" version "6.0.0"
+    id "com.github.node-gradle.node" version "3.0.1"
```
lhstrh, line 14:
```suggestion
```

org.lflang.lds/uf.py
```diff
@@ -0,0 +1,227 @@
...
+import os
+import argparse
+import subprocess
+import time
+import threading
+import tempfile
+import shutil
+
+# This is the path to the fat jar that is to be updated.
+FAT_JAR = os.path.join('vscode-extension', 'ls', 'lflang-lds.jar')
```
lhstrh, line 14: This directory no longer exists. This also confuses me, because the `gradle.build` that existed in the `vscode-extension` directory was copying jars from a different location in the tree into `vscode-extension/lf/lflang-lds.jar`. This suggests that `uf.py` is executed _after_ the gradle build script moves stuff into this location. But where is it called from?

lhstrh: Tagging @a-sr and @petervdonovan for visibility.

lhstrh: @petervdonovan explained that it was a utility script that has outlived its usefulness. I've removed it.

lhstrh (commented):
This looks ready to merge. There are a few FIXMEs left in the code, but those can be addressed later. After merging this, **we should not attempt to merge master into the `vscode` branch**, because this will break it. Instead, we should get the extension, now hosted [here](https://github.com/lf-lang/vscode-lingua-franca)) to build independently, after which we should update all documentation and remove the `vscode` branch altogether.

PR #792, opened 2021-12-09
Typescript server gcloud example

housengw:
Add a proof-of-concept example of deploying the Typescript Server example on Google Cloud Platform.

I have gone through all the steps to make sure it works, but I did not add CI yet. 

The added docker file is temporary. Once TS supports docker file generation, that docker file can be removed.

### DIFF
--- a/example/TypeScript/src/HTTPServerReactor/HTTPServer.lf
+++ b/example/TypeScript/src/HTTPServerReactor/HTTPServer.lf
@@ -8,6 +8,7 @@ main reactor {
         import * as https from "https"
         import * as http from "http"
     =}
+    state count:number(0);
     state server:{=http.Server | undefined=}({=undefined=});
     physical action serverRequest:{= [ http.IncomingMessage, http.ServerResponse ] =};
     reaction (startup) -> serverRequest {=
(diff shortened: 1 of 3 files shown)


edwardalee (approved):
Wow, this is a very nice milestone! There are a lot of steps, but that's understandable at this point. Would putting a timeout in the LF program help accidentally incurring charges?

Another experiment to try would be to deploy an RTI to Google Cloud and then create a web server that serves a web page to control an LF program running on your laptop. I.e., there would be two federates, one in the cloud that is a "digital twin" of the one on the laptop. Change something in the digital twin, and the same thing changes on the laptop, and vice versa.

housengw:
Theoretically, running this demo would not incur any charge, but it is a good practice to do clean up even if a timeout is set on the server. 

There are two components involved with pricing in this demo:
1. the kubernetes cluster (via Google Kubernetes Engine).
2. the docker image of the LF Typescript server (via Google Container Registry)

The kubernetes clusters has a flat rate management fee of $0.10 per hour, but the Google's free tier provides $74.40 in monthly credits per billing account. See [Google's pricing docs](https://cloud.google.com/kubernetes-engine/pricing#cluster_management_fee_and_free_tier) for more details.

The container registry charges $0.10 per GB per month for docker images that are > 0.5GB and free for images <= 0.5GB. The image for the LF server used in the demo is 242MB. See [here](https://cloud.google.com/artifact-registry/pricing) for details.

PR #794, opened 2021-12-10
SMT generation via Uclid5

lsk567:
This draft PR documents the progress of UclidGenerator for LF. By automatically generating Uclid5 models that encode the LF semantics and k-induction formulas, users can prove invariants for an LF program.


lsk567:
@lhstrh Thanks for reminding me that this branch still exists. It is subsumed by #1271. I will close this one.

PR #795, opened 2021-12-10
Fix Eclipse Problems after VSCode Branch Merge

a-sr:
Commit [8080ed8](https://github.com/lf-lang/lingua-franca/commit/8080ed85a55910df537c23c0486361aaafdf3337) reverted a workaround that enabled error-free working in Eclipse in the presence of Kotlin files (PR #437). Consequently, the master currently has an error present only in Eclipse.

According to the commit message this was done in order to include Kotlin in the LS.
This PR adds Kotlin to the LS in a different way and re-applies the solution to resolve the Eclipse problem with the Kotlin dependency.

Additionally, it adds the sprotty update site (transient dependency of the Klighd LS) to the oomph setup because oomph seems to have problems resolving the dependency indirectly via Klighd.

### DIFF
--- a/oomph/LinguaFranca.setup
+++ b/oomph/LinguaFranca.setup
@@ -375,6 +375,8 @@
               url="https://github.com/fabioz/Pydev/releases/download/pydev_9_0_1/"/>
           <repository
               url="https://download.eclipse.org/wildwebdeveloper/releases/0.13.0/"/>
+          <repository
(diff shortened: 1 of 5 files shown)


lhstrh:
We are so lucky that you are keeping tabs on this, @a-sr! This problem had not even occurred to me yet. Thank you for the prompt fix!

lhstrh (approved):

PR #804, opened 2021-12-15
Adds validation for protected  main reactor parameters

Rodario:


org.lflang/src/org/lflang/validation/LFValidator.xtend
```diff
@@ -749,6 +749,19 @@ class LFValidator extends BaseLFValidator {
...
+        } 
+        if(this.target == Target.CPP) {
+            val container = param.eContainer
+            val reactor = container as Reactor
+            if(reactor.isMain){ 
+                // we need to check for the cli parameters that are always taken
+                val myList = #['t', 'threads', 'o', 'timeout', 'k', 'keepalive', 
+                                'f', 'fast', 'help']
+                if(myList.contains(param.name)){
+                    error("Cli parameter name is allready in use by Lingua Franca",
```
lhstrh, line 761:
```suggestion
                    error("Parameter '" + param.name + "' is already in use as command line argument by Lingua Franca,",
```

org.lflang/src/org/lflang/validation/LFValidator.xtend
```diff
@@ -749,6 +749,19 @@ class LFValidator extends BaseLFValidator {
             if (param.inferredType.isUndefined()) {
                 error("Type declaration missing.", Literals.PARAMETER__TYPE)
             }
+        } 
+        if(this.target == Target.CPP) {
+            val container = param.eContainer
+            val reactor = container as Reactor
+            if(reactor.isMain){ 
+                // we need to check for the cli parameters that are always taken
+                val myList = #['t', 'threads', 'o', 'timeout', 'k', 'keepalive',
```
lhstrh, line 758: I think it would be better to define these in the `Targets` class and to generalize this check (not only for C++).

lhstrh (commented):
I realize this is a draft, but I would just like to point out that it would be good to accompany this feature with a unit test.

org.lflang/src/org/lflang/validation/LFValidator.xtend
oowekyala: Just for reference, the Rust target avoids these naming conflicts by prefixing user-defined parameter names with `main-`. So if a user defines a `threads` parameter it ends up on the CLI as `--main-threads`. Maybe we should align both targets on either behavior.

Rodario: > Just for reference, the Rust target avoids these naming conflicts by prefixing user-defined parameter names with `main-`. So if a user defines a `threads` parameter it ends up on the CLI as `--main-threads`. Maybe we should align both targets on either behavior.

That sound like a good idea. 
I was not sure what our guideline is on that subject. 
In combination with @lhstrh  comment I will clean this up.

lhstrh:
Closing this PR because the proposed change set has been adopted in https://github.com/lf-lang/lingua-franca/commit/8e5092d23a923c33a252c0a9236862f792acadfd (see https://github.com/lf-lang/lingua-franca/pull/886)

Issue #810, opened 2021-12-20
`{=0=}` should not be a valid time literal


cmnrd:
While working on the handling of time values within our compiler, I found many inconsistencies, redundant code and weird constructs. The code regarding handling of time values (ASTUtils) and their validation (LFValidator) should be revised. One particular change that is needed is the handling of `{=0=}`. Currently, we interpret this as the time 0. However, we should refrain from evaluating values escaped with `{=...=}`.

lhstrh:
Why? The brackets are there predominantly for situations where we can't parse what's inside, but if it turns out we can, why shouldn't we? I never interpreted the funny brackets to denote "don't look inside of this" but rather "what you find in here is target-specific."

cmnrd:
Exactly, the brackets denote target code. But `0` is not a valid time literal in C++. So `{=0=}` should also not be a valid time in LF. We say that we treat everything inside `{=...=}` as a black box and I think we should stick to this rule. Otherwise, we will get all sorts of weird behavior. For instance, why can I write `timer t({=0=})` but not `timer t({=foo()=})`? Why can I write `timer t({=0=})` which is illegal in C++, but cannot write `timer t({=0s=}` which would be proper C++ code?

oowekyala:
Related to #657

cmnrd:
Right, I remembered this issue but did not find it anymore. Thanks!

oowekyala:
I am closing this as a duplicate because #657 has more references than this one (eg in todos in the code, and in Github issues)

Issue #803, opened 2021-12-15
After 0 causes NullPointerException in diagram synthesis


cmnrd:
This might be related to #801, but I decided to open a separate issue for this, as I suspect that the underlying cause is different. Since #620, we support `after 0` (omitting the time unit) on connections. However, using this in an LF program will trigger a NullPointerException in the diagram synthesis. When writing `after 0s` instead (with the time unit), the diagram synthesis works correctly.

PR #811, opened 2021-12-20
Fix rendering of '0' time in diagram synthesis

### LINKED ISSUE #801: Diagram synthesis is broken for many tests and examples
I noticed that many of our examples do not produce valid diagrams in Epoch anymore. Instead, only an error message regarding a NullPointerException is shown. I suspect that this is caused by the recent refactorings in #759, but it could also be any other of the recent large PRs.

Examples and tests that do not work include:
- example/C/src/Delay.lf
- example/C/src/Patterns/Chain_02_Pipeline.lf
- example/C/src/Patterns/Loop_01_Single.lf
- example/C/src/Patterns/Loop_02_SingleDelay.lf
- test/C/src/ActionWithNoReaction.lf
- test/C/src/After.lf
- test/C/src/Alignment.lf
- test/C/src/CountTest.lf
- test/C/src/DanglingOutput.lf
- test/C/src/Deadline.lf
- test/C/src/DelayedAction.lf
- test/C/src/ToReactionNested.lf
- test/C/src/TriggerDownstreamOnlyIfPresent2.lf
- test/C/src/UnconnectedInput.lf

Note that this list is not exhaustive. There are many more files that do not produce valid diagrams. I was not able to identify a clear pattern of which language elements produce valid or invalid diagrams.

I think this shows that we need to be more thorough with testing. The correct operation of our diagram synthesis is currently not validated at all. I will open a separate issue for this (see #802).

### LINKED ISSUE #802: Diagram synthesis should be tested
In some recent PR merges, the diagram synthesis broke for many of our tests and examples (see #801). That this could slip through unnoticed, shows that there is a big gap in our test infrastructure. We currently do not test the correct operation of the diagram package.

A simple test infrastructure for diagrams could simply generate diagrams for all of our tests and examples and see if any exceptions occur. While this is certainly not perfect and the output is not validated, this alone should give a high coverage and prevent issues such as #801. However, I suspect in order to make this work, we need to fix the compilation issues outlined in #398.

cmnrd:
This fixes various instances of nullpointer exceptions for LF times without a unit (i.e. the literal `0`).

 While working on this, I noticed a lot of issues with the code that handles time. However, while I started to implement a few fixes, I realized that part of this clean up is already done in #544. As to avoid further conflicts, this PR only fixes the issues at hand. To keep track of the needed clean-up, I opened #810.

Closes #801, #803

### DIFF
--- a/org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.xtend
+++ b/org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.xtend
@@ -428,11 +428,11 @@ class LinguaFrancaShapeExtensions extends AbstractSynthesisExtensions {
 		).boldLineSelectionStyle
 		
 		val labelParts = newArrayList
-		if (timer.offset !== TimerInstance.DEFAULT_OFFSET) {
-			labelParts += timer.offset.toString
+		if (timer.offset !== TimerInstance.DEFAULT_OFFSET && timer.offset !== null) {
+        labelParts += timer.offset.toString
 		}
-		if (timer.period !== TimerInstance.DEFAULT_PERIOD) {
-			labelParts += timer.period.toString
+		if (timer.period !== TimerInstance.DEFAULT_PERIOD && timer.period !== null) {
+        labelParts += timer.period.toString
 		}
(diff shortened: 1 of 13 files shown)


edwardalee (approved):
These changes look good to me.

a-sr (approved):
Looks ok.

PR #813, opened 2021-12-28
Vscode handle build result

petervdonovan:
This PR, along with the accompanying PR [here](https://github.com/lf-lang/vscode-lingua-franca/pull/22), adds the following features to the VS Code extension:
* Progress reporting for both complete and partial builds
  * For complete builds triggered by the "Build" command, this appears as a toast notification.
  * For partial builds triggered by file saves, this appears in the blue progress bar at the bottom of the editor.
    * This feature exists for C++ only, but it should soon be added to other targets
* User cancellation for complete builds only.
* A "Build and Run" command that builds the current LF file and runs the executable in VS Code's built-in terminal.

It also corrects the following bugs, which were encountered while testing the above:
* Partial builds for C++ had surprising (non)behavior in that they require a complete build to have been done already
* URIs were constructed incorrectly in VS Code, causing failures on paths with escaped characters such as colons
* Cancellation of subprocesses was not working reliably.

It also includes a refactoring in how we handle the generator context, which was necessary in order to implement progress reporting cleanly.

These changes have been tested on Ubuntu and Windows. The VS Code extension has certain deficiencies in how it works with Windows and MSVC, but those issues existed before this PR and will be addressed separately.


lhstrh:
I hadn't heard of a "toast notification" before. For those wondering: https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=builder-syntax

lhstrh (approved):
This is a real leap forward in the handling of the build process and really impressive work! I've left some (mostly stylistic) comments.

org.lflang/src/org/lflang/generator/cpp/CppGenerator.kt
```diff
@@ -59,27 +56,32 @@ class CppGenerator(
...
         val codeMaps = generateFiles(fsa)
 
         if (targetConfig.noCompile || errorsOccurred()) {
             println("Exiting before invoking target compiler.")
-        } else if (cppFileConfig.compilerMode == Mode.LSP_MEDIUM) {
-            CppValidator(cppFileConfig, errorReporter, codeMaps).doValidate(context.cancelIndicator)
+            context.finish(GeneratorResult.GENERATED_NO_EXECUTABLE.apply(codeMaps))
+        } else if (context.mode == Mode.LSP_MEDIUM) {
+            context.reportProgress(
+                "Code generation complete. Validating...", IntegratedBuilder.GENERATED_PERCENT_PROGRESS
```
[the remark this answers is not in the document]
petervdonovan: I should have written, "Validating generated code." I will add comments to make it clearer what is going on here.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -805,54 +793,32 @@ public static Path findFile(String fileName, Path directory) {
...
+    /**
+     * Return the name of the RTI executable.
+     * @return
+     */
+    public String getRTIBinName() {
+        return nameWithoutExtension(srcFile) + RTI_BIN_SUFFIX;
+    }
+
+    public File getRTIBinFile() {
+        return this.binPath.resolve(getRTIBinName()).toFile();
```
lhstrh, line 806: These methods also still need some JavaDoc.

org.lflang/src/org/lflang/generator/LFGeneratorContext.java
```diff
@@ -0,0 +1,140 @@
...
+public interface LFGeneratorContext extends IGeneratorContext {
+
+    /**
+     * Return the mode of operation, which indicates how the compiler has been invoked 
+     * (e.g., from within Epoch, from the command line, or via a Language Server).
+     */
+    Mode getMode();
+
+    /**
+     * Returns any arguments that will override target properties.
```
lhstrh, line 28:
```suggestion
     * Return any arguments that will override target properties.
```

org.lflang/src/org/lflang/generator/MainContext.java
```diff
@@ -0,0 +1,127 @@
...
+import org.lflang.DefaultErrorReporter;
+import org.lflang.ErrorReporter;
+import org.lflang.FileConfig;
+import org.lflang.TargetConfig.Mode;
+import org.lflang.generator.IntegratedBuilder.ReportProgress;
+
+/**
+ * A {@code MainContext} is a context that is not nested in
+ * any other generator context. There is one {@code MainContext}
+ * for every build process.
```
lhstrh, line 17: It would be good to add a note here to describe what the context is used for...

org.lflang/src/org/lflang/generator/MainContext.java
```diff
@@ -0,0 +1,127 @@
...
+        this.cancelIndicator = cancelIndicator;
+        this.reportProgress = reportProgress;
+        this.args = args;
+        this.hierarchicalBin = hierarchicalBin;
+        this.constructErrorReporter = constructErrorReporter;
+        this.hasConstructedErrorReporter = false;
+    }
+
+    @Override
+    public CancelIndicator getCancelIndicator() {
```
lhstrh, line 83: JavaDoc missing in the remainder of this file.

org.lflang/src/org/lflang/generator/SubContext.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param endPercentProgress The percent progress of the containing build process when this
+     *                           nested process ends.
+     */
+    public SubContext(LFGeneratorContext containingContext, int startPercentProgress, int endPercentProgress) {
+        this.containingContext = containingContext;
+        this.startPercentProgress = startPercentProgress;
+        this.endPercentProgress = endPercentProgress;
+    }
+
+    @Override
```
lhstrh, line 37: Please add missing JavaDoc here, too.

org.lflang/src/org/lflang/generator/cpp/CppGenerator.kt
```diff
@@ -28,13 +28,10 @@ package org.lflang.generator.cpp
...
 import org.eclipse.emf.ecore.resource.Resource
 import org.eclipse.xtext.generator.IFileSystemAccess2
-import org.eclipse.xtext.generator.IGeneratorContext
 import org.lflang.*
 import org.lflang.TargetConfig.Mode
 import org.lflang.Target
-import org.lflang.generator.CodeMap
-import org.lflang.generator.GeneratorBase
-import org.lflang.generator.TargetTypes
+import org.lflang.generator.*
```
lhstrh, line 34: I think most coding guidelines suggest to have specific imports without wildcards.

lhstrh (commented):
Thanks for all the changes so far! I've added a few more notes for a final clean up. Finally, please add your author information to all the new files you created.

org.lflang/src/org/lflang/generator/cpp/CppGenerator.kt
petervdonovan: That makes sense. I'll configure IntelliJ so that it doesn't do that.

org.lflang/src/org/lflang/generator/MainContext.java
petervdonovan: That is true, more explanation was needed somewhere. In addition to adding emphasis to the fact that `MainContext` implements the `LFGeneratorContext` interface, I have added detail to the JavaDoc for `LFGeneratorContext`.

org.lflang/src/org/lflang/FileConfig.java
petervdonovan: Thanks for the catch -- I just pushed a commit that adds JavaDoc here. For some reason the diff that appears here on GitHub isn't updating, but the changes are definitely pushed.

org.lflang/src/org/lflang/generator/MainContext.java
petervdonovan: As indicated by the `@Override` annotation, the JavaDoc here would just be a duplicate of the documentation that appears for the interface. The interface fully specifies the API.

It's no problem to add it -- it's just copy-paste -- but I just want to double check for confirmation that that is really what we want. It would leave us with more documentation to maintain, and the two copies would have to be kept consistent with each other. (For what it's worth, the Google style guide [allows JavaDoc to be omitted](https://google.github.io/styleguide/javaguide.html#s7.3.4-javadoc-non-required) when `@Override` is present.)

PR #815, opened 2021-12-29
Refactor the containerization of the RTI

housengw:
Major changes:
- Put RTI inside the same docker-compose.yml as other federates. Previously, the containerized RTI had its own folder under src-gen and has its own docker-compose.yml. 
- Instead of generating a RTI folder under src-gen, the RTI image needs to be prebuilt.
- Use `docker compose` for nonfederated docker tests instead of `docker run`.


org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -1368,98 +1363,32 @@ class CGenerator extends GeneratorBase {
      * @param the content of the "services" section of the docker-compose.yml file.
      * @param the name of the federate to be added to "services".
      * @param the name of the federate's Dockerfile.
-     * @param the federationID.
      */
-    def appendFederateToDockerComposeServices(StringBuilder dockerComposeServices, String federateName, String dockerFileName, int dockerComposeFederationID) {
+    def appendFederateToDockerComposeServices(StringBuilder dockerComposeServices, String federateName, String context, String rtiName, String dockerFileName) {
```
Soroosh129, line 1367: I think this might be as good an opportunity as any to put these functions in a separate `CDockerGenerator.java` file.

org.lflang/src/org/lflang/generator/ts/TSDockerGenerator.kt
```diff
@@ -17,4 +17,19 @@ class TSDockerGenerator (
         """
         return dockerFileContent.trimMargin()
     }
+
+    fun generateDockerComposeFileContent(): String {
```
Soroosh129, line 21: Comment is missing for this function.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -181,11 +181,13 @@ class TSGenerator(
             fsa.generateFile(fileConfig.srcGenBasePath.relativize(tsFilePath).toString(),
                 tsCode.toString())
             
-            if (targetConfig.dockerOptions != null) {
+            if (targetConfig.dockerOptions != null && !isFederated) {
```
Soroosh129, line 184: If it is federated, I think an error message should be thrown that this is unsupported (if I understand the intention here correctly).

Soroosh129 (approved):
This looks great!

I have two minor comments:

- I think we can use this opportunity to move docker-related functions to a separate `CDockerGenerator.java` file (or is it better to do it in a separate PR?).
- Is there a way that we could put the `rti.Dockerfile` on Docker Hub to make the user experience smoother? (I assume putting it on the hub eliminates the extra step of building the RTI locally).

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -1368,98 +1363,32 @@ class CGenerator extends GeneratorBase {
...
-        // If a dockerfile exists, remove it.
-        var file = new File(dockerFile)
-        if (file.exists) {
-            file.delete
-        }
-        if (this.mainDef === null) {
-            return
+        dockerComposeServices.append('''«tab»«tab»command: -i 1«System.lineSeparator»''')
+        if (isFederated) {
+            dockerComposeServices.append('''«tab»«tab»depends_on: [«rtiName»]«System.lineSeparator»''')
```
Soroosh129, line 1375: Nice find! This is pretty cool.

housengw: One thing to note though, "depends_on" only makes sure the RTI container is "started" before the federate containers, so there is still a chance that the federates would start later, but try to connect to the RTI before it is ready..

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
housengw: you're correct

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
Soroosh129: I guess it's about time we fixed the underlying issue :-)

housengw:
seems like the github action tests are failing because the tests are using .yml files from master.

PR #816, opened 2021-12-29
Override enableSupportForSerialization in TSGenerator to fix an error in federated TS.

hokeun:
Override enableSupportForSerialization in TSGenerator to fix an issue of "Serialization is target-specific  and is not implemented for the TypeScript target." in federated TypeScript LF program.

java.lang.UnsupportedOperationException: Serialization is target-specific  and is not implemented for the TypeScript target.
        at org.lflang.generator.GeneratorBase.enableSupportForSerialization(GeneratorBase.java:974)
        at org.lflang.generator.GeneratorBase.doGenerate(GeneratorBase.java:465)
        at org.lflang.generator.ts.TSGenerator.doGenerate(TSGenerator.kt:109)
        at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:181)
        at org.lflang.lfc.Main.runGenerator(Main.java:301)
        at org.lflang.lfc.Main.main(Main.java:205)


org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -501,6 +502,22 @@ class TSGenerator(
...
+     * enable the requested serializations in 'enabledSerializations'
+     */
+    override fun enableSupportForSerialization(cancelIndicator: CancelIndicator?) {
+        for (serializer in enabledSerializers) {
+            when (serializer) {
+                SupportedSerializers.NATIVE -> {
+                    // No need to do anything at this point.
+                    println("Native serializer is enabled.")
+                }
+                else -> println("Serializer $serializer is enabled.")
```
lhstrh, line 516: I'm not 100% sure what this method is meant to do, but I think that only for serialization methods that are indeed supported it should print "is enabled" and throw a `UnsupportedOperationException` otherwise. Maybe @Soroosh129 can comment on this. It looks like the implementation in `PythonGenerator` is similarly flawed.

lhstrh (commented):
This fix gets rid of the error, but it looks like it only masks the problem...

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
hokeun: Right, this followed the implementation in `PythonGenerator`. By the way, TS federated execution fails without this fix, so I'd like to get this checked in to fix that...

Soroosh129: I agree with @lhstrh. It is better to throw an exception here if the `serializer` is not NATIVE. As best as I could remember, the solution in `PythonGenerator` was meant as a temporary measure because adding ROS 2 serialization was relatively straightforward. I haven't gotten around to that yet though.

hokeun: Thanks, Marten and Soroush. I made this thrown an exception in https://github.com/lf-lang/lingua-franca/pull/816/commits/1dc1ddb6b057a22be967be9f8c27f828bbb8e95c. Hope this makes more sense.

lhstrh (approved):
:+1:

PR #817, opened 2021-12-30
Refactor CI configuration with reusable workflows

lhstrh:
This PR refactors our Github Actions configuration. The existing `build.yml` had become too large, so I split workflows into separate files and did some tweaks to promote code reuse. Having done this, satellite repositories like `reactor-c` and `reactor-ts` can easily invoke integration tests without having to copy any code. I've also augmented these reusable workflows with a parameter to specify an alternative runtime, which is particularly useful for CI on said satellite repositories.


.github/actions/checkout-ref/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Checkout a particular ref
```
petervdonovan, line 1: [actions/checkout@v2](https://github.com/actions/checkout) does appear to provide the option to check out nested repositories. Was that considered as an alternative to this and the `fetch-depth: 0` workaround?

.github/workflows/benchmark-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+        default: Cpp
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Java JDK
+        uses: actions/setup-java@v1.4.3
+        with:
+          java-version: 14
```
petervdonovan, line 18: Do we use Java 11 or 14? This inconsistency already existed in `master`, but I wonder why we have it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - 'nightly'
+
+env:
+  # 2020.11
+  vcpkgGitRef: 0bf3923f9fab4001c00f0f429682a0853b5749e0
+
+jobs:
+  # Run the unit tests.
+  unit-tests:
+    uses: lf-lang/lingua-franca/.github/workflows/unit-tests.yml@modular-ci
```
petervdonovan, line 21: I assume we will have to delete the `@modular-ci` part after this is merged(?)

petervdonovan (approved):

.github/actions/checkout-ref/action.yml
lhstrh: It does check them out, but what we want is specify _which ref_ to check out, independent from what the configuration of the main repository says; `actions/checkout@v2` does not provide this functionality...

.github/workflows/benchmark-tests.yml
lhstrh: I wondered about this, too. Maybe @cmnrd can comment on this?

.github/workflows/ci.yml
lhstrh: We could just change it to `@master` if we want to remove the `modular-ci` branch.

.github/actions/checkout-ref/action.yml
petervdonovan: I did not mean the `submodules: true` option -- I meant like this:
```
- name: Checkout
  uses: actions/checkout@v2

- name: Checkout tools repo
  uses: actions/checkout@v2
  with:
    repository: my-org/my-tools
    path: my-tools
```

lhstrh: But does it accept a `ref`?

lhstrh: I suppose we could _not_ automatically check out submodules but do it manually and provide a ref if one was specified... That might work.

lhstrh:
I'm a bit puzzled by the federated tests for Python on Mac... Not sure what's going on there.

.github/actions/checkout-ref/action.yml
petervdonovan: I had thought that, but since we already have something that works, I guess we don't need to worry about it.

lhstrh: It would amount to a simpler solution though, so I might give that a shot.

lhstrh: OK, I found a way to make it work with `actions/checkout@v2`. I'll remove the custom action.

.github/workflows/benchmark-tests.yml
lhstrh: Changed it to 11, and it appears to work fine.

PR #819, opened 2022-01-03
Add digital twin demo

housengw:
Add a digital twin demo where two federates, one running locally and the other running on the cloud, are federated through an RTI running on the cloud. 

- [x] add bash script to simplify demo
- [x] allow user to override hostname, port and username of rti as a flag to `lfc`
- [x] specify RTI ip in runtime instead of compile time #842 

I'll add changes to specify RTI ip in runtime instead of compile time in another pull request since this PR is getting a bit large.


example/Python/src/DigitalTwin/KeyFob/KeyFobDemo.lf
```diff
@@ -0,0 +1,113 @@
+target Python {
```
Soroosh129, line 1: A short description here at the top would be helpful I think with a pointer to the README file.

example/Python/src/DigitalTwin/KeyFob/KeyFobDemo.lf
```diff
@@ -0,0 +1,113 @@
...
+    docker: true
+};
+
+preamble {=
+    import curses
+    import threading
+=}
+
+reactor KeyFob {
```
Soroosh129, line 10: Comment missing for this reactor.

example/Python/src/DigitalTwin/KeyFob/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Make sure you have a Google Cloud Platform (GCP) account and a project set up. 
+
+To set up an account, see here:
+https://console.cloud.google.com
+
+To set up a GCP project, see here:
+https://cloud.google.com/resource-manager/docs/creating-managing-projects
+
+We will be working with the Google Cloud SDK. Mac users can download it using brew.
```
Soroosh129, line 15:
```suggestion
We will be working with the Google Cloud SDK. For installation instructions, see here:
https://cloud.google.com/sdk/docs/quickstart#installing_the_latest_version

Mac users can download the SDK using brew:
```

example/Python/src/DigitalTwin/KeyFob/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Clean up
+
+Run the clean up script:
+```bash
+user$ ./cleanup.sh
+```
+
+### Conclusion
+
+Congrats! You just federated a LF program using an RTI hosted on the cloud.
```
Soroosh129, line 86:
```suggestion
Congrats! You just executed a federated LF program using an RTI hosted on the cloud.
```

org.lflang.lfc/src/org/lflang/lfc/Main.java
```diff
@@ -111,7 +111,10 @@
         THREADS("t", "threads", true, false, "Specify the default number of threads.", true),
         OUTPUT_PATH("o", "output-path", true, false, "Specify the root output directory.", false),
         RUNTIME_VERSION(null, "runtime-version", true, false, "Specify the version of the runtime library used for compiling LF programs.", true),
-        EXTERNAL_RUNTIME_PATH(null, "external-runtime-path", true, false, "Specify an external runtime library to be used by the compiled binary.", true);
+        EXTERNAL_RUNTIME_PATH(null, "external-runtime-path", true, false, "Specify an external runtime library to be used by the compiled binary.", true),
+        RTI_HOST(null, "host", true, false, "Specify the hostname or IP address of the RTI.", true),
```
Soroosh129, line 115: I think using a single argument called `--rti` is more descriptive than having separate `host`, `port`, and `user` arguments without a context of what they are for.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
```diff
@@ -1564,29 +1565,37 @@ abstract class GeneratorBase extends AbstractLFValidator implements TargetTypes
             if (mainReactor.host !== null) {
```
Soroosh129, line 1565: Does this if statement interfere with the intended object of the added code? It seems like if no `at` clause is present on the main/federated reactor, the following code will not execute.

Soroosh129 (changes requested):
This demo looks great. Thanks for taking on this project. It is an impressive feat!

I have a few comments:
- I think since @edwardalee had requested a few action items, it would be great if he could also review this PR before merging in
- Using `--rti user@host:port` instead of `--host`, `--user`, and `--port` might be more descriptive
- Would you please add tests for these new cli arguments to [here](https://github.com/lf-lang/lingua-franca/blob/master/.github/scripts/test-lfc.sh).

housengw:
I prefer to write a separate demo that imports and reuses the code in the key fob demo to address @edwardalee's feedback. The new demo will have more states.

Soroosh129:
> I prefer to write a separate demo that imports and reuses the code in the key fob demo to address @edwardalee's feedback. The new demo will have more states.

Okay, sounds good to me.

edwardalee (approved):
This looks great to me!  Excellent README files. I agree with @Soroosh129 that the .lf files need pointers to the README files.

org.lflang/src/org/lflang/generator/GeneratorBase.xtend
housengw: ah, you're right. good catch.

org.lflang.lfc/src/org/lflang/lfc/Main.java
housengw: I'll add prefix `rti` (ex. `--rti_host`) for now.

Soroosh129: I don't personally think that `--rti_host` looks intuitive enough. I really think that we should aspire to come up with an elegant permanent solution rather than coming up with temporary fixes. If you don't want to tackle this, I can take the implementation of the `--rti` cli argument as an action item.

housengw:
not sure why code coverage dropped by so much.

lhstrh:
It didn't, but #845 fixes error marker issue

Soroosh129 (approved):
Looks good to me!

PR #820, opened 2022-01-03
build rti image in github actions

housengw:
Reason for this pull request:

GitHub action uses .yml files from the master branch to run tests, and #815 requires building a docker image for the RTI to pass tests.


.github/workflows/c-tests.yml
```diff
@@ -50,6 +50,11 @@ jobs:
       - name: Install RTI
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
+      - name: Build RTI docker image;
```
lhstrh, line 53: I think it would be better to put this action in `./github/actions/install-rti` (or to create a separate action called `./github/actions/rti-docker`).

lhstrh (commented):
Thanks for adapting to the new Github Actions config! I've left a suggestion for reducing code replication.

Soroosh129:
Would this still be needed if we put a Dockerfile for the RTI on Docker Hub?

housengw:
I think so. This would make sure that the RTI image is functional before putting it on Docker hub

.github/actions/build-rti-docker/action.yml
```diff
@@ -0,0 +1,7 @@
+name: Build RTI docker image
+description: Build and install
+runs:
+  using: "composite"
+  steps: 
+    - run: .github/actions/install-rti/build-rti-image.sh
+      shell: bash
```
lhstrh, line 7:
```suggestion
      shell: bash
      
```

.github/actions/install-rti/action.yml
```diff
@@ -4,4 +4,4 @@ runs:
   using: "composite"
   steps: 
     - run: .github/actions/install-rti/install.sh
-      shell: bash
+      shell: bash
```
lhstrh, line 7:
```suggestion
      shell: bash
      
```

lhstrh (approved):
Looks good! Just added some new lines.

PR #823, opened 2022-01-04
fix build rti path

housengw:
the path to the .sh in `.github/actions/build-rti-docker` is wrong, which caused tests to fail in master. 
This PR corrects the path.

### DIFF
--- a/.github/actions/build-rti-docker/action.yml
+++ b/.github/actions/build-rti-docker/action.yml
@@ -5,4 +5,3 @@ runs:
   steps: 
     - run: .github/actions/build-rti-docker/build-rti-image.sh
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/build-rti-docker/action.yml
```diff
@@ -3,6 +3,6 @@ description: Build and install
 runs:
   using: "composite"
   steps: 
-    - run: .github/actions/install-rti/build-rti-image.sh
+    - run: .github/actions/build-rti-docker/build-rti-image.sh
       shell: bash
```
lhstrh, line 8:
```suggestion

```

lhstrh (approved):

PR #830, opened 2022-01-09
CodeCov

petervdonovan:
Because [CodeCov automatically merges multiple reports from the same CI run](https://docs.codecov.com/docs/merging-reports), I do not think that we need the `CodeGenCoverage` class.

(In fact, our method of reporting coverage was apparently causing artificially low code coverage statistics to be reported. This is because [`unit-tests.yml`](https://github.com/lf-lang/lingua-franca/blob/45fff4da288b8f3ed88021a44f65f25f63c821ac/.github/workflows/unit-tests.yml#L17) does not fetch submodules in `master`, leading to `IOException`s that aborted code generation. These exceptions went unnoticed because they were swallowed by a [universal `catch`](https://github.com/lf-lang/lingua-franca/blob/4d76a85ca03d6f8bc6ec66ad9cea4b770fa3c07d/org.lflang.tests/src/org/lflang/tests/TestBase.java#L723) clause. Then we [skipped checking](https://github.com/lf-lang/lingua-franca/blob/45fff4da288b8f3ed88021a44f65f25f63c821ac/org.lflang.tests/src/org/lflang/tests/TestBase.java#L191) that the test passed.)


lhstrh (approved):
Thanks for these simplifications! This looks great, and the 75% that is not reported definitely sounds more like it :-)

PR #841, opened 2022-01-12
Fix Compilation Error in Diagrams Package due to Missing Dependency

### LINKED ISSUE #840: Error in Eclipse
After a clean Oomph install of Eclipse, I get the following error:

`The type org.eclipse.sprotty.xtext.ls.SyncDiagramLanguageServer cannot be resolved. It is indirectly referenced from required .class files`

Seems like a missing dependency, but I'm not sure how to add it.

a-sr:
Added dependency indirectly referenced from required class file

Fixes #840

### DIFF
--- a/org.lflang.diagram/META-INF/MANIFEST.MF
+++ b/org.lflang.diagram/META-INF/MANIFEST.MF
@@ -21,5 +21,6 @@ Require-Bundle: de.cau.cs.kieler.klighd;bundle-version="2.0.0",
  org.lflang.ide;bundle-version="0.1.0",
  org.eclipse.xtext.ide;bundle-version="2.25.0",
  org.eclipse.lsp4j;bundle-version="0.10.0",
- org.eclipse.lsp4j.jsonrpc;bundle-version="0.10.0"
+ org.eclipse.lsp4j.jsonrpc;bundle-version="0.10.0",
+ org.eclipse.sprotty;bundle-version="0.9.0"
 Bundle-Vendor: University of California, Berkeley


lhstrh:
Independent of this fix, we run into this problem again where the URL `https://jcenter.bintray.com/com/google/guava/guava/maven-metadata.xml` hit and causing a "bad gateway" error. This appears to be a Gradle problem, but that, too, I don't know how to fix. I grepped the entire code base and it doesn't have the substring `bintray` in it...

As to your fix to `MANIFEST.MF`, how did you determine which line to add? Specifically, how do you know that `org.eclipse.sprotty;bundle-version="0.9.0"` exists in P2? Can it be queried somewhere? I've been finding it very difficult to get good information about this.

lhstrh:
Looks like the bintray problem has gone away by itself...

lhstrh (approved):

a-sr:
> As to your fix to `MANIFEST.MF`, how did you determine which line to add? Specifically, how do you know that `org.eclipse.sprotty;bundle-version="0.9.0"` exists in P2? Can it be queried somewhere? I've been finding it very difficult to get good information about this.

I knew because I added the p2 update site to the target platform ([9c31069](https://github.com/lf-lang/lingua-franca/commit/9c31069d833b3fcb964de3f1c0d398ebaeb1ff71)). The reason for the error was me not setting up the new dependencies properly.
In general, you can open the "Target Platform State" view in Eclipse to inspect all plugins available in the current configuration of your target platform.

PR #842, opened 2022-01-13
use runtime argument in digital twin demo

housengw:
- [x] use `--rti` instead of `--host`, `--port`, `--user` in run time arguments


example/Python/src/DigitalTwin/KeyFob/README.md
```diff
@@ -67,9 +67,9 @@ user@twin-vm ~ $ docker container attach CONTAINER_ID
 
 ### Running the local key fob
 
-Open another terminal in the directory where the `docker-compose.yml` is located. Run:
+Open another terminal in the directory where the `docker-compose.yml` is located. You might have to export `$RTI_IP` again. Run:
```
Soroosh129, line 70: I think this new sentence can be confusing because it's not clear what "again" here is referring to. I can see that `setup.sh` is supposed to export `RTI_IP`, so I'm assuming this sentence is a reference to that. But I think the script is just setting a local `RTI_IP` variable, not an environment variable.

It would also be helpful here I think if `-i` and `--host` are explained here as well.

Soroosh129 (changes requested):
Looks good to me!

I left a suggestion on the [accompanying](https://github.com/lf-lang/reactor-c/pull/30) PR on reactor-c that could result in minor changes.

example/Python/src/DigitalTwin/KeyFob/README.md
housengw: the user might have to export `RTI_IP` again because they opened a new terminal.

Soroosh129: Right, but exporting the `RTI_IP` environment variable is not mentioned earlier at all. This is the first time that it is mentioned as far as I can see, or am I missing a step?

housengw: oh wait, `RTI_IP` is exported in `setup.sh`. I'll specify that.

housengw: I added a script that the user can just run to both re-export `RTI_IP` and start the local copy

Issue #838, opened 2022-01-11
Port Xtend classes to Java


lhstrh:
We are well underway in our effort to port Xtend classes to Java. The following classes still need to be ported before we can cut our dependence on Xtend:

## `org.lflang.diagram` package (@housengw) 
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/postprocessor/ReactionPortAdjustment.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/CycleVisualization.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/ReactorIcons.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/NamedInstanceUtil.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/UtilityExtensions.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/InterfaceDependenciesVisualization.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/util/SynthesisErrorReporter.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/SynthesisRegistration.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaStyleExtensions.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/AbstractSynthesisExtensions.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/action/MemorizingExpandCollapseAction.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/action/AbstractAction.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/action/FilterCycleAction.xtend`
- [x]  `org.lflang.diagram/src/org/lflang/diagram/synthesis/action/ShowCycleAction.xtend`
[...]

PR #844, opened 2022-01-14
Ported `org.lflang.tests` from Xtend to Java

lhstrh:
See https://github.com/lf-lang/lingua-franca/issues/838.


petervdonovan (approved):
I see nothing amiss here. (The change in `Progress.java` breaks the VS Code extension, of course, but it is necessary, and it will be addressed separately (#846).)

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaASTUtilsTest.java
```diff
@@ -0,0 +1,299 @@
...
+         */
+        
+        model.eAllContents().forEachRemaining((obj) -> {
+            if (obj instanceof Parameter) {
+                Parameter parameter = (Parameter)obj;
+                if (parameter.getName() == "x") {
+                    var values = ASTUtils.initialValue(parameter, null);
+                    Assertions.assertEquals(values.get(0).getLiteral(), "1");
+                    
+                    values = ASTUtils.initialValue(parameter, list_a1);
```
oowekyala, line 263:
```suggestion
                    values = ASTUtils.initialValue(parameter, List.of(a1));
```
I would suggest removing all those list variables and replacing them with `List.of` like this. Alternatively you could also use `List.of` in the definition of those variables, which would save lines compared to LinkedList. I don't think using variables here nets any readability though

oowekyala (commented):
Given the absence of text blocks these tests are surprinsingley readable! Cheers

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaASTUtilsTest.java
lhstrh: Thanks for the suggestion. Will do!

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaASTUtilsTest.java
```diff
@@ -0,0 +1,295 @@
...
+                    a2 = inst;
+                } else if (inst.getName().equals("b1")) {
+                    b1 = inst;
+                } else if (inst.getName().equals("b2")) {
+                    b2 = inst;
+                }
+            }
+        }
+        
+        return Map.of("a1", a1, "a2", a2, "b1", b1, "b2", b2);
```
oowekyala, line 180:
```suggestion
       return asStream(model.eAllContents())
                .filter(obj -> obj instanceof Instantiation)
                .map(obj -> (Instantiation) obj)
                .collect(Collectors.toMap(Instantiation::getName, it -> it));
```

PR #848, opened 2022-01-14
Compile button now creates a separate resource for each compilation.

### LINKED ISSUE #746: Epoch compile button does not recreate the AST
It turns out that with the new Epoch button to compile the program, the button does not cause a new AST to get created.  So code gets generated on the basis of the previous AST that was present, including any AST transformations that were previously applied.

This is a serious problem for federated execution, but could be a problem for non-federated programs using `after`. One way to see the effect is to compile a program that has an `after` in it. Then close the diagram view and re-open the diagram view using the `Window->Show View` menu. You will see the transformed program instead of the original one.

For federated programs, this results in code being generated that has no communication between federates.

Not sure how to solve this problem...

a-sr:
This fixes #746 as the modifications to ecore model by the code generator no longer leak into the editor's model, the diagram, or future compilation.


edwardalee (approved):
Looks good to me. Thanks!

PR #850, opened 2022-01-15
Make the TypeScript target work on Windows

petervdonovan:
This ensures that TypeScript integration tests pass on Windows.


org.lflang/src/org/lflang/util/LFCommand.java
```diff
@@ -318,23 +316,52 @@ public static LFCommand get(final String cmd, final List<String> args, Path dir)
...
+                .map(String::strip).map(File::new).filter(File::canExecute).collect(Collectors.toList());
         } catch (InterruptedException | IOException e) {
             e.printStackTrace();
-            return false;
+            return null;
+        }
+    }
+
+    /**
+     * Attempt to start the execution of this command.
```
lhstrh, line 339:
```suggestion
     * Attempt to start the execution of this command.
     * 
     * First collect a list of paths where the executable might be found,
     * then select an executable that successfully executes from the 
     * list of paths. Return the {@code Process} instance that is the 
     * result of a successful execution, or {@code null} if no successful
     * execution happened.
```

lhstrh (approved):
These changes look great! I only provided a suggestion for a more elaborate comment.

PR #854, opened 2022-01-16
Epoch: Repair regex.

petervdonovan:
This corrects a mistake that I inserted into the C generator three months ago (in [this](https://github.com/lf-lang/lingua-franca/commit/7b563035c38db648a8e09bf6851196511c4ba9be) commit).


cmnrd (approved):
Looks good! Do you think it would be possible to unit test the error parsing somehow?

petervdonovan:
Yes, it is possible. In the `vscode-validate-generated-code` branch, this functionality is tested as part of integration tests for all targets except C. When I next touch this aspect of the C target, I will make sure that we have something to guard against regressions.

To be expedient I think maybe it's okay to merge this first, since Epoch is broken right now.

PR #856, opened 2022-01-18
Always run CI workflows from the current branch

cmnrd:
Currently the `ci.yml` workflow loads other sub-workflows, but hard-codes the workflow version to master. As a consequence, we always need to adjust `ci.yml` whenever we want to make changes to the sub-workflows in some branch. This PR is an attempt to avoid this hurdle, by using variable to automatically insert the correct branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,44 +18,44 @@ env:
 jobs:
   # Run the unit tests.
   unit-tests:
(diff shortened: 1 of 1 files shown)


lhstrh:
I tried similar things, to no avail. I agree that it is extremely annoying... Maybe worth providing feedback to GitHub about this?

Issue #859, opened 2022-01-19
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input


Soroosh129:
I get the following exception on `master` on commit b7d6115d636e315710c2717220f215a7e8362ff3 when compiling any of the Python tests:

```bash
$ lfc test/Python/srcMinimal.lf 
Generating code for: file:/home/soroush/lingua-franca-master/git/lingua-franca/test/Python/src/Minimal.lf
******** Mode: STANDALONE
******** Generated sources: /home/soroush/lingua-franca-master/git/lingua-franca/test/Python/src-gen/Minimal
******** Using 1 threads.
Generating setup file to /home/soroush/lingua-franca-master/git/lingua-franca/test/Python/src-gen/Minimal/setup.py
--- Current working directory: /home/soroush/lingua-franca-master/git/lingua-franca/test/Python/src-gen/Minimal
--- Executing command: pylint --output-format=json Minimal.py
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (String)""; line: 1, column: 0]
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4688)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4586)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3548)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3516)
	at org.lflang.generator.python.PythonValidator$2.lambda$getOutputReportingStrategy$2(PythonValidator.java:249)
	at org.lflang.generator.Validator.doValidate(Validator.java:64)
	at org.lflang.generator.python.PythonGenerator.doGenerate(PythonGenerator.java:1809)
	at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:163)
	at org.eclipse.xtext.generator.GeneratorDelegate.doGenerate(GeneratorDelegate.java:44)
	at org.eclipse.xtext.generator.GeneratorDelegate.generate(GeneratorDelegate.java:35)
	at org.lflang.lfc.Main.runGenerator(Main.java:277)
	at org.lflang.lfc.Main.main(Main.java:206)
[...]

PR #860, opened 2022-01-19
vscode-validate-generated-code cleanups

petervdonovan:
Yesterday, I merged a branch called `vscode-validate-generated-code` that included many changes. There were some things that I forgot to check first. This caused Eclipse errors, broke the Maven build, and caused #859. Fixes for the first two of these problems have already been merged. However, #859 still stands, and there are a couple other odds and ends that are cleaned up in this branch as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
 
   # Run language server tests.
   lsp-tests:
-    uses: lf-lang/lingua-franca/.github/workflows/lsp-tests.yml@master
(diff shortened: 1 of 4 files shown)


lhstrh:
The main problem here is that our test suite currently doesn't catch these problems. I created an issue (#861) proposing a fix that would prevent our nightly build from breaking unexpectedly, but for the Eclipse build and the other issue we encountered it's not clear to me how we can improve things.

lhstrh (approved):
Looks good. Shall we merge this?

PR #867, opened 2022-01-20
Do not break Eclipse build

petervdonovan:
This allows the Maven build and Eclipse build to work at the same time. (I have only tested the Maven build, but the Eclipse build "should" be fine.) It breaks the language server, but that is fine.


lhstrh (approved):
Looks good to me. Which ref shall we use for `vscode-lingua-franca`?

petervdonovan:
> Which ref shall we use for vscode-lingua-franca?

My best guess is d16cb04, unless I find out that there are even more problems with the recent merge.

PR #871, opened 2022-01-21
Clean up digital twin demo

housengw:
- [x] add a version that has 3 states (locked, drivers door unlocked, all doors unlocked)
- [x] factor out logger and curses window into a utils file.
- [x] include youtube link to a recorded demo in the readme file.


example/Python/src/DigitalTwin/KeyFob/dev/README.md
```diff
@@ -28,30 +28,37 @@ NAME    ZONE           MACHINE_TYPE   PREEMPTIBLE  INTERNAL_IP  EXTERNAL_IP
 rti-vm  us-central1-a  n1-standard-1               10.128.0.7   34.133.143.163  RUNNING
 ```
 
+Export the `EXTERNAL_IP` of the RTI:
+```bash
+RTI_IP=`gcloud compute instances list | grep 'rti-vm' | awk '{print $5}'`
```
lhstrh, line 33: Is there an `export` missing here?

lhstrh (approved):

example/Python/src/DigitalTwin/KeyFob/dev/README.md
housengw: assignment `=` sets a variable local to the current bash process, so `export` is optional

PR #876, opened 2022-01-22
Xtend to java validation test

housengw:


housengw:
I didn't port the following functions due to an unknown `it` variable:
1. `private List<String> synthesizeExamples(UnionType type, boolean correct)`
2. `private List<String> synthesizeExamples(DictionaryType type, boolean correct)`
3. `public void checkTargetProperties()`
4. `public void checkCargoDependencyProperty()`

For example, in `public void checkCargoDependencyProperty()`, the function does not take any arguments but there is an `it` variable in the xtend code:

```
public void checkCargoDependencyProperty() {
         val prop = TargetProperty.CARGO_DEPENDENCIES;
         val knownCorrect = #[ "{}", "{ dep: \"8.2\" }", "{ dep: { version: \"8.2\"} }", "{ dep: { version: \"8.2\", features: [\"foo\"]} }" ]
         knownCorrect.forEach [
            prop.createModel(  >>>   it   <<<   ).assertNoErrors()
        ]
```

PR #879, opened 2022-01-24
change lsp ci yaml to master

housengw:
The ci-yaml in master points to a nonexistent branch for lsp tests.


petervdonovan:
Thanks for fixing this! I am lucky that you took care of it so quickly.

PR #882, opened 2022-01-25
Make linting optional in standalone mode

petervdonovan:
This change disables validation of generated code when in standalone mode but allows the user to request validation explicitly using the `lint` compiler option. For example,
```
bin/lfc --lint test/TypeScript/src/SimpleDeadline.lf
```
may fail with the error
```
lfc: error: Unexpected var, use let or const instead. [SimpleDeadline.ts:167:25]
```
whereas the file might compile without problems without the `--lint` flag.

This affects the **Python** and **TypeScript** targets. For the compiled languages, any errors from the compilation process should still be reported.
* All forms of syntax checking are disabled by default for Python in standalone mode.
* Type checking using `tsc` is not disabled. However, linting using `eslint` is disabled by default in standalone mode.

Incidentally, this may slightly speed up CI, but the validators will still be tested by LSP tests.

### DIFF
--- a/.github/scripts/test-lfc.sh
+++ b/.github/scripts/test-lfc.sh
@@ -47,6 +47,10 @@ bin/lfc --federated --rti rti test/C/src/Minimal.lf
 # -h,--help                          Display this information.
 bin/lfc --help
 
+# -l, --lint                         Enable linting during build.
+bin/lfc -l test/Python/src/Minimal.lf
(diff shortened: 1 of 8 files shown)


Soroosh129 (approved):
Looks good to me!

If possible, please add tests for the new lfc option to .github/scripts/test-lfc.sh.

cmnrd (approved):
Looks good! And I like the idea :)

Issue #883, opened 2022-01-25
Revise project refresh (@a-sr's comment on #875)


lhstrh:
Extracting some location pattern from the generated code seems quite complicated.
If it is sufficient to update only the project the source LF file is located in, assuming that the generated files will be in the same project, it can be done via the Resource that is available in the generator.
~~~
URI uri = resource.getURI();
if (uri.isPlatformResource()) { // This condition should normally be met when running Epoch
    IResource member = ResourcesPlugin.getWorkspace().getRoot().findMember(uri.toPlatformString(true));
    member.getProject().refreshLocal(IResource.DEPTH_INFINITE, null);
}
~~~

_Originally posted by @a-sr in https://github.com/lf-lang/lingua-franca/pull/875#discussion_r790913040_

PR #884, opened 2022-01-25
Per #883, simplify refreshing the project after generating code

edwardalee:
This implements the suggestion from @a-sr to refresh the resource in Epoch so that, after generating code, the generated code files become visible in the package browser. This seems to work on Mac, and although I don't have any way to test on other platforms, I see no reason why it wouldn't work there.


lhstrh (approved):
Looks good to me!

Issue #340, opened 2021-05-19
Reactors can't have names starting with numbers and the error message is cryptic


edwardalee:
If I create a file named 01_Test.lf with:
```
target C;
main reactor {
   reaction(startup) {=
      printf("Hello World.\n");
   =}
}
```
the result is cryptic compile error.  If I change it to
```
target C;
main reactor 01_Test {
   reaction(startup) {=
      printf("Hello World.\n");
   =}
}
```
then I get a validator error, "mismatched input '01' expecting '{'".
So I guess we don't allow reactor names to start with numbers, but we don't check filenames.

PR #890, opened 2022-01-26
validator: Report an error if an LF file name starts with a number

cmnrd:
This is a quick fix for #340.

This change conflicts with #886, but I think the change is small enough to merge it now and backport to #886.


housengw (approved):
lgtm. I'll make the corresponding change in #886

lhstrh (approved):
Thanks for the fix!

PR #892, opened 2022-01-26
Code emitter

edwardalee:
Following @cmnrd 's suggestion, I have factored out pr() and related methods for printing indented code into a new Java class, CodeBuilder, and removed them from the generator base classes.


lhstrh (commented):
This looks great! I think it includes changes from another PR, however, so let's get that through review first.

petervdonovan (approved):
All of this looks good to me!

lhstrh (approved):

Issue #855, opened 2022-01-16
NORMAL validation is no longer invoked on saving the file


edwardalee:
It used to be that when saving a file in Epoch, any validation function labeled NORMAL would be invoked. This is no longer the case (in master). LFValidator has this code:
```
    @Check(FAST)
    def checkModel(Model model) {
        // Since we're doing a fast check, we only want to update
        // if the model info hasn't been initialized yet. If it has,
        // we use the old information and update it during a normal
        // check (see below).
        if (!info.updated) {
            info.update(model, errorReporter)
        }
    }

    @Check(NORMAL)
    def updateModelInfo(Model model) {
        info.update(model, errorReporter)
    }
```
You can set a breakpoint in the second function and verify that it is not invoked on file save. It is invoked when you invoke the code generator using the button that @a-sr added.

With the current behavior, the above functions are incorrect.  The FAST checkModel function will not update the model when you make a change, making it fairly useless.

a-sr:
It seems the Xtext validation that runs when saving, is implemented as an automatic build job. Hence, if "Project -> Build automatically" is off, it is no longer triggered.
Since we want to have this option disabled, to keep Epoch usable with many models, we should only use fast validation rules.
I think in case of this specific validation rule, the handling of the ModelInfo may be better located in a model specific resource implementation. There it could hook into parsing and saving events directly.
As always with Xtext, it is a bit tricky to get it working. I could give it a try and create a draft for the redesign, if that is ok for you.

lhstrh:
What's the status of this?

PR #894, opened 2022-01-26
Do not swallow errors in imported files. Report them.

### LINKED ISSUE #446: Cycle detection in validator does not work if involved reactors are imported
Consider this minimal example:
* `Bar.lf`:
```
target C;

import Foo from "Foo.lf"
import Baz from "Baz.lf"

main reactor {
    foo = new Foo();
    baz = new Baz();
    foo.y -> baz.x;
    baz.y -> foo.x;
}
```
* `Foo.lf`:
```
target C;

reactor Foo {
    input x:int;
    input z:int;
    output y:int;
    
    reaction (x) {=
    =}
    
    reaction (z) -> y {=
    =}
}
```
* `Baz.lf`:
```
target C;

reactor Baz {
    input x:int;
    output y:int;
	
    reaction (x) -> y {= 
    =}
}

```
The cycle is not caught during validation. Instead (for the C target), the cycle is detected in the code generator when levels are assigned.

If only one of the reactors is imported (and the other defined in the same file as the main reactor), the problem is not present.

edwardalee:
When cycles involve imported files, as in issue #446, the error reporting was funky. All it tells you is that there are "unresolved compilation problems" in the imported file, but nowhere does it tell you what the problems are.  This small PR fixes that and closes #446.

There is still the bug in issue #855, which is that the validator doesn't get fully run upon saving the file, only when you invoke the code generator, so cycles are not reported at all until you code generate (they are, however, displayed in red in the diagram).

### DIFF
--- a/org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
+++ b/org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
@@ -21,6 +21,7 @@ import org.eclipse.xtext.generator.IGeneratorContext;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.validation.CheckMode;
 import org.eclipse.xtext.validation.IResourceValidator;
+import org.eclipse.xtext.validation.Issue;
 import org.lflang.ErrorReporter;
 import org.lflang.FileConfig;
 import org.lflang.Target;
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
🚀🚀🚀

PR #895, opened 2022-01-27
port `InstantiationGraph.xtend` to java

hokeun:


org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,179 @@
...
+ * `a = new A();`. This can result in many instances of reactor
+ * class A (if the containing reactor class is instantiated multiple times).
+ *
+ * In addition to the graph, this class keeps track of the instantiations
+ * that induce the dependencies. These can be retrieved using the method
+ * `getInstantiations(Reactor)`.
+ *
+ * @author{Marten Lohstroh <marten@berkeley.edu>}
+ */
+@SuppressWarnings("all")
```
lhstrh, line 62: Why do we need this annotation?

hokeun: I don't think this is necessary and it seems to be part of the auto-generated code. I just removed it.

hokeun: Thanks for pointing this out!

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+ * `getInstantiations(Reactor)`.
+ *
+ * @author{Marten Lohstroh <marten@berkeley.edu>}
+ */
+public class InstantiationGraph extends PrecedenceGraph<Reactor> {
+    /**
+     * A mapping from reactors to the sites of their instantiation.
+     */
+    protected final HashMultimap<Reactor, Instantiation> reactorToInstantiation =
+        HashMultimap.<Reactor, Instantiation>create();
```
petervdonovan, line 67: The `<Reactor, Instantiation>` part can be left to type inference -- should we remove it? The same comment applies to many other calls to generic methods throughout.

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * A mapping from reactor classes to their declarations.
+     */
+    protected final HashMultimap<Reactor, ReactorDecl> reactorToDecl =
+        HashMultimap.<Reactor, ReactorDecl>create();
+
+    protected final HashSet<Object> flaggedReactors = CollectionLiterals.<Object>newHashSet();
```
petervdonovan, line 75: This looks to be unused. Should we remove it?

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.eclipse.emf.common.util.EList;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.xtext.xbase.lib.CollectionLiterals;
+import org.eclipse.xtext.xbase.lib.Functions.Function1;
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
+import org.eclipse.xtext.xbase.lib.IteratorExtensions;
```
petervdonovan, line 38: At least some of the methods used from these imports have non-Xtext equivalents. In porting from Xtend, should we also be trying to edge Xtext out as a dependency? I am genuinely unsure about the right policy.

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public InstantiationGraph(final Resource resource, final boolean detectCycles) {
+        final Iterable<Instantiation> instantiations = Iterables.<Instantiation>filter(
+            IteratorExtensions.<EObject>toIterable(resource.getAllContents()), Instantiation.class);
+        final Reactor main = IterableExtensions.<Reactor>findFirst(
+            Iterables.<Reactor>filter(
+                IteratorExtensions.<EObject>toIterable(resource.getAllContents()), Reactor.class),
+            (Reactor it) -> {
+                return Boolean.valueOf((it.isMain() || it.isFederated()));
+            });
```
petervdonovan, line 123: I suspect that the most idiomatic way to do this whole operation (from line 116 down to here) would be to use Java's Streams API. But if we do keep using these XText library functions, then this might be more concise.
```suggestion
                (Reactor it) -> it.isMain() || it.isFederated()
            );
```

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    /**
+     * Traverse the AST and build this precedence graph relating the
+     * encountered instantiations. Also map each reactor to all
+     * declarations associated with it and each reactor to the sites of
+     * its instantiations.
+     * @param instantiation
+     * @param graph
```
petervdonovan, line 159: The `graph` parameter is gone now. This is private, so just deleting the `@param` comments might be an acceptable solution.
```suggestion
```

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.lflang.graph;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Iterables;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.eclipse.emf.common.util.EList;
```
petervdonovan, line 32: This looks unused.
```suggestion
```

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.xtext.xbase.lib.CollectionLiterals;
+import org.eclipse.xtext.xbase.lib.Functions.Function1;
```
petervdonovan, line 36:
```suggestion
```

org.lflang/src/org/lflang/graph/InstantiationGraph.java
```diff
@@ -0,0 +1,178 @@
...
+     * encountered instantiations. Also map each reactor to all
+     * declarations associated with it and each reactor to the sites of
+     * its instantiations.
+     * @param instantiation
+     * @param graph
+     */
+    private void buildGraph(final Instantiation instantiation, final Set<Instantiation> visited) {
+        final ReactorDecl decl = instantiation.getReactorClass();
+        final Reactor reactor = ASTUtils.toDefinition(decl);
+        final Reactor container = ((Reactor) instantiation.eContainer());
```
petervdonovan, line 164:
```suggestion
        final Reactor container = (Reactor) instantiation.eContainer();
```

petervdonovan (approved):
I left trivial comments, mostly on things that are immediately flagged by IntelliJ. Regardless, this is already much better now that it is not in Xtend!

org.lflang/src/org/lflang/graph/InstantiationGraph.java
lhstrh: Absolutely. Basing the port on the generated Java code has the disadvantage of inheriting these quirky dependencies. We should eliminate them.

org.lflang/src/org/lflang/graph/InstantiationGraph.java
hokeun: Great point thanks!

org.lflang/src/org/lflang/graph/InstantiationGraph.java
hokeun: Great catch!

org.lflang/src/org/lflang/graph/InstantiationGraph.java
hokeun: Thanks, removed them.

org.lflang/src/org/lflang/graph/InstantiationGraph.java
hokeun: Thanks, I was able to replace some of these with Java native API here: https://github.com/lf-lang/lingua-franca/pull/895/commits/f39126ea340e535bd7315c1f727042a592178a5e

hokeun: OK, now we have no dependence on xtext in this class!

hokeun:
> I left trivial comments, mostly on things that are immediately flagged by IntelliJ. Regardless, this is already much better now that it is not in Xtend!

Thank you for the thorough review!

org.lflang/src/org/lflang/util/IteratorUtil.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Iterator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import org.lflang.lf.Reactor;
+
+/**
+ * A utility class for Iterator.
+ *
+ * @author Clément Fournier
```
lhstrh, line 12:
```suggestion
```
I wrote the other two functions, so I'm sure attribution.

lhstrh (approved):
Looks great! Thanks!

PR #901, opened 2022-01-27
Carla intersection updates

Soroosh129:
This PR:
- Updates the Cara Intersection demo in LF to allow for a federated version
- Introduces the ROS version of the same demo


housengw (approved):
LGTM!

PR #909, opened 2022-01-28
port `ASTUtils.xtend` to java

housengw:
port `ASTUtils.xtend` to java

see #838


org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -1556,60 +1578,68 @@ static def withoutQuotes(String s) {
...
-    def static TargetDecl targetDecl(Resource model) {
-        return model.allContents.filter(TargetDecl).head
+    public static TargetDecl targetDecl(Resource model) {
+        return IteratorExtensions.head(Iterators.filter(model.getAllContents(), TargetDecl.class));
     }
 
+    /**
+     * Returns the list if it is not null. Otherwise return an empty list.
+     */
+    private static <T> List<T> convertToEmptyListIfNull(List<T> list) {
```
lhstrh, line 1642: I'm OK with leaving this here for now, but it should probably go into a separate `ListUtil` class. There's also `ListUtils.emptyIfNull(List<T> list)` in Apache Collections 4. I'm not sure if we already have that as a dependency in the project, but if we do, we should just use that.

lhstrh (approved):
This overall looks very good. I think this class could use some attention in order to weed out some things that we don't need and ensure that we have good code coverage, but we can do that at a later point. Let's merge this in to avoid folks from running into conflicts with this port, which is quite likely. Great job!

org.lflang/src/org/lflang/ASTUtils.java
housengw: The project currently does not have apache collections4 as a dependency

PR #915, opened 2022-01-30
Comment out Rust's guided search benchmark, because it locks up sometimes

jhaye:
See: https://github.com/lf-lang/reactor-rust/issues/11

This is to ensure everything runs smoothly when running all benchmarks for all targets.


jhaye:
A fix was found, so that should be merged instead: https://github.com/lf-lang/lingua-franca/pull/917

Issue #908, opened 2022-01-28
invalid time literal causees IllegalArgumentException


housengw:
minimal example:
```
target C;
main reactor {
    timer t(0 m)
}
```

output
```
wonghouseng@beefy-desktop-linux:~/Desktop/lingua-franca/test/C$ lfc src/exp.lf 
0    [main] ERROR text.validation.CompositeEValidator  - Error executing EValidator
java.lang.IllegalArgumentException
	at org.lflang.JavaAstUtils.toTimeValue(JavaAstUtils.java:160)
	at org.lflang.JavaAstUtils.getLiteralTimeValue(JavaAstUtils.java:213)
	at org.lflang.generator.ReactorInstance.getTimeValue(ReactorInstance.java:635)
	at org.lflang.generator.TimerInstance.<init>(TimerInstance.java:65)
	at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:816)
```

housengw:
fixec by #916

PR #916, opened 2022-01-30
Error reporting

edwardalee:
This PR started as a fix to #908, where an invalid time unit was causing an exception rather than a validation error. However, the fix exposed a more serious problem, which was that validation errors were being ignored when generating code using the 'Trigger LF code generation' button that @a-sr added to Epoch.  Fixing that exposed a further problem, which was that federated programs were failing validation at this stage because the validator did not recognize that the added reactions were unordered.

Since we don't have a way in the AST to mark reactions unordered, I've implemented a hack, which is to put a magic string at the start of a reaction body to mark it unordered.  Please don't tell me this is a hack. I know it is. I have no better way to solve this problem short of adding an 'unordered' keyword to the language to specify that a reaction is unordered. Feel free to implement any alternative solution.


org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -218,6 +227,14 @@ public ReactionInstance(
      */
     public TimeValue deadline = new TimeValue(TimeValue.MAX_LONG_DEADLINE, TimeUnit.NANO);
 
+    /**
+     * Sadly, we have no way to mark reaction "unordered" in the AST,
+     * so instead, we use a magic comment at the start of the reaction body.
+     * This is that magic comment.
+     */
+    public static String UNORDERED_REACTION_MARKER
+            = "// **** This reaction is unordered.";
```
Soroosh129, line 236: I think this is assuming that `//` is a comment indicator, which is fine for all of our target languages (except for Python, but it doesn't have a need for unordered reactions written in the Python language). Maybe this is a strong enough reason to introduce the `unordered` syntax?

lhstrh: No, we should only every introduce new syntax if we can agree that this improves the utility of the programming language, not because of some internal compiler issue. So far, I have not seen a compelling use case or argument for such syntax. The only thing that would be needed to fix _this_ particular issue is to specify for each target what the single-line comment character sequence is and use that instead of a hard-coded string. But as you already pointed out, it isn't really an issue yet. My point is, despite this being a hack, it would generalize just fine. Another solution that generalizes well and is not a hack, is to put unordered reactions in separate reactors.

edwardalee: This is a good idea. The 'magic string' could be made target-specific. How to do that?

org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
```diff
@@ -222,8 +227,27 @@ public static void validateImports(
             if (
                 bad.contains(resource) || issues.size() > 0
             ) {
+                // Report the error on this resource.
+                // Need a path, which, sadly, seems impossibly difficult to get reliably.
+                // Be very paranoid here because we don't want errors to occur while error reporting.
+                // FIXME: Is this how to get a path?
```
lhstrh, line 233: `fileConfig.srcFile` should do the trick: https://github.com/lf-lang/lingua-franca/blob/b7dc90e2c2df3c9dcba3eeae462ecc6edf22b6c0/org.lflang/src/org/lflang/FileConfig.java#L121)

org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
```diff
@@ -222,8 +227,27 @@ public static void validateImports(
...
+                for (Issue issue : issues) {
+                    errorReporter.reportError(path, issue.getLineNumber(), issue.getMessage());
+                }
+                
+                // Report errors on resources that import this one.
                 for (Reactor downstreamReactor : instantiationGraph.getDownstreamAdjacentNodes(reactor)) {
                     for (Import importStatement : ((Model) downstreamReactor.eContainer()).getImports()) {
+                        // FIXME: This will report the error on ALL import statements in
+                        // file doing the importing, not just the one importing this resource.
+                        // I have no idea how to determine which import statement is the right one.
```
lhstrh, line 250: The classes on the left-hand side of the `from` of an import statement should be linked to a class definition in some resource by our scope provider. If the resource matches the resource currently being validated, then that statement should be marked as erroneous.

org.lflang/src/org/lflang/generator/ReactionInstance.java
lhstrh: We can easily specify this in the `Target` class.

edwardalee: It's not clear to me what should go in the Target class. For now, I've removed the comment prefix '// ' and use String.contains() rather than String.startsWith() (which is much less efficient because it has to search the whole string when there is no match). If someone wants to take over the use of Target, I would be grateful.

org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
edwardalee: Thanks! That does it.

org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
edwardalee: I don't understand this suggestion. The resource will not match the resource currently being validated because it is an imported file. In any case, this bug is not new to this PR, but was in the code before. I just added the FIXME. I'll open an issue for this and move on.

lhstrh: OK, that works too.

org.lflang/src/org/lflang/generator/ReactionInstance.java
lhstrh: You're hiding something in the comments, assuming that `//` is indeed recognized as the start of a comment. This is not true for all targets. If, in the `Target` enum you specify for each target what their single-line comment character sequence is, then you can use that instead so that the trick generalizes across targets. That said, the concern @Soroosh129 raised isn't really a concern right now, because the trick isn't being used in Python yet. I'm also fine deferring this.

Soroosh129: I agree. This shouldn't be a concern for a while. For this to be a concern, a new target language must come along that doesn't have `//` as a comment annotation and implement federated execution using that target language natively.

Soroosh129: I was more trying to clarify if others think we should add the `unordered` keyword or not based on the raised concern.

edwardalee:
The Rust test that is causing test failures here is, in fact, invalid LF code. It has this:
```
    reaction() clock {= panic!("Should not be executed") =}
```
The validator rejects reactions with no trigger. This was masked by the non-invocation of the validation that this PR fixes.

lhstrh:
> 

Does it? I thought no triggers meant that it would execute when _any_ input or action was present. In retrospect, I think that (especially considering the fact that contained reactors' outputs can be triggers) that was a bad idea.

Currently, the validator gives a warning, not an error. How should we move forward?

PR #917, opened 2022-01-31
Fix Rust's Guided Search benchmark locking up, and other fixes

jhaye:
Some small fixes regarding idiomacy and performance, but mainly fixing the random softlock.

This had to do with creating the grid from a variable random seed that would sometimes create grids with no path from source to target, resulting in the search loop to run forever. To quote from the commit:

> It seems like with a variable seed, sometimes a grid is created with no
> path from start to target, so we just have to use a static seed that
> creates a usable grid every time. The C++ benchmark does the same thing.

### DIFF
--- a/benchmark/Rust/Savina/src/lib/guided_search.rs
+++ b/benchmark/Rust/Savina/src/lib/guided_search.rs
@@ -136,7 +136,7 @@ impl Grid {
             }
         }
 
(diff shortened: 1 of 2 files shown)


lhstrh (approved):
Looks good to me! Thanks for fixing this!

PR #918, opened 2022-01-31
Destructors

edwardalee:
This PR removes bogus destructor functions that are currently generated but are hopelessly incomplete and replaces them with generic functions to allocate and free memory.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.xtend
@@ -3761,9 +3761,14 @@ class CGenerator extends GeneratorBase {
                 initializeTriggerObjects.pr('''
                     «portRefName»_width = «output.width»;
                     // Allocate memory for multiport output.
-                    «portRefName» = («portStructType»*)_lf_allocate(«output.width», sizeof(«portStructType»), (self_base_t*)«reactorSelfStruct»); 
-                    «portRefName»_pointers = («portStructType»**)_lf_allocate(«output.width», sizeof(«portStructType»*), (self_base_t*)«reactorSelfStruct»);
-                    // Assign each output port pointer to be used in reactions to facilitate user access to output ports
+                    «portRefName» = («portStructType»*)_lf_allocate(
+                            «output.width», sizeof(«portStructType»),
+                            &«reactorSelfStruct»->base.allocations); 
+                    «portRefName»_pointers = («portStructType»**)_lf_allocate(
+                            «output.width», sizeof(«portStructType»*),
+                            &«reactorSelfStruct»->base.allocations); 
+                    // Assign each output port pointer to be used in
+                    // reactions to facilitate user access to output ports
                     for(int i=0; i < «output.width»; i++) {
                          «portRefName»_pointers[i] = &(«portRefName»[i]);
                     }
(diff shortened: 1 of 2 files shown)


org.lflang/src/org/lflang/generator/c/CGenerator.xtend
```diff
@@ -2041,25 +1985,20 @@ class CGenerator extends GeneratorBase {
...
-            ''')
-        }
+        generateReactionAndTriggerStructs(body, decl, constructorCode, federate)
+        
+        // The first field has to always be a pointer to the list of
+        // of allocated memory that must be freed when the reactor is freed.
+        // This means that the struct can be safely cast to self_base_t.
+        code.pr('''
+            typedef struct {
+                struct allocation_record_t *allocations;
```
Soroosh129, line 1998: Can this directly be `struct self_base_t* base` instead of `struct allocation_record_t* allocations`? Maybe it would be more future-proof and allow for expansion of `self_base_t`.

Soroosh129 (approved):
Looks great to me! I only had a minor comment.

org.lflang/src/org/lflang/generator/c/CGenerator.xtend
edwardalee: > Can this directly be `struct self_base_t* base` instead of `struct allocation_record_t* allocations`? Maybe it would be more future-proof and allow for expansion of `self_base_t`.

I don't understand this suggestion.  The typedef here is defining the self struct of a specific reactor class, which is different for every reactor class. The code here ensure that the first thing in that struct is a pointer to a list of allocations. Since this is the case for all self structs, then in C, you can cast the specific reactor class struct to `self_base_t` and then access the list of allocations.  This is what enables a generic destructor. It takes a specific self struct as an argument, casts it, and frees all the memory on the list.

Soroosh129: Yes, but couldn't the struct definition for a specific reactor class be:
```C
typedef struct {
    struct self_base_t* base;
    // Rest of the code
}
```

edwardalee: Sorry if I'm being dense, but I still don't understand. Your struct would have to be cast to some struct type whose first element is `self_base_t* base`. What would we call that type? `self_base_base_t`?

Soroosh129: I think the struct that I am proposing can still be cast to `self_base_t` using the same `(self_base_t)bar` casting notation, but if we now add members to the definition of the `self_base_t` struct, these members will also be included in all new instances of `foo_type_t` as well. This is "simulating" inheritance in C.

A good example of this type of inheritance is explained here in the context of extending the base PyObject type in the C Python API to create new types: https://docs.python.org/3/extending/newtypes_tutorial.html

edwardalee: Ah, got it!  The `*` is what threw me off. In fact, I don't think it would work with the `*`.

Soroosh129: I think we could make it work, but it would require an additional step to dereference the pointer and then cast it if I'm not mistaken. In any case, I think not using a pointer is the way to go.

lhstrh:
This looks good, but not critical for the release. Should we wait merging this in until after we tag `0.1.0-beta`? The `error-reporter` branch on the other hand, I think we should urgently get ready and merge in...

edwardalee:
Oops, sorry, didn't see your message.  I think it is critical to get it in the beta because otherwise someone might run `leaks` on our C code and assume we have been sloppy.  Also, the bogus destructors were a bit embarrassing.

edwardalee:
As for the error-reporter branch, we can't merge this until we figure out how to fix the failing tests it exposed. The Cpp one is easy: The validator does not allow reactors in Cpp to be named `Preamble`, but there is test called `Preamble.lf`.  This is just one example of deeper problems that might be lurking in the `lsp` and `cli` tests, where validation errors have been systematically masked by the bug fixed by the error-reporter branch.

lhstrh:
It's really only the Cpp and Rust tests that have validation in them. I think that once those are fixed the LSP tests will pass as well. Then the only thing left is the NPE in the CLI tests.

PR #924, opened 2022-02-02
Bump versions excludes

petervdonovan:
Locally, I have found that the version bump script detects a large number of manifest files from build directories in `org.lflang.rca` and `org.lflang.lds`. We do not track or maintain these manifest files, for they belong to other projects that we depend on. This addresses that.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,8 +4,8 @@ version=0.1.0-SNAPSHOT
 
 [versions]
 commonsCliVersion=1.4
+guiceVersion=5.0.1
 jacocoVersion=0.8.7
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -4,8 +4,8 @@ version=0.1.0-SNAPSHOT
 
 [versions]
 commonsCliVersion=1.4
+guiceVersion=5.0.1
 jacocoVersion=0.8.7
-juiceVersion=5.0.1
```
lhstrh, line 8: Definite typo :-)

lhstrh (approved):
Thanks for these fixes!

PR #933, opened 2022-02-03
Fixes #932

### LINKED ISSUE #932: Very strange error in C tests
See https://github.com/lf-lang/lingua-franca/actions/runs/1787860560
This failure deserves attention...

Soroosh129:


Soroosh129:
Should we update the error message in the deadline handler? It should be much rarer for a "misfire" of the deadline handler to occur, but not impossible.

edwardalee (approved):
Soroush is right that this test can legitimately fail nondeterministically. In particular, if one of the deadline misses that is supposed to occur is bad enough (it misses by much more than expected), then the whole execution lag (physical time - logical time) gets large enough that subsequent reactions that are not supposed to miss the deadline will nevertheless miss the deadline.  As @Soroosh129  suggested, we should probably update the error message.  Perhaps  instead of "Deadline miss handler invoked without deadline violation" we should say "Unexpected invocation the deadline miss handler. This could be indicative of true error, but it can also occur nondeterministically if the execution lag (physical time - logical time) gets large enough that the system cannot catch up."

lhstrh:
A very similar failure has currently broken the build: https://github.com/lf-lang/lingua-franca/pull/933/files

PR #934, opened 2022-02-03
Fix gradle setup by setting kotlin version to 1.4.30

### LINKED ISSUE #930: Kotlin classes do not get build by `buildLfc` task _unless_ the `clean` task is also run ...
... or if there are no build artifacts present from previous builds.

`./gradlew buildLfc` runs successfully and outputs a jar, but the dynamically loaded Kotlin classes cannot be found at runtime.

cmnrd:
According to the [gradle documentation](https://docs.gradle.org/current/userguide/upgrading_version_6.html), gradle 7.0 supports kotlin 1.4.30. Specifying this kotlin version in gradle.properties fixes the build problems described in #930.

Closes #930


petervdonovan (approved):
Thank you @cmnrd! As far as I know the motivation for the Kotlin version bump had been that the default JVM target was too old, but since we set the JVM target to 11, this looks like it should be no problem.

Issue #929, opened 2022-02-02
Epoch fails with IllegalStateException when compiling C++ programs via the code generation button


cmnrd:
Stack trace:
```
java.lang.IllegalStateException: This must only be called once on a given progress monitor instance.
	at org.eclipse.core.runtime.SlicedProgressMonitor.beginTask(SlicedProgressMonitor.java:41)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:483)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:449)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.syncIfNecessary(EclipseResourceFileSystemAccess2.java:545)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.syncIfNecessary(EclipseResourceFileSystemAccess2.java:538)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.getTraceFile(EclipseResourceFileSystemAccess2.java:531)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.generateFile(EclipseResourceFileSystemAccess2.java:212)
	at org.eclipse.xtext.generator.AbstractFileSystemAccess.generateFile(AbstractFileSystemAccess.java:100)
	at org.lflang.generator.cpp.CppGenerator.generateFiles(CppGenerator.kt:119)
	at org.lflang.generator.cpp.CppGenerator.doGenerate(CppGenerator.kt:70)
	at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:163)
	at org.eclipse.xtext.generator.GeneratorDelegate.doGenerate(GeneratorDelegate.java:44)
	at org.eclipse.xtext.generator.GeneratorDelegate.generate(GeneratorDelegate.java:35)
	at org.lflang.ui.actions.CompileActionHandler$1.run(CompileActionHandler.java:216)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

This seems to be related to the usage of the `AbstractFileSystemAccess` interface of the C++ generator, but I am not sure why this error is triggered.

lhstrh:
@petervdonovan is working on a fix; I will merge that in, re-tag `v0.1.0-beta`, and upload new build artifacts.

Issue #149, opened 2020-04-24
Unify the output behavior of code generators


cmnrd:
I have noticed that the file generation across targets uses different approaches which also leads to generated files being placed in different locations. I want to start a discussion about this and hope that we can decide for a common approach.

Consider the source file `<project-root>/foo/Bar.lf` where `<project-root>` refers to a project directory in the eclipse workspace. When compiling the file `Bar.lf`, this is what currently happens in the different targets:

- **C**: Places generated source code in `<project-root>/foo/src-gen` (In the directory containing `Bar.lf`. This happens when compiling from Eclipse and when compiling via `lfc`.
- **C++**: Uses the [IFileSystemAccess2](https://archive.eclipse.org/modeling/tmf/xtext/javadoc/2.9/org/eclipse/xtext/generator/IFileSystemAccess2.html) interface provided by Xtext to generate files. This places files in the poject root when compiling from Eclipse: `<project-root>/src-gen`. When compiling via `lfc`, it places the generated files in the current working directory: `<CWD>/src-gen`.
- **TS**: Uses a similar mechanism as the C target, but does not create a src-gen directory. It creates a directory `<project-root>/foo/Bar/` when compiling from both Eclipse and `lfc`.

I think in any case, we should stick to the usage of `src-gen`. The TS target can create subdirectories as needed within (`src-gen/Bar`). This is also what the C++ target does.

But where should the `src-gen` directory be placed? I personally prefer the Xtext default which is used in the C++ target. When running `lfc`, I would expect it to operate in the CWD and not to create new files next to the specified input file. When compiling from Eclipse, it creates `src-gen` in the project root. I am not a regular Eclipse user, but I think that is also what most Eclipse users would expect. It acknowledges that Eclipse has the notion of a project to group related things. If the generated code of two LF files should not be placed in the same src-gen directory, then those two files should actually not be part of the same project.

What is your opinion on this matter?

Independent of what the default behavior should be, I think it would make sense for `lfc` to accept an optional command line argument that allows the user to specify the output directory.

lhstrh:
I agree that the behavior should be the same across targets, and I think your suggestion to put generated code in `<project-root>/src-gen` when running in `INTEGRATED` mode, and in `<cwd/src-gen>` when running in `STANDALONE` mode indeed is the most sensible solution. Unless anyone disagrees, I'll make the change as soon as I get to it. Also, :+1: for the suggestion to add an output path to `lfc` :-)

[...]

PR #938, opened 2022-02-04
Refactoring: Purge code generators of IFileSystemAccess2.

petervdonovan:
#929 is not the first bug that was related to `IFileSystemAccess2` (see acf98595). In #929, writing to a file caused an error in `org.eclipse.core.runtime.SubMonitor`, a progress reporting class which I could not find the source code for. It is annoying for trivial operations like writing to files to cause `IllegalStateException`s in seemingly unrelated parts of the framework whose state is only ever touched in other, distant parts of our code base.

For this reason, this PR eliminates `IFileSystemAccess2` wherever possible. In  #149, @cmnrd gave a valid reason why we should use the `IFileSystemAccess2` interface (it specifies the `src-gen` directory), but this PR preserves the code that gets the location of the `src-gen` directory from the `IFileSystemAccess2`.

**EDIT:** This change has been tested in Epoch by compiling at least one test program in Epoch for each target.

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/TestBase.java
+++ b/org.lflang.tests/src/org/lflang/tests/TestBase.java
@@ -379,7 +379,7 @@ public abstract class TestBase {
         }
 
         fileAccess.setOutputPath(FileConfig.findPackageRoot(test.srcFile, s -> {}).resolve(FileConfig.DEFAULT_SRC_GEN_DIR).toString());
-        test.fileConfig = new FileConfig(r, fileAccess, context);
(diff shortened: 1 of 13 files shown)


petervdonovan:
No, I do not know why integration tests are not running. I have heard that switching the base branch and switching back can force some things to update on GitHub, but that did not work here. The CI workflow isn't written to be manually triggered either, though I wonder if it should be. In order to trigger CI, I may have to close this PR and open a new one...

cmnrd:
Looks like that did the trick!

Issue #928, opened 2022-02-02
lfc is broken on master and in our release artifacts


cmnrd:
lfc fails for the C++, Rust, and TypeScript targets, because it cannot load the Kotlin classes. I am clueless on what might be causing this, but something in the build setup must have changed recently. For this reason, the lfc 0.1.0-beta release artifacts to not work currently.

I also wonder why none of our tests revealed this problem.

petervdonovan:
I think this problem actually did appear in CI in the `release` branch, and it seemed to be a problem with Gradle's incremental build feature. Our "solution" was to use `./gradlew clean`.

housengw:
maybe we could add `try ./gradlew clean` in the error message as a pointer...

lhstrh:
Confirming that the problem how the artifacts were built (without `clean`). There is nothing actionable w.r.t. the code. I will create a separate issue to address apparent deficiency in our Gradle configuration.

lhstrh:
> maybe we could add `try ./gradlew clean` in the error message as a pointer...

I think we should just fix the build configuration (as per #930). While the suggested error message might be useful for us, it won't be useful for third-party users (who don't build from source).

PR #939, opened 2022-02-05
Update kotlin version

oowekyala:
I was having trouble building, the kotlin compilation was failing with this error:
```
> Task :org.lflang:compileKotlin FAILED
e: java.lang.NoClassDefFoundError: Could not initialize class org.jetbrains.kotlin.com.intellij.pom.java.LanguageLevel
        at org.jetbrains.kotlin.com.intellij.core.CoreLanguageLevelProjectExtension.<init>(CoreLanguageLevelProjectExtension.java:26)
        at org.jetbrains.kotlin.com.intellij.core.JavaCoreProjectEnvironment.<init>(JavaCoreProjectEnvironment.java:42)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreProjectEnvironment.<init>(KotlinCoreProjectEnvironment.kt:26)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$ProjectEnvironment.<init>(KotlinCoreEnvironment.kt:121)
        at org.jetbrains.ko
```

After some digging I found that our gradle wrapper uses kotlin 1.5.31:
```
$ ./gradlew --version
------------------------------------------------------------
Gradle 7.3.3
------------------------------------------------------------

Build time:   2021-12-22 12:37:54 UTC
Revision:     6f556c80f945dc54b50e0be633da6c62dbe8dc71

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.2 (Azul Systems, Inc. 17.0.2+8-LTS)
OS:           Linux 5.13.0-27-generic amd64
```

This PR updates the kotlin version we use to be in sync with that of gradle, which resolves the error for me.


cmnrd (commented):
I updated this branch to use the kotlin version specified in gradle.properties.

Unfortunately, with Kotlin 1.5.31 the issue #935 appears again (this is the reason why we downgraded to 1.4.30). Essentially, if you run build-lfc twice, it will produce a jar without Kotlin classes (for whatever reason...). Maybe you can try again with Kotlin version 1.4.30 and setting the plugin version explicitly as in this change? Also, let me know if you have any ideas regarding the cause of #935.

oowekyala:
@cmnrd Thanks for the review and context. I can't reproduce the build failures I was experiencing... Maybe it was because of my jdk version 17, though today it works fine... :shrug: I guess we can just close this.

I'm not sure what to do about  the issue you're referring to (is it really #935?), but it sounds like a caching problem maybe?

petervdonovan:
> I'm not sure what to do about the issue you're referring to

(I think he probably meant #928 and #930.)

cmnrd:
Yes, that's what I meant. Not sure what happened there with the numbers :laughing:

PR #940, opened 2022-02-05
Ported Xtend code in diagram package to Java

housengw:
Ported the diagrams package to java.

Also see #838.


lhstrh:
This is a great step forward! I think we need to be very careful, however, with #501, which has been a long time in the making and has recently become ready for review. Perhaps a sensible strategy here would be to consider the diff between the `master` and `modal-models` and carefully migrate the changes to the new Java files. If we then push the affected files to `modal-models`, we can rest assured that no changes had gotten lost. @a-sr, what do you think?

lhstrh:
> Sounds reasonable. Most of the adjustments for modes are located in a separate file and changes to the existing synthesis are rather limited. I can go ahead and merge this branch into `modal-models`, this way I can also review and test the changes done here. As long as we merge #940 before #501, everything should be fine, right?

That sounds perfect. Thanks for doing that, @a-sr!

a-sr (approved):
I tested the branch and fixed a few bugs that I encountered.
All diagrams I tested look normal.
The code isn't going to win any beauty contests, but that's no wonder since the entire Klighd API is designed to be used as Xtend extensions, and having those in plain Java is horrible in any case.

I'd say it's ready to merge.

lhstrh (approved):
I think that with some refactoring, some of this code can be made a bit more elegant, but I agree with @a-sr and think this is ready to merge.

PR #943, opened 2022-02-07
Get C benchmarks working again.

petervdonovan:
I have not done a thorough review of the recent changes to the C benchmarks, but this resolves the issues I have encountered:
* The configuration file needed to be updated so that benchmarks are run `numIterations` times instead of `numIterations`^2 times
* `stdout` was flooded by the Banking benchmark
* `numIterations` was not cogged for Banking


Soroosh129 (approved):
Looks good! I would suggest replacing the commented out `printf` statements with `info_print`s, which should not show up in the output if `logging: warn` is set.

PR #944, opened 2022-02-07
Scale back LSP tests

petervdonovan:
This scales back LSP tests because they have been taking a little more computational resources on GitHub Actions than is ideal. Additionally, the "build and run" tests are now redundant.

### DIFF
--- a/.github/workflows/c-tests.yml
+++ b/.github/workflows/c-tests.yml
@@ -51,7 +51,7 @@ jobs:
         if: ${{ runner.os == 'Linux' }}
       - name: Perform tests for C target with default scheduler
         run: |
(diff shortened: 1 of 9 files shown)


petervdonovan:
OK, this PR has changed the tests slightly so that the random number generator's state is different, and now it is blocked by test failures that cannot be reproduced locally. PyLint `2.12.2` seems to be giving different output for the same input and the same OS on GitHub's end in comparison to my machine. This should be fixed, but I am out of ideas for the moment.

lhstrh (commented):
Looks good, but did you mean to touch `reactor-c` as part of this PR?

petervdonovan:
> did you mean to touch reactor-c as part of this PR?

Definitely not -- I missed that while I was rebasing.

lhstrh (approved):

PR #949, opened 2022-02-08
fix port absent

housengw:
This PR fixes the bug where a federate in centralized coordination advance time too quickly before the RTI can send a PTAG, which prevents it from sending port absent messages to earlier tags.


edwardalee (commented):
Very interesting. This looks right to me, but it would good for @Soroosh129 to have a look.

Soroosh129 (approved):
This looks right to me as well.
Thanks for submitting this PR!

PR #950, opened 2022-02-09
 Port  `org.lflang.federated` to java

housengw:
see #838


Soroosh129 (approved):
The port overall looks good to me.

For some reason, GitHub is not showing me the ported FedLauncher file in a diff, so it is hard to verify that the logic matches the old Xtend implementation. I suspect the remote scripts in FedLauncher were broken even before this port, since we have no tests for them. However, just to be sure, would you please try to set the RTI and a federate to a remote host (e.g., 127.0.0.1 should work) and see if the `_distribute` and launch scripts work?

You would have to have password-less ssh set up on your local machine (so that `ssh 127.0.0.1` doesn't ask for a password, or to store any keys).

At some point, I think we either need to remove these remote scripts and go all in to Docker, or we should add tests for these to make sure they are not broken, but this is obviously beyond the scope of this PR.

housengw:
The distribute script is indeed broken before the port. The problem is that `compileCommandForFederate` returns an empty string, so the compile commands starts with a file, which returns a `Permission denied` error

PR #952, opened 2022-02-09
Fixes #768

### LINKED ISSUE #768: Extra folders generated for main reactors in imported files in the Python target
suppose we have two programs:

`HelloWorld.lf`
```
target Python;

reactor HelloWorld2 {
    reaction(startup) {=
        # do nothing
    =}
}

main reactor {
    a = new HelloWorld2();
}
```

`b.lf`
```
target Python;

import HelloWorld2 from "HelloWorld.lf";

main reactor {
    a = new HelloWorld2();
}
```

When I compile `b.lf` with lfc, a folder for the main reactor in the imported file also gets generated, which it shouldn't.

housengw:
Fixes #768


housengw:
ok, seems like there is a couple of flaky tests in the CI that randomly fails...

Issue #953, opened 2022-02-09
LSP tests randomly failing


housengw:
The `Uninstall packages MacOS` of the lsp test fails randomly with the same error message.

```
Run brew uninstall gcc
Error: Refusing to uninstall /usr/local/Cellar/gcc/11.2.0_3
because it is required by openblas and r, which are currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies gcc
Error: Process completed with exit code 1.
```

PR #954, opened 2022-02-09
attempt to fix 953

housengw:
See #953 

This PR adds the `--ignore-dependencies` flag to the uninstall GCC instruction according to the error message.


petervdonovan (approved):
Thank you! This seems like a good solution if CI passes.

PR #956, opened 2022-02-10
Clean up the file writing mechanism

cmnrd:
The path should be specified as an actual Path object.
Also Java provides simpler and more efficient mechanisms for writing files.

### DIFF
--- a/org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
+++ b/org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
@@ -314,9 +314,7 @@ public class JavaGeneratorUtils {
      * @param path The file to write the code to.
      */
     public static void writeToFile(String text, Path path) throws IOException {
-        if (!path.getParent().toFile().mkdirs()) {
-            throw new IOException("Could not create the directory " + path);
-        }
+        path.getParent().toFile().mkdirs();
         Files.write(path, text.getBytes());
     }
 


org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
```diff
@@ -312,13 +313,20 @@ public static LFResource getLFResource(
...
         }
+        Files.write(path, text.getBytes());
+    }
+
+    /**
+     * Write text to a file.
+     * @param text The text to be written.
+     * @param path The file to write the code to.
+     */
+    public static void writeToFile(CharSequence text, Path path) throws IOException {
```
petervdonovan, line 328: (For context, the motivation for using `CharSequence` here instead of `String` was that the C generator was having out-of-memory errors because of its large strings, and `toString()` does create a copy. However, I guess that is not important anymore since the C generator has been repaired by `scalability-banks`.)

org.lflang/src/org/lflang/generator/JavaGeneratorUtils.java
```diff
@@ -312,13 +313,20 @@ public static LFResource getLFResource(
...
+        writeToFile(text.toString(), path);
```
petervdonovan, line 329: (For context, the original motivation for using `CharSequence` here instead of `String` was to avoid calling `toString()` or `getBytes()`. The C generator was having out-of-memory errors because of its large strings, and `toString()` and `getBytes()` do seem to create copies. However, that is not important anymore since the C generator has been repaired by `scalability-banks`.)

PR #957, opened 2022-02-11
C benchmark repairs

petervdonovan:
This patches some flaws that recently appeared in a few C benchmarks. With the changes in this branch, I think that no obviously incorrect benchmark results remain.


cmnrd (approved):
Thanks! Looks great

Issue #403, opened 2021-07-11
Limitations of deadlines


edwardalee:
The deadline construct in LF is capable of detecting deadline violations only after the fact. If a reaction runs longer than expected, this can be detected by downstream reactor that receives an input from an output produced by that reaction, but the detection occurs only after the upstream reaction has completed execution. Suppose instead that we want to react as soon as the upstream reaction has exceeded its time budget?  One way to do this might be with the following syntax:
```
reactor Downstream {
    input trigger;
    input expected;
    reaction (trigger) {=
         printf("Expecting input within 10 msec.\n");
    =} deadline(expected) 10 msec {=
         printf("PANIC: Expected input did not arrive on time!\n");
    =}
}
```
Notice the additional argument to the `deadline` keyword, which is the input port `expected`.  The meaning of this could be as follows:

Let *t* be the logical time and *T* be the physical time at which the `trigger` reaction fires.  If *T* > *t* + 10 msec, then the PANIC body is invoked, as it would be for an ordinary deadline instruction. Otherwise, the ordinary reaction is invoked, but *in addition*, a physical timer is set to expire after 10 msec - (*T* - *t*). When that physical timer expires, a reaction is put on the reaction queue at highest priority. When that executes, it checks to see whether the `expected` input is known (absent or present), and if not, it invokes the PANIC reaction. It would also have to check whether logical time has advanced past the one at which the timer was started, in which case, there is no need to PANIC.

This seems kind of complicated, but I think it would work, at least in a multithreaded environment that either has multiple cores or has an underlying preemptive priority-driven thread scheduler.

Why is it so complicated?  First, I think the detection has to be done in a separate reactor from the one where the violation occurs.  Suppose we were to try this:

```
reactor Upstream {
    input trigger;
    output result;
    reaction(trigger) -> result {=
        // Long-running computation.
    =} completion_deadline 10 msec {=
         printf("PANIC: Result is not ready on time!\n");
    =}
}
```
In this case, the PANIC reaction body would be invoked concurrently with the Long-running computation, a violation of the mutual exclusion principle in Lingua Franca.  To allow this violation, both the Long-running computation and the PANIC reaction bodies would have to mutexes to access outputs and state variables, which is a lot to expect from programmers and is impossible to enforce.

[...]

PR #960, opened 2022-02-12
Add C code generation to store currently executed reaction, add AnytimePrime.lf as a demo for check_deadline().

hokeun:
Relevant PR in reactor-c: https://github.com/lf-lang/reactor-c/pull/38
Relevant issue: https://github.com/lf-lang/lingua-franca/issues/403

This demo program was built with the help of @edwardalee and @lhstrh.
The demo program's name is inspired by @lhstrh.


experimental/C/src/AnytimePrime.lf
```diff
@@ -0,0 +1,67 @@
+/**
+ * This program is a concept demonstration showing how the check_deadline()
+ * function works. This program performs "anytime computation" (https://en.wikipedia.org/wiki/Anytime_algorithm)
+ * to find the largest prime within the given deadline. The check_deadline()
+ * function is called after checking each number if it is prime. If the
```
edwardalee, line 5:
```suggestion
 * function is called after checking whether each number is prime. If
```

experimental/C/src/AnytimePrime.lf
```diff
@@ -0,0 +1,67 @@
...
+ * check_deadline() is called after the deadline has passed, the function calls
+ * the deadline handler and returns true. Then the program outputs the largest
```
edwardalee, line 7:
```suggestion
 * the deadline handler and returns true. Then the reaction outputs the largest
```

experimental/C/src/AnytimePrime.lf
```diff
@@ -0,0 +1,67 @@
...
+};
+
+preamble {=
+#include "vector.c"
+=}
+
+reactor Prime {
+    output out:{=long long=};
+    reaction(startup) -> out {=
+    int num_primes = 1;
```
edwardalee, line 29: Indentation is odd. Should use spaces, not tabs.

edwardalee (approved):
Looks great!  This is a significant improvement on our deadline mechanism...

experimental/C/src/AnytimePrime.lf
```diff
@@ -0,0 +1,67 @@
...
+ * @author Marten Lohstroh (marten@berkeley.edu)
+ */
+target C {
+    fast: true,
+    files: ["/lib/c/reactor-c/core/utils/vector.h",
+            "/lib/c/reactor-c/core/utils/vector.c"]
+};
+
+preamble {=
+#include "vector.c"
```
lhstrh, line 23: We should include the `.h` file, not the `.c` file.

experimental/C/src/AnytimePrime.lf
```diff
@@ -0,0 +1,67 @@
...
+    reaction(in) {=
+        printf("Largest prime found within the deadline: %d\n", in->value);
+    =}
+}
+
+main reactor AnytimePrime {
+    p = new Prime();
+    d = new Print();
+    p.out -> d.in;
+}
```
lhstrh, line 67: `\n` at end of file missing.

lhstrh (changes requested):
Only minor issues.

experimental/C/src/AnytimePrime.lf
edwardalee: Including the .h files requires additional steps, apparently:
1. Create a CMake file.
2. Include the CMake file.
This seems like quite a bit of extra hassle to me.  Any way to streamline this?

lhstrh: I agree that it's a hassle. I'm not sure how to improve this, but @cmnrd might have ideas. I think that we should go through the hassle or make improvements to reduce the hassle, not take shortcuts and settle on a bad practice because it's quick.

experimental/C/src/AnytimePrime.lf
hokeun: Thanks for catching!

experimental/C/src/AnytimePrime.lf
hokeun: Thanks!

experimental/C/src/AnytimePrime.lf
hokeun: I fixed this issue by adding the cmake for vector in reactor-c, as I can imagine the resizable vector can be useful in other LF programs as well as this one.

cmnrd: I think that for C and C++ requiring this additional cmake file isn't such a big hassle. One would need to do the same when writing a standalone C/C++ application.

We could think about adding syntax (or target properties) that make the inclusion of extra files simpler, but as always with C/C++ it is impossible to cover all the corner cases. I am in favor of giving programmers the tools they are used to rather than having oversimplified built-in support.

Issue #903, opened 2022-01-27
weird error message when STP or deadline is out of bound.


housengw:
minimal program:
```
target C;

reactor A {
    input i:int;
    reaction(i) {=

    =} STP(2147483648 nsec) {=

    =}
}
```

Error message:
```
For input string: "2147483648"
```

Same thing happens for deadlines

housengw:
I suspect this issue is an integer overflow issue caused by the fact that `TimeValue.java` uses `long` for its time values and `Time.java` uses `int` for its time values.

housengw:
I think we should parse the time values as strings in the grammar file and then do validation on the strings instead of ints or longs to prevent overflow

edwardalee:
I tried to fix this but couldn't find a way that didn't make the grammar ambiguous.  I suggest we not fix this. I don't think it's worth it given that we have time units.

PR #961, opened 2022-02-13
Validator cleanups

edwardalee:
This PR cleans up LFValidator:
* Organize methods for easier browsing.
* Make fields private that are not used anywhere else.
* Remove fields that are not used anywhere at all.
* Rename private static constants to use consistent convention.

This should be merged quickly because of the major file structure reorganization, which will likely lead to conflicts.

There is only one minor functional change: The checkSTPOffset() method was not actually performing any check. Now it is (so there is a supporting change in ModelInfo).

### DIFF
--- a/org.lflang/src/org/lflang/ModelInfo.java
+++ b/org.lflang/src/org/lflang/ModelInfo.java
@@ -43,6 +43,7 @@ import org.lflang.lf.Instantiation;
 import org.lflang.lf.Model;
 import org.lflang.lf.Parameter;
 import org.lflang.lf.Reactor;
+import org.lflang.lf.STP;
 
 
 /**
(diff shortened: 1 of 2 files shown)


lhstrh (commented):
I don't see any red flags here, but the diff is impossible to read because many things were moved around. I would be great if you could also add a unit test for the newly created check.

Soroosh129 (approved):
Looks good to me.

Thank you for adding the check for the STP offset.

Would this also solve #903?

**Edit:** I agree with @lhstrh that adding tests for the newly added checks would really help.

edwardalee:
This does not fix #903. In fact, the check of STP still does not work (it didn't work before). Neither does the check for overflowing deadlines, which uses the same architecture.  It needs to be a separate PR to fix those checks. This PR does not change any functionality at all.

Issue #907, opened 2022-01-28
reactor extends itself causes StackOverflowError


housengw:
minimal example 
```
target C;
reactor A extends A {}
```

output
```
wonghouseng@beefy-desktop-linux:~/Desktop/lingua-franca/test/C$  lfc src/exp.lf
0    [main] ERROR text.validation.CompositeEValidator  - Error executing EValidator
java.lang.StackOverflowError
	at org.eclipse.emf.ecore.util.EcoreEList.resolveProxy(EcoreEList.java:206)
	at org.eclipse.emf.ecore.util.EcoreEList.resolve(EcoreEList.java:161)
	at org.eclipse.emf.ecore.util.EObjectResolvingEList.resolve(EObjectResolvingEList.java:60)
	at org.eclipse.emf.common.util.BasicEList.get(BasicEList.java:348)
	at org.eclipse.emf.common.util.AbstractEList$EIterator.doNext(AbstractEList.java:698)
	at org.eclipse.emf.common.util.AbstractEList$EIterator.next(AbstractEList.java:685)
	at org.lflang.ASTUtils.allParameters(ASTUtils.java:615)
	at org.lflang.ASTUtils.allParameters(ASTUtils.java:616)
...
	at org.lflang.ASTUtils.allParameters(ASTUtils.java:616)
lfc: fatal error: An unexpected error occurred:
java.lang.NullPointerException
	at org.lflang.FileConfig.toIPath(FileConfig.java:707)
	at org.lflang.FileConfig.toPath(FileConfig.java:690)
	at org.lflang.lfc.Main.getValidatedResource(Main.java:336)
	at org.lflang.lfc.Main.runGenerator(Main.java:271)
	at org.lflang.lfc.Main.main(Main.java:209)
```

edwardalee:
This is because InstantiationGraph no longer detects cycles. Perhaps because of porting to Java?

housengw:
ok, seems like `ASTUtils.allParameters` is called before `LFValidator.checkReactor` is called and `InstantiationGraph` is initialized, and since `allParameters` is recursive, it causes a stack overflow before any checks take place.

Below is the function `allParameters`. `definition.getSuperClasses()` will always be the same reactor if the reactor extends itself. 
```
public static List<Parameter> allParameters(Reactor definition) {
        System.out.println("Doing param");
        List<Parameter> result = new ArrayList<>();
[...]

PR #962, opened 2022-02-13
Inheritance cleanups

edwardalee:
This PR changes the way inheritance is handled to fix two problems identified in #907.

### DIFF
--- a/org.lflang/src/org/lflang/ASTUtils.java
+++ b/org.lflang/src/org/lflang/ASTUtils.java
@@ -32,6 +32,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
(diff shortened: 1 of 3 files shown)


lhstrh (approved):
Looks good! I believe there are some unit tests that @housengw commented out because of the buggy behavior that is now fixed. If my recollection is correct, we should re-enable those unit tests.

lhstrh:
Looks like my changes introduced a bug. Sorry!

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -611,6 +542,29 @@ public static String getUniqueIdentifier(Reactor reactor, String name) {
...
+     * a given reactor definition.
+     * @param definition The reactor definition.
+     * @param elements A function that maps a reactor definition to a list of
+     *                 elements of type T.
+     * @param <T> The type of elements to collect (e.g., Port, Timer, etc.)
+     * @return
+     */
+    public static <T> List<T> collectElements(Reactor definition, Function<Reactor,List<T>> elements) {
+        List<T> result = new ArrayList<T>();
+        // Add elements of the current reactor.
```
cmnrd, line 556: This change looks great! The only "mistake" that I could spot is that the elements of the current reactor are added before for the elements of the super classes. In the original code, it was added after the super class elements. I don't know whether the order within the list has any meaning in this context. If it should not have meaning, then maybe the bug lies somewhere else.

cmnrd (commented):
The changes look good to me.

Indeed, the reordering in `collectElements` fixed the tests. As stated in my comment above, the fact that the test outcome depends on the order of elements within the resulting list, might indicate a bug further down the line.

edwardalee:
The order very definitely has meaning.  Not only should the base class elements be added first, but also when a reactor extends multiple base classes, those elements need to be added in the order that the extensions are declared. So there is only exactly one order that is correct.

Consider a reaction that writes to an output.  A subclass can "override" the output of a base class by overwriting it.

I'm surprised my test caught this bug. I didn't design it to check for this...

PR #967, opened 2022-02-16
Epoch error reporting

### LINKED ISSUE #966: Epoch fails to highlight errors on codegen
Minimal example:
```
target C;
main reactorx {}
```
In Epoch, when code generating, the EclipseErrorReporter fails to highlight errors with the following message:
```
WARNING: Setting markers in the IDE failed:
org.eclipse.core.internal.resources.ResourceException: Resource '/CTests/src' does not exist.
```
Curiously, although Eclipse reports no compile errors, line 122 of EclipseErrorReporter.java has a compile error. The line is:
```
                IMarker marker = iResource.createMarker(IMarker.PROBLEM);
```
The error is:
```
The type org.eclipse.core.runtime.jobs.ISchedulingRule cannot be resolved. It is 
 indirectly referenced from required .class files
```
In branch epoch-error-reporting, I've added a dependency to `org.eclipse.core.runtime` to MANIFEST.MF. I hope this doesn't create other problems...

In addition, `JavaGeneratorUtils.java` has an error where it is trying to mark errors on the **directory** containing the source file rather than the source file itself.

edwardalee:
This fixes #966 . It modifies MANIFEST.MF, which makes me nervous.


org.lflang/META-INF/MANIFEST.MF
```diff
@@ -22,7 +22,8 @@ Require-Bundle: org.eclipse.xtext,
  org.eclipse.lsp4j;bundle-version="0.12.0",
  com.fasterxml.jackson.core.jackson-core,
  com.fasterxml.jackson.core.jackson-annotations,
- com.fasterxml.jackson.core.jackson-databind
+ com.fasterxml.jackson.core.jackson-databind,
+ org.eclipse.core.runtime;bundle-version="3.23.0"
```
lhstrh, line 26:
```suggestion
 org.eclipse.core.runtime;bundle-version="3.24.0"
```
Adding this dependency should be fine. Our Gradle/Maven configurations, however, use `3.24.0` already (see https://github.com/lf-lang/lingua-franca/blob/master/gradle.properties), so we should match it.

lhstrh (commented):
I'm surprised to see that it has been reporting errors on the path (not the file). I think @petervdonovan should review. Adding the dependency is fine, but we should use the right version (see suggestion).

petervdonovan:
> I'm surprised to see that it has been reporting error on the path (not the file).

It looks like this change happened during refactoring only 2 weeks ago (commit b57c4021), so maybe it is not that surprising that the problem was only found recently.

What I do find strange is that, directory/file confusion aside, I am not sure that the idea behind b57c4021 was correct. Before b57c4021, we reported errors on the resource that had just been validated. Now, the error is reported on the root file of the code generation process.

lhstrh:
> > I'm surprised to see that it has been reporting error on the path (not the file).
> 
> It looks like this change happened during refactoring only 2 weeks ago (commit [b57c402](https://github.com/lf-lang/lingua-franca/commit/b57c40213d412b47ac6b78b979a7f743e50da01b)), so maybe it is not that surprising that the problem was only found recently.
> 
> What I do find strange is that, directory/file confusion aside, I am not sure that the idea behind the change introduced in [b57c402](https://github.com/lf-lang/lingua-franca/commit/b57c40213d412b47ac6b78b979a7f743e50da01b) was correct. Before [b57c402](https://github.com/lf-lang/lingua-franca/commit/b57c40213d412b47ac6b78b979a7f743e50da01b), we reported errors on the resource that had just been validated. Now, the error is reported on the root file of the code generation process.

Ah, you're right -- very good detective work! I recall [b57c402](https://github.com/lf-lang/lingua-franca/commit/b57c40213d412b47ac6b78b979a7f743e50da01b). Looks like we should just revert the changes on line 231. We should probably answer the `FIXME` that was in that code, however. If the code is correct, we should probably put it in a static method of `FileConfig` for convenient reuse.

edwardalee:
> I'm surprised to see that it has been reporting errors on the path (not the file). I think @petervdonovan should review. Adding the dependency is fine, but we should use the right version (see suggestion).

I committed your suggestion, and it cause CI to fail with this error:
```
Error:  Cannot resolve project dependencies:
[97](https://github.com/lf-lang/lingua-franca/runs/5218982626?check_suite_focus=true#step:8:97)
Error:    Software being installed: org.lflang 0.1.0.qualifier
[98](https://github.com/lf-lang/lingua-franca/runs/5218982626?check_suite_focus=true#step:8:98)
Error:    Missing requirement: org.lflang 0.1.0.qualifier requires 'osgi.bundle; org.eclipse.core.runtime 3.24.0' but it could not be found
[99](https://github.com/lf-lang/lingua-franca/runs/5218982626?check_suite_focus=true#step:8:99)
```
So we should revert this commit.

lhstrh:
Let me downgrade the dependency in Gradle/Maven and see if that works...

lhstrh (approved):
Looks good to me, but I haven't had a chance to check it with Epoch, so please check with the latest changes before merging.

lhstrh:
@a-sr: this PR is now in a state where there is a remaining problem that we don't seem able to fix. 

If an error marker is added through validation triggered by the Compile button, then the marker will not be removed if the error is fixed (even if the document is saved) until the Compile action is invoked again. It appears that validator instances only remove markers that they have added, leaving markers added by others untouched. I attempted removing the markers explicitly in `LFValidator.checkModel()` but the available `eResource` is apparently not an `iResource` and can therefore not be cast to an `iFile` (which has the `deleteMarkers()` method). Any help with this would be greatly appreciated!

cmnrd:
The EclipseErrorReporter has a method [`clearMarkers`](https://github.com/lf-lang/lingua-franca/blob/14e26feddb2e2770596c22b2e9a2397c53d468dd/org.lflang/src/org/lflang/generator/EclipseErrorReporter.java#L236) which should clear all the error markers.

As far as I know, this method is called in `doGenerate` of `GeneratorBase`. Probably it should be called instead by some hook that is invoked on file save.

PR #968, opened 2022-02-16
Fix port type mismatch in AnytimePrime.lf

hokeun:
Match the type of Print's in port to Prime's out port.


edwardalee (approved):
Ah yes, looks right to me.

hokeun:
> Ah yes, looks right to me.

Thanks for the approval!

PR #970, opened 2022-02-17
Simplify conditional fed codegen

edwardalee:
This PR is a small step towards improving the C codegen structure. It includes the following:
* Disallow ports in top-level reactors and remove such ports from one test that had them.
* Merge nearly identical methods in CGenerator into one consolidated method.
* Homogenize the way that the generator determines which components are included in a federate.
* A few other small fixes.

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
@@ -484,6 +484,45 @@ public class LinguaFrancaValidationTest {
 //             reactor Foo {
 //                 input in:int;
 //             }
+//             reactor Bar {
+//                 input in:int;
+//                 x1 = new Foo();
+//                 x2 = new Foo();
+//                 in -> x1.in;
+//                 reaction(startup) -> x2.in {=
+//                 =}
+//             }
+//         """
+// Java 11:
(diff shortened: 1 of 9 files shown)


Soroosh129:
It looks like [a few](https://github.com/lf-lang/lingua-franca/blob/f62128963c2c620cf61c9d54fbdf791781c1ea34/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java#L497) validation tests have ports in the main reactor.

Soroosh129 (approved):
This PR makes the CGenerater code a lot cleaner!
I think we should perhaps wait for @housengw to see if these changes would interfere with his work in #958.

One note: It looks like I had missed the addition of `currentFederate` as a class variable in a previous PR. I think relying on `currentFederate` could make it harder to break the CGenerator into multiple, smaller generators (like the CppGenerator).

housengw:
the code changes that overlap with #958 are minor. Is there a reason behind using `currentFederate` instead of passing a `federate` parameter to the code generating functions?

edwardalee:
> the code changes that overlap with #958 are minor. Is there a reason behind using `currentFederate` instead of passing a `federate` parameter to the code generating functions?

Every function needs it. We could pass it as a parameter, but it gets pretty tedious.

housengw (approved):
this PR looks good. We just need to keep track of where `currentFederate` can be modified..

PR #971, opened 2022-02-17
Add reactor-cpp as a submodule and reorganize the C++ build process

cmnrd:
This PR brings the following changes:
1. Make reactor-cpp a submodule of LF.
2. Overhaul the C++ build process (This has become necessary to support 1. while still allowing mixing of multiple reactor-cpp versions)
3. Make C++ code generation smarter by only overwriting files if they actually change. This enables cmake to properly detect that it does not need to rebuild a file and significantly speeds up compilation times after making changes to an LF file.

## The new C++ build process

Consider this LF project structure:
```
src/
    A.lf
    B.lf
```
Where `A.lf` and `B.lf` both contain a main reactor.

Previously, compiling `A.lf` would create a completely isolated cmake project in `src-gen/A/`. With this PR, this structure is created instead:
```
src-gen/
    A/
        CMakeLists.txt
        <source files>
    CMakeLists.txt
    reactor-cpp-lfbuiltin
```
The internals of `src-gen/A/` are similar to before. However, we get a new top-level CMakeLists.txt that creates a top-level cmake project for the entire LF project. `A/` is automatically included as a subproject. The directory `reactor-cpp-lfbuiltin` is automatically copied from the `reactor-cpp` submodule. The `-lfbuiltin` suffix indicates that it is the built-in reactor-cpp version of LF.

Let's assume we now compile `B.lf` and it sets this target property: `runtime-version: foo`. This will modify the src-gen directory as follows:
```
src-gen/
    A/
        CMakeLists.txt
        <source files>
    B/
        CMakeLists.txt
        <source files>
    CMakeLists.txt
    reactor-cpp-foo
    reactor-cpp-lfbuiltin
```
The `reactor-cpp-foo` directory will be automatically fetched via git. The sub-project `B` is configured to depend on `reactor-cpp-foo`, while `A` still depends on `reactor-cpp-lfbuiltin`.

Overall, this change makes the build-process a lot more robust. CMake is aware of all the files within a project and knows exatly what needs to be compiled when.

I am not sure if these changes might interfere with the LSP linting. @petervdonovan could you check?


cmnrd:
While working on this, I noticed that many methods in `FileConfig` should better be factored out in a `FileUtils` class, as they only provide file system access without depending on any configuration. I will make this change in a separate PR as to avoid hiding the changes made in this PR.

petervdonovan:
This sounds like a big improvement! Cloning the Git repository has definitely been a slow point in the language server. From your description it sounds like it probably will break `CppValidator` since the `CppValidator` depends on variables in `CMakeCache.txt`. I will take a look as soon as I can.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -439,32 +504,90 @@ public static void copyFile(Path source, Path destination)  throws IOException {
...
+     *
+     *  This also creates new directories for any directories on the destination
+     *  path that do not yet exist.
+     *
+     *  @param source The source directory as a path relative to the classpath.
+     *  @param destination The file system path that the source directory is copied to.
+     *  @param skipIfUnchanged If true, don't overwrite the file if its content would not be changed
+     *  @throws IOException If the given source cannot be copied.
+     */
+    public void copyDirectoryFromClassPath(final String source, final Path destination, final boolean skipIfUnchanged) throws IOException {
```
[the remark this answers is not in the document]
cmnrd: Probably not, I still have to test things in the IDE.

petervdonovan (commented):
With the changes I just pushed, the language server seems to work -- I checked by merging this into the `lsp` branch locally. LSP tests also pass for me locally, so hopefully there are no more blockers on that front.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -422,13 +428,72 @@ public static void copyDirectory(Path src, Path dest) throws IOException {
...
+     * path that do not yet exist.
+     *
      * @param source The source file path.
      * @param destination The destination file path.
+     * @param skipIfUnchanged If true, don't overwrite the destination file if its content would not be changed
      * @throws IOException if copy fails.
      */
-    public static void copyFile(Path source, Path destination)  throws IOException {
+    public static void copyFile(Path source, Path destination, boolean skipIfUnchanged)  throws IOException {
+        Files.createDirectories(destination);
```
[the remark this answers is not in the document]
cmnrd: Avoiding the code duplication is a good idea.

Could you clarify what you mean by "the subtle difference in how the destination argument is handled"? It should be the same across the two methods, but maybe I miss something. Or do you mean the effects of the `skipIfUnchanged` parameter?

cmnrd:
Thanks for your fixes! I will clean things up then and get it to work in Epoch.

org.lflang/src/org/lflang/FileConfig.java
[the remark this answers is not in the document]
petervdonovan: Here, it says, `Files.createDirectories(destination)`, but below, it says, `Files.createDirectories(destination.getParent())`. On second thought, it looks like it does not matter, but I just thought it odd.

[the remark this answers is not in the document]
cmnrd: Oh, interesting, that was not on purpose. I think it should be `Files.createDirectories(destination.getParent())` in both cases.

cmnrd:
As far as I am concerned, this is ready to be merged. The copy mechanism now also works from Epoch and was (manually) tested. It would be great if we could include this PR in the 0.1.0 release. I would only need an approval ;)

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -439,32 +507,115 @@ public static void copyFile(Path source, Path destination)  throws IOException {
...
-                    "A required target resource could not be found: " + source + "\n" +
-                        "Perhaps a git submodule is missing or not up to date.\n" +
-                        "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.\n"
-                        +
-                        "Also try to refresh and clean the project explorer if working from eclipse.");
+        if (sourceStream == null) {
+            throw new IOException(
+                "A required target resource could not be found: " + source + "\n" +
+                    "Perhaps a git submodule is missing or not up to date.\n" +
+                    "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.\n"
```
lhstrh, line 521:
```suggestion
                    "To initialize, run `git submodule init`; to update, run `git submodule update`.\n"
```

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -439,32 +507,115 @@ public static void copyFile(Path source, Path destination)  throws IOException {
...
-                throw new IOException(
-                    "A required target resource could not be found: " + source + "\n" +
-                        "Perhaps a git submodule is missing or not up to date.\n" +
-                        "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.\n"
-                        +
-                        "Also try to refresh and clean the project explorer if working from eclipse.");
+        if (sourceStream == null) {
+            throw new IOException(
+                "A required target resource could not be found: " + source + "\n" +
+                    "Perhaps a git submodule is missing or not up to date.\n" +
```
lhstrh, line 520:
```suggestion
                    "Perhaps a git submodule is not initialized and/or not up to date.\n" +
```

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -439,32 +507,115 @@ public static void copyFile(Path source, Path destination)  throws IOException {
...
+                    "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.\n"
+                    +
+                    "Also try to refresh and clean the project explorer if working from eclipse.");
```
lhstrh, line 523:
```suggestion
                    "You may also need to refresh and clean your build system or IDE.");
```

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -439,32 +507,115 @@ public static void copyFile(Path source, Path destination)  throws IOException {
...
+    public void copyDirectoryFromClassPath(final String source, final Path destination, final boolean skipIfUnchanged) throws IOException {
+        final URL resource = getClass().getResource(source);
+        if (resource == null) {
             throw new IOException(
-                "A required target resource could not be copied: " + source + "\n" +
+                "A required target resource could not be found: " + source + "\n" +
                     "Perhaps a git submodule is missing or not up to date.\n" +
-                    "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.",
-                ex);
+                    "See https://github.com/icyphy/lingua-franca/wiki/downloading-and-building#clone-the-lingua-franca-repository.\n"
```
lhstrh, line 564: Note to self: remove duplicate string.

org.lflang/src/org/lflang/FileConfig.java
petervdonovan: (We might want `git submodule update --init` with `--` before the "init.")

lhstrh: I agree that in a situation like this `git submodule update --init` might be a preferable shorthand. I'll adjust it.

lhstrh (approved):
Look very good! Would the directory name `reactor-cpp-default` maybe be better than `reactor-cpp-lfbuiltin`? The latter looks a bit odd and difficult to read, IMO.

cmnrd:
Yes, probably default is better

PR #972, opened 2022-02-18
Matrix benchmark fix

jhaye:
This properly adds up the intermediate results when calculating each matrix cell.

This increases execution runtime from ~50ms to ~80ms for the single-threaded case, but it's still an order of magnitude faster than C/C++ with ~800ms.

Using the validation function, this benchmark calculates reproducible results (i.e. no data race like for C++), but the results are still wrong. Does anyone know of a benchmark implementation that calculates results correctly?


cmnrd (approved):
Thanks!

PR #977, opened 2022-02-24
Attributes

lsk567:
This is a draft pull request tracking the progress of an annotation system in LF.

This PR extends the syntax to support "attributes," which has the form `@<attrName>(<paramName1> = <paramValue1>, <paramName2> = <paramValue2>, ...)`. `<paramName>` is optional, though @oowekyala pointed out that it would be good to have named parameters to conform with Java annotation grammar.

As a proof of concept, the label mechanism is implemented using an `@label` attribute.

Example: Label.lf
```
target C {
    threads: 1,
    keepalive: true
};

@label(name="This is a controller.")
reactor Controller {
    
    output out:bool;
    
    @label("startup reaction")
    reaction(startup) -> out {==}
}

@label("This is a vision system.")
reactor Vision {
    @label("A timer")
    timer t;
    
    input _in:bool;
    output out:bool;
    state ramp_exists:bool(false);
    
    @label(name="Action") 
    logical action act;
    
    @label(name="Processing visual input")
    reaction(_in) -> out, act {==}
}

reactor Door {
    input _in:bool;
    state door:bool;
    
    reaction(_in) {==}
}

@label("Top-level reactor")
main reactor {
    
    controller  = new Controller();
    vision      = new Vision();
    door        = new Door();
    
    controller.out -> vision._in;
    vision.out -> door._in;
}
```
label

TODOs:
- [ ] Address a `Duplicate AttrParm 'name'` issue.
- [ ] Support a clean way to incorporate DSLs, such as logic, into attributes based on the attribute name.

### DIFF
--- a/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.xtend
+++ b/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.xtend
@@ -92,7 +92,11 @@ import org.lflang.generator.ReactorInstance
 import org.lflang.generator.TimerInstance
 import org.lflang.generator.TriggerInstance
 import org.lflang.generator.TriggerInstance.BuiltinTriggerVariable
+import org.lflang.lf.Action
 import org.lflang.lf.Model
+import org.lflang.lf.Reaction
+import org.lflang.lf.Reactor
+import org.lflang.lf.Timer
 
 import static extension org.eclipse.emf.ecore.util.EcoreUtil.*
 import static extension org.lflang.ASTUtils.*
@@ -1078,8 +1082,31 @@ class LinguaFrancaSynthesis extends AbstractDiagramSynthesis<Model> {
 	
 	private def Iterable<KNode> createUserComments(EObject element, KNode targetNode) {
 		if (SHOW_USER_LABELS.booleanValue) {
-			val commentText = ASTUtils.findAnnotationInComments(element, "@label")
-			
+			var String commentText
+			// The list of elements where labels could be placed.
+			switch element {
+				Reaction : {
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+				Reactor : {
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+				Action : {
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+				Timer : {
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+			}
+
 			if (!commentText.nullOrEmpty) {
 				val comment = createNode()
 		        comment.setLayoutOption(CoreOptions.COMMENT_BOX, true)
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -85,7 +85,7 @@ QualifiedNameWithWildcard:
  * Declaration of a reactor class.
  */
 Reactor:
-    {Reactor} ((federated?='federated' | main?='main')? & realtime?='realtime'?) 'reactor' (name=ID)? 
+    {Reactor} (attributes+=Attribute)* ((federated?='federated' | main?='main')? & realtime?='realtime'?) 'reactor' (name=ID)? 
     ('<' typeParms+=TypeParm (',' typeParms+=TypeParm)* '>')?
     ('(' parameters+=Parameter (',' parameters+=Parameter)* ')')?
     ('at' host=Host)?
@@ -138,7 +138,7 @@ TargetDecl:
  *  - if the `time` type is specified, either a proper time interval and unit
  *  must be given, or a literal or code that denotes zero.
  */
- StateVar:
+StateVar:
     'state' name=ID (
         (':' (type=Type))? 
         ((parens+='(' (init+=Value (','  init+=Value)*)? parens+=')')
@@ -170,7 +170,7 @@ Output:
 // E.g. (0) or (NOW) or (NOW, ONCE) or (100, 1000)
 // The latter means fire with period 1000, offset 100.
 Timer:
-    'timer' name=ID ('(' offset=Value (',' period=Value)? ')')? ';'?;
+    (attributes+=Attribute)* 'timer' name=ID ('(' offset=Value (',' period=Value)? ')')? ';'?;
 
 Boolean:
     TRUE | FALSE
@@ -187,6 +187,7 @@ Boolean:
 // For all actions, minSpacing is the minimum difference between
 // the tags of two subsequently scheduled events.
 Action:
+    (attributes+=Attribute)*
     (origin=ActionOrigin)? 'action' name=ID 
     ('(' minDelay=Value (',' minSpacing=Value (',' policy=STRING)? )? ')')?
     (':' type=Type)? ';'?;
@@ -255,6 +256,9 @@ Attribute:
 ;
 
 AttrParm:
+    (name=ID '=')? value=AttrParmValue;
+
+AttrParmValue:
     STRING
 ;
 
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -192,6 +192,7 @@ Action:
     (':' type=Type)? ';'?;
 
 Reaction:
+    (attributes+=Attribute)*
     ('reaction')
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')?
     (sources+=VarRef (',' sources+=VarRef)*)?
@@ -246,6 +247,15 @@ Delay:
 // Chooses the serializer to use for the connection
 Serializer:
     'serializer' type=STRING
+;
+
+/////////// Attributes
+Attribute:
+    '@' attrName=AttrName ('(' attrParms+=AttrParm (',' attrParms+=AttrParm)* ')')?
+;
+
+AttrParm:
+    STRING
 ;
 
 /////////// For target parameters
@@ -487,6 +497,9 @@ enum ActionOrigin:
 
 enum Visibility:
     NONE | PRIVATE='private' | PUBLIC='public';
+
+enum AttrName:
+    LABEL='label';
 
 // Note: time units are not keywords, otherwise it would reserve
 // a lot of useful identifiers (like 's' or 'd').
--- a/org.lflang/src/org/lflang/validation/LFValidator.java
+++ b/org.lflang/src/org/lflang/validation/LFValidator.java
@@ -56,6 +56,8 @@ import org.lflang.TimeValue;
 import org.lflang.lf.Action;
 import org.lflang.lf.ActionOrigin;
 import org.lflang.lf.Assignment;
+import org.lflang.lf.Attribute;
+import org.lflang.lf.AttrParm;
 import org.lflang.lf.Connection;
 import org.lflang.lf.Deadline;
 import org.lflang.lf.Host;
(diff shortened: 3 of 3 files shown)


org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.xtend
```diff
@@ -1078,8 +1082,31 @@ class LinguaFrancaSynthesis extends AbstractDiagramSynthesis<Model> {
...
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+				Timer : {
+					var list = element.getAttributes.filter[ it.attrName.toString === 'label' ]
+					if (list.length != 0)
+						commentText = list.get(0).getAttrParms.get(0).getValue.replaceAll("^\"|\"$", "")
+				}
+			}
```
lsk567, line 1108: Not sure how to resolve this code duplication... Xtend's support for polymorphism seems insufficient to avoid it.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -488,6 +502,9 @@ enum ActionOrigin:
 enum Visibility:
     NONE | PRIVATE='private' | PUBLIC='public';
 
+enum AttrName:
+    LABEL='label';
```
lsk567, line 506: Make this `ID` and hardcode in the validator.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -248,6 +250,18 @@ Serializer:
     'serializer' type=STRING
 ;
 
+/////////// Attributes
+Attribute:
+    '@' attrName=AttrName ('(' attrParms+=AttrParm (',' attrParms+=AttrParm)* ')')?
```
lsk567, line 255: Trailing commas?

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -248,6 +250,18 @@ Serializer:
...
+;
+
+AttrParm:
+    (name=ID '=')? value=AttrParmValue;
```
lsk567, line 259: Attributes must be named if there are multiple. Check in the validator.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1442,6 +1444,82 @@ public void checkVarRef(VarRef varRef) {
         }
     }
 
+    @Check(CheckType.FAST)
+    public void checkReactionAttributes(Reaction reaction) {
```
lsk567, line 1448: Use `Attribute` as the parameter.

org.lflang/src/org/lflang/LinguaFranca.xtext
oowekyala: done

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -55,6 +57,7 @@
 import org.eclipse.xtext.validation.Check;
 import org.eclipse.xtext.validation.CheckType;
 import org.eclipse.xtext.validation.ValidationMessageAcceptor;
+import org.jetbrains.annotations.NotNull;
```
lhstrh, line 60: I don't think we want to introduce this dependency. It will break things for Eclipse users.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1199,6 +1206,122 @@ public void checkVarRef(VarRef varRef) {
         }
     }
 
+
+    /**
+     * Specification of the structure of an annotation.
+     */
+    static class AttributeSpec {
```
lhstrh, line 1213: This seems like a good candidate to factor out into its own file.

lhstrh: Everything until line 1296 could be moved there.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1199,6 +1206,122 @@ public void checkVarRef(VarRef varRef) {
...
+
+            paramSpecByName = params.stream().collect(Collectors.toMap(it -> it.name, it -> it));
+
+        }
+
+        /**
+         * Check that the attribute conforms to this spec. The
+         * attr has the correct name.
+         */
+        public void check(LFValidator validator, Attribute attr) {
```
lhstrh, line 1229: Javadoc is missing for these new methods.

lhstrh (changes requested):
Looks good! There is some code in the validator that still needs some attention.

lhstrh:
Shall we punt on the DSLs in the interest of getting this merged in soon? They should be a proper extension, anyway, so this can be addressed at a later time.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -1747,16 +1748,73 @@ public static String findAnnotationInComments(EObject object, String key) {
...
+    }
+
     /**
      * Search for an `@label` annotation for a given reaction.
-     * 
+     *
      * @param n the reaction for which the label should be searched
      * @return The annotated string if an `@label` annotation was found. `null` otherwise.
      */
     public static String label(Reaction n) {
```
oowekyala, line 1800: note that this overload could be removed if binary compatibility is not important, which seems likely.

lhstrh: Sorry, I didn't get this comment. What binary compatibility?

oowekyala: If someone has compiled a class that depends on this method and we remove it, their class will break. Minor releases generally preserve binary compatibility. I'm not sure whether this java API is publicly supported since you've started public releases... If the only ide extensions that use the java API are in this repo then we can most likely remove it (we're also pre-1.0.0)

org.lflang/src/org/lflang/validation/LFValidator.java
lsk567: Factoring out seems a bit tricky, since `check()` (L1229) calls `error()`, which is a protected method inherited from `AbstractDeclarativeValidator`. We could move the check functions back to the validator class, but that would sort of invalidate the refactoring because most of the code is still in the validator class.

org.lflang/src/org/lflang/validation/LFValidator.java
lsk567: Javadoc added in the new commits.

org.lflang/src/org/lflang/validation/LFValidator.java
lsk567: I build the eclipse product, and it looks like Epoch isn't broken by this dependency.

lhstrh: :point_up:

lsk567: But it also seems like this can be safely removed.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1688,11 +1713,162 @@ private boolean sameType(Type type1, Type type2) {
...
+
+    /** A map from a string to a supported AttributeSpec */
+    private static final Map<String, AttributeSpec> ATTRIBUTE_SPECS_BY_NAME = new HashMap<>();
+
+    //////////////////////////////////////////////////////////////
+    //// Inner classes.
+    /**
+     * Specification of the structure of an annotation.
+     */
+    static class AttributeSpec {
```
lhstrh, line 1732: I think I raised this point before, but why aren't we putting this class in its own file?

lhstrh (changes requested):

org.lflang/src/org/lflang/validation/LFValidator.java
lhstrh: How did you build Epoch? In Eclipse?

org.lflang/src/org/lflang/validation/LFValidator.java
lhstrh: Could pass in a lambda?

org.lflang/src/org/lflang/validation/LFValidator.java
lhstrh: Just now read your reply https://github.com/lf-lang/lingua-franca/pull/977#discussion_r902077973 (let's continue there and resolve this comment.

lhstrh:
Also note that there still is an unresolved conflict in `LFValidator.java`.

org.lflang/src/org/lflang/validation/LFValidator.java
lhstrh: The point is not that it would break Epoch, it would prevent developers from building Epoch _in Eclipse_. This is because Eclipse doesn't use Maven or Gradle but it's own build system that is severely deprived from otherwise generally accessible dependencies like this one. We've run into this issue several times before...

org.lflang/src/org/lflang/validation/LFValidator.java
oowekyala: We can override the error function and just call super, this will make it visible

org.lflang/src/org/lflang/validation/LFValidator.java
lsk567: Okay, got it...

org.lflang/src/org/lflang/validation/LFValidator.java
lsk567: Works like a magic. Thanks for the great advice.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1394,31 +1418,42 @@ public void checkStateResetWithoutInitialValue(StateVar state) {
...
     /**
      * Return a list of error messages for the target declaration.
      */
     public List<String> getTargetPropertyErrors() {
         return this.targetPropertyErrors;
     }
 
+    /**
+     * Generate an error message for an AST node.
+     * Overriding the parent method here to make the method public,
```
lhstrh, line 1445:
```suggestion
```
This type of info doesn't really belong in Javadoc because it doesn't help a programmer decide whether (or what for) to use the method.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1394,31 +1418,42 @@ public void checkStateResetWithoutInitialValue(StateVar state) {
...
+     * so that it can be called from the AttributeSpec class.
```
lhstrh, line 1446:
```suggestion
```
Methods have no designated call site...

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1698,11 +1733,27 @@ private boolean sameType(Type type1, Type type2) {
...
+
     private static List<String> SPACING_VIOLATION_POLICIES = List.of("defer", "drop", "replace");
-    
+
     private static String UNDERSCORE_MESSAGE = "Names of objects (inputs, outputs, actions, timers, parameters, "
             + "state, reactor definitions, and reactor instantiation) may not start with \"__\": ";
-    
+
     private static String USERNAME_REGEX = "^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\\$)$";
+
```
lhstrh, line 1743: Lines `1743`:`1758` belong in the `AttributeSpec` class.

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * The type of attribute parameters currently supported
+     */
+    enum AttrParamType {
+        STRING,
+        INT,
+        BOOLEAN,
+        FLOAT
+    }
+}
```
lhstrh, line 171:
```suggestion
}

```

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.lflang.lf.AttrParm;
+import org.lflang.lf.Attribute;
+import org.lflang.lf.LfPackage.Literals;
+
+/**
+ * Specification of the structure of an annotation.
+ */
```
lhstrh, line 41: Author tags missing.

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+            // Check whether a parameter conforms to its spec.
+            parmSpec.check(validator, parm);
+            seen.add(parm.getName());
+        }
+        return seen;
+    }
+
+    /**
+     * The specification of the attribute parameter
```
lhstrh, line 117:
```suggestion
     * The specification of the attribute parameter.
```

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -0,0 +1,171 @@
...
+                                        Literals.ATTRIBUTE__ATTR_NAME);
+                    }
+                    break;
+            }
+            
+        }
+    }
+
+    /**
+     * The type of attribute parameters currently supported
```
lhstrh, line 163:
```suggestion
     * The type of attribute parameters currently supported.
```

lhstrh (changes requested):
Much better! Left some more comments/requests.

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1394,31 +1418,42 @@ public void checkStateResetWithoutInitialValue(StateVar state) {
...
         return this.targetPropertyErrors;
     }
 
+    /**
+     * Generate an error message for an AST node.
+     * Overriding the parent method here to make the method public,
+     * so that it can be called from the AttributeSpec class.
+     */
+    @Override
+    public void error(java.lang.String message,
```
oowekyala, line 1447: Technically you can also leave it protected. Protected visibility is subclasses + current package, so it would be visible in AttributeSpec, which is in this package. This is a detail though

lsk567:
The current Attribute syntax generates 2 warnings from Antlr.
```
warning(200): //home/runner/work/lingua-franca/lingua-franca/org.lflang/../org.lflang/srcgen/org/lflang/parser/antlr/internal/InternalLF.g:5661:3: Decision can match input such as "RULE_ID '-' RULE_ID '@' {RULE_ID, 'physical'}" using multiple alternatives: 1, 2
As a result, alternative(s) 2 were disabled for that input

warning(200): //home/runner/work/lingua-franca/lingua-franca/org.lflang/../org.lflang.ide/srcgen/org/lflang/ide/contentassist/antlr/internal/InternalLF.g:13820:2: Decision can match input such as "RULE_ID '-' RULE_ID '@' {RULE_ID, 'physical'}" using multiple alternatives: 1, 2
As a result, alternative(s) 2 were disabled for that input
```

It seems that the leading `@` sign in the Attribute syntax is confusing the parser since `@` is also used in IPV4Host, IPV6Host, and NamedHost. Changing `@` to `@@` in the Attribute syntax got rid of the warnings, which confirms the previous observation, but we need a cleaner solution.

Any thoughts on this? @lhstrh @oowekyala

Soroosh129:
There are two possible solutions that comes immediately to my mind:
- Get rid of the `at` notation entirely and use attributes instead if needed
- Change `Host` to just a string and check it it in the validator

Under both solutions, IPV4Host, IPV6Host, and NamedHost are no longer needed.

lhstrh:
> Any thoughts on this? @lhstrh @oowekyala

The ambiguity comes from the "at" clause of an instantiation in combination with the annotations that can be featured inside of a reactor definition (i.e., above declarations of inputs, outputs, reactions, etc.).

Once the parser encounters something like `x = new X() at foo` followed by an `@bar`, its doesn't know whether the `@bar` should be part of the host or, instead, is an annotation of the subsequent declaration (such as a `input y:baz`).

oowekyala:
@lhstrh is right. One solution would be to mandate a semicolon to terminate all declarations. I also like Soroush's idea to change Host back to a string, the productions are really complicated and anyway some checking is still done in the validator:
https://github.com/lf-lang/lingua-franca/blob/b24e5b5a52f8a51b5fd2914f1228b9576f6d06a1/org.lflang/src/org/lflang/LinguaFranca.xtext#L432-L446

Soroosh129:
I personally think that an attribute instead of `at` is the more elegant solution. Where a federate resides I believe _is_ an attribute anyway.

How about something like this:
```
@Host(foo_user@192.168.1.1:1000)
foo = new Foo($
```
and
```
@Host(foo_user@192.168.1.1:1000)
federated reactor {}
```

lhstrh:
I just want to emphasize that we don't need a semicolon on _all_ declarations, just on instantiations that have an at clause. The impact of this change would be minimal.

Soroosh129:
>  just on instantiations that have an at clause.

That would be a bit weird.

> I think we might want to more carefully evaluate the 'at' clause and its use before changing it.

I'm not proposing to change its meaning in this PR. Just how it can be declared. If we convert it to an attribute, then we can of course later still have a discussion about whether to keep it or change what it does.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -1747,16 +1748,73 @@ public static String findAnnotationInComments(EObject object, String key) {
         return null;
     }
 
+    /**
+     * Return the value of the {@code @label} attribute if
+     * present, otherwise return null.
+     *
+     * @throws IllegalArgumentException If the node cannot have attributes
+     */
+    public static String findLabelAttribute(EObject node) {
```
Soroosh129, line 1757: Would it make sense if this was in a separate class like `AttributeUtils`? Same for other stuff here that relate to attributes.

lhstrh:
> I just want to emphasize that we don't need a semicolon on _all_ declarations, just on instantiations that have an at clause. The impact of this change would be minimal.

Concretely, my suggestion for a quick fix to the ambiguity problem is to replace the following line on `LinguaFranca.xtext:233`:

```
    ')' ('at' host=Host)? ';'?;
```

with

```
    ')' (('at' host=Host ';') | ';'?);
```

org.lflang/src/org/lflang/ASTUtils.java
lsk567: Yes, I like this idea.

lsk567:
Looks like we are ready to go?

lhstrh (approved):
Looks good! Thanks for making all the requested changes :-)

PR #978, opened 2022-02-25
FileConfig cleanups

cmnrd:
This PR cleans up FileConfig which currently seems to mix a variety of concerns. Despite some other simplifications and cleanups, it factors out functions that are purely related to file system access to a new static class FileUtils.

~Note that this branch is based on #971 and hence include it's changes. The diff should clear up once #971 is merged.~


edwardalee (approved):
I did not check this thoroughly, but on a spot check, all looked good to me.  It's a big improvement!  Except that it looks like there is a compile error with packaging Epoch...

cmnrd:
As far as I can see, this should be ready to be merged. I also added this PR to the 0.1.0 milestone. Given that there are no major concerns by reviewers, we should be able to merge this for 0.1.0.

org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -554,6 +621,47 @@ static public String triggerRefNested(PortInstance port, String runtimeIndex, St
...
+     * If there is no LF_CLASSPATH environment variable, use CLASSPATH,
+     * if it is defined.
+     * The first file found will be returned.
+     *
+     * @param fileName The file name or relative path + file name
+     * in plain string format
+     * @param directory String representation of the director to search in.
+     * @return A Java file or null if not found
+     */
+    public static Path findFile(String fileName, Path directory) {
```
lhstrh, line 636: Why was this moved to `CUtil`? I don't see what's C-specific about this.

org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -591,13 +699,52 @@ public static void runBuildCommand(
...
     }
 
+
+    /**
+     * Remove files in the bin directory that may have been created.
+     * Call this if a compilation occurs so that files from a previous
+     * version do not accidentally get executed.
+     * @param fileConfig
+     */
+    public static void deleteBinFiles(FileConfig fileConfig) {
```
lhstrh, line 716: This, too, doesn't look C-specific to me.

lhstrh (commented):
I think it's a sensible choice to factor out all the static methods of `FileConfig` into a separate `FileUtil` class. There are some methods that ended up in `CUtil`, however, which seem out of place there.

org.lflang/src/org/lflang/generator/c/CUtil.java
cmnrd: Also this function is only used by the C-generator. Honestly, I don't quite understand what it does, but it seems to make various assumptions as to where files are placed and how they are named. Hence, I figure it will not be easily applicable to other targets. I also don't quite understand why it is needed since we have methods for cleaning up the entire bin dir.

lhstrh: So the function is "quarantined" here with the eventual goal of eliminating it?

cmnrd:
My goal is to make FileConfig and FileUtil classes focus on one specific task each. FileConfig provides details about where certain files should be located and is meant to configure the generators. FileUtil provides generic static methods for file system access (including access to files and directories in jars or bundles). By generic FS access I mean operations like writing a file, deleting a file/directory or copying a file/directory.

The two methods in CUtil don't fit the scope of either of those classes. They are not configuration related, and they don't provide generic filesystem access. Instead, they implement very specific algorithms. These algorithms are intervened with the code generation process (delete bin in fact analysis the resource contents and deletes files where it assumes the generator has placed them, findFile makes assumptions as to where the source files for the compilation are located). Therefore, I argue that the methods should be placed in `GeneratorUtils` or a similar class. However, since they are only used in C and I currently don't see that they would be useful for other targets I placed them in `CUtil`. I might be wrong though and if people feel `GeneratorUtils` would be the better place, I am happy to move it over there.

org.lflang/src/org/lflang/generator/c/CUtil.java
cmnrd: Also see my comment in the main thread. I don't know the C Generator well enough to understand what this method does, why it is needed, and whether it can be eliminated. I only know that the other generators don't use it ;)

lhstrh:
> My goal is to make FileConfig and FileUtil classes focus on one specific task each. FileConfig provides details about where certain files should be located and is meant to configure the generators. FileUtil provides generic static methods for file system access (including access to files and directories in jars or bundles). By generic FS access I mean operations like writing a file, deleting a file/directory or copying a file/directory.

Yes, I'm fully on board with this.
 
> The two methods in CUtil don't fit the scope of either of those classes. They are not configuration related, and they don't provide generic filesystem access. Instead, they implement very specific algorithms. These algorithms are intervened with the code generation process (delete bin in fact analysis the resource contents and deletes files where it assumes the generator has placed them, findFile makes assumptions as to where the source files for the compilation are located). Therefore, I argue that the methods should be placed in `GeneratorUtils` or a similar class. However, since they are only used in C and I currently don't see that they would be useful for other targets I placed them in `CUtil`. I might be wrong though and if people feel `GeneratorUtils` would be the better place, I am happy to move it over there.

I understand. We can leave them in `CUtil` for now. I believe that the `findFile` method implements the policy all code generators should use for locating files that are either imported or listed in the `files` target property. If they don't, we should bring all targets in alignment, and if they do we have a case of code duplication. We might create an issue for this and move on?

lhstrh (approved):

org.lflang/src/org/lflang/generator/c/CUtil.java
edwardalee: This method implements a simple functionality that all targets should implement: If a code generation or compilation fails for a program Foo.lf, then there should be no binary file bin/Foo.  If there is, it is a bug. If the other targets don't implement this, then it is a bug.

PR #982, opened 2022-02-25
Carla intersection refactoring

housengw:
Refactored the intersection example to allow for more effective comparison between the implementation in LF and ROS.

Both implementations now uses the same underlying vehicle, rsu and carla simulator logic.


Soroosh129 (approved):
This refactoring greatly improves our ability to do a proper A/B testing. The changes look great!

PR #984, opened 2022-02-28
Enable generation of ROS2 nodes directly from C++ code

cmnrd:
This PR provides basic support for compilation of ROS2 nodes directly from LF programs in the C++ target. There are still many features missing that will be part of subsequent PRs. Also, it is unclear to me how the new ROS2 specific generator should interact with the LSP implementation. However, I would also address this in a future PR.

With this PR basic compilation works in lfc and Epoch. To validate it, I added a ros2 CI flow which compiles all the C++ tests for the ros2 platform.

~Note that this PR is based on #978 and hence currently includes all changes from this PR in the diff.~


example/Cpp/src/ROS/MinimalPublisher.lf
```diff
@@ -0,0 +1,31 @@
...
+}
+
+public preamble {=
+    #include "rclcpp/rclcpp.hpp"
+    #include "std_msgs/msg/string.hpp"
+=}
+
+main reactor {
+    private preamble {=
+        // FIXME: forward declaration to make the node visible
```
edwardalee, line 12: Is this really necessary? Can't you generate this declaration?

edwardalee (approved):
This looks pretty cool to me. It seems to offer a nice simple way to make an LF program into a ROS node.

example/Cpp/src/ROS/MinimalPublisher.lf
```diff
@@ -0,0 +1,31 @@
...
+
+    reaction(startup) {=
+        publisher = lf_node->create_publisher<std_msgs::msg::String>("topic", 10);
+    =}
+
+    reaction(t) {=
+        auto message = std_msgs::msg::String();
+        message.data = "Hello, world! " + std::to_string(count++);
+        reactor::log::Info() << "Publishing: " << message.data;
+        publisher->publish(message);
```
Soroosh129, line 29: I don't think the default rclcpp executor is thread-safe, so calling publish for the same topic from multiple LF worker threads will be problematic without acquiring a mutex.

org.lflang/src/lib/cpp/lf_timeout.hh
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#pragma once
+
+#include "reactor-cpp/reactor-cpp.hh"
+
+class __lf_Timeout : public reactor::Reactor {
+ private:
+  reactor::Timer timer;
+  reactor::Reaction r_timer{"r_timer", 1, this, [this]() { environment()->sync_shutdown();} };
```
Soroosh129, line 32: How is timeout coordinated with rclcpp's [shutdown](https://docs.ros2.org/bouncy/api/rclcpp/namespacerclcpp.html#a529038885d9d1dbecc5f48244123035e)?

Soroosh129 (approved):
Looks great!

I agree with @edwardalee that this provide a simple way of turning an LF program into a ROS node.

I left a few comments based on my knowledge of ROS 2.

example/Cpp/src/ROS/MinimalPublisher.lf
cmnrd: Unfortunately, thread-safety is not discussed in the documentation. According to [this post](https://answers.ros.org/question/280108/ros2rclcpp-threadsafety-of-nodecreate_publisher/), however, these operations should be thread-safe.

org.lflang/src/lib/cpp/lf_timeout.hh
cmnrd: The node invokes rclcpp's shutdown after the reactor execution terminated (i.e. after all shutdown reactions were processed and the main scheduler thread terminates).

org.lflang/src/org/lflang/generator/cpp/CppRos2PackageGenerator.kt
```diff
@@ -0,0 +1,113 @@
...
+                |
+            ${" |"..(includeFiles?.joinToString("\n") { "include(\"$it\")" } ?: "")}
+            """.trimMargin()
+        }
+    }
+
+    fun generateBinScript(): String {
+        val relPath = fileConfig.binPath.relativize(fileConfig.outPath).toUnixString()
+
+        return """
```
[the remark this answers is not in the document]
cmnrd: Thanks for the pointer! I was vaguely aware of launch files, but did not think of them when creating the script.

example/Cpp/src/ROS/MinimalSubscriber.lf
```diff
@@ -0,0 +1,32 @@
...
+
+    state subscription: {= rclcpp::Subscription<std_msgs::msg::String>::SharedPtr =}
+    state count: unsigned(0)
+
+    physical action message: std::string;
+
+    reaction(startup) -> message {=
+        subscription = lf_node->create_subscription<std_msgs::msg::String>(
+            "topic", 10, [&message](const std_msgs::msg::String::SharedPtr msg) { message.schedule(msg->data); } );
+        // FIXME: Why can't we use a reference type in the lambda argument?
```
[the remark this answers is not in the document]
cmnrd: I guess so, but I am not sure. I tried to use `const std_msgs::msg::String::SharedPtr& msg` in the signature, but that fails with a lot of weird C++ errors. I just realized after checking the documentation again that galactic uses `const std_msgs::msg::String& msg` in the signature and foxy uses `const std_msgs::msg::String::SharedPtr msg`. So I guess  `const std_msgs::msg::String::SharedPtr& msg` is not supported, but `const std_msgs::msg::String& msg` is.

org.lflang/src/org/lflang/generator/cpp/CppRos2PackageGenerator.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    @Suppress("LocalVariableName") // allows us to use capital S as variable name below
+    private val S = '$' // a little trick to escape the dollar sign with $S
+
+    fun generatePackageXml(): String {
+        return """
+            |<?xml version="1.0"?>
+            |<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+            |<package format="3">
+            |  <name>${fileConfig.name}</name>
```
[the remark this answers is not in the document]
cmnrd: Right, I don't think automagically converting the name is a good idea. My current idea is to add a `package-name` property to the `ros2` section. There we could also add other details such as the license and maintainer information required in the `package.xml`.

org.lflang/src/org/lflang/generator/cpp/CppRos2PackageGenerator.kt
```diff
@@ -0,0 +1,113 @@
...
+            |  <maintainer email="todo@todo.com">Todo</maintainer>
+            |  <license>Todo</license>
+            |
+            |  <buildtool_depend>ament_cmake</buildtool_depend>
+            |  <buildtool_depend>ament_cmake_auto</buildtool_depend>
+            |  
+            |  <depend>rclcpp</depend>
+            |  <depend>rclcpp_components</depend>
+            |  <depend>std_msgs</depend>
+            |  <depend>$reactorCppName</depend>
```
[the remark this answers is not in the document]
cmnrd: Not yet, but it is planned and I would also add this to the `ros2` property.

example/Cpp/src/ROS/MinimalPublisher.lf
cmnrd: This is just a temporary workaround until I found a better solution. I am trying to keep the code for generating reactor classes agnostic of the target platform, as it keeps the code generator clean. However, there needs to be some way of extending the generated code. I will fix this mechanism as soon as I have found a satisfactory solution.

.github/workflows/cpp-ros2-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+          ref: ${{ inputs.compiler-ref }}
+      - name: Check out specific ref of reactor-CPO
+        uses: actions/checkout@v2
+        with:
+          repository: lf-lang/reactor-cpp
+          path: org.lflang/src/lib/cpp/reactor-cpp
+          ref: ${{ inputs.runtime-ref }}
+        if: ${{ inputs.runtime-ref }}
+      - name: Setup ROS2
+        uses: ros-tooling/setup-ros@0.2.2
```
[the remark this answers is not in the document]
cmnrd: We could do something similar as in the benchmark tests, by adding a target parameter and making the jobs conditional. Feel free to add the C tests (but probably in another PR).

cmnrd:
Thanks for your suggestions! I consider this PR just as the start and there will be more PRs in the future to extend functionality and fix things where needed.

Issue #887, opened 2022-01-26
Docker generation does not support the `files` target property


Soroosh129:
Currently, the `files` target property is ignored when a `.Dockerfile` is generated.

As a result, a program such as this would not work:
```Python
target Python {files: ../../include/hello.py, timeout: 2 secs, docker: true};

import Source, Print from "../StructAsType.lf"

preamble {=
    import hello
=}

federated reactor at rti {
    s = new Source();
    p = new Print();
    s.out -> p._in;
}
```

This issue affects C, Python, and TypeScript targets.

PR #987, opened 2022-02-28
Docker files property

housengw:
This PR adds tests to verify that the docker file generation works with "files" target property. 
see #887.

The Python target copies everything in `src-gen` to the docker container, so the above functionality works by default.
The C target needs a little tweak since docker does not allow copying of files outside of the build context into the container. I have implemented a fix.
The Typescript target does not seem to use the "files" target property at all during code generation. On top of that, the current docker implementation also copies everything from `src-gen` to the container. So I take that as resolved.

### DIFF
--- a/test/Python/src/docker/FilesPropertyContainerized.lf
+++ b/test/Python/src/docker/FilesPropertyContainerized.lf
@@ -1,0 +1,34 @@
+target Python {
+    files: "../include/hello.py",
+    docker: true
+};
+
+preamble {=
+    try:
+        import hello
+    except:
+        request_stop()
+=}
+
+main reactor {
+    preamble {=
+        try:
+            import hello
+        except:
+            request_stop()
+    =}
+    state passed(false);
+    timer t(1 msec);
+    
+    reaction(t) {=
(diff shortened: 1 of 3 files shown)


Soroosh129 (approved):
Looks great!

Do we have an active issue for this that would be closed by merging?

test/Python/src/docker/FilesPropertyContainerized.lf
```diff
@@ -0,0 +1,34 @@
+target Python {
+    files: "../include/hello.py",
+    docker: true
+};
+
+preamble {=
+    try:
+        import hello
+    except:
+        request_stop()
```
Soroosh129, line 10: I like this approach, although I'm not sure how `request_stop()` would behave at the time of execution of this code, especially in federated execution, since the program might not be fully initialized yet.

housengw: good point. I am down to settle for this approach for now, since the purpose of the test is to make sure files specified in the "files" target property are also available within the container during build time.

PR #990, opened 2022-03-02
ignore .classpath and .project files

housengw:


org.lflang.tests/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Uncomment this line if you wish to ignore the project description file.
+# Typically, this file would be tracked if it contains build/dependency configurations:
+.project
+
+.classpath
```
lhstrh, line 5: `\n` missing

org.lflang/.gitignore
```diff
@@ -1 +1,7 @@
 model/
+
+# Uncomment this line if you wish to ignore the project description file.
+# Typically, this file would be tracked if it contains build/dependency configurations:
+.project
+
+.classpath
```
lhstrh, line 7: `\n` missing

lhstrh:
Looks like this approach won't work. Apparently, you can solve the problem of being suggested to check in changes to these files by locally by doing `git update-index --skip-worktree FILENAME`, but there is no way to configure the repository to have these files but don't track further changes. @edwardalee, @a-sr: what is the reason these files were checked in again? I recall it being something Eclipse/Oomph related...

a-sr:
> @a-sr: what is the reason these files were checked in again? I recall it being something Eclipse/Oomph related...

.project files are required for importing folders as projects into eclipse (during but not limited to the oomph setup).
.classpath, well configures the class path, but these should not be ignored because they are configured specifically for eclipse-plugins and do not represent the default.

PR #993, opened 2022-03-02
Replace the 'threads' target property with the 'workers' and 'threading' properties

### LINKED ISSUE #1020: List modal reactor tests in own category

cmnrd:
This PR implements the solution discussed in #290.

> Based on our discussion yesterday, the proposal is to have the following two target properties: threading = on | off (if supported, default is on) and workers = [int]. We will deprecate threads because it is ambiguous and its meaning overloaded. By the time that we end up implementing preemption (where we the size of the thread pool has to outnumber the workers), we could add an extra parameter for that (may be thread-pool-size = [int]).

I implemented the change for all the targets. However, there are some remaining issues that I don't know how to resolve.:

- [x] The number of worker threads is configured in various places throughout the C generator and even in duplicated code in the Python generator (see the diff). In the Python generator, even contradicting values are assigned. This is a large code smell to me, but I don't know how to fix it.
- [x] Before, the unthreaded runtime has been the default. Now, `threading=true` and `workers=0` are the default. I don't know if the C runtime will interpret `workers=0` correctly, as before it was not possible to have the threaded runtime with zero workers configured.
- [x] ~Since the default changed, this influences all the C tests that don't configure threads. Should we set `threading=false` in all tests that don't specify something else?~
- [x] ~In `org.lflang.tests.Configurators` the method `useFourThreads` reconfigures a test to use 4 workers. Should we force enable threading in this case?~
- [x] Remove `workers: N` entirely from all of our tests
- [x] Test the `workers` property (done in C and C++)

Unrelated to the above, this PR also fixes #1020.


lhstrh:
>     * Since the default changed, this influences all the C tests that don't configure threads. Should we set `threading=false` in all tests that don't specify something else?
> 
>     * In `org.lflang.tests.Configurators` the method `useFourThreads` reconfigures a test to use 4 workers. Should we force enable threading in this case?

I think we can either specify `threading=false` in all the generic tests, or we can leave it out and instead use a `useSingleThread` configurator to impose that property (instead of `useFourThreads`). The goal is to ensure that all tests are run using both runtime implementation. What is the default value of `workers`? I think it would make sense to match the number of available cores by default.

cmnrd:
We don't know the number of available cores on the target machine. In fact, the same binary could be run on different machines.

To quote from the comment in TargetConfig.java:
>  The number of worker threads to deploy. The default is zero, which indicates that the runtime is allowed to freely choose the number of workers.

I think this is the most reasonable default and it is also what is currently implemented for C++ (and I think for Rust). For zero workers, the C++ runtime uses the number of hardware threads available on the executing machine.

Soroosh129:
> For zero workers, the C++ runtime uses the number of hardware threads available on the executing machine.

I don’t know of any reliable method for this that works on all platforms for C. Do you happen to know any? Maybe CMake has something?

cmnrd:
No, I don't know. But it is also not needed to implement the exact same strategy for C. To me, the key is that `workers=0` means unspecified. This gives the targets the freedom to implement whatever strategy seems feasible. It could use a hard-coded number of workers (e.g. 4)., it could use the numbers of cores of the machine executing the compiler, it could check the number of cores at runtime, or use some heuristic to identify the best number of workers for a given application.

cmnrd:
I have updated the test configurators to force `threading=false` for all single threaded tests as suggested by Marten.

Minus the code smells in the C and Python generators and the unclear meaning of `workers=0`, this PR is ready. I have tested C programs with `workers=0` and currently they don't execute any reactions at all.

cmnrd:
In #1001, Marten suggested to not hardcode the number of workers in our tests. I think this is a good idea and added it as a bullet point to the PR description. I think currently the `workers` and `threading` properties are overridden anyway for all the test by the test configurators.

However, I think we should somehow test that `workers` and `threading` properties do work. So we should have tests where the values are hardcoded and we don't overwrite them in the test runner.

Soroosh129:
> I don't know if the C runtime will interpret workers=0 correctly, as before it was not possible to have the threaded runtime with zero workers configured.

It doesn't as far as I know. 
As I asked above, I'm not sure how we should interpret `workers=0`.
If we need to determine the number of CPU cores, we have a huge task ahead of us.

Soroosh129:
> [E]ven in duplicated code in the Python generator

The code duplication in `generatePreamble` is going away in #1009.

lhstrh:
> > I don't know if the C runtime will interpret workers=0 correctly, as before it was not possible to have the threaded runtime with zero workers configured.
> 
> It doesn't as far as I know. As I asked above, I'm not sure how we should interpret `workers=0`. If we need to determine the number of CPU cores, we have a huge task ahead of us.

How about a default value `>1` (maybe `4` or `8`?) in lieu of a mechanism that detects the number of cores? That said, I don't think it would be a "huge task" to [determine the number of cores](https://stackoverflow.com/questions/4586405/how-to-get-the-number-of-cpus-in-linux-using-c/4586471) but it will require some care craft a mechanism that works reliably across platforms. In this case, too, I would say the default number is a reasonable fallback option.

Soroosh129:
> Now, threading=true and workers=0 are the default. 

Is this just limited to tests?

It occurs to me that this might not be the best default in general for the Python target.
There is no true multi-threaded concurrency in Python.
Setting threading=true will add a lot of overhead by using the threaded reactor-c and  requiring GIL locks in the body of reactions.

Maybe the same is true for the TypeScript target?

Soroosh129:
We still have a weird case for federated execution.

As mentioned before, the CGenerator attempts to adjust the number of workers to account for all the blocking network input control reactions (there should be at least one worker available for each at all times). This is simply done by adding the number of network inputs to the number of workers during code generation.

The problem is that the C runtime will see a non-zero number of workers for federated programs, even if `workers=0`.

What do you think we should do here? The solution to this is not very clear to me.

org.lflang.tests/src/org/lflang/tests/Configurators.java
```diff
@@ -46,24 +46,35 @@
...
+     * do not distinguish threaded and unthreaded runtime, the number
+     * of workers is set to 1.
      *
      * @param test The test to configure.
      * @return True if successful, false otherwise.
      */
     public static boolean useSingleThread(LFTest test) {
-        test.context.getArgs().setProperty("threads", "0");
+        test.context.getArgs().setProperty("threading", "false");
+        test.context.getArgs().setProperty("workers", "1");
```
lhstrh, line 61: This line seems unnecessary. Isn't `workers` ignored if `threading` is `false`?

org.lflang.tests/src/org/lflang/tests/Configurators.java
```diff
@@ -46,24 +46,35 @@
...
+     * For targets that provide a threaded and an unthreaded runtime,
+     * this configures using the threaded runtime. For all targets,
+     * the number of workers is set to 4.
      *
      * @param test The test to configure
      * @return True if successful, false otherwise.
      */
     public static boolean useFourThreads(LFTest test) {
-        test.context.getArgs().setProperty("threads", "4");
+        test.context.getArgs().setProperty("threading", "true");
```
lhstrh, line 76: This preserves meaning, but can't we remove this method altogether?

example/C/src/Deadline.lf
```diff
@@ -10,7 +10,7 @@
  * @author Edward A. Lee
  */
 target C {
-    threads: 1,
+    workers: 1,
```
lhstrh, line 13:
```suggestion
```

example/C/src/Patterns/Chain_02_Pipeline.lf
```diff
@@ -15,7 +15,7 @@
  * @author Edward A. Lee
  */
 target C {
-    threads: 4,
+    workers: 4,
```
lhstrh, line 18: I think here it makes sense to leave the parameter so that it shows how the number of workers can be adjusted.

example/C/src/ROS/ROSBuiltInSerialization.lf
```diff
@@ -24,7 +24,7 @@
  target CCpp {
     cmake: true,  // Only CMake is supported
     cmake-include: "include/CMakeListsExtension.txt",
-    threads: 1,
+    workers: 1,
```
lhstrh, line 27:
```suggestion
```

example/C/src/ReflexGame.lf
```diff
@@ -12,7 +12,7 @@
  * @author Marten Lohstroh
  */
 target C {
-    threads: 1,
+    workers: 1,
```
lhstrh, line 15:
```suggestion
```

example/C/src/ReflexGame/ReflexGameTest.lf
```diff
@@ -1,5 +1,5 @@
 target C {
-    threads: 1,
+    workers: 1,
```
lhstrh, line 2:
```suggestion
```

example/C/src/Rhythm/Rhythm.lf
```diff
@@ -29,7 +29,7 @@
  * @see RhythmDistributedNoUI.lf
  */
 target C {
-    threads: 2,
+    workers: 2,
```
lhstrh, line 32:
```suggestion
```

example/C/src/Rhythm/SensorSimulator.lf
```diff
@@ -3,7 +3,7 @@
  * This has no audio output, but just tests the ncurses interface.
  */
 target C {
-    threads: 2,
+    workers: 2,
```
lhstrh, line 6:
```suggestion
```

example/C/src/TrainDoor/TrainDoor.lf
```diff
@@ -1,4 +1,4 @@
-target C {threads: 1, keepalive: true};
+target C {workers: 1, keepalive: true};
```
lhstrh, line 1:
```suggestion
target C {keepalive: true};
```

lhstrh (approved):
This looks great! Thanks!

org.lflang.tests/src/org/lflang/tests/Configurators.java
lhstrh: Never mind this. I thought we were running all the tests multi-threaded by default and had a special case for single-threaded, but the way it's setup currently we force all tests to be single-threaded and have multi-threaded as the special case...

lhstrh (approved):

PR #995, opened 2022-03-03
Fix eclipse error reporting

### LINKED ISSUE #994: Error reporting in Epoch is broken
Maybe due to recent changes in FileUtil and related classes, error reporting is broken in Epoch.  If I attempt to generate code for the following:
```
target C;
federated reactor {
    initial mode A { }
}
```
I get a null-pointer exception.  I should instead get this:
```
The currently selected code generation or target configuration does not support modal reactors!
```
(Modal models are not yet supported for federated reactors).

cmnrd:
This change reverts a small part of #978 in order to fix the broken error reporting in Epoch.

Closes #994

### DIFF
--- a/org.lflang/src/org/lflang/util/FileUtil.java
+++ b/org.lflang/src/org/lflang/util/FileUtil.java
@@ -371,7 +371,7 @@ public class FileUtil {
     /**
      * Get the specified uri as an Eclipse IResource or null if it is not found.
      *
-     * Also returns null if this is not colles from within a running Eclipse instance.
+     * Also returns null if this is not called from within a running Eclipse instance.
      *
      * @param uri A java.net.uri of the form "file://path".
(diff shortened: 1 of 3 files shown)


org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -371,20 +371,25 @@ public static IResource getIResource(Path path) {
     /**
      * Get the specified uri as an Eclipse IResource or null if it is not found.
      *
+     * Also returns null if this is not colles from within a running Eclipse instance.
```
edwardalee, line 374:
```suggestion
     * Also returns null if this is not called from within a running Eclipse instance.
```

edwardalee (approved):
Looks good to me!  Thanks for fixing.

PR #998, opened 2022-03-03
Check for null reactor to avoid NPE

edwardalee:
This seems to fix the second problem in #994.


cmnrd:
I pushed a change that should fix #994 for good.

Issue #83, opened 2020-01-08
The C++ target should be able to process LF files without a main reactor


cmnrd:
Currently the build fails when there is no main reactor. We could generate a library instead.

PR #1000, opened 2022-03-03
[C] do not generate code without main reactor

### LINKED ISSUE #913: Files without main don't compile if they have imports
The file `test/C/lib/Imported.lf` gives compile errors when compiled with lfc or Epoch. The reason is that it has an `import` statement that is ignored. This should be easy to fix (?).

housengw:
Do not generate code when there is no main reactor

Fixes #913

### DIFF
--- a/org.lflang/src/org/lflang/generator/GeneratorUtils.java
+++ b/org.lflang/src/org/lflang/generator/GeneratorUtils.java
@@ -56,9 +56,7 @@ public class GeneratorUtils {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
-        TargetDecl targetDecl = null;
-            return findAll(resource, TargetDecl.class).iterator().next();
-        return targetDecl;
+        return findAll(resource, TargetDecl.class).iterator().next();
     }
 
     /**
--- a/org.lflang/src/org/lflang/generator/GeneratorUtils.java
+++ b/org.lflang/src/org/lflang/generator/GeneratorUtils.java
@@ -56,16 +56,14 @@ public class GeneratorUtils {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
+        int cnt = 0;
         TargetDecl targetDecl = null;
         for (TargetDecl t : findAll(resource, TargetDecl.class)) { // getAllContents should never return null.
-            if (targetDecl != null) {
-                throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
-            }
             targetDecl = t;
+            cnt++;
         }
-        if (targetDecl == null) {
-            throw new InvalidSourceException("No target found!");
-        }
+        if (cnt > 1)  throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
+        if (cnt == 0) throw new InvalidSourceException("No target found!");
         return targetDecl;
     }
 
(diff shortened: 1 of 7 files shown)


lhstrh:
How does this PR relate to #913? Not generating code when there is no main reactor means that we cannot generate any pure library code. Is that really what we want?

housengw:
The error outlined in #913 is due to `CGenerator` not checking if main reactor exist before generating code in `doGenerate`.

The `Cpp` target also doesnt generate code when there is no main reactor. It might make sense to standardize code generation behavior across targets.

cmnrd:
I think we should really generate a library if there is no main reactor. In fact, I think we should create a library for all LF files and only generate a small binary that uses this library if there is a main reactor. Note that for C++ we have a long open issue: #83.

For the short term solution, though, it might be better to just forbid compilation without a main reactor. Without a package system, it does not make much sense to compile LF programs as libraries.

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -55,16 +56,14 @@ private GeneratorUtils() {
...
-            if (targetDecl != null) {
-                throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
-            }
             targetDecl = t;
+            cnt++;
         }
-        if (targetDecl == null) {
-            throw new InvalidSourceException("No target found!");
-        }
+        if (cnt > 1)  throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
```
lhstrh, line 65:
```suggestion
```
I don't think our grammar permits more than one target declaration, so this check is unnecessary altogether.

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -55,16 +56,14 @@ private GeneratorUtils() {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
+        int cnt = 0;
```
lhstrh, line 59:
```suggestion
```

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -55,16 +56,14 @@ private GeneratorUtils() {
...
         TargetDecl targetDecl = null;
         for (TargetDecl t : findAll(resource, TargetDecl.class)) { // getAllContents should never return null.
-            if (targetDecl != null) {
-                throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
-            }
             targetDecl = t;
+            cnt++;
```
lhstrh, line 63:
```suggestion
```

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -55,16 +56,14 @@ private GeneratorUtils() {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
+        int cnt = 0;
         TargetDecl targetDecl = null;
         for (TargetDecl t : findAll(resource, TargetDecl.class)) { // getAllContents should never return null.
-            if (targetDecl != null) {
-                throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
-            }
             targetDecl = t;
```
lhstrh, line 62:
```suggestion
            return t;
```

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -55,16 +56,14 @@ private GeneratorUtils() {
...
+            cnt++;
         }
-        if (targetDecl == null) {
-            throw new InvalidSourceException("No target found!");
-        }
+        if (cnt > 1)  throw new InvalidSourceException("There is more than one target!"); // FIXME: check this in validator
+        if (cnt == 0) throw new InvalidSourceException("No target found!");
```
lhstrh, line 66:
```suggestion
```
Our grammar _requires_ a target declaration.

lhstrh (approved):
Looks like a good quick fix. Like @cmnrd said, more work would be needed for the generation of library code. I've made suggestions for the removal of checks that I think are unnecessary.

lhstrh:
Oh, and congrats on filing PR #1000 :partying_face:

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -56,14 +56,10 @@ private GeneratorUtils() {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
-        int cnt = 0;
         TargetDecl targetDecl = null;
         for (TargetDecl t : findAll(resource, TargetDecl.class)) { // getAllContents should never return null.
```
lhstrh, line 60:
```suggestion
```

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -56,14 +56,10 @@ private GeneratorUtils() {
...
-            targetDecl = t;
-            cnt++;
+            return t;
         }
```
lhstrh, line 62:
```suggestion
```

org.lflang/src/org/lflang/generator/GeneratorUtils.java
```diff
@@ -56,14 +56,10 @@ private GeneratorUtils() {
      * Return the target declaration found in the given resource.
      */
     public static TargetDecl findTarget(Resource resource) {
-        int cnt = 0;
         TargetDecl targetDecl = null;
         for (TargetDecl t : findAll(resource, TargetDecl.class)) { // getAllContents should never return null.
-            targetDecl = t;
-            cnt++;
+            return t;
```
lhstrh, line 61: Wouldn't this also work?
```suggestion
            return findAll(resource, TargetDecl.class).iterator().next();
```

lhstrh (approved):
One more suggestion.

org.lflang/src/org/lflang/generator/GeneratorUtils.java
housengw: this would also work.

lhstrh (approved):

PR #1005, opened 2022-03-05
Better error reporting

edwardalee:
This is a rather small PR for a change that somehow got forgotten that improves error reporting in imported files.


Soroosh129 (approved):
It seems like these changes are already in master?
The "Files changed" tab is empty for me.

edwardalee:
Very odd. A comparison with master showed me two changes that I had indeed thought were already in master, but the results seemed to show they were not. But I guess they are!

PR #1008, opened 2022-03-08
Grammar cleanup

cmnrd:
This PR deletes various grammar rules that are not used.

Note: The `Expr` rule seems useful, but making it work requires some effort. We will introduce expressions as part of #544.


edwardalee (approved):
LGTM!

PR #1011, opened 2022-03-09
Refactor `CGenerator.xtend`

housengw:
As part of the effort to port `CGenerator.xtend` to java, I have refactored some of it into separate files. 

This PR has reached a good stopping point and is ready to merge.

PR #1014, opened 2022-03-09
Removed unnecessary xtend-gen folder from classpath.

soerendomroes:
There is no xtend in lignua franca, therefore, references to the xtend-gen folder can be safely removed.

### DIFF
--- a/oomph/LinguaFranca.setup
+++ b/oomph/LinguaFranca.setup
@@ -255,14 +255,7 @@
         targetURL="${git.clone.lingua.franca.location|uri}/org.lflang.ui.tests/src-gen/.dummy">
       <content></content>
     </setupTask>
-    <setupTask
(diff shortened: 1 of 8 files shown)


a-sr (approved):
Looks good to me.

@lhstrh  This cleans up some xtend leftovers in the Eclipse configuration that are no longer necessary since all the xtend files in these projects were replaced by Java. A problem with these configurations only occurred in a fresh Oomph installation. @soerendomroes noticed this during a setup and fixed it right away. Thanks for that.

lhstrh (approved):
This looks good to me. We should repeat these steps for the `org.lflang` package once #1011 has been completed.

lhstrh:
Thanks for these fixes! I've enabled CI (which was necessary because this was your first PR); let's merge once the tests have passed.

PR #1015, opened 2022-03-09
Improve layout options

soerendomroes:
This requires #1014 to be merged first.

The layout option `setLayoutOption(node, LayeredOptions.THOROUGHNESS, 100);` is debatable since it may increase the time needed for automatic layout. However, it seems to be quick enough for all examples I tested and is still in millisecond range.

### DIFF
--- a/oomph/LinguaFranca.setup
+++ b/oomph/LinguaFranca.setup
@@ -255,14 +255,7 @@
         targetURL="${git.clone.lingua.franca.location|uri}/org.lflang.ui.tests/src-gen/.dummy">
       <content></content>
     </setupTask>
-    <setupTask
-        xsi:type="setup:ResourceCreationTask"
-        excludedTriggers="BOOTSTRAP"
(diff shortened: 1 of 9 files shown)


a-sr (changes requested):
This PR is mixed up with changes from #1014.

Do these layout changes require the upcoming elk release or are they compatible with the currently used release?

soerendomroes:
> This PR is mixed up with changes from #1014.
> 
> Do these layout changes require the upcoming elk release or are they compatible with the currently used release?

They work with the current release.

a-sr:
@soerendomroes I just noticed that your changes to the `SEPARATE_CONNECTED_COMPONENTS` have a negative effect on the aesthetics of certain models.

SCC on:
cutter-scc-on

SCC off:
cutter-scc-off
Here the startup reaction chain is unpleasantly stretched out through the layers of the other reactors.

Could you comment on is. Is there a way to we can have separate but ordered connected components?

PR #1023, opened 2022-03-10
Replace Value with Expression in the grammar

cmnrd:
This makes a first step at implementing the spec in #986 by creating an extensible Expression syntax in replacement of the old `Value` rule. In consequence of the grammar changes, this PR also includes a ton of refactoring.

~This PR is based on #1022 and includes its changes.~


org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -285,24 +285,32 @@ VarRefOrModeTransition returns VarRef:
...
-    ((parens+='(' (init+=Value (','  init+=Value)*)? parens+=')')
-        | (braces+='{' (init+=Value (','  init+=Value)*)? braces+='}')
+    ((parens+='(' (init+=Expression (','  init+=Expression)*)? parens+=')')
+        | (braces+='{' (init+=Expression (','  init+=Expression)*)? braces+='}')
     )?
 ;
 
-Value:
-    (parameter=[Parameter] | time=Time | literal=Literal | code=Code);
+Expression:
```
edwardalee, line 304: I'm confused by this. I would expect an Expression to somehow different from a Value.  This looks to me the same as what we used to call a Value.

edwardalee (commented):
I'm not understanding this PR. It seems to be mostly about renaming the Value AST node to Expression without changing what it means. Is this meant to be a stepping stone towards supporting expressions?

cmnrd:
Yes, this a first step at introducing an expression syntax. This PR only converts what we now call a `Value` to an `Expression`. However, at the same time the resulting model is adjusted such that we have a base class `Expressions` and the different types of expressions (`ParameterReferecen`, `Literal`, `Code`, `Time`) are child classes. This enables easy addition of new expression rules in the future. Also, this allows for a better design of our code generators where we can use the Visitor pattern (in Java) or when blocks (in Kotlin) to differentiate the different expression types. This, however, is not part of this PR yet.

org.lflang/src/org/lflang/LinguaFranca.xtext
lhstrh: I think the idea here is that @cmnrd plans to augment this rule later on. This PR is just meant to set the stage for that.

lhstrh (approved):
Great work! I quite like how this has improved the readability of the code.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -285,33 +277,35 @@ VarRefOrModeTransition returns VarRef:
...
+    {Literal} literal = Literal
+    | Time
+    | ParameterReference
+    | Code
+;
 
-Time:
-    (interval=INT unit=TimeUnit)
+ParameterReference:
```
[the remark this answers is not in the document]
lhstrh: Same with `Expression` vs. `Expr` maybe. I don't know. No strong preference here, except for maintaining consistency.

lhstrh (approved):

petervdonovan:
The LSP tests have been a hindrance for this branch... sorry.

This time, the LSP tests seem to be failing because the body of the STP handler in `test/Python/src/federated/DistributedCountDecentralizedLateDownstream.lf` is not being code-generated. This does not seem right to me. Maybe @Soroosh129 could explain what is going on?

Soroosh129:
> This time, the LSP tests seem to be failing because the body of the STP handler in test/Python/src/federated/DistributedCountDecentralizedLateDownstream.lf is not being code-generated. This does not seem right to me. Maybe @Soroosh129 could explain what is going on?

I can't seem to replicate this issue. The STP handler is being generated for me on `master` and this branch:
#### Python
image

#### C
image

petervdonovan:
I was mistaken; thanks to Soroush for patiently explaining. The real reason for the most recent test failure in this branch is that instead of having `toText` insert tags while `toUntaggedText` does not insert tags, we have a `toText` function that does not insert tags and a `toTaggedText` function that does insert tags. This makes inserting tags more like an "opt-in" than an "opt-out."

I don't feel strongly about this, and I am sure there is a reason for this change. However, I am not convinced that it is the best choice. In `ASTUtils.toUntaggedText` (which is now deleted), I had left a comment stating that `toUntaggedText` only existed because we were analyzing and patching the contents of code blocks using regular expressions, which is a practice that probably should be discouraged. I think that verbatim code should nearly always have tags, so with the behavior of `toText` that is in this branch, I will probably have to do another pass over all the code generators to change most mentions of `toText` to `toTaggedText`.

lhstrh:
I think I lean towards reverting the mechanism to opt-out instead of having opt-in.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -769,77 +762,27 @@ public static <T> List<T> collectElements(Reactor definition, EStructuralFeature
...
+     * Translate the given code into its textual representation
+     * with {@code CodeMap.Correspondence} tags inserted.
+     * @param node AST node to render as string.
      * @return Textual representation of the given argument.
      */
-    public static String toText(Code code) {
-        return CodeMap.Correspondence.tag(code, toUntaggedText(code), true);
+    public static String toText(EObject node) {
+        if (node == null)
+            return null;
```
lhstrh, line 772: Should it be null or `""`? The latter seems safer.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -769,77 +762,27 @@ public static <T> List<T> collectElements(Reactor definition, EStructuralFeature
...
-        if (e.getLiteral() != null) {
-            str = withoutQuotes(e.getLiteral()).trim();
-        }
-        if (e.getId() != null) {
-            str = e.getId();
-        }
-        return str;
+    public static String toUntaggedText(EObject node) {
+        if (node == null)
+            return null;
```
lhstrh, line 784: Same question here.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -1499,14 +1374,15 @@ public static boolean belongsTo(EObject eobject, Reactor reactor) {
...
         int result = 0;
-        for (Value value: values) {
-            if (value.getLiteral() == null) { 
+        for (Expression expr: expressions) {
+            if (!(expr instanceof Literal)) {
                 return null;
             }
             try {
-                result += Integer.decode(value.getLiteral());
+                // FIXME: why does this sum the values in the list??
```
lhstrh, line 1384: :thinking:

lhstrh (approved):
This looks good to me! I guess the only minor aesthetic issue I see that not all developers might initially be aware of what tagging is and whether they should use `toText` or `toUntaggedText`. Perhaps we can come up with a better name for `toUntaggedText` that highlights its typical use case rather than the presence of tags? Maybe `toOriginalText`? Or `toVerbatimText`? IDK...

org.lflang/src/org/lflang/ASTUtils.java
petervdonovan: It seems non-obvious to me that the latter is safer. Is the AST node supposed to be null, or not? If it is OK for it to be null, then that should probably be documented. Since it's not documented, my best guess is that it's not OK to be null, in which case returning the empty string might be worse than doing nothing because converting `null` to `""` hides the original failure. We want to fail fast, right?

This is not the first time I have said something along these lines -- maybe we should discuss it to make sure I understand this design principle properly.

org.lflang/src/org/lflang/ASTUtils.java
petervdonovan: This method is a little weird, but at least the Javadoc states that this is the desired behavior. I deleted the FIXME.

org.lflang/src/org/lflang/ASTUtils.java
lhstrh: I haven't looked at all the call sites, but in general, references to other AST nodes _could_ be null. I'm not sure what the right approach is either, but since this is used in code generation, returning an empty string seems relatively safe. Not sure though.

lhstrh (approved):

cmnrd:
Thanks for finishing this up! With regard to the discussion about `toText` and `toTaggedText`, I found that toText is also used a lot outside of the generators and there the additional tags are pretty confusing. However, I like the new to `toOriginalText` name for the function and it should be pretty clear now.

PR #1025, opened 2022-03-10
[CI] Bring back PNPM cache.

petervdonovan:
The PNPM cache step was deleted [here](c7011df59059c402b0d0c0b3d0ce2c1ed9c7004a) because we were hashing lock files that did not exist and later hashing an enormous number of lock files. This adds back a modified version in an attempt to prevent transient failures like [this](https://github.com/lf-lang/lingua-franca/runs/5492769558?check_suite_focus=true).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
   # Run language server tests.
   lsp-tests:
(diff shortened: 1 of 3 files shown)


lhstrh (approved):
Looks good! Thanks for the fix!

PR #1028, opened 2022-03-10
Automatically cancel old workflow actions

Soroosh129:
This PR adds automatic cancellation of old GitHub Actions workflows triggered by an event on the same branch.

See [here](https://github.com/marketplace/actions/cancel-workflow-action) for more information.

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,4 +1,4 @@
-name: Cancel
+name: Cancel Previous Jobs
 
 on:
   workflow_call:
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   cancel:
-    name: 'Cancel Previous Runs'
+    name: run
     runs-on: ubuntu-latest
     timeout-minutes: 3
     steps:
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,0 +1,15 @@
+name: Cancel
+
+on:
+  workflow_call:
+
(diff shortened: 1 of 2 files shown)


.github/workflows/cancel.yml
```diff
@@ -0,0 +1,15 @@
+name: Cancel
```
lhstrh, line 1:
```suggestion
name: Cancel Previous Jobs
```

.github/workflows/cancel.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  cancel:
+    name: 'Cancel Previous Runs'
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
+      - uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          all_but_latest: true
+          access_token: ${{ github.token }}
```
lhstrh, line 15:
```suggestion
          access_token: ${{ github.token }}

```

lhstrh (approved):
This is a welcome change. Thanks!

.github/workflows/cancel.yml
```diff
@@ -0,0 +1,15 @@
+name: Cancel Previous Jobs
+
+on:
+  workflow_call:
+
+jobs:
+  cancel:
+    name: 'Cancel Previous Runs'
```
lhstrh, line 8:
```suggestion
    name: run
```
This is to confirm with the other workflows and how it is rendered in GH Actions GUI.

Soroosh129: I think `Cancel Previous Runs` was more descriptive than just `run`:
image

PR #1029, opened 2022-03-10
Broaden a pylint ignore.

petervdonovan:
This is just a one-line patch. I am not sure why it's necessary (it worked before!), but it seems harmless in the sense that it won't conceal any errors that would otherwise have been detected.

This change has been tested in Epoch.


lhstrh (approved):

Soroosh129:
Could it be, perhaps, that somehow the installed Python module for a Python-target program (which is installed via `pip`) cannot be discovered within the Epoch environment?

petervdonovan:
> Could it be, perhaps, that somehow the installed Python module for a Python-target program (which is installed via pip) cannot be discovered within the Epoch environment?

I do not think that the Python module needed to be installed in order for Pylint to work. The Pylint check was supposed to be fast, and installing the module was one of the slowest parts of compilation. Therefore, if the module is required, then there is a bug.

PR #1038, opened 2022-03-12
[error reporting] Correct bugs reported on Epoch.

### LINKED ISSUE #1030: EclipseErrorReporter reports errors in the wrong file
I am not sure if this problem is on my end or if it is on the Eclipse side. The path I provide to `EclipseErrorReporter` is `/resource/temp/src/TraceTesting.lf`, but the resource that comes from invoking `FileUtil.getIResource` is `null`.

This causes warnings to be reported on the main file on the wrong lines in cases where they should be reported on the imported file instead. This can be reproduced by opening the Python version of [ModalActions.lf](https://github.com/lf-lang/lingua-franca/blob/python-modal-models/test/Python/src/modal_models/ModalActions.lf) in Epoch.

petervdonovan:
Fixes #1030.

These changes have been tested in Epoch and address two bugs: A missing leading '/' character in C error reporting, and the fact that for a reason unknown to me, some of the paths (sourced originally from the Ecore model, I believe) were prefixed with a segment named `resource`. The latter problem was patched downstream of its source because I was not optimistic that there would be a cleaner solution.

Thank you @cmnrd for answering some questions on this.


org.lflang/src/org/lflang/generator/EclipseErrorReporter.java
```diff
@@ -153,6 +146,21 @@ private String report(String message, Severity severity, Integer line, Path file
         return header + ": " + message;
     }
 
+    private IResource bestEffortGetIResource(Path path) {
+        IResource iResource = null;
+        if (path != null) {
+            iResource = FileUtil.getIResource(path);
+            if (iResource == null) {
+                // This operation is risky and may fail with a IllegalStateException or NPE.
```
edwardalee, line 154: Shouldn't this be guarded by a try...catch then?  It's really bad when error reporting throws an exception because this results in hanging threads and useless messages about us caught exceptions.

edwardalee (approved):
Looks good except for the suggested try...catch.

org.lflang/src/org/lflang/generator/EclipseErrorReporter.java
```diff
@@ -109,21 +111,12 @@ private String report(String message, Severity severity, Integer line, Path file
...
-        if (iResource == null) {
-            iResource = fileConfig.iResource;
-        }
+            System.err.println(header + ": " + file + " line " + line + "\n" + message);
 
         // Create a marker in the IDE for the error.
         // See: https://help.eclipse.org/2020-03/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2FresAdv_markers.html
         try {
-            IMarker marker = iResource.createMarker(IMarker.PROBLEM);
+            IMarker marker = bestEffortGetIResource(file).createMarker(IMarker.PROBLEM);
```
lhstrh, line 119: I al ready mentioned this, but for the record: this solution should be factored into the existing `FileUtil.getIResource` method.

lhstrh (changes requested):

org.lflang/src/org/lflang/generator/EclipseErrorReporter.java
petervdonovan: Ah, sorry, I forgot to push.

org.lflang/src/org/lflang/generator/EclipseErrorReporter.java
petervdonovan: Now any `IllegalStateException` will fail silently. Our existing implementation assumes the workspace root will be non-null, so hopefully that will be safe.

lhstrh (approved):
Looks good to me; thanks for the fix!

PR #1040, opened 2022-03-14
Improve Stability of Positioning Edges connected to Reactions in Diagrams

a-sr:
If a reaction has multiple input/output dependencies, sometimes the edges do not properly end at the border of the shape but the bounding box. This PR fixes this issue for most (or at least more) cases.

Also fixes the 'bundle dependencies' synthesis option for reactions that was broken in the translation from xtend to java.


edwardalee (approved):
LGTM, except that the indentation in these files is wonky, which makes the GitHub diffs very hard to read. Looks like there is a mix of tabs and spaces in the files.

lhstrh:
> The indentation is indeed very inconsistent throughout the entire diagram package. I will create a separate PR with all tabs converted to spaces or should I include it in this one?

Let's fix indentation in a separate PR for the sake of readability of the diff. I'll go ahead and merge this as is.

PR #1043, opened 2022-03-15
Fix errors in tracing util.

hokeun:


Soroosh129 (approved):
Looks great!

Thanks for fixing this.
This shows that we should add tests for these utilities.

PR #1045, opened 2022-03-16
refactor `CGenerator.xtend`

housengw:
This PR constitutes another part of the effort to port `CGenerator.xtend` to java. 
After this, I think I will complete the port.
See #838


lhstrh (commented):
Looking great so far! Some newly created functions still need JavaDoc, but this being a draft, I'm guessing you're aware of that :-)

lhstrh (approved):
🚀🚀🚀

PR #1050, opened 2022-03-17
Removed broken dependency on `org.eclipse.xpand.feature.group`

lhstrh:
There was a spurious dependency in the Oomph configuration that was causing the build to fail. This PR removes that dependency.


edwardalee (approved):
LGTM

PR #1049, opened 2022-03-16
Port `CGenerator.xtend` to java

housengw:
see #838

PR #1051, opened 2022-03-17
Removed all Xtend dependencies

housengw:
Since the last xtend file has been ported to java in #1049, all Xtend dependencies can be removed.


lhstrh (commented):
This looks good to me. There are two things I believe might not be addressed by this:
 - I believe that the Oomph config added empty `xtend-gen` directories somewhere, but this might have gone away in the PR submitted by @soerendomroes; and
 - `XtendFormatterEclipse.xml` is no longer needed and should also be removed.

lhstrh (approved):

PR #1053, opened 2022-03-18
The files target property: Add support for directories

Soroosh129:
This adds support for including entire directories in the `files` target property (e.g., `files: ["/etc"]`).

I found this to be very convenient. Imagine a game that requires a number of images that are located in a folder.
It would be convenient to be able to use `files: ["images"]` instead of listing each image file one by one.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -442,9 +442,9 @@ public class CGenerator extends GeneratorBase {
             var actions = Iterables.filter(IteratorExtensions.toIterable(resource.getAllContents()), Action.class);
             for (Action action : actions) {
                 if (Objects.equal(action.getOrigin(), ActionOrigin.PHYSICAL)) {
-                    // If the unthreaded runtime is requested, use the threaded runtime instead
+                    // If the unthreaded runtime is not requested by the user, use the threaded runtime instead
                     // because it is the only one currently capable of handling asynchronous events.
-                    if (!targetConfig.threading) {
+                    if (!targetConfig.threading && !targetConfig.setByUser.contains(TargetProperty.THREADING)) {
                         targetConfig.threading = true;
                         errorReporter.reportWarning(
                             action,
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -1028,7 +1028,7 @@ public class CGenerator extends GeneratorBase {
     }
     
     /**
-     * Copy all files listed in the target property `files` and `cmake-include` 
+     * Copy all files or directories listed in the target property `files` and `cmake-include` 
      * into the src-gen folder of the main .lf file
      * 
      * @param targetConfig The targetConfig to read the `files` and `cmake-include` from.
--- a/org.lflang/src/org/lflang/generator/c/CUtil.java
+++ b/org.lflang/src/org/lflang/generator/c/CUtil.java
@@ -553,66 +553,76 @@ public class CUtil {
     }
 
     /**
-    * Copy the 'fileName' from the 'srcDirectory' to the 'destinationDirectory'.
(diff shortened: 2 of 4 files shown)


org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -553,51 +553,68 @@ static public String triggerRefNested(PortInstance port, String runtimeIndex, St
...
+     * procedure: 
+     *     1- Search in LF_CLASSPATH. @see findFile() 
+     *     2- Search in CLASSPATH. @see findFile() 
+     *     3- Search for 'fileName' as a resource. That means the `fileName` 
+     *        can be '/path/to/class/resource'. @see java.lang.Class.getResourceAsStream()
+     *
+     * @param fileName Name of the file or directory.
+     * @param srcDir   Where the file or directory is currently located.
+     * @param dstDir   Where the file or directory should be placed.
+     * @return The name of the file or directory in destinationDirectory.
```
edwardalee, line 569:
```suggestion
     * @return The name of the file or directory in destinationDirectory or an empty string on failure.
```

edwardalee (approved):
LGTM! Useful feature. Let's merge.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -442,9 +442,9 @@ public void accommodatePhysicalActionsIfPresent() {
             var actions = Iterables.filter(IteratorExtensions.toIterable(resource.getAllContents()), Action.class);
             for (Action action : actions) {
                 if (Objects.equal(action.getOrigin(), ActionOrigin.PHYSICAL)) {
-                    // If the unthreaded runtime is requested, use the threaded runtime instead
+                    // If the unthreaded runtime is not requested by the user, use the threaded runtime instead
                     // because it is the only one currently capable of handling asynchronous events.
-                    if (!targetConfig.threading) {
+                    if (!targetConfig.threading && !targetConfig.setByUser.contains(TargetProperty.THREADING)) {
```
lhstrh, line 447: I'm a little confused about this change. What does it do? I seems entirely unrelated to the PR.

org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -553,51 +553,68 @@ static public String triggerRefNested(PortInstance port, String runtimeIndex, St
...
+     *
+     * @param fileName Name of the file or directory.
+     * @param srcDir   Where the file or directory is currently located.
+     * @param dstDir   Where the file or directory should be placed.
+     * @return The name of the file or directory in destinationDirectory or an empty string on failure.
+     */
+    public static String copyFileOrResource(String fileName, Path srcDir,
+            Path dstDir) {
+        // Try to copy the file or directory from the file system.
+        Path file = findFileOrDirectory(fileName, srcDir);
```
lhstrh, line 574: I would rename `file` to `path` here.

org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -553,51 +553,68 @@ static public String triggerRefNested(PortInstance port, String runtimeIndex, St
...
+                }
+                return file.getFileName().toString();
+            } catch (IOException e) {
+                // Failed to copy the file or directory, most likely
+                // because it doesn't exist. Will try to find it as a
+                // resource before giving up.
+            }
+        }
+
+        String filenameWithoutPath = fileName;
```
lhstrh, line 592: The following takes the tail of a path (e.g., `foo/bar/baz.lf` yields `baz.lf`; `foo/bar` yields `bar`) and just looks for it on the class path. This doesn't seem right to me. I don't think there are any tests for this, and I think the `FIXME` below alludes to this. I realize this `FIXME` preexists this PR, but I think we should fix and not ignore it.

lhstrh (commented):
I think the added feature makes sense, but I had some comments in the pipeline that I didn't get to share before the PR was merged.

PR #1055, opened 2022-03-20
Improved Support for Imported Reactors in Diagrams

### LINKED ISSUE #889: Navigate to imported reactors in diagrams
The current behavior is implemented in the klighd framework and it is not really a bug but intended this way.
The basic UX concept is that the editor defines the context for the diagram view. If the selection of a diagram element that is defined in a different resource would jump to the textual source, it would switch to the respective editor and display the diagram for that editor, losing the previous context.
I think this behavior would be annoying because it happens rather often that you select elements without the intent to look at the source code, e.g. to navigate in the model, highlight them for presentation, or expand them.

I could try to add an alternative action, e.g. ctrl+click, to navigate to the source code of reactors in other files. This way, losing context would be intentional but still inconsistent with the default.

@lhstrh I agree, if we want this behavior, we should add a new ticket because this will take some careful consideration in order to provide a pleasant user experience.

_Originally posted by @a-sr in https://github.com/lf-lang/lingua-franca/issues/341#issuecomment-1012034929_

a-sr:
A normal selection of diagram elements that originate from code outside the current editor (imported reactors), will now highlight the respective import instead of nothing.
If the Alt key is down during selection of such diagram elements, the active editor will change to the one containing the code. This will also cause the diagram view to update its content this reflect the model of the new active editor. 

Closes #889

Since this is a UI specific feature, this current implementation only affects the Epoch IDE. The LF VSC extension requires a separate implementation which I am not familiar with.

### DIFF
--- a/org.lflang.ui/META-INF/MANIFEST.MF
+++ b/org.lflang.ui/META-INF/MANIFEST.MF
@@ -25,7 +25,10 @@ Require-Bundle: org.lflang,
  org.eclipse.core.runtime,
  org.eclipse.core.resources,
  org.eclipse.pde.core,
- org.eclipse.ui.forms
+ org.eclipse.ui.forms,
+ de.cau.cs.kieler.klighd;bundle-version="2.1.0",
+ de.cau.cs.kieler.klighd.ui;bundle-version="2.1.0",
+ de.cau.cs.kieler.klighd.ui.view;bundle-version="2.1.0"
 Import-Package: org.apache.log4j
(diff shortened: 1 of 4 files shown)


edwardalee (approved):
Wow, this is more complicated than one would have hoped! But this is great functionality! Let's merge it.

lhstrh (approved):
This definitely requires a level of knowledge about Xtext/Ecore/Eclipse that I don't posses, so my review isn't of much use. The feature is very useful though -- thanks for implementing it!

PR #1061, opened 2022-03-26
Fix network sender reaction multiport

Soroosh129:
This fixes an issue related to the way network sender reactions are triggered when their trigger is a multiport output from a federate or a bank of federates.

In this fix, a check is added to make sure a given port is present before the body of the network sender reaction reads it.


edwardalee (approved):
LGTM! Let's merge.

PR #1062, opened 2022-03-27
Update and add federated execution tests for TypeScript target.

hokeun:
Relevant PR: https://github.com/lf-lang/reactor-ts/pull/91


edwardalee (approved):
Tests look OK to me. One minor issue is that we settled on the following comment style:
```
/**
 * Text goes here...
 */ 
```
rather than this:
```
/** Text shouldn't go here.
 *   more text.
 */ 
```

hokeun:
> Tests look OK to me. One minor issue is that we settled on the following comment style:
> 
> ```
> /**
>  * Text goes here...
>  */ 
> ```
> 
> rather than this:
> 
> ```
> /** Text shouldn't go here.
>  *   more text.
>  */ 
> ```

Thank you @edwardalee ! Fixed these in https://github.com/lf-lang/lingua-franca/pull/1062/commits/7e048703fda71631f3392dcc0fa6040a222e85ea.

PR #1065, opened 2022-03-31
Commented out annotation of outputs on transitions

edwardalee:
I'm not sure whether this is a good idea, but as we were working with modal models today, it struck us as odd that transitions were annotated with effects. But these effects (like outputs) have nothing much to do with the transitions. Is there a reason that we are missing for these effects annotations?


a-sr (approved):
I agree, these effects do not really relate to transitions and can be removed.

lhstrh:
Shall we just remove the functionality rather than leave it commented out?

PR #1066, opened 2022-04-01
Federated Modal Models

a-sr:
Activated federated code generation for modal reactors.
Also added tests for both C and Python


a-sr:
@Soroosh129 Could you help me here?
I created test models for federates with modes in C and Python but I wasn't able to test them in a distributed way on my machine, so I am not sure if I configured them correctly. C has a `Segmentation fault` but Python seems to run, at least on Ubuntu.
What do you think, is this a problems with the tests, the runtime implementation, or a larger conceptual one?

Soroosh129:
I was able to resolve the segfault in C, but the Python tests and the decentralized test still fail.

For Python, I think the same strategy employed in 0c8c9877e3be0c61130551071349deaf16218f77 should work. 
I.e., we need to make sure that only timers, state variables, etc. that actually belong to a federate are considered while generating code for that federate. 

I will have a bit free time later this week to work on this if needed.

PR #1084, opened 2022-04-07
Minor Improvements to Diagrams

a-sr:
- Fixed offset in physical connection rendering position
- Improved positioning of startup and shutdown triggers
- Modes now have local startup and shutdown triggers


edwardalee (approved):
LGTM!

PR #1085, opened 2022-04-08
Do not create the temporary dependency link if a connection between f…

edwardalee:
…ederates has a delay.  This fixes a problem that a federated program with feedback was exhibiting spurious cycle detection errors.

After fixing this, two bugs surfaced that required the following two fixes:

- Fixed an issue where control reactions were being triggered even though the federate had no upstream/downstream zero-delay logical connections.

- Removed the problematic optimization where all federates and the RTI assumed that all federates already have a PTAG at startup.

See lf-lang/reactor-c#83 for the accompanying PR.


Soroosh129:
I think we might have fixed the original issue discussed in #1086. 
I can't reproduce the false cycle error anymore.
To confirm, I added a simplified version of the test ([FeedbackDelaySimple.lf](https://github.com/lf-lang/lingua-franca/blob/79fe28e0087c90f2657c59156b1a18a210e5bfe5/test/C/src/federated/FeedbackDelaySimple.lf)), which for me executes to completion.
I also don't get a cycle error for FeedbackDelay.lf. 

Nonetheless, it seems like there is a deadlock occurring in FeedbackDelay, which causes it to timeout.
This looks like an unrelated issue to #1086.

I suggest we keep the FeedbackDelaySimple test in this PR and merge without FeedbackDelay (if all tests pass), and subsequently create a new issue for the deadlock happening in FeedbackDelay.

Soroosh129:
I believe this PR is ready for review. I'm not sure why the Cpp tests timed out, but I'm pretty sure that's unrelated to the changes in this branch.

Soroosh129:
@edwardalee Could you please kindly take a look at the changes here to see if they look OK? I could not ask for a review since you are the original creator of this PR.

lhstrh (approved):
Looks OK to me. Be sure to update the refs in `ci.yml` after merging this (and before deleting this branch ;-))

.github/workflows/ci.yml
```diff
@@ -54,12 +54,12 @@ jobs:
 
   # Run the C integration tests.
   c-tests:
-    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@master
+    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@federated-delay-fix
```
edwardalee, line 57: Wait, we probably don't want these changes in ci.yml to be committed when this is merged to master, right?

edwardalee (commented):
These changes look right to me, except perhaps the ci.yml change.

.github/workflows/ci.yml
lhstrh: We do. After committing, it's a matter of changing these to `master`. This will work as long as the referenced branch exists.

PR #1087, opened 2022-04-10
Fixed an issue with bank_index

Soroosh129:
Fixed an issue where bank_index was not being generated as a reactor parameter unless the user had explicitly added it as a parameter.


housengw (approved):
LGTM

PR #1088, opened 2022-04-11
Switched to building the Python extension module in-place

Soroosh129:
It has been a long-time quirk of the Python target that it installs the generated Python C extension module system-wide, to ensure compatibility.

However, I think I have found a workaround, inspired by the issues @edwardalee was having with the [Pac-Man](https://github.com/lf-lang/experimental-lingua-franca/tree/main/Python/src/Pac-Man) game.

This fix should dramatically reduce the clutter in terms of installed packages in a user's local environment.

It should also make working with virtual environments easier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
   
   # Run the Python integration tests.
   py-tests:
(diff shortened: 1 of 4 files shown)


housengw (approved):
LGTM!

PR #1094, opened 2022-04-12
Java configuration bumped to version 17

### LINKED ISSUE #1079: Remove logic for handling examples from testing framework
After the examples were relocated to a separate repo, the logic for handling examples in the testin framework is no longer needed.

lhstrh:
This PR brings the minimum Java requirement from 11 to 17. This has several advantages:
 - better syntax for code generators written in Java (string templates)
 - compatibility with Apple Silicon
 - long-term support (17 is the latest LTS release)

In addition, this PR fixes #1079.


housengw (approved):
LGTM! We can use this PR as a reference of all the `JavaSE-*` version that need to be updated to bump the java version. Thank you for doing this!

lhstrh:
> LGTM! We can use this PR as a reference of all the `JavaSE-*` version that need to be updated to bump the java version. Thank you for doing this!

Yes, we could. That said, I think the intention here is not to change this until another LTS version appears.

PR #1097, opened 2022-04-12
Update `SET`, `schedule` and `tag` APIs in the C target

housengw:
Please refer to issue #1108 for the names of the APIs before and after the change.


test/C/src/SetWithCustomDestructor.lf
```diff
@@ -0,0 +1,58 @@
...
+    output out:complex_t*;
+    logical action a;
+    reaction(startup) -> a {=
+        for (size_t i = 0; i < 10; i++) {
+            schedule(a, MSEC(50));
+        }
+    =}
+    
+    reaction(a) -> out {=
+        complex_t* val = (complex_t*) malloc(sizeof(int));
```
edwardalee, line 29:
```suggestion
        complex_t* val = (complex_t*) malloc(sizeof(complex_t));
```

test/C/src/SetWithCustomDestructor.lf
```diff
@@ -0,0 +1,58 @@
...
+        val->size = 1000;
```
edwardalee, line 30:
```suggestion
        val->size = 10;
```
(No need to make a test more expensive to run).

edwardalee (changes requested):
Small error detected. Approve modulo changing function names to lower case.

housengw:
Seems like a Windows specific problem is keeping `SetWithCustomConstructor` from passing on Windows

Soroosh129 (approved):
LGTM!

Before merging this, let's make sure everyone is aware of this change (maybe by announcing it on lf-interest) and that the website and/or the wiki are up-to-date.

housengw:
I can make an announcement. With regards to the website, I think we should merge the PR first. Otherwise we would have documentation that refer to code not in master

edwardalee (approved):
LGTM!  We can immediately update the website repo because that website is not live. But anyway, I think we should forge ahead with this...

PR #1098, opened 2022-04-14
Add support for threading compiler flag to Rust target

jhaye:
Based on the given threading flag, the `parallel-runtime` feature is enabled for the Rust runtime in the generated code.

The way it is implemented right now allows one to override the flag by setting the feature manually in the LF source. There is however no option to disable it the same way.


cmnrd:
> In that case we would need a similar feature for the runtime that reports back the number of workers.

Do you think this would be possible to add with reasonable effort?

What do you mean by "If the parallel-runtime feature is enabled from a file"? If threading is set to false, then parallel execution should be disabled. If the parallel runtime feature is requested manually and threading is set to false, then we should probably stick to the manual specification, but I think it would be a good idea to also issue a warning.

jhaye:
> Do you think this would be possible to add with reasonable effort?

Yes, probably. I already opened an issue for it.

> What do you mean by "If the parallel-runtime feature is enabled from a file"?

It looks like this:
```
target Rust {
    build-type: Release,
    cargo-features: [ "cli" ],
    rust-include: "../lib/pseudo_random.rs",
    cargo-dependencies: {
        reactor_rt: {
            features: ["parallel-runtime"]
        },
    }
};
```
Right now this would enable the parallel runtime independent of the `threading` flag.

cmnrd:
Alright, I think this is Ok and we cannot really avoid it. However, if `threading: false` is set in your example, then we should print a warning.

oowekyala (commented):
(nitpicks)

jhaye:
> print a warning

How can I do that?

PR #1099, opened 2022-04-14
Support inline code arguments on main reactor for Rust target

jhaye:
Previously inline arguments for the main reactor weren't handled in a way that was appropriate for clap, the CLI library used for Rust reactor programs. 
Basically, it couldn't handle arguments like this: `max_value: u64({= 1 << 60 =})`.

This PR upgrades to a stable version of clap and handles inline arguments correctly. This depends on #1098, so that should be merged first.

Ref: https://github.com/lf-lang/benchmarks-lingua-franca/issues/5

### DIFF
--- a/org.lflang/src/org/lflang/generator/rust/RustCargoTomlEmitter.kt
+++ b/org.lflang/src/org/lflang/generator/rust/RustCargoTomlEmitter.kt
@@ -51,7 +51,7 @@ object RustCargoTomlEmitter : RustEmitterBase() {
             |[dependencies]
             |env_logger = "0.9"
             |log = { version = "0.4", features = ["release_max_level_info"] }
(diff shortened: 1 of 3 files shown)


cmnrd (approved):
Cool! I converted this to draft and will merge this as soon as we have merged #1098.

cmnrd (changes requested):
I think this needs to be rebased to the new master. It seems to still include some changes from #1098.

Also, could you add a test for this added functionality (or extend an existing test)?

jhaye:
I added a test, but not sure if it's automatically added to the bulk test run.

cmnrd (approved):
Yes, all the lf files in the test directories are run automagically. Thanks!

PR #1100, opened 2022-04-14
Added support for state variables in diagrams

a-sr:
Added option to display state variables in reactors. Will be off by default.

Also includes some minor unrelated improvements in the diagrams package.

### DIFF
--- a/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
+++ b/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
@@ -64,6 +64,7 @@ import org.eclipse.xtext.xbase.lib.ListExtensions;
 import org.eclipse.xtext.xbase.lib.Pair;
 import org.eclipse.xtext.xbase.lib.StringExtensions;
 import org.lflang.ASTUtils;
+import org.lflang.InferredType;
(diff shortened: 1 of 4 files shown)


lhstrh:
If we show the funny brackets we might as well show what's between them, no?

a-sr:
I agree with @lhstrh. I don't think we should omit or replace target code.
In the end, this is an optional feature, so we do not bloat the diagram by default.

I will revise the PR accordingly; also adjusting reactor parameters to be displayed consistently.

lhstrh:
Does anything need to change in the VS Code extension to enable this feature (besides updating the submodule)?

a-sr:
> Does anything need to change in the VS Code extension to enable this feature (besides updating the submodule)?

No, this feature will work on both platforms out-of-the-box.
When I create a PR with changes that only work in Epoch, such as #1055, I will make sure to add a comment. Or do we want add a specific workflow, e.g. adding a dedicated label?

lhstrh:
> > Does anything need to change in the VS Code extension to enable this feature (besides updating the submodule)?
> 
> No, this feature will work on both platforms out-of-the-box. When I create a PR with changes that only work in Epoch, such as #1055, I will make sure to add a comment. Or do we want add a specific workflow, e.g. adding a dedicated label?

That's great! What you're doing now works well. Note that we already have an `epoch` label.

cmnrd (approved):
Great!

PR #1101, opened 2022-04-18
Allow bank_index in initializers.

edwardalee:
A common problem with banks is that members of the bank may need to each have its own set of parameter values. The only mechanism we currently offer requires modifying the reactor class itself, which needs to set state variables (rather than parameters) from a global table. This means that the reactor can only be used in a bank and must be written to work in bank.

This PR adds to the C and Python targets the ability to initialize parameters using a table lookup and the `bank_index` variable. Here is an example:

```
target C;
preamble {=
    int table[] = {4, 3, 2, 1};
=}
reactor Source(bank_index:int(0), value:int(0)) {
    output out:int;
    reaction (startup) -> out {=
        SET(out, self->value);
    =}
}
main reactor {
    source = new[4] Source(value = {= table[bank_index] =});
}
```

The `table` defined in the preamble determines the parameter values to use for each bank member.

This change was motivated by @Soroosh129's PacMan game example, where I have made the ghosts into a bank rather than four individual reactor instances.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -2231,6 +2231,8 @@ public class CGenerator extends GeneratorBase {
      */
     public void generateParameterInitialization(ReactorInstance instance) {
         var selfRef = CUtil.reactorRef(instance);
+        // Declare a local bank_index variable so that initializers can use it.
+        initializeTriggerObjects.pr("int bank_index = "+CUtil.bankIndex(instance)+";");
         for (ParameterInstance parameter : instance.parameters) {
             // NOTE: we now use the resolved literal value. For better efficiency, we could
             // store constants in a global array and refer to its elements to avoid duplicate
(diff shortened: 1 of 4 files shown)


Soroosh129 (approved):
This is a neat idea. It's very useful. In the past, we had to resort to lambda functions, such as [here](https://github.com/lf-lang/experimental/blob/e420d206c3c13f328d9d465552f3e2ee8e72a67b/Python/src/Intersection/Intersection.lf#L409).

To clarify, I made ghosts into individual instances instead of members of a bank based on the comment that each ghost should have its own personality. The idea was then to subsequently create a reactor class for each ghost.

Alternatively, we could pursue the idea of enabling reactor classes to be passed as parameters. Then, the same Ghost reactor could be used with a parametrized Behavior reactor.

Or, there might be a third approach to creating unique behaviors for each ghost, while still enjoying the compactness that banks bring to the table.

cmnrd:
This is a great idea and I like that it is very general. Instead of a table lookup, we could also invoke a function and pass the bank index as argument for some computation.

We should also add this for the other targets. I am not sure what would be best: open an issue to not forget about this or attempt at adding support for all targets in this PR?

cmnrd (approved):

PR #1103, opened 2022-04-19
Update `SET`, schedule and tag APIs in the C target

housengw:


cmnrd:
I don't know the background of this change, but I would like to comment on another aspect of naming in C. "set" is a very common term and likely used in many C libraries. By using such a common function name (`SET` or `set`) we are at risk of getting name conflicts with external libraries. Since there is an API change anywhere, I think it is worth considering using a prefixed form `lf_set` to make name conflicts less likely in C.

lhstrh:
> Undefining the macros will have the side effect that `set` cannot be invoked in any function defined in a `preamble`. I think this is OK, but we need to understand that this is a restriction.  If we go with a more proprietary naming convention, as @cmnrd suggests, like `lf_set`, then it become less imperative to undefined the macros. Let's discuss.

Not being able to invoke `set` in a preamble sounds like a great feature to me, actually.

Soroosh129:
Another way to avoid this conflict once and for all is to put `set` and `schedule` (as pointers to functions) inside a global variable, such as `ctx` (inspired by the Rust target) or `lf`, or even `self`.

The programmer would then access them via `ctx->set`, etc.

housengw:
`lf->set` is less preferable in C than `lf_set` IMO, and the fact that `SET` is a macro might make it difficult to put inside a struct

Soroosh129:
> Not being able to invoke set in a preamble sounds like a great feature to me, actually.

I agree that this is a feature. 
We want to prevent asynchronous calls to `set` in the preamble, and I think it's just safer to disallow `set `outside reaction bodies altogether. However, no one can stop a programmer from passing a pointer to `set` to an asynchronous thread (or just an ordinary function defined in the preamble).

Soroosh129:
> [T]he fact that SET is a macro might make it difficult to put inside a struct.

You're right.

PR #1115, opened 2022-04-25
Improve run-time API for getting bank ids

lhstrh:
The changes in `reactor-ts` incorporated by the PR make it easier to find the bank id of an ancestor reactor instance.


edwardalee (commented):
The getBankIndex(number) function looks problematic to me. It will fail if the number is greater than the depth of the reactor. Also, why return -1 rather than 0 when it's not in a bank? Seems like an ordinary reactor is a special case of a bank of width 1.  Also, I'm not sure what the use case for this is. Is this function meant to be called from user reactor code?

lhstrh:
In `reactor-ts` there is no reserved parameter for `bank_index`. Instead, reaction code can just retrieve it using this `getBankIndex()` function. The generalization of this `getBankIndex` method was inspired by the example code you wrote. Specifically, it allows for the follow code to be simplied:
```
target TypeScript;
reactor Child (
    parentBankIndex:number(0)
) {
    output out:number;
    reaction(startup) -> out {=
        out = parentBankIndex * 2 + this.getBankIndex();
    =}
}
reactor Parent {
    c = new[2] Child(parentBankIndex = {= this.getBankIndex() =});
    reaction(c.out) {=
        for (let i = 0; i < c.length; i++) {
            console.log(`Received ${c[i].out} from child ${i}`)
        }
    =}
}
main reactor {
    p = new[2] Parent();
}
```
which turns into:
```
target TypeScript;
reactor Child {
    output out:number;
    reaction(startup) -> out {=
        out = this.getBankIndex(1) * 2 + this.getBankIndex();
    =}
}
reactor Parent {
    c = new[2] Child();
    reaction(c.out) {=
        for (let i = 0; i < c.length; i++) {
            console.log(`Received ${c[i].out} from child ${i}`)
        }
    =}
}
main reactor {
    p = new[2] Parent();
}
```

lhstrh:
> The getBankIndex(number) function looks problematic to me. It will fail if the number is greater than the depth of the reactor. 

What do you mean by "fail"? A top-level reactor is never in a bank, so in that case the return value will be `-1`.

> Also, why return -1 rather than 0 when it's not in a bank? 

Because `0` could be a valid index, whereas `-1` could not.

lhstrh:
Having a second look, I noticed the possibility of a NPE. I'll fix that.

edwardalee:
I don't think reactor code should reference a bank index other than its own. This breaks modularity. What if the reactor class gets used at the top level?

lhstrh:
But then why do we have an example in the documentation that does exactly this?

edwardalee:
I think the example you are referring to is this:
```
target C;
reactor Child (
    bank_index:int(0),
    parent_bank_index:int(0)
) {
    reaction(startup) {=
        printf(
            "My bank index: %d. My parent's bank index: %d.\n",
            self->bank_index, self->parent_bank_index
        );
    =}
}
reactor Parent (
    bank_index:int(0)
) {
    c = new[2] Child(parent_bank_index = bank_index);
}
main reactor {
    p = new[2] Parent();
}
```
Here, `parent_bank_index`, as far as `Child` is concerned, is just a parameter. The name is irrelevant. This reactor can be instantiated anywhere and that parameter can be set to anything.  The parameter gets associated with the parent's bank index on the line that instantiates `Child`, where it should be, because it is there that the code knows about the hierarchy.

lhstrh:
The code prints `"My parent's bank index: ...."` if that doesn't presuppose facts about the hierarchy that the reactor is embedded in, then I don't know what does :-D

edwardalee:
OK, you've convinced me that this is a bad example of this usage.  The parameter should be renamed `foo` rather than `parents_bank_index` and the print statement should say "foo = %d%". Then the confusion will go away.  The pattern is essential. I've needed it several times building examples. Without it, nested banks are pretty useless.

lhstrh:
But if the functionality is essential, then why rely on a pattern instead of an API like the one I propose? To me, the latter is cleaner.

cmnrd:
@lhstrh in your example, the Child reactor needs to make the assumption that the relevant parent is one level up in the hierarchy. If we modify the example by adding a reactor Foo to the hierarchy (between Parent and Child), this assumption does not hold anymore. With the pattern used by Edward, the problem is easily solved:
```
target TypeScript;
reactor Child (
    parentBankIndex:number(0)
) {
    output out:number;
    reaction(startup) -> out {=
        out = parentBankIndex * 2 + this.getBankIndex();
    =}
reactor Parent {
    c = new[2] Child(parentBankIndex = {= this.getBankIndex() =});
    reaction(c.out) {=
        for (let i = 0; i < c.length; i++) {
            console.log(`Received ${c[i].out} from child ${i}`)
        }
    =}
}
reactor Foo(parentBankIndex:number(0)) {
  c = new Child(parentBankIndex=parentBankIndex)
  output out:number
  c.out -> out
}
main reactor {
    p = new[2] Parent();
}
```

With the proposed API, the lookup depth could be provided as a parameter to Child. But then the code wouldn't be simpler anymore and the depth looks like a pretty weird parameter to me.

Note that in Edward's example the parameter `parent_bank_index` can have an arbitrary meaning and name. Let's call it `foo` for the sake of the argument. In the example, we happen to instantiate Child such that the bank index of the parent instance is assigned to `foo`. We could also choose any other assignment. Most importantly, we can instantiate Child outside of a bank and still assign a meaningful value to `foo`. With the proposed API, however, Child needs to make the assumption that it is instantiated from within a bank. This limits the reusability of Child and will cause unexpected runtime errors when used in other contexts.

lhstrh:
I had no use case for this feature until this example was introduced. I should stress that this was meant as an _additional convenience_, not as a replacement mechanism for parameters. Parameter work as expected in `reactor-ts` and the example currently checked into the `website-lingua-franca` uses this mechanism to pass in the `parent_id`. I removed the feature and will close this PR.

PR #1118, opened 2022-04-28
Baremetal spike support

sberkun:
Add C support for the risc-v spike platform. Also comes with code generation support for any of the 4 platforms (mac, windows, linux, spike) by setting the "platform" flag.

Also, no longer pulls in every platform support file for all platforms, only pulls in the specific support files needed. Could be used in the future to facilitate cross-compilation and expand support to more platforms. In addition, this could be used to replace the OS macros in https://github.com/lf-lang/reactor-c/blob/main/core/platform.h, which would make custom platforms significantly easier.

### DIFF
--- a/org.lflang/src/org/lflang/TargetProperty.java
+++ b/org.lflang/src/org/lflang/TargetProperty.java
@@ -1342,6 +1342,7 @@ public enum TargetProperty {
         ARDUINO("Arduino"),
         LINUX("Linux"),
         MAC("Darwin"),
+        SPIKE("Spike"),
         WINDOWS("Windows");
 
         String cMakeName;
--- a/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
@@ -146,6 +146,7 @@ class CCmakeGenerator {
         cMakeCode.pr("${LF_MAIN_TARGET}");
         sources.forEach(cMakeCode::pr);
         cMakeCode.pr("${CoreLib}/platform/${LF_PLATFORM_FILE}");
+        cMakeCode.pr("${LF_ADDITIONAL_SOURCES}");
         additionalSources.forEach(cMakeCode::pr);
         cMakeCode.unindent();
         cMakeCode.pr(")");
--- a/org.lflang/src/org/lflang/generator/c/CCoreFilesUtils.java
+++ b/org.lflang/src/org/lflang/generator/c/CCoreFilesUtils.java
@@ -23,6 +23,7 @@ public class CCoreFilesUtils {
         List<String> coreFiles = new ArrayList<>();
         coreFiles.addAll(getBaseCoreFiles());
         coreFiles.addAll(getPlatformFiles());
+        coreFiles.addAll(getSpikeFiles());
         if (isFederated) {
             coreFiles.addAll(getFederatedFiles());
         }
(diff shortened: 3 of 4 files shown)


Soroosh129:
Hi @sberkun 

Thanks for submitting this PR.

Could you please explain the following change:

> Also, no longer pulls in every platform support file for all platforms, only pulls in the specific support files needed.

To provide a bit of a background, we put CMake in charge of choosing the platform support files because of federated (distributed) programs. The idea is that you could have a federation where a federate runs on macOS and a federate runs on Linux. With the current design, you could invoke `lfc example_federate.lf` and then move the generated code for each federate to whatever platform that you want (as long as it is supported).

Also, would you please create an accompanying PR in reactor-c with your changes?
Moreover, I noticed that it looks like the reference to reactor-cpp is being changed in your PR. Is that also part of your proposed changes?

cmnrd (changes requested):
Looks like reactor-cpp got reverted to some older ref. This should be fixed before merging

sberkun:
Hi @Soroosh129, thank you for the feedback.

Since part of the functionality needed to support the spike platform is a way to conditionally include support files (in particular, the spike platform has a Makefile that should not be included when generating code for other platforms), I thought it would be nice to use that functionality for the platform files as well. 

However, I didn't know that the federated programs supported code generation for mixed platforms, so I'll go back and remove this change. 

I did not mean to change the reference to reactor-cpp, I probably made a mistake when merging in the recent changes from master. Thank you for reminding me about making a PR to reactor-c, I'll do that once I fix the issues here.

sberkun:
Sorry, it's been a while, so I've rebased my changes off of the current master branch. Corresponding reactor-c PR is https://github.com/lf-lang/reactor-c/pull/84.

This time, the generator always pull in spike support files rather than conditionally pulling them in; they're all contained to the core/spike/ folder (previously, I had the Makefile be copied to the root of the generated code).

I make a folder for the spike tests (tests/C/src/spike/). Given that anyone testing the spike platform will need riscv64-unknown-elf-gcc and the spike simulator installed, should we try to integrate these tests into the testing suite? Or should we just leave these as "optional" tests?

Soroosh129:
> Sorry, it's been a while, so I've rebased my changes off of the current master branch. Corresponding reactor-c PR is https://github.com/lf-lang/reactor-c/pull/84.
>
> This time, the generator always pull in spike support files rather than conditionally pulling them in; they're all contained to the core/spike/ folder (previously, I had the Makefile be copied to the root of the generated code).

Thanks! This looks great. What is the procedure now for building the included tests manually? It seems like instructions and comments are missing on those files. I think adding user instructions (maybe a new page on the [website](https://github.com/lf-lang/website-lingua-franca)?) is the only important remaining to-do before merging this.

> I make a folder for the spike tests (tests/C/src/spike/). Given that anyone testing the spike platform will need riscv64-unknown-elf-gcc and the spike simulator installed, should we try to integrate these tests into the testing suite? Or should we just leave these as "optional" tests?

Is there a ready-made Docker image that already has what we need? In any case, creating a specific CI task for Spike tests is definitely possible. It is also possible to install `riscv64-unknown-elf-gcc` in our existing CI test for C. For the moment, you can move the added tests to the [experimental](https://github.com/lf-lang/experimental-lingua-franca) (or [examples](https://github.com/lf-lang/examples-lingua-franca)) repo.

lhstrh:
@lsk567 and @sberkun: what is the status of this PR?

sberkun:
@lhstrh sorry for the late reply: feels closer, see discussion in the corresponding [reactor-c pr](https://github.com/lf-lang/reactor-c/pull/84).

lhstrh:
OK, left some comments there.

lhstrh:
@sberkun and @lsk567, what should we do with this PR?

lsk567:
This PR depends on the corresponding PR in reactor-c, right? @sberkun

PR #1119, opened 2022-04-28
diagram: Add categories explicitly to the synthesis options.

soerendomroes:
Since the LSP does not allow sending cyclic structures via JSON the
category of a synthesis option is only retrievable by its id. If the
category is not also added as a synthesis option, klighd-vscode omits all

options for which the category is not retrievable. Adding the categories
explicitly solves this.
Bad:

Screenshot from 2022-04-28 12-59-08
Good:
Screenshot from 2022-04-28 13-00-03


a-sr (approved):
LGTM

lhstrh (approved):
Great! Thanks for fixing this!

PR #1121, opened 2022-04-28
Add support for the `link-libs` target property

### LINKED ISSUE #1120: Specifying libraries to link is too difficult
A common use case is that an imported LF reactor requires an external library, such as the math library in C. Unfortunately, we currently have no good way for the imported file to be self contained and specify this requirement.  Moreover, the preferred way we have currently requires that the user of the imported file create a `math-include.cmake` file with this:
```
target_link_libraries( ${LF_MAIN_TARGET} m )
```
and then specify a target property:
```
   cmake-include: "math-include.cmake”
```
Alternatively, if the user is using gcc or clang, then just the following target property will work:
```
   flags: "-lm"
```
This falls pretty low on the usability scale.

I suggest a new target property, something like:
```
   link-libraries: ["m"]
```
This could be a list of required libraries. This would be put in the imported file and would need to be translated into the appropriate lines in the generated cmake file or into compiler flags if not.  If multiple imported files specify the same libraries, these would be consolidated.

Other (better) ideas?

Soroosh129:
Fixes #1120.


Soroosh129:
I'm not sure how to update the reference to the benchmarks to point here: lf-lang/benchmarks-lingua-franca#11

@lhstrh do you happen to know how this can be done?

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
   # Run the C benchmark tests. 
   c-benchmark-tests:
-    uses: lf-lang/benchmarks-lingua-franca/.github/workflows/benchmark-tests.yml@main
+    uses: lf-lang/benchmarks-lingua-franca/.github/workflows/benchmark-tests.yml@link-libraries
```
lhstrh, line 40: This will only use a different version of the workflow definition, which hasn't changed, so this is not what you want to do.
I'm a little unsure as to what we do want to do; we're currently not using the workflow parameters, which are:
```
      compiler-ref:
        required: false
        type: string
      benchmarks-ref:
        required: false
        type: string
        default: main
```
I think we may want to specify `compiler-ref` to be `${{ github.ref }}`. I think it's fine not to specify `benchmarks-ref` and just test against `main`, but if you'd like to test against `link-libraries` you could do that. The problem is that after merging it needs to be reverted back to `main`...

cmnrd (commented):
Please see my comment in https://github.com/lf-lang/lingua-franca/issues/1120

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -227,6 +228,22 @@ CodeBuilder generateCMakeCode(
         }
         cMakeCode.newLine();
         
+        // Add the link libraries (if any)
+        for (var lib : targetConfig.linkLibraries) {
+            // Need a special case for libm on Windows
+            if (lib.trim() == "m") {
```
cmnrd, line 234: This special casing underlines my concern raised in https://github.com/lf-lang/lingua-franca/issues/1120#. As far as we know, this solution only works for the math library (as we have no tests for other libraries), and yet it already needs additional logic to work on all platforms. What happens if we support more libraries than just math? Should we have special cases for all of them?

petervdonovan:
Incidentally, the I actually do not understand why the LSP tests are failing in this branch. I tried to reproduce the Python validation test failure on my machine but could not, and instead, I had a test failure in Rust validation tests, which I should think would be unaffected by these changes. I thought that the random seed was set up so that this would not happen. I wonder if the LSP tests should be disabled temporarily until I find a way to make them less complex.

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
edwardalee: I think that handling special cases in this way is appropriate and could lead to an overall better user experience.  Use of the math library is very common in C programs, and it really unfortunate that the it has to handled differently on different platforms.  But once we understand those needs, why leave it users to have rediscover these quirks themselves?

lhstrh:
I would like to point out that there is already a field `compileLibraries` in `TargetConfig`. There appears to be a proliferation of competing mechanisms...

lhstrh:
One odd thing about this design is that `m` isn't, in fact, the library. The library is `math`. It's just that the common compiler flag to link it is `m`. I think that if we were to generalize, we should use that library names...

lhstrh:
Also, it appears that math is actually the _only_ standard library that requires an extra flag (at least with gcc), so I think we've already created excess generality with this target property. It seems that a property `link-math` could possibly be sufficient. I'm now running tests to see what flags are really needed for standard libraries across the platforms we support.

test/C/src/target/LinkLibrariesNoCMake.lf
```diff
@@ -0,0 +1,15 @@
+/**
+ * Test that the link-libraries target property is working correctly.
+ * The failure for this test is failure to compile.
+ */
+target C {
+    link-std-libs: "m",
```
petervdonovan, line 6: I am confused -- please forgive me. On my machine, this test compiles even if this line is deleted. It even compiles using a version of `lfc` that is not from this branch. What is this test supposed to verify?

(Incidentally, `LinkLibraries.lf` also compiles on my machine even if I delete the preamble, albeit with a warning. I also do not understand why `link-std-libs: "m"` is commented out there.)

lhstrh: I commented out that line to verify whether it would only compile on Linux or also on other platforms. The test was contributed by @Soroosh129 without that line not commented out. But when I commented it out it still worked, so I'm trying to figure out what problem we're trying to fix here...

lhstrh:
I think the only reasonable thing to do is to not make this a feature of `0.2.0` and first make sure this solution is solid. We can always publish a patch version that includes this feature.

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -247,4 +250,36 @@ CodeBuilder generateCMakeCode(
...
+     * @param cb The code builder to append to.
+     */
+    private void addLinkedLibraries(CodeBuilder cb) {
+        StringBuffer block = new StringBuffer();
+        for (var lib : targetConfig.linkLibraries) {
+           var findLib = "find_library( LF_%1$s_LIB %1$s )".formatted(lib);
+           var linkLibs = """
+                   target_link_libraries( ${LF_MAIN_TARGET} "${LF_%s_LIB}")""".formatted(lib);
+           switch (lib.trim()) {
+                case "m":
```
housengw, line 265: If we are only special casing the math library, this function might be an overkill. 
If we plan to special case other libraries, I think going with`cmake-include` is a better solution (as mentioned [in #1120](https://github.com/lf-lang/lingua-franca/issues/1120#issuecomment-1113009360)).

cmnrd:
> I think that handling special cases in this way is appropriate and could lead to an overall better user experience. Use of the math library is very common in C programs, and it really unfortunate that the it has to handled differently on different platforms. But once we understand those needs, why leave it users to have rediscover these quirks themselves?

The problem that I see is that this knowledge base will never be complete, and it needs to be maintained (as libraries and their usage change over time). If we promise users that they can just specify the library they want to link to and it works on any platform, then we need to fulfill this promise somehow. The effort required to do this seems extensive to me. If we really want to go down that path, we should at least be transparent about precisely which libraries are supported. We should have tests for every library we claim to support and issue validation errors (or at least warnings) if users try to use the target property for linking other libraries.

Soroosh129:
I see that there has been a lively discussion here in my absence.
I apologize for my late response.

In terms of a special handling for `libm`, the unfortunate situation is that some systems require the `-lm` flag and some don't. This goes back to older days where some implementations of the C standard library (`libc`) separated out the math functions' implementation into a separate library. Newer implementations of `libc` seem to already incorporate the math functions, so there is no need to link against `libm`. Most newer compilers, including GCC and Clang, just ignore the `-lm` if you pass it, but, unfortunately, the MSVC compiler (which is a compiler that we support) patently refuses to compile if `libm` is linked.

If you want to use the math functions in your code and ensure maximum compatibility, you need to add the following to your [cmake-includes](https://www.lf-lang.org/docs/handbook/target-declaration#cmake-include):
```cmake
IF (NOT WIN32)
  find_package(m REQUIRED)
  target_link_libraries(${LF_MAIN_TARGET} m)
ENDIF()
``` 
The math library is really an oddity in the C standard library. As far as I know, there is no other standard library for C that needs to be specifically linked. Hence, I think it was justified to provide an easy shortcut for this in the form of `link-libraries: ["m"]`, as suggested originally in #1120. An alternative would be to provide a built-in `mlib.cmake` to allow for `cmake-include: ["mlib.cmake"]`.

In terms of the usefulness of `link-libraries` in general, it is really a shorthand for the following:
- If `cmake: false`, pass the specified libraries via the `-l` flag.
- If `cmake: true`, pass the specified libraries via
  ```cmake
  find_library( LF_FOO_LIB foo )
  target_link_libraries( ${LF_MAIN_TARGET} "${LF_FOO_LIB}")
  ```

This is providing a shorthand for a scenario that can arise many times. Note that this mechanism supports `cmake: false` as well. Of course, cmake-include will always be there if a more advanced situation arises (although, you would have to leave `cmake: false` behind). Think of `link-libraries` as the easy mode for a common pattern.

Finally, I envisioned `link-libraries` as a method that also allows for inclusion of our own [utility libraries](https://github.com/lf-lang/reactor-c/tree/main/util) (e.g., `link-libraries: ["deque"]`), but this could wait until a future PR.

I see that the the `link-libraries` is renamed to `link-std-libs`. This sounds wrong to me. I think if we want to keep it, the `link-libraries` name sounded more appropriate.

lhstrh:
My suggestion would be to wait with finalizing this until @cmnrd returns from his vacation so we can reach agreement on what the final solution should look like.

Soroosh129:
Tagging @cmnrd now that he's back to see if we can reach a consensus. @cmnrd, please see my [comment](https://github.com/lf-lang/lingua-franca/pull/1121#issuecomment-1134515134) above.

cmnrd:
With regard to the math library, why not add `-lm` by default? Is there a case where we really don't want to add the math library?

Ok, so if we see `link-libraries` as a shortcut for a specific piece of cmake code, then I think it is reasonable to include it. We should then specifically describe in the documentation which cmake code it produces, in which scenarios it works, and how the user can add manual cmake code when it doesn't work. From the initial proposal, I understood `link-libraries` as a general solution to the problem of finding and linking dependencies, which I don't think it is possible. If we label this as a limited solution that works in some (or many?) cases, I am fine with including it.

org.lflang/src/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
```diff
@@ -133,6 +133,21 @@ class CppStandaloneCmakeGenerator(private val targetConfig: TargetConfig, privat
                 |)
                 |target_link_libraries($S{LF_MAIN_TARGET} $reactorCppTarget)
                 |
+            ${" |"..targetConfig.linkLibraries.joinToString("\n") {
+                    buildString {
+                        // Need a special case for libm on Windows
+                        if (it.trim() == "m") {
+                            append("if(NOT MSVC)\n")
```
cmnrd, line 140: Do we need to include libm for C++?

org.lflang/src/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
```diff
@@ -133,6 +133,21 @@ class CppStandaloneCmakeGenerator(private val targetConfig: TargetConfig, privat
...
+                        }
+                        append("find_library(LF_${it}_LIB $it)\n")
+                        append("target_link_libraries($S{LF_MAIN_TARGET} \"$S{LF_${it}_LIB}\")\n")
+                        if (it.trim() == "m") {
+                            append("endif()\n")
+                        }
+                    }.toString()
+                }
+            }
+                |
```
cmnrd, line 150: The added code snippet really breaks with the style of the class. It would be best to factor out this code in a separate (Java?) function and call it here.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1261,7 +1261,7 @@ protected void initializeClockSynchronization() {
                     // FIXME: This is a linker flag not compile flag but we don't have a way to add linker flags
```
Soroosh129, line 1261:
```suggestion
```

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -198,7 +198,16 @@ CodeBuilder generateCMakeCode(
         for (String compilerFlag : targetConfig.compilerFlags) {
             switch(compilerFlag.trim()) {
                 case "-lm":
+                    // Need a special case for libm on Windows
```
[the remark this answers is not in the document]
Soroosh129: Right, but we do support MSVC. I don't think we should abandon that just for this library.

org.lflang/src/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
Soroosh129: You're right, we don't.

org.lflang/src/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
Soroosh129: I updated the code and removed `find_library` since it appears to be unnecessary.

Soroosh129:
The c-benchmark tests will fail because lf-lang/benchmarks-lingua-franca#11 is not merged yet. I have no idea how to proceed. Could you help me @lhstrh, please?

PR #1126, opened 2022-04-29
Fix set executing_reaction in self struct

### LINKED ISSUE #1123: Segfault when calling `_lf_check_deadline`
The root cause of the segfault is that`self->executing_reaction` is `NULL` inside `_lf_check_deadline`, so when the check deadline function tries to dereference `self->executing_reaction` to get the deadline, it dereferences a `NULL` pointer.

Minimal Example to reproduce:
```
target C;

main reactor {
    reaction(startup) {=
        lf_check_deadline(self, true);
    =}
}
```

Running this code gives a segfault.
This bug has caused the `AnytimePrime.lf` example in the experimental repo to fail.

housengw:
Fixes #1123

### DIFF
--- a/test/C/src/DeadlineAnytime.lf
+++ b/test/C/src/DeadlineAnytime.lf
@@ -1,0 +1,30 @@
+// Test whether the lf_check_deadline function works.
+target C;
+
+reactor A {
+    state i:int;
+    logical action a;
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
Nice! Thanks for adding the test also.

PR #1131, opened 2022-05-02
Fixed a few calls to deprecated functions that were missed

edwardalee:
The C and Python code generators were still generating some calls to deprecated functions like `error_print_and_exit` which causes release 2.0 to generate code that issues warnings when compiled. This PR fixes the ones I have found.


housengw (approved):
LGTM! Thanks for the change

PR #1133, opened 2022-05-02
Add support for interleaved operator to Rust target

jhaye:
This adds support for the `interleaved` operator to the Rust target, including a unit test.

The test passes but has to be compiled with threading disabled, otherwise the output gets garbled.


cmnrd (approved):
Looks good!

cmnrd:
Regarding the garbled output: We should introduce a thread-safe logging API for Rust (see https://github.com/lf-lang/lingua-franca/issues/1138).

PR #1134, opened 2022-05-02
Update references to deprecated print apis in python target

housengw:
This PR updates some missed references to deprecated print APIs in the python target.


edwardalee (approved):
LGTM.

PR #1135, opened 2022-05-03
Let package_lfc script put files in appropriately named directory

lhstrh:


cmnrd (approved):
Looks good! Is the version change in `gradle.properties` intended? And should we simply remove the 'SNAPSHOT' from the version name?

lhstrh:
The `SNAPSHOT` suffix is a Maven convention that I think we should abide by as long as we're tied to Maven and OSGi in any fashion. Basically, what's in `master` will always be "snapshot"; only tagged revisions in `release` will be actual releases and thus not labeled as a snapshot version. To make things extra confusing, OSGi uses the suffix `.qualifier` to mean the same thing. If any of these things go out of sync, the build breaks. I know. Don't get me started...

PR #1136, opened 2022-05-03
Check for Java 17 in lfc scripts and properly exit on error

cmnrd:
This updates the build and launch lfc scripts to check for a Java 17 runtime (instead of 11). Also it introduces a small fix that let's the script exit gracefully if no Java 17 runtime was found.

### DIFF
--- a/bin/lfc.ps1
+++ b/bin/lfc.ps1
@@ -42,8 +42,8 @@ if (-not (Test-Path $java_cmd)) {
 
 # check for correct java version
 $java_version = (Get-Command java | Select-Object -ExpandProperty Version).toString()
-if ([version]$java_version -lt [version]"11.0") {
-    throw "JRE $java_version found but 11.0 or greater is required."
+if ([version]$java_version -lt [version]"17.0") {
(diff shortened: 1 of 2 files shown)


lhstrh (changes requested):
Nice catch! We should update the ps1 script accordingly as well...

petervdonovan (approved):
I can confirm that this works on Windows, correctly producing the following error:
```
JRE 11.0.12.0 found but 17.0 or greater is required.
At C:\lingua-franca\bin\lfc.ps1:46 char:5
+     throw "JRE $java_version found but 17.0 or greater is required."
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (JRE 11.0.12.0 f...er is required.:String) [], RuntimeException
    + FullyQualifiedErrorId : JRE 11.0.12.0 found but 17.0 or greater is required.
```

lhstrh:
Excellent. Thank you for checking, @petervdonovan!

lhstrh (approved):

Issue #1096, opened 2022-04-12
Smarter default number of threads


edwardalee:
Whenever I run an LF program on my laptop, I get by default 16 worker threads. For a small program, this adds a lot of overhead.  It would be easy to analyze the program while calculating levels to come up with a smarter default. That analysis already calculates the depth of the reaction graph.  It could also calculate the width and cap the default number of threads with that number.

erlingrj:
The issue remains for C++ and Rust. For Cpp it would be possible to adress the issue in a similar way. But then we need to create the ReactionInstanceGraph in the CppGenerator also.

In another issue you indicated that this might not be desirable since the Cpp target does that at runtime during boot.

lhstrh:
Inviting @cmnrd and @jhaye to comment on this.

jhaye:
In the Rust runtime we use [rayon](https://crates.io/crates/rayon) to spread work across threads. If no `workers` argument is supplied, it creates a thread pool with the number of threads equal to the number of logical CPUs.

[It is documented](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.num_threads) that this behaviour might change in the future, where if the number of threads is not specified for the tread builder, it could automatically manage the number of threads to maximise performance.

> If num_threads is 0, or you do not call this function, then the Rayon runtime will select the number of threads automatically. At present, this is based on the RAYON_NUM_THREADS environment variable (if set), or the number of logical CPUs (otherwise). In the future, however, the default behavior may change to dynamically add or remove threads as needed.

So we might just have to wait a bit for this issue to fix itself.

Soroosh129:
> Currently the strategy is to use as many workers as there are hardware threads. 

Well, this used to be the strategy in the C runtime as well. The new smarter strategy proposed by @edwardalee and implemented by @erlingrj looks at the maximum width of the reaction graph for a given level and chooses the minimum of that and the number of hardware threads as the default.

I don't think the strategy in the C++ runtime addresses this issue.

cmnrd:
No, indeed it does not. However, I also currently don't see need to implement this strategy as, so farm I didn't see a significant penalty for having more threads. So I am opting for simplicity here. If someone is willing to implement other strategies and shows that there are indeed benefits, I am happy to accept PRs :)

edwardalee:
I think you may not be seeing the overhead because of the benchmark runner. The memory and startup overhead could be significant, particularly for embedded platforms.

PR #1139, opened 2022-05-03
Smarter default number of workers

erlingrj:
I thought I'd address some  minor issues as a way of familiarizing myself with the codebase. 

This is adressing issue with a smarter default number of workers #1096. Currently only implemented for C. In case the user does not specify the number of worker threads, we use the minimum of the number of cores and the breadth of the Reaction graph. With this proposal the C (and in the future the C++ and Rust) runtime will always be configured with a specific number of workers.

Questions:
1. Is it a reliable approach to find number of cores through Java runtime?
2. Could this be done higher in the class hiererachy? Such that we dont have to reimplement it for all target languages? A problem I saw is that the ReactionInstanceGraph does not seem to be built in the abstract GeneratorBase class but rather in the derived classes
3. In this approach we will suggest 4 threads if there is a single Reactor with 4 reactions. This might not be what we want as they will never execute in concurrently. 

When questions are resolved I can port to the other target language generators


Soroosh129:
Hi @erlingrj 

Thanks for submitting this PR.

The threaded C runtime determines the number of workers at [runtime](https://github.com/lf-lang/reactor-c/blob/2ceec60bc3d9b3420553f66813c40790d1e28c85/core/threaded/reactor_threaded.c#L1092). I personally prefer the existing approach because we use the number of cores of the machine that executes the program, not the number of cores on the machine that code generates the program.

In keeping with the current scheme of things, the calculated breadth would simply have to be conveyed to the C runtime.

But, if you think we should determine the number of cores statically at code generation time, we should remove the duplicate logic in the C runtime.

erlingrj:
Another alternative would be to pass the breadth as a flag through CMake. Then the runtimes could do the comparison of available cores and breadth of reaction graph themselves.

In my opinion, it would be better to not do the reaction-graph analysis during run-time.  But I think you have a more qualified opinion than me

Soroosh129:
That sounds like a great strategy.

I think your code can slightly be adjusted to accommodate this design. In [createMainReactorInstance](https://github.com/lf-lang/lingua-franca/blob/08825e6bbeb524a284e39b2d787a9ad45edeb088/org.lflang/src/org/lflang/generator/c/CGenerator.java#L2740), you could add the following logic:

```diff
if (this.mainDef != null) {
    if (this.main == null) {
        // Recursively build instances. This is done once because
        // it is the same for all federates.
        this.main = new ReactorInstance(toDefinition(mainDef.getReactorClass()), errorReporter, 
            this.unorderedReactions);
        var reactionInstanceGraph = this.main.assignLevels();
        if (reactionInstanceGraph.nodeCount() > 0) {
            errorReporter.reportError("Main reactor has causality cycles. Skipping code generation.");
            return;
        }
-
+        if (targetConfig.workers == 0) {
+            targetConfig.compileDefinitions.put( 
+                "LF_REACTION_GRAPH_BREADTH",   
+                 reactionInstanceGraph.availableProcessors()
+             )
+         }
        // Force reconstruction of dependence information.
        if (isFederated) {
            // Avoid compile errors by removing disconnected network ports.
            // This must be done after assigning levels.  
            removeRemoteFederateConnectionPorts(main);
            // There will be AST transformations that invalidate some info
            // cached in ReactorInstance.
            this.main.clearCaches(false);                    
        }
    }   
}
```

The C runtime can then take `LF_REACTION_GRAPH_BREADTH` into consideration in [determine_number_of_workers](https://github.com/lf-lang/reactor-c/blob/2ceec60bc3d9b3420553f66813c40790d1e28c85/core/threaded/reactor_threaded.c#L1092) (similar to the way it considers `NUMBER_OF_WORKERS`).

edwardalee:
You say:

3. In this approach we will suggest 4 threads if there is a single Reactor with 4 reactions. This might not be what we want as they will never execute in concurrently.

But if the reaction graph breadth is getting calculated correctly, the breadth should be 1 in this case.

erlingrj:
I added your suggestion @Soroosh129. There was some confusion that maybe you can help me with.

What is it that ReactionInstanceGraph is referring to with numReactionsAtLevel? I am not really sure what it is counting, but it is not what I expected. To find the breadth of the graph I added my own logic into the assignLevels functions. And count the breadth with "numIndependentReactionsPerLevel".

An example of this is the following graph:
Screenshot from 2022-05-05 13-29-36

Where numReactionsPerLevel= [3,2,2,1]
and my newly added numIndependentReactionsPerLevel=[1,2,2,1]

edwardalee:
In this example, I'm pretty sure what we want is numIndependentReactionsPerLevel=[1,2,2,1].

erlingrj:
I will look into that

erlingrj:
I merged them now. It works on a few examples here.

Soroosh129 (approved):
Looks good to me! Thank you for taking care of this issue.

We should perhaps do a performance comparison before and after this change but I don’t think that should block merging this in.

erlingrj:
How do you structure PRs that include multiple git-repos? I have made a PR in reactor-c with the same name [here](https://github.com/lf-lang/reactor-c/pull/74)

Soroosh129:
> How do you structure PRs that include multiple git-repos? I have made a PR in reactor-c with the same name [here](https://github.com/lf-lang/reactor-c/pull/74)

You can update the hash that is pointing to a particular commit in reactor-c via a `git add org.lflang/src/lib/c/reactor-c` and a `git commit -m “Updated reactor-c”`.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -2744,22 +2745,35 @@ private void createMainReactorInstance() {
...
-                if (isFederated) {
-                    // Avoid compile errors by removing disconnected network ports.
-                    // This must be done after assigning levels.  
-                    removeRemoteFederateConnectionPorts(main);
-                    // There will be AST transformations that invalidate some info
-                    // cached in ReactorInstance.
-                    this.main.clearCaches(false);                    
+                // Inform the run-time of the breadth/parallelism of the reaction graph
+                var breadth = reactionInstanceGraph.getBreadth();
+                if(breadth == 0) {
```
lhstrh, line 2755:
```suggestion
                if (breadth == 0) {
```

lhstrh (approved):

lhstrh:
This looks good, but I see some conflicts arose in the meantime. If you can fix these, we can quickly go ahead and merge this in. Thanks!

erlingrj:
I merged the the PR locally.

PR #1140, opened 2022-05-04
Disable pylint `c-extension-no-member` warning

housengw:
After the API changes in the python target in #1116, the time API in the python target became `lf.time.<function_name>`.

Since `lf` is a c extension of python, pylint does not have python source code to reference the content of `lf` and complains about `Module 'lf' has no 'time' member`. 

A temporary fix is to disable `c-extension-no-member` warnings globally in the generated `.py` file.


lhstrh:
Seems reasonable as a stop-gap measure, but what would be a permanent look like?

petervdonovan (approved):
This looks right to me -- thank you! Later I may try to re-enable this feature.

PR #1150, opened 2022-05-10
Major refactor of the docker generators

housengw:
The purpose of this PR is to improve the abstraction of `DockerGeneratorBase` such that support for docker file generation in future target languages can be more easily implemented without changing the `DockerGeneratorBase` or other extended docker file generators.


org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
```diff
@@ -47,75 +57,88 @@ public class DockerGeneratorBase {
...
-        moduleNameToData.put(lfModuleName, k);
-        nFederates++;
-        appendFederateToDockerComposeServices(
-            composeServices,
-            (String) k.get(Key.DOCKER_COMPOSE_SERVICE_NAME),
-            (String) k.get(Key.DOCKER_BUILD_CONTEXT),
-            dockerFilePath.getFileName().toString());
+     * @param generatorData Data from the code generator.
+     */
+    public void addFederate(GeneratorData generatorData) {
```
lhstrh, line 137: Not sure we need federation lingo here. How about `addProcess` or `addProgram` or `addDeployment` or something in that vain?

lhstrh: Or just `addFile`...

org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
```diff
@@ -124,92 +147,103 @@ public void addFederate(
...
 
+    /**
+     * Throws a Runtime Exception if any field in "dockerData" is null.
+     */
+    private void validateNotNull(DockerData dockerData) {
+        if (dockerData.getFilePath() == null || dockerData.getFileContent() == null ||
+            dockerData.getComposeServiceName() == null ||
+            dockerData.getBuildContext() == null ||
+            dockerData.getContainerName() == null) {
+                throw new RuntimeException("Fields in DockerData cannot be null");
```
lhstrh, line 196: Based on the logic in this function, it appears that `filePath` or `fileContent` may be `null`. Either the message has to be made more specific (e.g., `filePath` or `fileContent` field must be set) or more generic (e.g., Missing fields in `DockerData` instance).

org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
```diff
@@ -2,18 +2,33 @@
...
+ *   interface of this class. This is the target code generator-facing API.
+ *
+ * The purpose of this abstraction design is to contain all
+ * docker-specific information inside docker generator classes and
+ * prevent docker-related information from polluting target code generators.
+ *
  * @author{Hou Seng Wong <housengw@berkeley.edu>}
  */
-public class DockerGeneratorBase {
+abstract public class DockerGeneratorBase {
```
lhstrh, line 31: Are different implementations of this class expected to differ much?

org.lflang/src/org/lflang/generator/c/CDockerGenerator.java
```diff
@@ -1,38 +1,106 @@
...
+         * The name of the LF module in CGenerator.
+         * Typically, this is "fileConfig.name + _ + federate.name"
+         * in federated execution and "fileConfig.name" in non-federated
+         * execution.
+         */
+        private String lfModuleName;
+        /**
+         * The value of "currentFederate.name" in CGenerator.
+         */
+        private String federateName;
```
lhstrh, line 33: Same concern here about federation lingo.

lhstrh (commented):
Generally looks good; left some comments.

org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
housengw: This is a good point. I will change the name to `addFile`

org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
housengw: good point. I will change the error message to what you suggested.

org.lflang/src/org/lflang/generator/DockerGeneratorBase.java
housengw: the main difference between different implementations is the way they populate the `DockerData` class based on `?GeneratorData` class, which is target language specific

org.lflang/src/org/lflang/generator/c/CDockerGenerator.java
housengw: this is OK since `CGeneratorData` is specific to docker file generation in the C target, and based on the current design of `CGenerator` the value `federate.name` is used to generate the docker file.

lhstrh (approved):

PR #1161, opened 2022-05-11
Fix segfault in python target argument handling

housengw:
This PR fixes the problem described in the email chain where federated execution in python was segfaulting.


Soroosh129 (approved):
LGTM!

It is a bit concerning that our CI tests were passing while memory was clearly being corrupted.

Also, the following warning (which I believe is new) is still present, but it is most likely due to an unrelated issue:
```
Federate 1: WARNING: Attempt to update the last known status tag of network input port 0 to an earlier tag was ignored.
```

PR #1164, opened 2022-05-12
Restore the ability to put the RTI on a remote machine

edwardalee:
Somewhere along the way, the ability to specify a remote machine for the RTI, e.g.:
```
federated reactor at wessel.eecs.berkeley.edu { ... }
```
got broken. The generated script was incorrect and would fail to start the RTI. Now it works, although there is a rather ad hoc five second wait for the RTI to start up.


Soroosh129 (approved):
LGTM!

The ad-hoc wait time is an unfortunate side-effect of the way federates actually look for the RTI (by rotating through thousands of network ports).

PR #1165, opened 2022-05-12
Add federated docker file generation for TS target

housengw:
Add federated docker file generation for TS target


lhstrh (approved):
Looks good to me!

Issue #79, opened 2019-12-18
Write conflicts between reactors and reactions


lhstrh:
When a connection of the form `c.x -> y` is made, there could potentially be a reaction that has `y` as an anti-dependency. In this case, I think, the connection should not be allowed (it could give rise to nondeterminism). We are currently not performing any such checks.

edwardalee:
I don’t understand.  Don’t we just have to define the order in which contained reactors react relative to contained reactions? Can you give a concrete example that isn’t resolved by, say, invoking reactions before contained reactors?

Edward

---------
Edward A. Lee
EECS, UC Berkeley
eal@eecs.berkeley.edu
http://eecs.berkeley.edu/~eal

lhstrh:
Reactors don't react; reactions do. In the construction of the precedence graph we abstract away reactors entirely. The only way to define order you suggest, is to introduce insert additional edges into the graph. We can do this (under the hood, at compile time) but this seems like a rather opaque solution to me. I think it's clearer to disallow it, just like we do with many-to-one connections between ports. In either case: right now, I think this situation remains unchecked. I'll add an example to illustrate my point, but I probably won't get to it before Friday.

lhstrh:
Erroneous commit message on c77b688e. This is the issue that was fixed.

Issue #68, opened 2019-12-02
Dependency on `cmake`


lhstrh:
The C++ code generator uses `cmake` in the build process. On Mac, `cmake` is put into `/opt/bin`, which is not on the default `PATH`, it turns out. The fix appears to be to add `opt/bin` to `/etc/paths`. We can improve error messages to make clear what the actual problem is, but we suspect that many Mac users will end up running into this issue. The question is: do we really need/want `cmake`?

cmnrd:
I personally want cmake and I think we actually need it. As I see it, there are
two alternatives to cmake:

1. We could fall back to plain Makefiles
2. We could use another advanced build system (scons for instance)

Both options are bad in my opinion. Option 1 is bad because writing makefiles
is pain. It works well for small projects that should run on a specific
platform, but as soon as you want to support multiple compilers on multiple
platforms linking to different versions of the STL and potentially other
libraries, things become complex. Cmake solves this. Of course there are some
smaller problems here and there (as we had with MacOS), but once the
configuration works a rather simple cmake script allows you to build your
project on any of the major plarforms using any of the major C++ compilers.

I think option 2 is bad because switching the build system would only move the
dependency. I expect similar (if not worse) problems with any of the
alternatives I know of. The advantage of cmake is that it is widely used in
C++ projects. This means that a C++ developer usually has cmake
installed. Moreover it allows LF programms to easily link to other cmake
projects (See AsyncCallback.lf and AsyncCallback.cmake in C++ tests)

In my opinion, it is not too much to require that cmake is installed and
available in PATH. Most C++ programmers will have cmake installed anyway.

Couldn't we simply resolve the path issue by checking if cmake is available (or
if Eclipse is running on MacOS) and try to run /opt/bin/cmake if it is not?

lhstrh:
We could manually add the directory where OSX installs `cmake` to the PATH variable given to the ProcessBuilder (if that dir exists). That shouldn't really get into the way of anything... What is the directory, Edward? Was is `/opt/bin` or `/opt/local/bin`?

edwardalee:
I’m pushing changes to CppGenerator that “solve” the cmake problem on at least my Mac and that print out more informative messages.  However, now the generated code does not compile with make. ??? I don’t think I changed anything that would have affected that…

Note that in our coding style, we are using four spaces for indentation, not tabs. CppGenerator now has a mix, unfortunately. Can I replace all tabs with four spaces?

[...]

Issue #67, opened 2019-11-26
Impose physical time limit on tests


lhstrh:
Since tests declare their own time limit in the run-time directives listed in the LF code, we don't want to override them in `run-lf-tests`. Instead, we should impose a generous upper bound on the execution time (say, 5 minutes?) to preempt tests that have failed to declare a time limit and mark them as failed.

Issue #18, opened 2019-04-15
Syntax for deadlines


lhstrh:
Deadlines are associated with reactions. How do we express them?
The [reaction decl](../wiki/syntax#reaction-decl) is a pretty crowded syntactically already, so it might be better to add an annotation in the [reactor decl](../wiki/syntax#reactor-decl), e.g.:

```
reactor Punctual {
  input x:int;
  deadline foo 10 ms;
  reaction foo(x) {
    // do something
  }
}
```

But perhaps it makes more sense for the deadline to be determined on the level of the composite; after all, the deadline is probably application specific and thus is likely unknown until integration.

If we opt for this latter approach, we could have:

```
x = new Punctual();
deadline(x.foo, 10 ms);
```

I think I'm in favor of the last option.

PR #1169, opened 2022-05-13
Implement new handling of startup/shutdown/reset reactions in modes

### LINKED ISSUE #999: startup reactions in modal models
Currently, a startup reaction in a modal model reacts once when the program starts, but only if the startup reaction is defined in the initial mode. If the startup reaction is in some other mode, it is ignored and never invoked.

Wouldn't a better behavior be to invoke a startup reaction whenever a mode is entered, which for an initial mode is at the start of execution, but again each time the mode is entered via a reset transition?  A history transition would not invoke a startup reaction unless the mode is being entered for the first time.

### LINKED ISSUE #1048: [Modal models] Validator needs to check for reactions triggered by `shutdown` inside of modes
As @Soroosh129 and @a-sr pointed out, there appears to be no clean solution for dealing with shutdown reactions inside of modes. Executing shutdown reactions of inactive modes would be awkward, and not executing them could lead to memory leaks (in case there is shutdown code intended to de-allocate memory). Instead, we should disallow shutdown reactions inside of modes altogether (or at least throw a warning).

a-sr:
Requires [reactor-c PR #79](https://github.com/lf-lang/reactor-c/pull/79)
Requires [benchmarks PR #15](https://github.com/lf-lang/benchmarks-lingua-franca/pull/15)
Requires [benchmarks PR #18](https://github.com/lf-lang/benchmarks-lingua-franca/pull/18)

Implies [website PR #67](https://github.com/lf-lang/website-lingua-franca/pull/67)
Implies [website PR #68](https://github.com/lf-lang/website-lingua-franca/pull/68)

Closes #999
Closes #1048

---

Implements the following semantics:
- Startup reactions:
    - Only triggered upon first entry into a mode
    - Implication: Some legacy reactors (without reset reactions) might stop working when entered via reset because an "initial token" is not recreated upon reset entry
- Shutdown reactions:
    - Triggered at shutdown time of reactor
    - Only for modes that were active anytime before
    - Executed even is mode is inactive
- New reset reactions:
    - Triggered at activation time of a mode if entered via reset
- New reset transition:
    - No longer resets state variables
        - Can be handled by reset trigger
        - Implication: Always requires manual reset. Legacy reactors (without reset reaction) keep state variables values upon reset, leading to previous "history" leaking into the current (otherwise reset) behavior (e.g. a counter)

---

TODO:
- [x] Fix Python support
- [x] Fix newly blocked keywords in tested models (#947)
- [x] Add a warning for reactor instantiations with state variables but without reset reactions


a-sr:
The implementation is complete and this PR and the related ones are ready for review.

The benchmark tests still fail because the fix in the benchmark repository is not yet merged.
The error in cli-tests on Mac looks weird but seems unrelated to this PR, since other tests succeed. @lhstrh Could you help me out with that one?

edwardalee (approved):
LGTM!  The website documentation will also need to be updated.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -134,7 +134,7 @@ TargetDecl:
         ((parens+='(' (init+=Expression (','  init+=Expression)*)? parens+=')')
             | (braces+='{' (init+=Expression (','  init+=Expression)*)? braces+='}')
         )?
-    ) ';'?
+    ) (reset?='reset')? ';'?
```
lhstrh, line 137: I'm confused about this syntax. What does it mean if a state variable of type `foo*` is marked as `reset`?

lhstrh: I also find it a bit strange to see this modifier at the end of the declaration. I think it would make more sense to have it at the beginning.

lhstrh: Also, I'm thinking that a keyword like `transient` might be better, because it is a clear adjective and a much less frequently used term in variable names.

edwardalee: Isn't `reset` already a keyword (for reset transitions)?

edwardalee: I think this needs to have a clear connection with a reset transition.  The state gets reset when a reset transition is taken.

a-sr: The problem with `transient` is that it inverts the modifier and consequently the default (if absent).
Our concept relies on the fact that by default we do not automatically reset a state variable, such that if a reactor was not written with modes in mind we to not leak memory by resetting state variables pointing to allocated memory. The approach with `transient` contradicts this in my opinion.

lhstrh: I see why that's desirable. I still think it's better to have `reset state ...` than `state ... reset`.

Do you know what the reset behavior is for state variables with a pointer types?

a-sr: I added the reset as a shorthand notation for resetting the state variable in a reset reaction, so I agree with @edwardalee that it should have a connection to it.

a-sr: In my mind, I do not consider the reset as an integral component of the state variable but rather a handler added to it, like a deadline or an after delay. I also did not want to draw to much attention to it by putting it first but if you think a `reset state` makes more sense, I can change that.

> Do you know what the reset behavior is for state variables with a pointer types?

Well, the variable is reset to the initial value provided by the user. I assumed that there always has to be one but if that was a misconception, I will add an error message in case `reset` is used without an initial value.

lhstrh: Handlers go at the end, sure, but this isn't really a handler but a modifier that labels the state as being automatically reset. I suggest we put the `reset` in front of the state variable declaration.

org.lflang.ui/src/org/lflang/ui/highlighting/LFSemanticHighlightingCalculator.java
```diff
@@ -61,23 +56,6 @@ public void provideHighlightingFor(XtextResource resource, IHighlightedPositionA
             return;
         }
         Model model = (Model) parseResult.getRootASTElement();
-        
-        // Provide keyword highlighting for special mode transitions
```
lhstrh, line 65: How come this is no longer necessary?

lhstrh (changes requested):
This looks good to me, except for the syntax (as pointed out in conversation). If also really like your suggestion to validate that any state variable that is marked as reset indeed has an initial state assigned to it.

org.lflang/src/org/lflang/LinguaFranca.xtext
a-sr: Ok

org.lflang.ui/src/org/lflang/ui/highlighting/LFSemanticHighlightingCalculator.java
a-sr: When there was only `reset(<mode>)` and `continue(<mode>)`, I was able to make a workaround that treated these keywords as strings in the grammar (#947) and then faked their role as keywords in the editor by adjusting the highlighting (which also only worked in Epoch). However, with `reset` now used in the enum rule for built-in triggers and as boolean modifier (reset?='reset') in state variables, it was not possible to treat is as a non-keyword. So I also rolled back the workaround in transition effect to make the grammar more consistent.

a-sr:
Ok
- added the validation error for reset state variables without initial values,
- moved the `reset` modifier in front of the `state` keyword, and
- adjusted the syntax examples on the website ([PR68](https://github.com/lf-lang/website-lingua-franca/pull/68))

a-sr:
Rust benchmark tests currently fail because there are newly added models using `reset` as variable name. [Benchmarks PR #18](https://github.com/lf-lang/benchmarks-lingua-franca/pull/18) fixes these models.

edwardalee:
Is this ready to merge?

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1313,6 +1308,79 @@ public void checkModeInstanceNamespace(Reactor reactor) {
         }
     }
 
+    @Check(CheckType.FAST)
+    public void checkMissingStateResetInMode(Reactor reactor) {
```
lhstrh, line 1312: We should probably add some unit tests for this...

test/C/src/modal_models/ModalStateResetAuto.lf
```diff
@@ -0,0 +1,98 @@
...
+    reaction(stepper) -> modal.next {=
+        lf_set(modal.next, true);
+    =}
+    
+    modal.mode_switch,
+    modal.count0,
+    modal.count1,
+    modal.count2
+    -> test.events
+}
```
lhstrh, line 98: Missing newline.

lhstrh (commented):
These changes are looking great! My only comment is that the new validator checks should have unit tests. Other than that, this looks ready to merge.

test/C/src/modal_models/ModalStateResetAuto.lf
a-sr: Added newline.

org.lflang/src/org/lflang/validation/LFValidator.java
a-sr: Added unit tests for all mode related validation rules.

test/C/src/modal_models/ModalStateResetAuto.lf
```diff
@@ -0,0 +1,98 @@
...
+    
+    state counter0:int(0);
+    
+    reaction(next) -> count0 {=
+        printf("Counter0: %d\n", self->counter0);
+        lf_set(count0, self->counter0++);
+    =}
+    
+    initial mode One {
+        state counter1:int(0);
```
lhstrh, line 28: It looks like the state variable isn't rendered quite correctly in the diagram:
image

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -1304,15 +1299,88 @@ public void checkModeInstanceNamespace(Reactor reactor) {
...
+                            instantiation.getName()), instantiation, Literals.INSTANTIATION__NAME);
                     }
                     names.add(instantiation.getName());
                 }
             }
         }
     }
 
+    @Check(CheckType.FAST)
+    public void checkMissingStateResetInMode(Reactor reactor) {
```
lhstrh, line 1312: It appears that a `reset` transition is the default, meaning that is no modifier is given in the transition effect, this is to mean that it is a reset transition. This could be confusing, because it may suggest that a "normal" transition is of a different kind than `reset` or `continue`. We had a similar problem with actions, and decided that it would be better to require a disambiguation from the programmer and thus only allow `logical action` or `physical action`, and not just `action`. I think we should take the same approach here, and enforce this in the validator so we can present a useful error message.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -451,6 +453,12 @@ enum ActionOrigin:
 enum Visibility:
     NONE | PRIVATE='private' | PUBLIC='public';
 
+enum BuiltinTrigger:
+    STARTUP='startup' | SHUTDOWN='shutdown' | RESET='reset';
+
+enum ModeTransition:
+    RESET='reset' | HISTORY='continue';
```
lhstrh, line 460: I'm still a bit on the fence about the `continue` keyword. If refer to this a "history transition" and label it with an "H" in the diagram, then why don't we just use the `history` keyword and avoid all confusion?

test/Python/src/modal_models/ModalStateReset.lf
```diff
@@ -29,6 +29,10 @@ reactor Modal {
         state counter1(0)
         timer T1(0msec, 250msec)
 
+        reaction(reset) {=
```
lhstrh, line 32: The rendering of the `reset` trigger seems a bit off:

image

I would actually like to propose an alternative visual syntax. How about something along these lines:
image

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -2221,7 +2222,153 @@ public void testUnrecognizedTarget() throws Exception {
             "Unrecognized target: Pjthon");
     }
 
-    
+    @Test
+    public void testInitialMode() throws Exception {
```
lhstrh, line 2226: Great! Thanks for putting these in!

lhstrh (approved):
This looks really good. I raised a couple points about syntax, validation, and visual representation. If you have the bandwidth, I would prefer we address these issues as part of this PR... I'm also fine with merging and adding fixes later.

lhstrh:
I'll go ahead and merge this in to avoid stalling things any further. I've created issues to track my comments.

PR #1170, opened 2022-05-13
Refactor `CGenerator` and `PythonGenerator`

housengw:
Refactor `CGenerator` and `PythonGenerator` such that the `fileConfig` property is not overwritten in the `for (federate : federates)` loop in the case of federated execution.


housengw:
A noteworthy observation:
- the target properties of imported reactors are inspected during code generation (namely in `doGenerate` -> `generateCodeForCurrentFederate` -> `generateReactorDefinitions` -> `generateReactorChildren` -> `inspectReactorEResource`). In `inspectReactorEResource` is where the `cmake-includes` files of imported reactors get copied to `src-gen`. However, if we want to get rid of reassigning `fileConfig`, going with this design means we have to pass a `fileConfig` into that chain of functions to get the right `src-gen` path for federated execution.

lhstrh:
Subsumed by the work in #1221

PR #1174, opened 2022-05-16
Logging cleanups for VS Code

petervdonovan:
This PR improves the textual output that is emitted by the language server. It should be merged before the corresponding PR [here](https://github.com/lf-lang/vscode-lingua-franca/pull/47).


lhstrh (approved):
Looks good! Bizarre that VS Code ignores standard out...

PR #1178, opened 2022-05-18
Add STP handler to network reaction

edwardalee:
This PR adds an STP handler to network receiver reactions that does the same thing as the regular reaction. This suppresses  error messages added to the federated-error-message-on-stp branch of reactor-c, which should be merged first.


edwardalee:
Oops, sorry. Just pushed.

Edward

——————
Edward A. Lee
EECS, UC Berkeley
http://eecs.berkeley.edu/~eal

org.lflang/src/org/lflang/federated/FedASTUtils.java
```diff
@@ -230,21 +229,32 @@ private static void addNetworkReceiverReaction(
...
-        ));
+        networkReceiverReaction.getCode().setBody(body);
+        
+        // Set the STP handler to be the same as the body.
+        // FIXME: This is a hack that assumes STP violations are passed
+        // on to the reactions that are triggered by this network receiver.
+        if (coordination == CoordinationType.DECENTRALIZED) {
+            networkReceiverReaction.setStp(factory.createSTP());
+            networkReceiverReaction.getStp().setCode(factory.createCode());
+            networkReceiverReaction.getStp().getCode().setBody(body);
```
Soroosh129, line 256: I don't think this would do what we want unless the STP handler body manually passes down the STP violation somehow.

edwardalee: I guess the issue here is that we don't want to silently invoke reactions at the wrong logical time when there is no STP handler, but then, when there is an STP violation, the network receiver, which doesn't have an STP handler, will trigger an error message even if there is a handler downstream.  I don't have a good solution for this. Suggestions?  I don't think going back to silently invoking reactions at the wrong time is the right solution.

edwardalee:
I believe this PR is obsolete and the problem has been dealt with correctly.

PR #1179, opened 2022-05-18
Address LSP test failures.

petervdonovan:
This simplifies the `pr` method in `CodeBuilder.java`. The work ostensibly done by this method is, I think, taken care of elsewhere.

I believe that this fixes the test failure [here](https://github.com/lf-lang/lingua-franca/runs/6487050913?check_suite_focus=true). I was able to reproduce the failure by running the tests locally with the same random seed, and the failure was gone after I made this change.

After this change, the generated code still looks mostly the same to me as before.

### DIFF
--- a/org.lflang/src/org/lflang/generator/CodeBuilder.java
+++ b/org.lflang/src/org/lflang/generator/CodeBuilder.java
@@ -98,31 +98,11 @@ public class CodeBuilder {
     }
 
     /**
-     * Append the specified text plus a final newline to the specified
-     * code buffer. This also replaces tabs with four spaces.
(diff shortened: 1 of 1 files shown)


petervdonovan:
> What was the bug?

In Python, if the very first line in a `{= =}` code block had incorrect indentation, then the entire code block would have the same incorrect indentation. I think that this has been causing LSP tests to fail by moving syntax errors out of the range covered by the code maps.

lhstrh:
Wow! Great detective work.

lhstrh (approved):
Very nice!

PR #1180, opened 2022-05-19
[epoch] Add default template that is just a skeleton

lhstrh:
Somehow, the description of the template is not rendered in the wizard. I have no clue why because all the other templates work exactly the same. Some extra :eyes: on this would be appreciated.


edwardalee (approved):
LGTM!

PR #1183, opened 2022-05-19
Fix whitespace in tests

petervdonovan:
LSP tests failed in the `expressions` branch (again!), this time because of mixed spaces and tabs. This style violation caused ESLint to throw an error, which ended compilation (`tsc` was not even executed, so no type checking). This prevented the TypeScript validator from catching the error that it was expected to catch.

I would like to merge this directly into @cmnrd's `expressions` branch, since that one will soon go into `master` anyway.


lhstrh (approved):
Holy cow! This is quite the diff! Assuming that this indeed only concerns white space changes, this looks good to me :-)

It would be great to have checks for the formatting of LF code as part of CI to avoid problems like these in the future.

PR #1189, opened 2022-05-23
Deadlock in Federated Execution

Soroosh129:
This adds a test that demonstrates a scenario where the federation deadlocks and fixes the issue.

The underlying issue was that network control reactions were being generated for connections that had after delays.


edwardalee (approved):
LGTM!

PR #1190, opened 2022-05-23
[Experimental] Quasi-static scheduler

lsk567:
This PR implements a quasi-static (QS) scheduler proposed in #1080. The QS scheduler generates a fixed static schedule at compile time and follows the fixed schedule at runtime.

To use this scheduler, the user needs to install two additional dependencies: [Uclid5](https://github.com/uclid-org/uclid) and [z3](https://github.com/Z3Prover/z3). Both need to be added to PATH and able to be invoked on the command line.

The scheduler is currently experimental (i.e. not production-ready). For small test cases such as `ThreadedThreaded.lf`, it can achieve comparable performance wrt that of the default NP scheduler. However, for large benchmarks such as `SleepingBarber.lf`, the QS scheduler currently exhibits a significant slowdown. We are yet to figure out the cause of this.


org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -225,8 +226,13 @@ CodeBuilder generateCMakeCode(
                     cMakeCode.pr("add_link_options( "+compilerFlag+")");
             }
         }
-        cMakeCode.newLine();
 
+        if (targetConfig.schedulerType == SchedulerOption.QS) {
```
Soroosh129, line 230: I think it's better to handle this [here](https://github.com/lf-lang/lingua-franca/blob/d561add3e18a2e65b8648f240a4c7609819eb653/org.lflang/src/org/lflang/generator/c/CGenerator.java#L864) with a 
```java
if (targetConfig.schedulerType == SchedulerOption.QS) {
    targetConfig.compileDefinitions.put("SCHEDULER_QS", "");
}
```

org.lflang/src/org/lflang/generator/c/CCoreFilesUtils.java
```diff
@@ -111,14 +112,15 @@ private static List<String> getThreadSupportFiles(
...
-                ) :
-                List.of("reactor.c");
+        List<String> supportFiles = new ArrayList<String>(Arrays.asList(
+            "threaded/scheduler.h",
+            "threaded/scheduler_instance.h",
+            "threaded/scheduler_sync_tag_advance.c",
+            "threaded/scheduler_" + scheduler + ".c",
+            "threaded/reactor_threaded.c"
+        ));
+        if (scheduler == SchedulerOption.QS)
```
Soroosh129, line 122: This would also be better handled [here](https://github.com/lf-lang/lingua-franca/blob/d561add3e18a2e65b8648f240a4c7609819eb653/org.lflang/src/org/lflang/generator/c/CGenerator.java#L864).

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -563,6 +564,22 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
                     dockerGenerator.fromData(lfModuleName, federate.name, fileConfig));
             }
 
+            System.out.println("Checking main instance in CGenerator.java.");
+            System.out.println(this.main);
+            System.out.println(main);
+
+            // Generate schedule.h if QS scheduler is used.
```
Soroosh129, line 571: Sorry to repeat this, but this should also be in [`pickScheduler`](https://github.com/lf-lang/lingua-franca/blob/d561add3e18a2e65b8648f240a4c7609819eb653/org.lflang/src/org/lflang/generator/c/CGenerator.java#L864) I think.

org.lflang/src/org/lflang/generator/c/CTriggerObjectsGenerator.java
```diff
@@ -161,6 +166,13 @@ public static String generateInitializeTriggerObjects(
             isFederated,
             clockSyncIsOn
         ));
+        if (targetConfig.schedulerType == SchedulerOption.QS) {
```
Soroosh129, line 169: Why not have this inside `generateSchedulerInitializer`?

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
lsk567: Great, thanks! This is indeed a cleaner approach.

org.lflang/src/org/lflang/generator/c/CGenerator.java
lsk567: Make sense. Will revise the code accordingly. Thanks.

org.lflang/src/org/lflang/generator/c/CTriggerObjectsGenerator.java
lsk567: Right, it makes sense to have this in `generateSchedulerInitializer`. Thanks for the catch!

lsk567:
> Please note that unless this eventually gets merged with a squash commit, these papers will continue to linger in Git's history. Perhaps now is the time to create a new branch (forked off master) and quash your changes in this branch into the new one (using a PR). You can then continue developing in that new branch and eventually submit a clean PR to master.

@lhstrh Okay, sounds good to me.

build.gradle
```diff
@@ -33,6 +33,9 @@ subprojects {
         implementation group: 'com.google.inject', name: 'guice', version: guiceVersion
         // https://mvnrepository.com/artifact/commons-cli/commons-cli
         implementation group: 'commons-cli', name: 'commons-cli', version: commonsCliVersion
+        // https://search.maven.org/artifact/tools.aqua/z3-turnkey/4.8.16/jar
+        implementation 'tools.aqua:z3-turnkey:4.8.16'
```
lhstrh, line 37: I noticed that this added dependency is not reflected in the OSGi or Maven configurations. This means that it would only work with `lfc`, which is built using Gradle, not with Epoch or the LDS server (and thus VS Code).

lhstrh: For Maven there is this:
```
<dependency>
    <groupId>tools.aqua</groupId>
    <artifactId>z3-turnkey</artifactId>
    <version>4.8.16</version>
</dependency>
```
I don't know how to fix the Eclipse configuration, to be honest. But we can't merge this until that is fixed because otherwise the project simply won't build in Eclipse.

lhstrh: Maybe @a-sr has any suggestions?

lhstrh: Add this, but Maven still doesn't work. I'm probably missing something obvious again...

org.lflang/src/org/lflang/generator/c/scheduling/uclid/SmtScheduleGenerator.java
```diff
@@ -0,0 +1,659 @@
...
+MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+***************/
+
+package org.lflang.generator.c;
```
lhstrh, line 27: Incorrect package declaration. That said, I'm not sure this is the right place to put this generator. It generates UCLID code, so I think it should be in a separate `uclid` package.

build.gradle
lhstrh: Alternatively, z3 could be included as a submodule. I expect this kind of problem to surface with other dependencies, too. And including a submodule for every library that isn't available as an OSGi bundle doesn't seem like a good solution either.

a-sr: Yes, all non-plugin dependencies in Eclipse are tricky. If z3 is a compile-time dependency for Java, I guess there is no way around putting it in a plugin to make it work in Eclipse.
We also had this issue in Klighd with the Sprotty library that is only available on maven central. We created a repository to wrap it into a plugin ([sprotty-mirror](https://github.com/kieler/sprotty-mirror)). The build automatically pulls the library, builds it into a plugin, and deploys it as an update site, such that we can use it in our Eclipse and Oomph setup as dependency. Maybe we need something similar for LF.
The official Eclipse Orbrit works with the same principle.

Regarding the dependency you added to the pom file, I do not know why this does not work or if that even will work.  In the maven configuration, the option to consider pom dependencies in addition to the target platform is enables but I do not know how it works in detail. The current configuration has something to do with the workaround for Kotlin, but to be honest I am not sure how all these configurations work together.
Sorry.

org.lflang/src/org/lflang/generator/c/scheduling/uclid/SmtScheduleGenerator.java
lsk567: Thanks. Will put that in a separate `uclid` package.

lsk567:
@Soroosh129 brought up the issue that naming the implementation a "quasi-static scheduler" might be too generic, since there can be different variants of the quasi-static scheduler (e.g. a QS scheduler with a schedule generation mechanism using some other constraint solvers). When specifying the QS scheduler, it would be useful to further parametrize the scheduler. For example, `scheduler: QS(smt)`.

lhstrh:
> When specifying the QS scheduler, it would be useful to further parametrize the scheduler. For example, `scheduler: QS(smt)`.

This strikes me as excess generality. Wouldn't something like `qs-uclid5` do?

PR #1195, opened 2022-05-26
[Federated] Fix segfault for internally disconnected output ports

Soroosh129:
This fixes a segmentation fault that arose whenever a federate had an output port that was not internally connected (e.g., it did not appear as an effect in any reaction). In the CGenerator, such an output port will be NULL, but the output control reactions were attempting to check the `->is_present` field of these ports. 

This PR adds an extra check for this.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CNetworkGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CNetworkGenerator.java
@@ -350,7 +350,11 @@ public class CNetworkGenerator {
             "LF_PRINT_LOG(\"Contemplating whether to send port \"",
             "          \"absent for port %d to federate %d.\", ",
             "          "+portID+", "+receivingFederateID+");",
-            "if (!"+sendRef+"->is_present) {",
+            "if ("+sendRef+" != NULL) { // Check that the output port is not NULL",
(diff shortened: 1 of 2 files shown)


org.lflang/src/org/lflang/generator/c/CNetworkGenerator.java
```diff
@@ -350,7 +350,11 @@ public static String generateNetworkOutputControlReactionBody(
             "LF_PRINT_LOG(\"Contemplating whether to send port \"",
             "          \"absent for port %d to federate %d.\", ",
             "          "+portID+", "+receivingFederateID+");",
-            "if (!"+sendRef+"->is_present) {",
+            "if ("+sendRef+" != NULL) { // Check that the output port is not NULL",
```
edwardalee, line 353: This could be simplified to:
```
if (sendRef == NULL || sendRef->is_present) {
    send_port_absent...
}
```

edwardalee (approved):
LGTM! Slight simplification suggested.

org.lflang/src/org/lflang/generator/c/CNetworkGenerator.java
Soroosh129: Okay, thank you. This is much cleaner.

PR #1197, opened 2022-05-27
[Python] Adjusted build to fall back to `python`

Soroosh129:
This PR adds a fallback mechanism for platforms that do not have the `python3` command but do have a `python` command that points to a Python 3 interpreter.


edwardalee (approved):
LGTM!

PR #1208, opened 2022-05-31
Use `toOriginalText` in `org.lflang.diagram`.

petervdonovan:
Some days ago I renamed a function in `ASTUtils` but forgot to update the function name in the `diagram` package. This amends that.


lhstrh (approved):
Thanks for catching that!

PR #1209, opened 2022-06-01
Failed attempt to add New Folder option to menus for Epoch

edwardalee:
Epoch is missing a "New Folder" option in the menus.  This draft PR is a failed attempt to add such an option. Does anyone know what other Eclipse magic is necessary to make something appear in the menus?  How was the New Lingua Franca File added?  I tried to emulate that, but the New Folder option does not appear.


lhstrh:
I'm closing this PR because the Epoch code is no longer hosted in this repo. Yet, I wonder whether @a-sr could help with this?

Issue #937, opened 2022-02-04
Publication of some or all modules to Maven Central


Rojods:
Would there be any interest on publishing some or all modules to Maven Central? This issue is somewhat related to #142 . 

The use case for such feature is the easy integration of LF tooling into 3rd party toolsets. Specially for those that wish to perform M2M or other advanced analysis while reusing the LF toolset. 

To give a concrete use case, I speak for myself. I am currently the main developer of [ForSyDe IO](https://github.com/forsyde/forsyde-io), which is majorly implemented as a Java library outside Eclipse's ecosystem. 
Doing my actual research requires parsing a LF model and doing a simple M2M transformation to ForSyDe IO. Naturally, my first thought is to just grab the LF compiler and extract it's internal IR that is an EMF model. Sadly, after a handful trials and errors, the only _reproducible_ solution to integrate it in my toolset was to copy the model submodule from the `git` repo into my own repo. This means that until I find better solutions, to achieve reproducibility I have to manually update the LF model in ForSyDe IO. This is not a problem per se, it just feels less robust when compared to a plain maven/gradle project experience.

I am also aware how hard it is to make the standard Eclipse ecosystem play nice with maven/gradle or other non-Eclipse tools in general. I have little experience, but a couple dabbles here and there taught me some painful lessons :). I just wonder, as there are already gradle (Oomph) builds in place, maybe it would not be a monumental effort to bring at least the core modules to maven central in the future. 

I would label this issue as "Enhancement" myself, but it seems like I'm not allowed to.

lhstrh:
Thanks for suggesting this, @Rojods. I wasn't aware that anyone was using parts of our toolchain as a library, but your use case certainly justifies looking into this. I'm not sure anyone on the team has prior experience with publishing to Maven Central, but as we're establishing a release cycle that will involve a fair amount of automation, publishing on Maven Central is something that might be relatively easy to add.

On a related note, this week we've published the first version of our new [VS Code extension](https://marketplace.visualstudio.com/items?itemName=lf-lang.vscode-lingua-franca); check it out :-)

Rojods:
Thanks for the reply @lhstrh !! I'll be sure to check it out. I have some limited experience with Maven central, and it's basically just a one time effort to setup everything. At least it was wit ForSyDe IO 🙂. 

[...]

PR #1211, opened 2022-06-02
A possible stub for publishing the LF internal model/AST to Github Pak.

Rojods:
This is a small stub to help a bit with the issue opened in #937 . I am fairly certain that if any necessary jars/modules are published in Github Packages, the outcome is just as good as doing it for Maven central. Plus, it seems like Github Packages requires less setting up than maven central (e.g. no need for given GPG key).

Since I am not an expert on the infrastructure surrounding LF, I've tried my best to use GH's stub publishing along the other GH actions I saw in the repo. Hopefully it does not destroy anything but gives an _almost correct_ workflow to publish the model/AST jars to GH's packages.

The tests still all passed, minus the repo_token one. I could not find a proper branch to try to merge in to, so I leave master for now. If this is inappropriate, I can change the request.

### DIFF
--- a/.github/workflows/publish-model-2-packages.yml
+++ b/.github/workflows/publish-model-2-packages.yml
@@ -1,0 +1,21 @@
+name: Publish org.lflang to GitHub Packages
+on:
+  workflow_dispatch:
+  release:
+    types: [created]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: Prepare build environment
+        uses: ./.github/actions/prepare-build-env
+      - name: Publish package
+        uses: gradle/gradle-build-action@0d13054264b0bb894ded474f08ebb30921341cee
+        with:
+          arguments: org.lflang:publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,8 @@ subprojects {
     repositories {
         mavenCentral()
     }
-    
+
+    apply plugin: 'java'
     apply plugin: 'kotlin'
     compileKotlin {
         destinationDir = compileJava.destinationDir
--- a/org.lflang/build.gradle
+++ b/org.lflang/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id 'maven-publish'
+}
+
 dependencies {
     implementation "org.eclipse.xtext:org.eclipse.xtext:${xtextVersion}"
     implementation "org.eclipse.xtext:org.eclipse.xtext.xbase.lib:${xtextVersion}"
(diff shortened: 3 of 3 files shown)


.github/workflows/publish-model-2-packages.yml
```diff
@@ -0,0 +1,21 @@
+name: Publish org.lflang to GitHub Packages
+on:
+  workflow_dispatch:
+  release:
+    types: [created]
```
lhstrh, line 5: I think this means it will also be triggered by the nightly build. Was this the intent? I'm thinking we should probably exclude `nightly`.

.github/workflows/publish-model-2-packages.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: Prepare build environment
+        uses: ./.github/actions/prepare-build-env
+      - name: Publish package
+        uses: gradle/gradle-build-action@0d13054264b0bb894ded474f08ebb30921341cee
```
lhstrh, line 17: Why use this specific hash? Maybe just use `v2.1.7`?

.github/workflows/publish-model-2-packages.yml
```diff
@@ -0,0 +1,21 @@
...
+        with:
+          arguments: org.lflang:publish
```
lhstrh, line 19: What gets published as a result of this action, exactly? Is there a way to do a dry run and find out?

lhstrh (commented):
This looks pretty cool! I left some comments...

.github/workflows/publish-model-2-packages.yml
Rojods: In my excitement I just put what the GH actions gradle tutorial gave me. I fully agree with your suggestion.

.github/workflows/publish-model-2-packages.yml
Rojods: I am not sure if there is a way to do a dry-run of this, sadly. I know a tool that can simulate all github actions offline. But I guess we don't get the GITHUB_TOKEN that is injected in the containers that run all actions. 

In theory this this command would publish the `jar` that is generated from the `org.lflang` submodule of the repo. It might need some adjustment when it comes to the dependencies of the MWE2 plugins are doing them proper, but I guess it is not much extra work! 

In any case, I think the only way to see if this is working is to get the developer machine (you, @lhstrh?) and try to upload the jars to GH packages. This is the default flow for Maven Central so I am not sure if it can be replicated. I will try to check it.

Rojods:
I have now put conditionals in the action so that it "ignores" `nightly`. For the moments it was the best solution I could think of, since the event trigger `release` does not seem to support filter like `push` does. I think it will work out even though it might feel a bit hackish.

.github/workflows/publish-model-2-packages.yml
lhstrh: OK, thanks! I'll tinker with this.

PR #1215, opened 2022-06-06
Added support for C methods

edwardalee:
I've added support for methods in C because I need this to implement LET.
The invocation of a method in target code is a bit awkward.
You have to do this:
```
    self->my_method(self, my_arguments);
```
But with this design, method names need not be unique across reactors.


Soroosh129:
There is an alternative way to do this using `#define` and `#undef` within the body of reactions.
```C
foo_reaction_function_1(...) {
    #define my_method(var1, var2) foo_my_method(self, var1, var2)

    #undef my_method
}
```

This would make the syntax for accessing methods slightly different though from the syntax for accessing state variables.

edwardalee:
Great idea!  I will do this.

lhstrh:
I personally don't find using the self struct to find a method all that awkward. In fact, that's were I would look for them given that's where state variables live...

Soroosh129 (approved):
Looks great!

Issue #1203, opened 2022-05-30
Deadline of 0 is interpreted as no deadline


edwardalee:
When you enter a deadline of 0 msec, the deadline never gets violated, at least in the C target.  This is counterintuitive, as a deadline of 0 might be used to test the deadline violation handler.  I suspect that 0 is being used as a flag for no deadline, but it would be better to use something else (e.g. FOREVER?).

lhstrh:
Fixed by @billy-bao (see #1217).

PR #1217, opened 2022-06-09
Generate code to set negative deadline if there is no deadline specified

billy-bao:
If no deadline is specified in the LF program, the C code generator will set the deadline field of the reaction to NEVER (a negative value), which is interpreted as no deadline in the runtime. This fixes issue #1203 along with an update to the C runtime (https://github.com/lf-lang/reactor-c/pull/86).


lhstrh (approved):
The reported failure appears to be due to recent changes in master that got merged in here (not your change), so let's go ahead and merge this.

PR #1221, opened 2022-06-10
Restructuring of mechanism for compiling federated programs

lhstrh:
**Edit.** Refactor logic related to federated execution out of GeneratorBase, CGenerator, PythonGenerator, and TSGenerator  into a new `org.lflang.federated.FedGenerator`.

- [x] Move `analyzeFederates()` out of GeneratorBase into FedGenerator (@Soroosh129)
- [x] Have FedGenerator create separate .lf files for each federate (see #1212) (@Soroosh129)
- [x] [TS](https://github.com/lf-lang/lingua-franca/blob/fed-gen/org.lflang/src/org/lflang/federated/extensions/TSExtension.java) (@lhstrh, @hokeun)
        - **Note:** For the TS target, the code for these interfaces is already in the TSGenerator (e.g., [here](https://github.com/lf-lang/lingua-franca/blob/71fcad47ecc324dd403d60457ca955f1062e1ee7/org.lflang/src/org/lflang/generator/ts/TSGenerator.kt#L522))
    - [x] Implement "executable" preambles (using a syntax such as `init` (suggested by @cmnrd)) for each target to take care of pre-start tag initializations:
        - [x] [C target](https://github.com/lf-lang/lingua-franca/blob/71fcad47ecc324dd403d60457ca955f1062e1ee7/org.lflang/src/org/lflang/federated/extensions/CExtension.java#L633) (@billy-bao)
        - [x] Create a mechanism that can convert TargetConfig to TargetDecl 
        - **Note:** For the C and TS targets, the code for pre-start tag initialization is in their respective code generators, usually surrounded by `if (isFederated)` checks. Implementing the executable preamble would mostly require moving these blocks to the federated generator.
- [x] ~~Output interface instead of using the `@interface` label (@lhstrh)~~ Add internal dependency information to reactions
- [x] Have FedGenerator call each code generator depending on the target of each generated .lf file (@billy-bao)
- [x] Fix the `--clean` flag, which does not work for federated programs because they are located in a separate `fed-gen` directory (@axmmisaka)
- [x] Fix Docker support for C (@lhstrh)
- [x] Fix Docker support for TypeScript (@lhstrh)
- [x] Fix exception throw when `bin` directory already exists:
```
Exception in thread "pool-1-thread-2" org.eclipse.xtext.util.RuntimeIOException: Failed to create the directory /home/marten/lingua-franca-master/git/lingua-franca/test/C/fed-gen/DistributedCount/bin
	at org.lflang.util.FileUtil.createDirectoryIfDoesNotExist(FileUtil.java:486)
	at org.lflang.generator.c.CGenerator.doGenerate(CGenerator.java:468)
	at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:181)
	at org.lflang.federated.generator.FedGenerator.lambda$compileFederates$0(FedGenerator.java:246)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```
- [x] Re-enable `LetInferenceTests`


petervdonovan:
Grr, GitHub does not allow for the source branch of a PR to be changed, so I will have to move the title and comments used here to #1227.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -242,16 +240,10 @@ class TSGenerator(
...
-        federate: FederateInstance,
         codeMaps: MutableMap<Path, CodeMap>,
         dockerGenerator: TSDockerGenerator
     ) {
         var tsFileName = fileConfig.name
-        // TODO(hokeun): Consider using FedFileConfig when enabling federated execution for TypeScript.
-        // For details, see https://github.com/icyphy/lingua-franca/pull/431#discussion_r676302102
-        if (isFederated) {
-            tsFileName += '_' + federate.name
-        }
```
hokeun, line 261: I have a general question regarding "isFederated" blocks that are being removed here. (Not those with FIXMEs)

Where is this (e.g., formatting the shell script's name) going to be handled once fed-gen is done? I guess in FedGenerator?

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
```diff
@@ -67,61 +66,62 @@ class TSConstructorGenerator (
...
+//            for ((key, _) in federate.dependsOn) {
+//                // FIXME: Get delay properly considering the unit instead of hardcoded BigInt(0).
+//                federateConfigurations.add("this.addUpstreamFederate(${key.id}, BigInt(0));")
+//            }
+//            for ((key, _) in federate.sendsTo) {
+//                federateConfigurations.add("this.addDownstreamFederate(${key.id});")
+//            }
+//        }
+//        return federateConfigurations.joinToString("\n")
+//    }
```
hokeun, line 124: General question for the commented code blocks "FIXME: Move this to FedGenerator":

Is TS generator going to somehow inherit the FedGenerator class? How can target-specific codes be included in the FedGenerator?

hokeun (commented):
Thanks for taking care of the TS generator!

I left a couple of comments to understand this change better.

Also, I have one more question:

I tried to Lfc-compile [test/TypeScript/src/federated/HelloDistributed.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/federated/HelloDistributed.lf), but I got an error like this:

```
******** Using 2 threads to compile the program.
lfc: error: Mixed triggers and effects from different federates. This is not permitted
 --> test/TypeScript/src/federated/HelloDistributed.lf:44:5
   |
43 | federated reactor HelloDistributed at localhost {
   | >>>>>>>>>>>>>>
44 |     reaction(startup) {=
45 |         console.log("Printing something in top-level federated reactor.");
46 |     =}
   | <<<<<< Mixed triggers and effects from different federates. This is not permitted
47 |     s = new Source();      // Reactor s is in federate Source

lfc: fatal error: Aborting due to previous error

```

It doesn't look related to the fed-gen changes. Are we supposed to get such errors?

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
Soroosh129: There is an extension for TS [here](https://github.com/lf-lang/lingua-franca/blob/fed-gen/org.lflang/src/org/lflang/federated/extensions/TSExtension.java). This extension is used by the FedGenerator to produce target-specific code in the generated .lf file for each federate.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
Soroosh129: > I guess in FedGenerator?

Yes, that's the plan at least...

Soroosh129:
@hokeun:
> I tried to Lfc-compile [test/TypeScript/src/federated/HelloDistributed.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/federated/HelloDistributed.lf), but I got an error like this:

You are using zero-delay logical connections between federates. This currently does not work because the causality interface (`@interface`, see the original comment on this PR) is not implemented yet.

Soroosh129:
I'm a bit stuck at the moment in my attempt to remove `currentFederate` from the CGenerator.

Specifically, there are multiple places in the CGenerator where running counts (that are class attributes) get incremented by `currentFederate.numRuntimeInstances(...)`. I have no idea what these are for and how I might be able to remove them. @edwardalee would you happen to remember what each of these running counts are for? I think I need your help to get rid of them.

Here are some examples where this pattern is used:  [1](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1419) [2](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1424) [3](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1436) [4](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1467) [5](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1595) [6](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1652) [7](https://github.com/lf-lang/lingua-franca/blob/109caa51d1f4d7718f8dae984e1068f0e2be6019/org.lflang/src/org/lflang/generator/c/CGenerator.java#L1837).

edwardalee:
> I'm a bit stuck at the moment in my attempt to remove `currentFederate` from the CGenerator.
> 
> Specifically, there are multiple places in the CGenerator where running counts (that are class attributes) get incremented by `currentFederate.numRuntimeInstances(...)`. I have no idea what these are for and how I might be able to remove them. @edwardalee would you happen to remember what each of these running counts are for? I think I need your help to get rid of them.

These counts are used to determine the sizes of statically allocated arrays, for example, the array of startup reactions that need to be invoked at startup.

The issue here is that one instance of `ReactorInstance` may represent some number of runtime instances of the reactor. The number depends on whether the reactor is instantiated as a bank, whether any of its containers is instantiated as a bank, and whether it's top-level container is in the federate.  The latter condition is particularly subtle because if the top-level container is a bank, then the width of the bank is irrelevant and should be treated as 1 because only one bank member will be in the federate.

Soroosh129:
> [...] no top-level container will be a bank, right?

Right. At least that's what I'm aiming for.

> In fact, you could update ReactorInstance.getTotalWidth() to not take a depth argument because I suspect it will only ever be called with depth == 0.

Thank you! I applied your solution in c75d02a. Luckily, there is already a version of `getTotalWidth` that takes not argument (and passes 0 as the depth to `ReactorInstance.getTotalWidth(int depth)`).

org.lflang/src/org/lflang/federated/extensions/FedTargetExtension.java
```diff
@@ -0,0 +1,136 @@
...
+     * @oaram srcOutputPort FIXME
+     * @param connection FIXME
+     */
+    String generateNetworkOutputControlReactionBody(
+        VarRef srcOutputPort,
+        FedConnectionInstance connection
+    );
+
+    /**
+     * FIXME
```
hokeun, line 122: Can you add a brief explanation of this method?

Soroosh129: Will do (the FIXMEs are temporary placeholders). [Here](https://github.com/lf-lang/lingua-franca/blob/1c7924c2d6833014a4664bd74651d061934fad86/org.lflang/src/org/lflang/generator/GeneratorBase.java#L717) is the old comment for this method (the functionality hasn’t changed so the comment should probably be along the same lines).

Nonetheless, I think it would be better if we only move existing code for now in this PR and not implement new features. Control reactions are an unimplemented feature in the TS target as far as I know.

Soroosh129: Sorry, I was looking at the wrong thing. Will add a comment shortly.

Soroosh129, line 122:
```suggestion
     * Return the type for the raw network buffer in the target language (e.g., `uint_8` in C). This would be the type of the 
     * network messages after serialization and before deserialization. It is primarily used to determine the type for the 
     * network action at the receiver.
```

org.lflang/src/org/lflang/federated/extensions/FedTargetExtension.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return
```
Soroosh129, line 123:
```suggestion
```

hokeun (approved):
I'll take another look at FIXMEs assigned to me early this week!

org.lflang/src/org/lflang/federated/generator/FedPreambleEmitter.java
```diff
@@ -13,8 +14,16 @@ public FedPreambleEmitter() {}
...
-    String generatePreamble(FederateInstance federate, LinkedHashMap<String, Object> federationRTIProperties, ErrorReporter errorReporter) {
-        return FedTargetExtensionFactory.getExtension(federate.target).generatePreamble(federate, federationRTIProperties, errorReporter);
+    String generatePreamble(FederateInstance federate, FedFileConfig fileConfig, LinkedHashMap<String, Object> federationRTIProperties, ErrorReporter errorReporter)
+        throws IOException {
+        String pre = FedTargetExtensionFactory.getExtension(federate.target).generatePreamble(
+            federate, fileConfig, federationRTIProperties, errorReporter);
+        if (pre.length() == 0) return "";
+        else return
+        """
+        preamble {=
```
Soroosh129, line 24: Thanks! This looks great.

However, the common `preamble` syntax unfortunately doesn't work with the Cpp target because it requires a `public` or `private` modifier for the preamble. See [here](https://www.lf-lang.org/docs/handbook/preambles?target=cpp) for some examples.

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -404,7 +406,12 @@ protected String generateTopLevelPreambles() {
         for (Model m : models) {
             pythonPreamble.pr(PythonPreambleGenerator.generatePythonPreambles(m.getPreambles()));
         }
-        return "";
+
+        // C preamble for federated execution setup
+        String ret = "";
+        if (targetConfig.fedSetupPreamble != null)
```
Soroosh129, line 412:
```suggestion
        if (targetConfig.fedSetupPreamble != null) {
```

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -404,7 +406,12 @@ protected String generateTopLevelPreambles() {
...
+            ret = "#include \"" + targetConfig.fedSetupPreamble + "\"";
```
Soroosh129, line 413:
```suggestion
            ret = "#include \"" + targetConfig.fedSetupPreamble + "\"";
        }
```

org.lflang/src/org/lflang/federated/generator/FedPreambleEmitter.java
billy-bao: OK, then I think I'll revert this change and handle C & Python specifically then.

Soroosh129: Okay, thank you so much.

org.lflang/src/org/lflang/generator/ts/TSConstructorGenerator.kt
```diff
@@ -67,61 +66,62 @@ class TSConstructorGenerator (
...
+//            // Default of 0 is an indicator to use the default port, 15045.
+//            if (port == 0) {
+//                port = 15045
+//            }
+//            return """
+//            super(federationID, ${federate.id}, ${port},
+//                "${federationRTIProperties()["host"]}",
+//                timeout, keepAlive, fast, success, fail);
+//            """
+//        }
```
hokeun, line 85: @lhstrh, there's an issue when generating the TypeScript code here.
Currently, `main reactor` is generated as `App` and `federated reactor` is generated as `FederatedApp`. After fed-gen, all the reactors will become `main reactor`, and will be generated as `App`. Thus, we won't be able to use federated specific functions in this case. Do you have any suggestions on how to address this issue?

@Soroosh129, is there a way to distinguish whether a .lf file was generated by `fed-gen` or manually written by user?

Soroosh129: > @Soroosh129, is there a way to distinguish whether a .lf file was generated by fed-gen or manually written by user?

Not currently, but we can easily add that if needed (it could be as simple as a special marker in the preamble).

hokeun: > Not currently, but we can easily add that if needed (it could be as simple as a special marker inside the preamble)
Thank you, @Soroosh129 ! I think this will be very helpful. Maybe we need some way to indicate the LF is meant to be converted to a federated reactor anyway. Do you think you could add that to this PR?

Soroosh129: > Do you think you could add that to this PR?

Do you mean we should add this for all targets? Why not just add a special marker for the TS target's preamble in TSExtension?

hokeun: > Why not just add a special marker for the TS target's preamble in TSExtension?

It's because we should have this information at the code generation time in the current design. Federated and main reactors are code-generated differently like these (they inherit different classes):

- `class HelloDistributed extends __FederatedApp`
- `class HelloDistributed extends __App`

> Do you mean we should add this for all targets?

I think this will be useful, what do you think?

Soroosh129: > I think this will be useful, what do you think?

We don't have a use case for that in the C and Python targets , at least for now.

> It's because we should have this information at the code generation time in the current design.

Right. The special marker is meant to be read by the code generator. For example:
```TypeScript
preamble {=
    // 0f6cf1 This program is federated
=}
```

axmmisaka:
Notes to myself:
Cleaning logic is here https://github.com/lf-lang/lingua-franca/blob/f1c4e65912cfdd103ca4b3fdf7f9c1aa509e54b0/org.lflang/src/org/lflang/generator/GeneratorBase.java#L383; 
I tried to fix `--clean` by simply overriding `FedFileConfig::doClean()`, but it appears that cleaning logic is using a `FileConfig` even if `--federated` is passed in (idk if this is expected, but should be?) (see https://github.com/lf-lang/lingua-franca/commit/bebc935bf5ba2e439c494622f29c61d066b3d643)
Gotta find some more elegant way to fix later today

lhstrh:
@oowekyala and @hokeun I merged changes coming mostly from you two in this commit: https://github.com/lf-lang/lingua-franca/pull/1221/commits/99f1657b2e611d43afc5c298f7c749a4212d7fdc

Please have a look at it to make sure I didn't botch anything... I don't have a lot of confidence in this merge.

hokeun:
> @oowekyala and @hokeun I merged changes coming mostly from you two in this commit: [99f1657](https://github.com/lf-lang/lingua-franca/commit/99f1657b2e611d43afc5c298f7c749a4212d7fdc)
> 
> Please have a look at it to make sure I didn't botch anything... I don't have a lot of confidence in this merge.

Thank you, @lhstrh, for your efforts. Is defaultFederateConfig still an issue? I should be able to take another look this weekend.

lhstrh:
`defaultFederateConfig` is still an issue. It looks like the code for generating the definition in the preamble is still present, but the preamble itself is missing from the generated code. Could potentially be an easy fix, but I don't know the code well enough to quickly find it...

oowekyala:
The problem is that the preamble that contains the `defaultFederateConfig` is generated as a top-level preamble:
```
// c.lf
...
preamble {=
    const defaultFederateConfig: __FederateConfig = {
        ...
    }
=}


@_fed_config(network_message_actions="")
main reactor  {
    c = new Count()

    @_fed_send
    reaction(c.out) {=
        if (c.out !== undefined) {
            this.util.sendRTITimedMessage(c.out, 1, 0);
        }
    =}
}
```

this is done here: https://github.com/lf-lang/lingua-franca/blob/71ef8d0701c6a9283905f42eb7736621d0424dae/org.lflang/src/org/lflang/federated/generator/FedPreambleEmitter.java#L41

which is target-independent code. However the TS generator only outputs preambles that are contained in the reactor (so not top-level ones): https://github.com/lf-lang/lingua-franca/blob/71ef8d0701c6a9283905f42eb7736621d0424dae/org.lflang/src/org/lflang/generator/ts/TSReactorGenerator.kt#L147

I'm not sure what the correct behavior is here. Is the scope of a top-level preamble defined in TS? Where should it be generated?

lhstrh:
@oowekyala thanks for the pointers! The fix is in [46f0dac](https://github.com/lf-lang/lingua-franca/pull/1221/commits/46f0dacf06171d74104182747d8d1b0a4d442827).

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -210,9 +239,26 @@ enum AttrParamType {
...
+            List.of(new AttrParamSpec(AttributeSpec.VALUE_ATTR, AttrParamType.LANGUAGE, null))
+        ));
+        // @_fed_recv
+        ATTRIBUTE_SPECS_BY_NAME.put("_fed_recv", new AttributeSpec(null));
+        // @_fed_send
+        ATTRIBUTE_SPECS_BY_NAME.put("_fed_send", new AttributeSpec(null));
+        // @_fed_inp_ctrl
+        ATTRIBUTE_SPECS_BY_NAME.put("_fed_inp_ctrl", new AttributeSpec(null));
+        // @_fed_out_ctrl
+        ATTRIBUTE_SPECS_BY_NAME.put("_fed_out_ctrl", new AttributeSpec(null));
```
cmnrd, line 253: The `_fed_recv`, `_fed_send`, `_fed_inp_ctrl` and `_fed_out_ctrl` are only used to mark reactions as unordered (see AttributeUtils). It would be cleaner to introduce an `@unordered` attribute instead.

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -210,9 +239,26 @@ enum AttrParamType {
...
+        // @_fed_config
+        ATTRIBUTE_SPECS_BY_NAME.put("_fed_config", new AttributeSpec(
```
cmnrd, line 255: The only use of this attribute appears to be to tell the TS generator that the main reactor is a federate. Why don't we simply use a `@federate` attribute without argument for this?

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -265,6 +266,9 @@ public class TargetConfig {
     /** Rust-specific configuration. */
     public final RustTargetConfig rust = new RustTargetConfig();
 
+    /** Path to a C file used by the Python target to setup federated execution. */
+    public String fedSetupPreamble = null;
```
cmnrd, line 270: Target specific members should be added only in derived classes.

org.lflang/src/org/lflang/ast/FormattingUtils.java
```diff
@@ -60,25 +77,25 @@ public static String render(EObject object, int lineLength) {
...
         if (object instanceof Model model) {
             var targetDecl = ASTUtils.targetDecl(model);
-            if (targetDecl != null && targetDecl.getName().toUpperCase().contains("PYTHON")) {
-                return "#";
+            if (targetDecl != null) {
+                return Target.fromDecl(targetDecl);
             }
         }
-        return "//";
+        return Target.C;
```
cmnrd, line 98: Why does this return the C target as default? This seems wrong to me

org.lflang/src/org/lflang/ast/FormattingUtils.java
```diff
@@ -48,10 +51,24 @@ public class FormattingUtils {
...
+
+    /**
+     * Return a String representation of {@code object}, with lines wrapped at
+     * {@code lineLength}, with the assumption that the target language is
+     * {@code target}.
+     */
+    public static String render(EObject object, int lineLength, Target target, boolean codeMapTags) {
         MalleableString ms = ToLf.instance.doSwitch(object);
-        String singleLineCommentPrefix = getSingleLineCommentPrefix(object);
+        String singleLineCommentPrefix = target == Target.Python ? "#" : "//";
```
cmnrd, line 69: This will not work for all targets. I couldn't find it, but I thought we had a method somewhere that returns the comment prefix

org.lflang/src/org/lflang/federated/OldFedFileConfig.java
```diff
@@ -26,36 +22,39 @@
...
+ * for federated execution. The code generator should create one instance of this class
  * for each federate.
- * 
- * @author Soroush Bateni
  *
+ * @author Soroush Bateni
  */
-public class FedFileConfig extends FileConfig {
-    
+@Deprecated
```
cmnrd, line 38: Shouldn't we just remove this instead of deprecating it? Why is it still needed?

org.lflang/src/org/lflang/federated/extensions/CppExtension.java
```diff
@@ -0,0 +1,66 @@
+package org.lflang.federated.extensions;
```
cmnrd, line 1: Its currently not planned to support federates (using this scheme) in C++. I suggest removing this stub

org.lflang/src/org/lflang/federated/extensions/RustExtension.java
```diff
@@ -0,0 +1,66 @@
+package org.lflang.federated.extensions;
```
cmnrd, line 1: Also this we should only add once we implement federates in Rust

org.lflang/src/org/lflang/federated/extensions/TSExtension.java
```diff
@@ -0,0 +1,163 @@
...
+            sendingPort.getContainer().getName(),
+            sendingPort.getVariable().getName(),
+            connection.getDstFederate().id,
+            connection.getDstFederate().networkMessageActions.size()
+        );
+    }
+
+    @Override
+    public String generateNetworkInputControlReactionBody(int receivingPortID, TimeValue maxSTP, CoordinationType coordination) {
+        return "// TODO(hokeun): Figure out what to do for generateNetworkInputControlReactionBody";
```
cmnrd, line 64: there are still a couple of todos here

org.lflang/src/org/lflang/federated/generator/FedConnectionInstance.java
```diff
@@ -0,0 +1,123 @@
...
+import org.lflang.generator.SendRange;
+import org.lflang.lf.Connection;
+
+/**
+ * Class representing a federated connection.
+ */
+public class FedConnectionInstance {
+
+    /**
+     * FIXME
```
cmnrd, line 15: Lots of FIXMEs

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -83,6 +85,10 @@ public ReactionInstanceGraph(ReactorInstance main) {
     public void rebuild() {
         this.clear();
         addNodesAndEdges(main);
+
+        // FIXME: Use {@link TargetProperty#EXPORT_DEPENDENCY_GRAPH}.
```
cmnrd, line 89: FIXME

org.lflang/src/org/lflang/generator/rust/CargoDependencySpec.java
```diff
@@ -28,6 +28,7 @@
 import java.util.LinkedHashMap;
```
cmnrd, line 28: Why do these Rust files change? Pinging @oowekyala to take a look at the changes

test/Cpp/src/federated/DistributedCount.lf
```diff
@@ -0,0 +1,18 @@
+target Cpp
```
cmnrd, line 1: Since there is no federation support in C++, we should remove this.

cmnrd (commented):
This is really a huge PR and tough to review. I made a pass over the changes and focused mostly on changes in the core code base (outside the federated and target code generators). Most of my comments should be relatively easy to address. I think we really need to reconsider the attributes added in this PR though.

org.lflang/src/org/lflang/federated/validation/FedValidator.java
```diff
@@ -0,0 +1,69 @@
+package org.lflang.federated.validation;
```
cmnrd, line 1: Wouldn't this be better located in `org.lflang.validation`?

petervdonovan:
I am having trouble with generated delay reactors in Python. Maybe @cmnrd knows more?

For context:
```
This still generates wrong code. For example, AfterNoTypes fails with a
segfault because the body of _lf_gendelay_0reaction_function_0
(everything after `cinit`, which ends around line 256) is wrong. But I
am having trouble tracing through the code generator to find out why.
```

lhstrh:
> > I am having trouble with generated delay reactors in Python. Maybe @cmnrd knows more?
> 
> 
> 
> No, sorry. In https://github.com/lf-lang/lingua-franca/pull/1508, I only moved code around, but did not touch the internals of the delay generation in the targets. Did something change in the way the delay bodies are generated in Python it this branch? I could imagine that if the delay bodies changed in this branch, merging with master simply deleted theses updated functions and added the old ones in the new place.
> 
> 
> 
> 

We should inspect the merge closely. Probably threw away chaness from fed-gen.

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -481,9 +481,7 @@ public static void writeToFile(CharSequence text, Path path) throws IOException
     }
 
     public static void createDirectoryIfDoesNotExist(File dir) {
-        if (dir.exists()) return;
-        if (dir.mkdirs()) return;
-        throw new RuntimeIOException("Failed to create the directory " + dir);
```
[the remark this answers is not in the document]
oowekyala: In fact the argument could be a Path, since callers convert a Path to a File to call this.

petervdonovan:
It looks like the C (and maybe CCpp) test failures that we are seeing now -- in particular, the STP violations and determinism violations -- existed even before we merged `fed-gen-wip`. I think they are the same as those mentioned in [my previous comment in the fed-gen-wip PR](https://github.com/lf-lang/lingua-franca/pull/1521#issuecomment-1377827620). I will probably have to do a binary search through the commit history.

org.lflang.tests/src/org/lflang/tests/TestRegistry.java
```diff
@@ -78,7 +78,7 @@ public Set<LFTest> getTests(Target t, TestCategory c) {
      * test file that has a directory in its path that matches an entry in this
      * array will not be discovered.
      */
-    public static final String[] IGNORED_DIRECTORIES = {"failing", "knownfailed", "failed"};
+    public static final String[] IGNORED_DIRECTORIES = {"failing", "knownfailed", "failed", "fed-gen"};
```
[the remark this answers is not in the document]
lhstrh: I agree. I think it should be left out.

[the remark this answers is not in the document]
lhstrh:
```suggestion
    public static final String[] IGNORED_DIRECTORIES = {"failing", "knownfailed", "failed"};
```

org.lflang/src/org/lflang/TargetConfig.java
lhstrh: Sounds reasonable, but there is some other target-specific stuff here already (i.e., the `rust` member right above). I don't understand why it's here at all, or what it is, actually. The comment also seems wrong.

org.lflang/src/org/lflang/ast/FormattingUtils.java
lhstrh: I agree, this isn't right. Will attempt a fix.

org.lflang/src/org/lflang/ast/FormattingUtils.java
lhstrh: Should it return `null`?

org.lflang/src/org/lflang/federated/OldFedFileConfig.java
lhstrh: I have no idea. I'm all for removing it. Let me see if it's used anywhere. If not, I'll chunk it.

org.lflang/src/org/lflang/federated/extensions/TSExtension.java
lhstrh: @hokeun?

org.lflang/src/org/lflang/federated/generator/FedConnectionInstance.java
lhstrh: It was left like this by the author. I'll make an attempt at fixing it.

lhstrh (commented):
Responded to some comments, but haven't made a full pass yet.

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -481,7 +481,7 @@ protected void generateReaction(Reaction reaction, ReactorDecl decl, int reactio
 
         // Reactions marked with a `@language(C)` attribute are generated in C
         var reactionLanguageAttr = AttributeUtils.findReactionLanguageAttribute(reaction);
-        if (reactionLanguageAttr != null && reactionLanguageAttr.equals(Target.C)) {
+        if (reactor.getName().contains(DelayBodyGenerator.GEN_DELAY_CLASS_NAME) || reactionLanguageAttr != null && reactionLanguageAttr.equals(Target.C)) {
```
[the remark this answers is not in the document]
petervdonovan: Resolved by ffa5f0f

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -306,6 +421,10 @@ public enum TargetProperty {
      * or a dictionary of options that includes the string name.
      */
     PLATFORM("platform", UnionType.PLATFORM_STRING_OR_DICTIONARY, Target.ALL, 
+            (config) -> {
+                // FIXME: add code to turn platform config into AST node
```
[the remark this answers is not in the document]
arengarajan99: I'll have a crack at it.

org.lflang/src/org/lflang/federated/OldFedFileConfig.java
lhstrh: Looks like it's still being used, but it's unclear to me why. It looks like nonsense to me. I'll just remove it and we'll see what breaks.

petervdonovan:
According to the core dump, the `FloatingPointException` which appears in `DistributedPhysicalActionUpstreamLong` (and a few other failing tests, I think) [is thrown from this line](https://github.com/lf-lang/reactor-c/blob/a03131a1e846fdd42632a1db94a7915362a3a58d/include/core/threaded/worker_assignments.h#L221), which is used only in the `adaptive` scheduler.

I don't know why we are using that scheduler in the first place, but anyway -- it means that we're trying to schedule a reaction at a level that the adaptive scheduler thinks should never have any reactions scheduled on it. Or we are trying to schedule a reaction at a level that is outside the bounds of the `num_workers_by_level` array. In any case, `num_workers_by_level[level]` is zero.

org.lflang/src/org/lflang/federated/extensions/CExtensionUtils.java
```diff
@@ -246,13 +246,25 @@ public static void handleCompileDefinitions(
...
         initializeClockSynchronization(federate, federationRTIProperties);
     }
 
+    /**
+     * The number of threads needs to be at least one larger than the input ports
+     * to allow the federate to wait on all input ports while allowing an additional
+     * worker thread to process incoming messages.
+     * @return The minimum number of threads needed.
+     */
+    public static int minThreadsToHandleInputPorts(FederateInstance federate) {
```
lhstrh, line 262: I'm amazed that you were able to pinpoint this problem and implement a fix! Well done! That said, this shows again that the approach is exceptionally brittle and won't scale well. We need a better solution...

org.lflang/src/org/lflang/federated/extensions/TSExtension.java
hokeun: If I remember correctly, these were added by @Soroosh129 as placeholders when he first created this file. I think we had some discussion on control reaction bodies with @lhstrh and @byeong-gil last fall (or summer), but I can't remember right now. Let me think about it again...

hokeun: Ah, I remembered this. These are related to port absent messages, and we currently don't have any test requiring port absent yet. Thus, we can leave this as is for now. I think we first need to have the port absent implementation in reactor-ts merged, like these two : https://github.com/lf-lang/reactor-ts/pull/119, https://github.com/lf-lang/reactor-ts/pull/117

petervdonovan:
[It looks like we have a deadline violation in the enclave tests.](https://github.com/lf-lang/lingua-franca/actions/runs/3922259891/jobs/6705087765#step:7:453) Do you know how we should handle this, @cmnrd? It seems unlikely that this issue resulted from the changes in this branch, but I'm not sure whether it is a real concurrency bug or just a result of the timing variability in CI.

petervdonovan:
This error is new:
```
    pnpm: Command failed with exit code 128: git ls-remote --refs git+ssh://git@github.com/lf-lang/reactor-ts.git fed-gen
    Host key verification failed.
    fatal: Could not read from remote repository.
```
I do not think that we introduced changes that would cause this. Maybe it is just a transient error.

lhstrh:
> This error is new:
> 
> ```
>     pnpm: Command failed with exit code 128: git ls-remote --refs git+ssh://git@github.com/lf-lang/reactor-ts.git fed-gen
>     Host key verification failed.
>     fatal: Could not read from remote repository.
> ```
> 
> I do not think that we introduced changes that would cause this. Maybe it is just a transient error.

It is transient. I have seen it before...

org.lflang/src/org/lflang/ast/FormattingUtils.java
petervdonovan: It is probably because sometimes (not in actual code generation but maybe in debugging), we try to format a snippet of the AST that is not a full `Model`. Therefore, the target declaration is unavailable, and so the target is unknown.

petervdonovan:
For the record, the newly failing CppRos2Tests probably should have been failing previously. It looks like the fix in the tests by @lhstrh exposed another bug in the tests, where the AST transformation that was used to set the `ros2` target property was not having any effect.

org.lflang/src/org/lflang/ast/FormattingUtils.java
lhstrh: Right. For that reason, I changed it so that a `RuntimeException` is thrown. Is this the right type of exception, or should it be something like an `IllegalArgumentException`?

org.lflang/src/org/lflang/ast/FormattingUtils.java
lhstrh: Fixed.

org.lflang/src/org/lflang/ast/FormattingUtils.java
petervdonovan: I would be fine with either, although an `IllegalArgumentException` would be more specific I suppose.

I suppose that if we really want to do it the "right" way, we would replace references to `EObject` with `Model` so that the object passed to the `render` function has to be a `Model`. Then the runtime exception would not be thrown. I do not know how much work it would be to make it that way however, and I would be fine deferring that refactoring until the current design actually becomes a problem.

org.lflang/src/org/lflang/AttributeUtils.java
```diff
@@ -159,6 +159,35 @@ public static boolean isSparse(EObject node) {
...
+        }
+        return null;
+    }
+
+    /**
+     * Return true if the reaction is unordered.
+     */
+    public static boolean isUnordered(Reaction reaction) {
+        for (var attribute : getAttributes(reaction)) {
+            if (
```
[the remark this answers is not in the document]
lhstrh: Fixed in 3afe50a78.

org.lflang/src/org/lflang/AttributeUtils.java
```diff
@@ -159,6 +159,35 @@ public static boolean isSparse(EObject node) {
         return findAttributeByName(node, "sparse") != null;
     }
 
+    /**
+     * Return the language attribute on a reaction, or null if none is defined.
+     */
+    public static Target findReactionLanguageAttribute(Reaction reaction) {
+        for (var attribute : getAttributes(reaction)) {
```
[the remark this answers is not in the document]
lhstrh: Fixed in 3afe50a78.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
lhstrh: I don't know what this is. I guess I'll just remove the `FIXME`.

lhstrh:
```suggestion
```

org.lflang/src/org/lflang/generator/rust/CargoDependencySpec.java
lhstrh: I don't know about this either, but I did see @oowekyala chime in recently. Are we good to go with this, @oowekyala?

cmnrd:
> [It looks like we have a deadline violation in the enclave tests.](https://github.com/lf-lang/lingua-franca/actions/runs/3922259891/jobs/6705087765#step:7:453) Do you know how we should handle this, @cmnrd? It seems unlikely that this issue resulted from the changes in this branch, but I'm not sure whether it is a real concurrency bug or just a result of the timing variability in CI.

It can be ignored here. I will try to further relax the timings on master. Apparently CI on MacOS is particularly unpredictable...

org.lflang/src/org/lflang/federated/extensions/TSExtension.java
hokeun: Here's ongoing PR for adding: NetworkOutputControlReactionBody https://github.com/lf-lang/lingua-franca/pull/1333/files I'll work with @byeong-gil to merge this PR soon.

org.lflang/src/org/lflang/validation/AttributeSpec.java
```diff
@@ -210,9 +239,26 @@ enum AttrParamType {
         ));
         // @sparse
         ATTRIBUTE_SPECS_BY_NAME.put("sparse", new AttributeSpec(null));
+        // @language(lang)
+        ATTRIBUTE_SPECS_BY_NAME.put("language", new AttributeSpec(
```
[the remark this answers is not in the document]
cmnrd: Fixed in https://github.com/lf-lang/lingua-franca/pull/1552

org.lflang/src/org/lflang/validation/AttributeSpec.java
cmnrd: Fixed in https://github.com/lf-lang/lingua-franca/pull/1552

org.lflang/src/org/lflang/validation/AttributeSpec.java
cmnrd: Fixed in https://github.com/lf-lang/lingua-franca/pull/1552

.idea/codeStyles/Project.xml
```diff
@@ -70,13 +69,16 @@
       <option name="USE_SCALADOC2_FORMATTING" value="true" />
     </ScalaCodeStyleSettings>
     <codeStyleSettings language="JAVA">
-      <option name="RIGHT_MARGIN" value="100" />
+      <option name="RIGHT_MARGIN" value="80" />
```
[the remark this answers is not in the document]
cmnrd: Fixed in https://github.com/lf-lang/lingua-franca/pull/1552

cmnrd (commented):
Resolved a couple of conversations that are fixed by https://github.com/lf-lang/lingua-franca/pull/1552

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -85,7 +84,7 @@ public abstract class TestBase {
     private static final PrintStream err = System.err;
 
     /** Execution timeout enforced for all tests. */
-    private static final long MAX_EXECUTION_TIME_SECONDS = 60;
+    private static final long MAX_EXECUTION_TIME_SECONDS = 300;
```
cmnrd, line 87: Do we really want to increase the timeout by factor 5? Why is this needed?

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -477,6 +474,7 @@ private GeneratorResult generateCode(LFTest test) throws TestError {
         try {
             generator.doGenerate(test.getFileConfig().resource, fileAccess, test.getContext());
         } catch (Throwable e) {
+            e.printStackTrace();
```
cmnrd, line 477: This should not be needed as we rethrow and the outer context will print the stacktrace.

petervdonovan: I think the reason why I agreed this was that the stack trace only went as far as the point where the error was rethrown. I don't think I could even see the name of the original exception. I could have missed something though.

I could be mistaken, but I perceive it as a recurrent problem that we take pains to make test output look nice, and in the process we make the framework complex and introduce bugs that swallow useful information.

org.lflang.tests/src/org/lflang/tests/TestBase.java
lhstrh: :man_shrugging: WDYT, @petervdonovan?

petervdonovan: I myself almost forgot that I did this. Here is the commit message that I wrote last week:
> Try to address failing Python docker tests.
> Warning: increase in TestBase.MAX_EXECUTION_TIME_SECONDS! This is
apparently necessary because of the quantity of dependencies that must
be downloaded in order to run Python Docker tests.

Apparently a test timed out once because of this. We can reset the timeout to what it was before, but there may be occasional test failures as a result that will waste our time.

I suppose the downside of increasing the timeout is that it will waste resources in terms of energy and CI runners if we let lots of tests time out e.g. with livelocks without noticing and cancelling everything. I'm not sure how serious an issue this would be.

lhstrh: Do you think `120` or `180` would suffice? I agree with @cmnrd that `300` seems excessive.

petervdonovan: I do not know, but probably.

org.lflang.tests/src/org/lflang/tests/TestBase.java
petervdonovan: Just now I ran into a similar problem at line 546. The only output I get is this:
```
Reported issues:
Exception during test execution.
org.lflang.tests.TestError: Exception during test execution.
	at org.lflang.tests.TestBase.execute(TestBase.java:546)
	at org.lflang.tests.TestBase.validateAndRun(TestBase.java:700)
	at org.lflang.tests.TestBase.runTestsAndPrintResults(TestBase.java:197)
	at org.lflang.tests.TestBase.runTestsFor(TestBase.java:244)
	at org.lflang.tests.TestBase.runTestsForTargets(TestBase.java:221)
	at org.lflang.tests.RuntimeTest.runDockerTests(RuntimeTest.java:143)
	at org.lflang.tests.runtime.PythonTest.runDockerTests(PythonTest.java:119)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:104)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:98)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:214)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:210)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:135)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:66)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:151)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:35)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:107)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:88)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.lambda$execute$0(EngineExecutionOrchestrator.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.withInterceptedStreams(EngineExecutionOrchestrator.java:67)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:52)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:114)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
	at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.processAllTestClasses(JUnitPlatformTestClassProcessor.java:99)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.access$000(JUnitPlatformTestClassProcessor.java:79)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor.stop(JUnitPlatformTestClassProcessor.java:75)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:61)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at jdk.proxy2/jdk.proxy2.$Proxy5.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$3.run(TestWorker.java:193)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:133)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:71)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
```

Whereas printing the stack trace gives a much more useful output:
```
java.io.IOException: Cannot run program "docker compose": error=2, No such file or directory
```

We could get clever by catching the `TestException` somewhere else, higher up in the call stack, and printing the stack trace of the original exception at some special region of the log, but I would argue that that would be the wrong direction to take. We should not value the beauty of the output over the more hidden advantage of simple, easy-to-verify code.

org.lflang.tests/src/org/lflang/tests/TestBase.java
petervdonovan: Timeout reduced to 180 seconds by [6ab91c2](https://github.com/lf-lang/lingua-franca/pull/1221/commits/6ab91c28e61082bd71f5f5de72a779b898e844b1); please unresolve if this is not satisfactory. It takes me 88 seconds (n=1) to run a Python Docker test, so I don't think it's excessive to set the timeout to be double that.

org.lflang.tests/src/org/lflang/tests/TestBase.java
cmnrd: > We could get clever by catching the TestException somewhere else, higher up in the call stack, and printing the stack trace of the original exception at some special region of the log, but I would argue that that would be the wrong direction to take. We should not value the beauty of the output over the more hidden advantage of simple, easy-to-verify code. 

This is exactly what we are doing since https://github.com/lf-lang/lingua-franca/pull/1498. There was only a small bug in the handling of `TestError` which lead to the wrong stacktrace being printed. I will push a fix. All in all, I would argue that this design is pretty robust and it was not all motivated by beatifying the output. The major advantage is that all the error handling is now done in one single place and not scattered all over the place. So overall the risk of introducing bugs should be much lower.

cmnrd: Fixed in https://github.com/lf-lang/lingua-franca/pull/1221/commits/c4c7dcfe57b580c9bdf49af43ea8c483a9e7fca2

cmnrd:
I just noticed that there are  some more weird issues with the Docker support. It seems that docker files are now generated for all federated programs, no matter if they have docker support enabled or not. Also, even containerized federated programs produce a then useless launch script.

lhstrh:
Error code `139` observed in [this CI run](https://github.com/lf-lang/lingua-franca/actions/runs/3951505115/jobs/6767511875) suggests a segfault due to wrong format specifiers in a `printf`. I have seen a bunch of warnings in the C compilation output of both federates and the RTI, but I'm not sure if they could be the source of the problem. Tagging folks who might be able to help: @erlingrj, @edwardalee, @petervdonovan.

lhstrh:
I notice that if I run the same federated containerized test in rapid succession, then after a few tries I inevitably run into this:
```
DistributedCountContainerized-rti  | WARNING: RTI failed to accept the socket. Resource temporarily unavailable. Trying again.
DistributedCountContainerized-p    | DEBUG: _lf_create_token: Allocated memory for token: 0x7f7b570ae930
DistributedCountContainerized-p    | Federate 1: LOG: Connecting to the RTI.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15045. Trying 15046.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15046. Trying 15047.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15047. Trying 15048.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15048. Trying 15049.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15049. Trying 15050.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-p    | Federate 1: Failed to connect to RTI on port 15050. Trying 15051.
DistributedCountContainerized-p    | Federate 1: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-rti  | WARNING: RTI failed to accept the socket. Resource temporarily unavailable. Trying again.
```

This log suggests to me that the problem is in the RTI, not with the federates. This brings me to another question: which version of the RTI is being used here? The RTI doesn't have any version numbers, so I guess nobody can answer this question, but where does the image come from? The `docker-compose.yml` file has an entry `image: lflang/rti:rti` which suggests that an image has been published somewhere. By whom? How do we update it? @housengw do you happen to know anything about this?

lhstrh:
OK, it looks like the docker image was published by @housengw: https://hub.docker.com/u/housengw, ten months ago.

edwardalee:
> Error code `139` observed in [this CI run](https://github.com/lf-lang/lingua-franca/actions/runs/3951505115/jobs/6767511875) suggests a segfault due to wrong format specifiers in a `printf`. I have seen a bunch of warnings in the C compilation output of both federates and the RTI, but I'm not sure if they could be the source of the problem. Tagging folks who might be able to help: @erlingrj, @edwardalee, @petervdonovan.

I put in quite a bit of effort to purge warnings, but never got to the RTI or federated code. Help would be welcome.

In [lf_tag_64_32.h](https://github.com/lf-lang/reactor-c/blob/main/include/core/platform/lf_tag_64_32.h) it says:

```
// Define PRINTF_TIME and PRINTF_MICROSTEP, which are the printf
// codes (like the d in %d to print an int) for time and microsteps.
// To use these, specify the printf as follows:
//     printf("Time: " PRINTF_TIME "\n", time_value);
// On most platforms, time is an signed 64-bit number (int64_t) and
// the microstep is an unsigned 32-bit number (uint32_t).
// Sadly, in C, there is no portable to print such numbers using
// printf without getting a warning on some platforms.
// On each platform, the code for printf if given by the macros
// PRId64 and PRIu32 defined in inttypes.h.  Hence, here, we import
// inttypes.h, then define PRINTF_TIME and PRINTF_MICROSTEP.
// If you are targeting a platform that uses some other type
// for time and microsteps, you can simply define
// PRINTF_TIME and PRINTF_MICROSTEP directly in the same file that
// defines the types _instant_t, _interval_t, and _microstep_t.
#include <inttypes.h>
#define PRINTF_TIME "%" PRId64
#define PRINTF_MICROSTEP "%" PRIu32

// For convenience, the following string can be inserted in a printf
// format for printing both time and microstep as follows:
//     printf("Tag is " PRINTF_TAG "\n", time_value, microstep);
#define PRINTF_TAG "(" PRINTF_TIME ", " PRINTF_MICROSTEP ")"
```

org.lflang/src/org/lflang/TargetConfig.java
lhstrh: https://issue.lf-lang.org/1558

org.lflang/src/org/lflang/federated/validation/FedValidator.java
lhstrh: I agree, but there a few examples of badly placed classes. I've created an issue so we can move on and do this later: https://github.com/lf-lang/lingua-franca/issues/1561

lhstrh (commented):
This has gotten to a point where I think it is ready to merge, modulo whatever errors might still come up in CI.

PR #1223, opened 2022-06-11
Eliminate self-> for state references

edwardalee:
This PR eliminates the need for `self->name` when referencing state variables in a reaction or method body. You can just references them by name directly. The change is backward compatible, so existing programs using `self->name` will still work.


edwardalee:
I have decided that this approach is really not viable because it can't distinguish between a reference to `foo->value` and an input with name `value`.

PR #1224, opened 2022-06-11
Machine-applicable refactorings

petervdonovan:
I did a quick (< 2 hour) pass through some files related to the C generator because the warnings appearing in the IDE were distracting. The only changes that should make any semantic difference to our code are replacement of `==` with `.equals` to compare strings.

In the interest of minimizing conflicts (and of avoiding controversy), I tried to do this quickly and neglected many more potential issues that IntelliJ flagged. Maybe some more "code health" passes of this sort should happen if we think it's a good idea.

Changes include:
* Replacing redundant type arguments with diamond types
* Removing redundant `toString()` calls in string concatenation
* Using pattern matching to avoid casting (after using `instanceof`)
* Fixing broken references in Javadoc comments
* Deleting or updating outdated method documentation
* **Replacing usages of == and != to compare strings!** This seems like a real bug, and it surprises me that it was not caught in CI.

Changes do not include:
* Removing unnecessary parentheses
* Always surrounding + with spaces when used for string concatenation
* Making members final
* Deleting unused class variables
* Deleting unused methods
* Deleting unused parameters
* Deleting useless local variables
* Addressing warnings that point to logical errors in the control flow of methods
* Addressing "wrong tag" warnings for comments of the form "@author{Name Surname <email.address@university.edu>}" that make "@author{Name" look like all one JavaDoc tag
* Refactoring "pseudo functional style code" that causes side effects, even though it uses APIs designed for functional programming

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -475,7 +475,7 @@ public class CGenerator extends GeneratorBase {
                 //  Visual Studio compiler is extensive.
                 return false;
             }
-            if (targetConfig.useCmake == false) {
+            if (!targetConfig.useCmake) {
                 errorReporter.reportError(
                     "Only CMake is supported as the build system on Windows. "+
                     "Use `cmake: true` in the target properties. Exiting code generation."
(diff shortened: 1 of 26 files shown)


org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1081,19 +1081,19 @@ private void generateReactorChildren(
     private void pickCompilePlatform() {
         var OS = System.getProperty("os.name").toLowerCase();
```
lhstrh, line 1082: The uppercase is a bit weird here. I'd write `osn` or something like that.

lhstrh (approved):
Thanks for these cleanups!

PR #1227, opened 2022-06-13
Automatic LF code formatter

petervdonovan:
Tasks:
 - [x] Implement missing cases in visitor pattern
 - [x] Implement line wrapping (Billy, Peter) -- (edit: @petervdonovan is making another pass at this)
 - [x] Preserve comments - @petervdonovan 
 - [x] Allow the user to disable formatting for specific sections of their code (see comments below) - @petervdonovan 
 - [x] Add CI check that applies formatter to all *`.lf` files in the `test` directory and checks for equivalence (@lhstrh)
 - [x] Add methods for testing semantic equivalence of AST subtrees
 - [x] Rename `org.lflang.lfc` to `org.lflang.cli` - @billy-bao 
 - [x] Create `lff` package and create `lff` application that applies formatter - @billy-bao 
 - [x] Factor out WIP toward new `FedGenerator` (to be submitted in a separate PR)


cmnrd (commented):
This looks like a great use for the switch classes! :) So far everything looks great and I am excited to see the formatted code.

One question: Do you think it would be possible to combine this with target language formaters and run, let's say, `clang-format` on the code bodies?

petervdonovan:
> Do you think it would be possible to combine this with target language formaters and run, let's say, clang-format on the code bodies?

Yes, I believe we plan to do that, although it may be a separate PR.

petervdonovan:
I just realized that this approach to formatting can make code like this unreadable:
```
    test = new TraceTesting(
        events_size = 5,
        trace_size = 165,
        trace = (
            0,0,0,1,1,0,0,0,0,0,0,
            500000000,0,0,0,1,1,1,0,0,0,0,
            250000000,0,0,1,1,0,1,0,0,0,0,
            250000000,1,1,0,1,0,1,0,0,0,0,
            0,0,1,0,1,0,1,1,1,0,0,
            500000000,0,1,0,1,0,1,0,1,1,1,
            250000000,0,1,0,1,0,1,1,1,0,1,
            250000000,1,1,0,1,0,1,0,1,0,1,
            250000000,0,1,0,1,1,1,0,1,0,1,
            250000000,0,1,1,1,0,1,0,1,0,1,
            500000000,1,1,0,1,1,1,0,1,0,1,
            0,0,1,0,1,0,1,1,1,0,1,
            500000000,0,1,0,1,0,1,0,1,1,1,
            250000000,0,1,0,1,0,1,1,1,0,1,
            250000000,1,1,0,1,0,1,0,1,0,1
        ), training = false)
```

The placement of newlines in this trace definition seems important, but if we are generating straight from the AST, that information is erased. It ends up looking like:
```
    test = new TraceTesting(
        events_size = 5,
        trace_size = 165,
        trace = (
            0,
            0,
            0,
            1,
            1,
            0,
            0,
            0,
            0,
            0,
            0,
            500000000,
            0,
            0,
            0,
            ...
```

My basic idea here was that the goal of the programmer is to craft an AST, not a text file (programs != ASCII art), but this is an exception.

We have the option to let the programmer include a comment to force the original format to be preserved. Something like:
```
        trace = (  // keep-format
            0,0,0,1,1,0,0,0,0,0,0,
            500000000,0,0,0,1,1,1,0,0,0,0,
            250000000,0,0,1,1,0,1,0,0,0,0,
            ...
```
This probably would be easy to implement, but it is not very pretty. I'm open to other ideas.

cmnrd:
I don't quite understand the code example. Is this LF code or target code?

Also other code formatters face the problem that you describe. Special comments that switch of the formatting are the standard solution. When it comes to formatting, there is also a lot of opinion involved about the right or best style. We could also consider making certain decisions configurable, but this is probably something we would tackle further down the road. However, personally, I like the philosophy of the Python formatter [black](https://github.com/psf/black), which does not allow configuration.

petervdonovan:
> I don't quite understand the code example. Is this LF code or target code?

It is LF code, copied directly from a [modal models test](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/modal_models/ModalStateResetAuto.lf).

Yes, I also like Black and was inspired by it. I do not really want configurability. I just brought up this issue because a near-term goal is to format _all_ of our tests with this formatter, and because I think even a Black fanatic would agree that the lists of numbers that appear in the modal models tests do benefit greatly from their particular use of whitespace.

cmnrd:
Ok, I see. I would be absolutely in favour of adding a comment that can switch formatting off. Black uses this: 
```Python
# fmt: off
np.array(
    [
        [1, 0, 0, 0],
        [0, -1, 0, 0],
        [0, 0, 1, 0],
        [0, 0, 0, -1],
    ]
)
# fmt: on
```

The main problem that I see, is that we need to be somehow able to switch formatting back on or limit the scope. In your proposed solution:
```
        trace = (  // keep-format
            0,0,0,1,1,0,0,0,0,0,0,
            500000000,0,0,0,1,1,1,0,0,0,0,
            250000000,0,0,1,1,0,1,0,0,0,0,
            ...
```
would the `// keep-format` only apply to the annotated AST node?

We could also consider using the new annotations for this purpose. This might be clearer than comments, espacially since it would clearly separate annotations from comments in target code blocks that influence target code formatting.

petervdonovan:
> would the `// keep-format` only apply to the annotated AST node?

Yes. As it turns out, having a single comment that applies to the annotated AST node is much easier to implement than using two comments to toggle the formatter on and off. This is because, in order for the formatter to work correctly, it has to associate comments with AST nodes regardless. I would like to keep the `// keep-format` approach unless anyone feels sure that a different approach is better.

Soroosh129:
I think force-pushing to this branch that has dependents is not a great idea. This created a lot of issues in the fed-gen branch because the history was lost.

petervdonovan:
Can we try to merge this soon?

I promise that the change I just made was trivial and that this has mostly just been awaiting review for 23 days.

lhstrh:
Sorry, this was not on my radar as my review wasn't requested. I'll have a look at this soon. I agree, we should get this merged...

.idea/codeStyles/Project.xml
```diff
@@ -47,6 +46,9 @@
           <emptyLine />
         </value>
       </option>
+      <option name="ALIGN_MULTILINE_RECORDS" value="false" />
```
lhstrh, line 49: Do these changes belong in this PR? What are they for?

org.lflang.cli/src/org/lflang/cli/CliBase.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+    }
+
+    /**
+     * Obtains a resource from a path. Returns null if path is not an LF file.
+     * @param path The path to obtain the resource from.
+     * @return The obtained resource. Set to null if path is not an LF file.
+     */
+    // visible in tests
```
lhstrh, line 129: Don't understand this comment...
```suggestion
```

org.lflang/META-INF/MANIFEST.MF
```diff
@@ -35,7 +35,8 @@ Export-Package: org.lflang,
  org.lflang.serializer,
  org.lflang.services,
  org.lflang.util,
- org.lflang.validation
+ org.lflang.validation,
+ org.lflang.formatting2
```
lhstrh, line 39: Why are we introducing a `formatting2` package?

org.lflang/src/org/lflang/StringsBundle.properties
```diff
@@ -0,0 +1 @@
+VERSION = 0.2.2-SNAPSHOT
```
lhstrh, line 1:
```suggestion
VERSION = 0.3.1-SNAPSHOT
```

org.lflang/src/org/lflang/ast/MalleableString.java
```diff
@@ -0,0 +1,567 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.function.ToLongFunction;
+import java.util.stream.Collector;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import com.google.common.collect.ImmutableList;
+
+public abstract class MalleableString {
```
lhstrh, line 20: This class needs a description in JavaDoc.

.idea/codeStyles/Project.xml
petervdonovan: No, they do not belong. They change the IntelliJ settings to prevent auto-alignment, since I feel that the kind of auto-alignment that IntelliJ uses is a waste of horizontal space. They also move the vertical lines in the editor to 80 columns and 100 columns in accordance with our decision to keep comments 80 columns wide and code 100 columns wide.

There are a few things we could do about this:
- Stop tracking the IntelliJ settings so that each programmer can adapt the settings to their personal preferences without losing the changes to the settings whenever they need to pull or switch branches;
- Move these changes into a separate PR, with the understanding that they reinforce certain code style expectations;
- Undo these changes, and wait until we have a Java style guide before applying them.

org.lflang/src/org/lflang/formatting2/LFFormatter.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.inject.Inject;
+
+public class LFFormatter implements IFormatter2 {
+
+    @Inject
+    private IResourceValidator validator;
+
+    @Override
+    public List<ITextReplacement> format(FormatterRequest request) {
+        // TODO: Use a CancelIndicator that actually cancels?
```
lhstrh, line 30: Could you please create an issue to track this `TODO`?

org.lflang/META-INF/MANIFEST.MF
petervdonovan: If I remember correctly, I think Xtext expects to find formatting-related code in a package called `formatting2`, in the same way that it expects the validator to be in a certain package. I am not sure if this is a hard requirement though.

org.lflang/src/org/lflang/util/StringUtil.java
```diff
@@ -79,70 +79,72 @@ public static String removeQuotes(String str) {
         }
         return str;
     }
-
```
lhstrh, line 82: This class appears to have an overlap in purpose with the new `org.lflang/src/org/lflang/ast/FormattingUtils.java`. Would it make sense to merge them or scope them more clearly?

test/C/src/ActionDelay.lf
```diff
@@ -1,28 +1,29 @@
 // Test logical action with delay.
-target C;
+target C
```
lhstrh, line 2: I recall that we wanted to use semicolons in the C target for consistency, but I'm not sure.

lhstrh (commented):
This looks great and almost ready to merge! :rocket: :rocket: :rocket:  I left a few comments. The formatting of the tests looks excellent to me, modulo a single bike shedding issue: semicolons. I believe it makes sense to enforce them in the LF code if the target requires them, and I recall @edwardalee expressing the same preference.

org.lflang/src/org/lflang/StringsBundle.properties
lhstrh: Note to self: https://github.com/lf-lang/release-tools/issues/15

org.lflang/META-INF/MANIFEST.MF
lhstrh: OK, weird. But if Xtext is the reason, then sure :shrug:

.idea/codeStyles/Project.xml
lhstrh: The changes make sense. I vote for moving them to a separate PR, but I'm also fine with sneaking them in -- I mostly just wanted to understand what these changes meant :-)

test/C/src/ActionDelay.lf
petervdonovan: @cmnrd explained why his preference was _not_ to use the semicolons, and I agreed with him. I'm not sure I remember exactly what he said, but reasons include
- It makes sense to distinguish between LF code and target code because the two are semantically different. This makes the situation quite unlike the expressions discussion where we wanted to make an LF tuple expression look like a Python tuple. Unlike those target-code-like expressions, the lines of LF code that we would terminate with semicolons do not mean the same, and do not try to mean the same, as the statements appearing inside target code blocks. Therefore, those LF expressions should not try to _look_ the same as target code.
- The semicolons are unnecessary. They don't encode semantic meaning. They don't provide informative visual cues to programmers. They don't provide useful information to the compiler (except possibly in the case of _bugs_ that should be _fixed_, maybe in the case of the IPv4 and IPv6 rules that pollute our grammar if I remember correctly). Therefore, the principle of parsimony in all things written forbids the use of semicolons.

Insertion of the semicolons in the formatter would also increase the number of special cases in our implementation.

petervdonovan: The higher-level comment here is that it isn't a serious problem to have semicolons, and we can insert them if we like, but because there is a cost to having them, there should be a reason for using them. The hard-headedness of C programmers who do not want to see any new syntax probably should not be the only reason.

lhstrh: This sounds reasonable to me. I'm personally fine with not having them...

org.lflang/src/org/lflang/formatting2/LFFormatter.java
petervdonovan: Done (#1342).

.idea/codeStyles/Project.xml
petervdonovan: > I vote for moving them to a separate PR

Done!

test/C/src/ActionDelay.lf
petervdonovan: Given that no one else has followed up on this, is this PR ready to merge?

lhstrh: Yes, let's do it. We should follow up on the required semicolon in the `at`clause and revisit that construct all together. It doesn't appear in the tests and is not very well documented, so I don't think we need to address it before merging this in.

PR #1231, opened 2022-06-14
Add infrastructure for line wrapping.

petervdonovan:
This commit is a partial attempt to replace `String`s with `MalleableString`s that each admit multiple possible `String` representations. `MalleableStrings` compose with each other; the intent is to implement line wrapping by starting at the top-level `MalleableString`, changing its representation to one that reduces the line length, and repeating the same procedure through its nested `MalleableString`s until all line lengths are within the specified limit. However, this algorithm is not implemented yet.


petervdonovan:
There is a known bug that is causing the unit tests to fail. I just pushed this to get the general idea visible; the bugs will be dealt with if we do choose to go down this route.

PR #1232, opened 2022-06-14
String-based line-wrapping for formatter

billy-bao:
This is a basic implementation for line-wrapping LF code. The formatter takes into account the current level of indentation when breaking lines for line-wrapping, but will always break at spaces (violating the line-wrap limit if it cannot find a space before the limit). Set the line-wrap limit with ToLf.instance.setLineWrap().
Part of #1227.

### DIFF
--- a/org.lflang/src/org/lflang/ast/ToLf.java
+++ b/org.lflang/src/org/lflang/ast/ToLf.java
@@ -712,7 +712,7 @@ public class ToLf extends LfSwitch<String> {
     }
     
     /**
-     * Wraps a multi-line String based on the current state of lineWrap and indentLvl.
+     * Wrap a multi-line String based on the current state of lineWrap and indentLvl.
      * If lineWrap is 0, returns orgString. Otherwise, breaks lines such that if possible, each line has less than
      * (lineWrap - (indentLvl * INDENTATION) % lineWrap) characters, only breaking at spaces.
      * @param orgString The String to wrap.
@@ -724,7 +724,7 @@ public class ToLf extends LfSwitch<String> {
     }
     
     /**
-     * Wraps a single line of String based on the current state of lineWrap and indentLvl. See wrapLines().
+     * Wrap a single line of String based on the current state of lineWrap and indentLvl. See wrapLines().
      * @param line The line to wrap.
      * @return The wrapped line.
      */
--- a/org.lflang/src/org/lflang/ast/ToLf.java
+++ b/org.lflang/src/org/lflang/ast/ToLf.java
@@ -69,6 +69,11 @@ public class ToLf extends LfSwitch<String> {
 
     /** The number of spaces to prepend to a line per indentation level. */
     private static final int INDENTATION = 4;
+    
+    // Number of characters at which to line wrap.
+    private int lineWrap = 0;
+    // Current indentation level.
+    private int indentLvl = 0;
 
     // FIXME: This class needs to respect comments, which are lost at the EObject level of abstraction and must be
     //  obtained using NodeModelUtils like this: https://github.com/lf-lang/lingua-franca/blob/c4dfbd9cebdb9aaf249508360e0bac1ce545458b/org.lflang/src/org/lflang/ASTUtils.java#L1692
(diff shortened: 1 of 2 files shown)


org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -691,6 +710,55 @@ public String defaultCase(EObject object) {
...
+        return orgString.lines().map(this::wrapIndividualLine).collect(Collectors.joining(System.lineSeparator()));
+    }
+    
+    /**
+     * Wraps a single line of String based on the current state of lineWrap and indentLvl. See wrapLines().
+     * @param line The line to wrap.
+     * @return The wrapped line.
+     */
+    private String wrapIndividualLine(String line) {
+        int wrapLength = lineWrap - (indentLvl * INDENTATION) % lineWrap;
```
petervdonovan, line 732: This use of `%` ensures that `wrapLength >= 0`, but it looks strange and I am not sure it implements a behavior that users would expect.

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -722,7 +794,8 @@ private String typeAnnotationFor(Type type) {
      * @return A string representation of {@code statementListList}.
      */
     private String indentedStatements(List<EList<? extends EObject>> statementListList, int extraSeparation) {
-        return statementListList.stream().filter(((Predicate<List<? extends EObject>>) List::isEmpty).negate()).map(
+        indentLvl++;
```
petervdonovan, line 797: I am not sure if `indentLvl` really should be a class variable. If it were passed around as a method argument, that would restrict its scope and make it more explicit that its value is affecting the behavior of the method being invoked -- both of which might be nice.

petervdonovan (approved):
This implementation is way simpler and less messy than what I had in mind.

I left some nitpicks. My only real concern with this approach is what I said on Slack -- that to do this at the level of string processing as opposed to AST processing, while respecting single-line comments and strings and Python code blocks etc., will effectively result in a re-implementation of part of our parser. [We have done something like that with semantic highlighting](https://github.com/lf-lang/vscode-lingua-franca/blob/c321a5febf356cb147662cc96b0eadfa9d1bcdb5/src/highlight.ts#L165), but it was a little messy and I am not sure if it was a good idea (as in, we might eventually delete those 500 lines of code and replace them with queries to the language server, which has the AST).

I advocate for merging this into the `pretty-printer` branch now so that we can move on to other things, with the understanding that if we do run into the problems that I have talked about, I am not in favor of investing a lot of time into getting this to work well without directly processing the AST.

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -691,6 +710,55 @@ public String defaultCase(EObject object) {
...
+     * @param orgString The String to wrap.
+     * @return The wrapped String.
+     */
+    private String wrapLines(String orgString) {
+        if (lineWrap == 0) return orgString;
+        return orgString.lines().map(this::wrapIndividualLine).collect(Collectors.joining(System.lineSeparator()));
+    }
+    
+    /**
+     * Wraps a single line of String based on the current state of lineWrap and indentLvl. See wrapLines().
```
lhstrh, line 727:
```suggestion
     * Wras a single line of String based on the current state of lineWrap and indentLvl. See wrapLines().
```

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -691,6 +710,55 @@ public String defaultCase(EObject object) {
             object.getClass().getName()
         ));
     }
+    
+    /**
+     * Wraps a multi-line String based on the current state of lineWrap and indentLvl.
```
lhstrh, line 715:
```suggestion
     * Wrap a multi-line String based on the current state of lineWrap and indentLvl.
```

org.lflang/src/org/lflang/ast/ToLf.java
billy-bao: While I agree that passing it around as an argument would be nicer, I think it would require large-scale changes (maybe modify LFSwitch so that doSwitch() passes an extra argument for everything so that caseCode() can have access to `indentLvl`?), which I'm not yet confident in making at my current level of familiarity. Any help or suggestions on this is appreciated.

petervdonovan: After reading your comment, I realized that I do not even understand why the indentation is being tracked at all. If you override `doSwitch` so that wrapping is applied at the very very end, you have the whole string right there, and you know exactly how long all the lines are, including all the indentation. That would also let you insert the line-wrapping logic into one single place instead of three places (`caseCode`, `list`, and `indentedStatements`).

billy-bao: The indentation is tracked to take advantage of existing code which breaks the input into indented blocks. If all line-wrapping is done at the end, I don't think there's any way to figure out which line should be indented by how much without redoing the parsing. Unless we cache the indentation levels for each line during the parsing? Then I think it'll be not very different from what I have now.

petervdonovan: Okay, let's just merge this into `pretty-printer` and move on.

Issue #1194, opened 2022-05-26
Cargo dependency names are parsed as identifiers


jhaye:
When defining the Rust target with a cargo dependency, that dependency's name is parsed as an identifier. This means that if it contains a keyword, LFC throws a parsing error.

For example, this target config:
```
target Rust {
    build-type: Release,
    cargo-features: ["cli"],
    cargo-dependencies: {
        cpu-time: "1",
    }
};
```
Throws the following error:
```
'time' is a reserved keyword which cannot be used as an identifier.
```

I'm not familiar with the parser, so I don't know how to fix this.

cmnrd:
Maybe @lhstrh can comment? But I am afraid that the only real fix for this is not using xtext...

cmnrd:
I don't know if it works, but one idea could be to use `Token` instead of `ID` for the keys.

lhstrh:
See #1238.

PR #1238, opened 2022-06-20
Widen set of permitted identifiers in target properties

lhstrh:
This PR is in response to issue https://github.com/lf-lang/lingua-franca/issues/1194.

For this fix to be viable, the following things still need to be done:
- [ ] adjust `getSyntaxErrorMessage` so that a parse errors in the target properties do not get mislabeled as an inappropriate use of keywords; and
- [ ] disable highlighting of keywords in target properties (I'm not sure how to do this; @a-sr do you have a suggestion?).


lhstrh:
This doesn't appear to work. I think this is because enums in Xtext are declarations of nonterminals that implicitly take precedence over other tokenization rules. I remember removing them from the grammar for this reason, but it looks like they reappeared (my guess as a recent PR related modal models). It looks like if we want to go down this road some refactoring is required...

a-sr:
I have never altered the default highlighting, I only added new highlighting but this is a different mechanism (semantic).
The [Xtext docmumentation for code highlighting](https://www.eclipse.org/Xtext/documentation/310_eclipse_support.html#highlighting) describes two types of highlighting: lexical and semantic. Default keyword highlighting is a result of the lexical highlighter. However, this mechanism seems to be context-free, probably for performance reasons. Adjusting the default lexical highlighting to detect the target property context and stop highlighting keywords there might do the trick. But only for Epoch, syntax highlighting in VSC is done independently from Xtext.

jhaye (approved):
Based on my testing this fixes the issue.

cmnrd (approved):
LGTM! Should we get this merges and resolve the syntax highlighting later (and open a new issue)?

lhstrh:
@jhaye, are you interested in taking a look at the syntax highlighting issue?

oowekyala:
Maybe we should just allow `ID("-" ID)*` like before but also add an option to use a string literal? This would circumvent this highlighting issue. The key-value pairs also already look like JSON so I don't think it's surprising
```yaml
target Rust {
    build-type: Release,
    cargo-features: ["cli"],
    cargo-dependencies: {
        "cpu-time": "1",  # here
    }
};
```

cmnrd:
I am not sure what the status of this PR/issue is, but this might become obsolete since we plan to move the target properties out of LF code and into an external toml file managed by a build tool.

lhstrh:
> Maybe we should just allow `ID("-" ID)*` like before but also add an option to use a string literal? This would circumvent this highlighting issue. The key-value pairs also already look like JSON so I don't think it's surprising

I agree, this might be a better solution. Want to push a fix to this branch?

PR #1240, opened 2022-06-21
Standalone code formatter

billy-bao:
In-progress implementation of lff, a standalone code formatter. Code is based on lfc. Part of #1227.
Test contributions are welcome.


org.lflang.cli/build.gradle
```diff
@@ -61,3 +74,43 @@ task runLfc(type: JavaExec) {
...
+    // The working directory will be the root directory of the lingua franca project
+    // CLI arguments can be passed to lff by using --args. Note that you need
+    // to escape cli flags which start with --.For instance --args ' --help'.
+    // Otherwise they're parsed as arguments to the Gradle CLI, not lfc.
+    description = "Build and run lff, use --args to pass arguments"
+    group = "application"
+    classpath = sourceSets.main.runtimeClasspath
+    mainClass = 'org.lflang.cli.Lff'
+    workingDir = '..'
+}
```
lhstrh, line 116:
```suggestion
}

```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+ * Based on lfc.
+ *
+ * @author {Marten Lohstroh <marten@berkeley.edu>}
+ * @author {Christian Menard <christian.menard@tu-dresden.de>}
+ * @author {Billy Bao <billybao@berkeley.edu>}
+ */
+public class Lff {
+
+    /// current lff version as printed by --version
+    private static final String VERSION = "0.2.2-SNAPSHOT";
```
lhstrh, line 51: Rather than replicating this string, we should move this elsewhere (probably a class in the root of the `org.lflang` package) and reference it from both `Lff` and `Lfc`.

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+     */
+    public static void main(final String[] args) {
+        final ReportingBackend reporter = new ReportingBackend(new Io(), "lff: ");
+
+        // Injector used to obtain Main instance.
+        final Injector injector = new LFStandaloneSetup(new LFRuntimeModule(), new LFStandaloneModule(reporter))
+            .createInjectorAndDoEMFRegistration();
+        // Main instance.
+        final Lff main = injector.getInstance(Lff.class);
+        // Apache Commons Options object configured to according to available CLI arguments.
```
lhstrh, line 150:
```suggestion
        // Apache Commons Options object configured according to available CLI arguments.
```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * Checks all given input paths and the output path, then invokes the formatter on all files given.
+     */
+    private void runFormatter(List<Path> files) {
+        String pathOption = CLIOption.OUTPUT_PATH.option.getOpt();
+        Path outputRoot = null;
+        if (cmd.hasOption(pathOption)) {
+            outputRoot = Paths.get(cmd.getOptionValue(pathOption)).toAbsolutePath().normalize();
+            if (!Files.exists(outputRoot)) { // FIXME: Create it instead?
```
lhstrh, line 198:
```suggestion
            if (!Files.exists(outputRoot)) {
```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+            }
+        } catch (ParseException e) {
+            reporter.printFatalError("Unable to parse commandline arguments. Reason: " + e.getMessage());
+            formatter.printHelp("lff", options);
+            System.exit(1);
+        }
+    }
+
+    /**
+     * Checks all given input paths and the output path, then invokes the formatter on all files given.
```
lhstrh, line 191:
```suggestion
     * Check all given input paths and the output path, then invokes the formatter on all files given.
```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+                } else {
+                    formatSingleFile(path, outputRoot.resolve(path.getFileName()));
+                }
+            }
+        }
+        reporter.printInfo("Done formatting.");
+    }
+
+    /**
+     * Invokes the formatter on all files in a directory recursively.
```
lhstrh, line 235:
```suggestion
     * Invoke the formatter on all files in a directory recursively.
```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+                    }
+                }
+            }
+        } catch (IOException e) {
+            reporter.printError("Error reading directory " + curDir + ": " + e.getMessage());
+        }
+    }
+
+    /**
+     * Loads and validates a single file, then formats it and outputs to the given outputPath.
```
lhstrh, line 261:
```suggestion
     * Load and validates a single file, then formats it and outputs to the given outputPath.
```

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+        if (issueCollector.getErrorsOccurred() ) {
+            // if there are errors, don't print warnings.
+            List<LfIssue> errors = printErrorsIfAny();
+            String cause = errors.size() == 1 ? "previous error"
+                                              : errors.size() + " previous errors";
+            reporter.printFatalErrorAndExit("Aborting due to " + cause);
+        }
+    }
+
+    // visible in tests
```
lhstrh, line 305: Javadoc missing.

org.lflang.cli/src/org/lflang/cli/Lff.java
```diff
@@ -0,0 +1,351 @@
...
+                                                  issue.getLineNumber(), issue.getColumn(),
+                                                  issue.getLineNumberEnd(), issue.getColumnEnd(),
+                                                  issue.getLength(), FileUtil.toPath(uri)));
+            } catch (IOException e) {
+                reporter.printError("Unable to convert '" + uri + "' to path." + e);
+            }
+        }
+        return resource;
+    }
+
```
lhstrh, line 339: Javadoc missing.

lhstrh (changes requested):
This looks great! I'm only requesting minor changes. Unsurprisingly, there is some overlap between `Lff` and `Lfc`. Especially since we're planning to also create an `Lfd`, I think it makes sense to create a base class so we can share code between these classes.

org.lflang.cli/src/org/lflang/cli/Lff.java
billy-bao: Not sure where's a good place to put this. Would org.lflang.FileConfig work?

lhstrh: Not sure that's the right place either. I imagine it would be useful to gather strings in a single file to make it easy to style them a certain way and eventually do internationalization. The version number could be in a class like that. This might be the right time to start thinking about this. I believe Java uses property files and `ResourceBundle` classes to deal with collections of strings...

billy-bao: Did that!

org.lflang/src/org/lflang/StringsBundle.properties
```diff
@@ -0,0 +1 @@
+VERSION = 0.2.2-SNAPSHOT
```
lhstrh, line 1:
```suggestion
VERSION = 0.2.2-SNAPSHOT

```

lhstrh (approved):
Awesome! Very clean :rocket: :rocket: :rocket:

PR #1241, opened 2022-06-22
Adjust reset symbols in diagrams

### LINKED ISSUE #1219: Minor issues with diagrams of modal reactors
It looks like the state variable isn't rendered quite correctly in the diagram:
image

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1169#discussion_r894757345_

Also:

The rendering of the `reset` trigger seems a bit off:

image

I would actually like to propose an alternative visual syntax. How about something along these lines:
image

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1169#discussion_r894771808_

a-sr:
Fixes use of unicode symbols in state variables and revises reset trigger figure.
Closes #1219

Result:
reset
for
```
reactor R (param:int(0)){
    state s:int(0)
    reset state rs:int(0)
    reaction(reset) {==}
}
```


lhstrh:
This looks very cool! One concern I have is that the reset arrow might be hard to spot on the state. Would it make sense to wrap the reset arrow all the way around the bullet rather than have it dangle off the side?

a-sr:
I also thought about this but my feeling is that this would make the "bullet point" too large, resulting in a very inconsistent appearance with others.
Furthermore, I did not want to draw too much attention to it because, same as state variables and parameters in general, its relevance on the diagram level is rather limited because we do not show reactions bodies where this would have an effect.

lhstrh:
The aspect ratio is not quite as I meant, but here is a mock up of what it could look like:

image

I'm basically suggesting that the outer circle around the filled dot can be replaced by an arrow. That way, it would take up about as much space as an ordinary state icon.

a-sr:
Ah, now I get it. Yes, this is a good a idea. I will redesign it accordingly.

a-sr:
Adjusted appearance. Now it looks like this:
reset2

edwardalee:
Maybe lose the dot in the middle of the reset bullet? This might make it look cleaner and distinguish it better (and also match the event icon better).

lhstrh:
Hm, I'm unsure about this one. I agree that the new version relates more obviously to the `reset` trigger, but it looks less like a state variable, too, which might be confusing. My idea was for the bullet to represent the notion of "state variable" and the ring around it to indicate whether it gets reset (broken, with arrow) or not (solid).

a-sr:
I agree with @lhstrh, in the previous version the similarities between the sate variables were stronger, which also was my initial intention, to make the reset "only" a minor modification.
I can revert the last change, if we want to go with the previous design.

edwardalee:
Yes, I see this point... let's revert.

lhstrh (approved):
Nice!

PR #1246, opened 2022-06-23
Add platform target property

Soroosh129:
This is based on the work started by @sberkun in #1118.

### DIFF
--- a/test/C/src/target/Platform.lf
+++ b/test/C/src/target/Platform.lf
@@ -7,6 +7,6 @@ target C {
 
 main reactor {
     reaction(startup) {=
-        lf_print("Hello, Linux!");
+        lf_print("Hello, Mac!");
     =}
 }
--- a/test/C/src/target/Platform.lf
+++ b/test/C/src/target/Platform.lf
@@ -1,0 +1,12 @@
+/**
+ * Test that the cross-compile capability works as expected.
+ */
+target C {
+    platform: "Mac"
+};
+
+main reactor {
+    reaction(startup) {=
+        lf_print("Hello, Linux!");
+    =}
+}
--- a/org.lflang/src/org/lflang/TargetConfig.java
+++ b/org.lflang/src/org/lflang/TargetConfig.java
@@ -190,7 +190,9 @@ public class TargetConfig {
     public boolean noRuntimeValidation = false;
 
     /**
-     * Set the target platform to compile for.
+     * Set the target platform.
+     * This tells the build system what platform-specific support
+     * files it needs to incorporate at compile-time.
      *
      * @author Samuel Berkun (sberkun@berkeley.edu)
      */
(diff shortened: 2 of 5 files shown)


test/C/src/target/Platform.lf
```diff
@@ -0,0 +1,12 @@
+/**
+ * Test that the cross-compile capability works as expected.
```
lhstrh, line 2: What does this test, exactly? Isn't this expected to only work on macOS?

Soroosh129:
It should be noted that the `platform` target property is only part of the picture. There is [more to cross-compiling](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html) something than just setting the platform.

test/C/src/target/Platform.lf
Soroosh129: No, I'm hoping that this works on Linux too. The target tests for C only run on Mac and Linux.

lhstrh (approved):
This looks great, but I'm not sure about the test...

test/C/src/target/Platform.lf
lhstrh: > The target tests for C only run on Mac and Linux.

That's exactly why I'm asking. I would expect compilation to work, but the produced executable to not work on Linux or Windows... If not, when what does the target property do at all?

Soroosh129: As I commented below, there is a lot more to cross-compiling than just setting the platform. In that sense, the `platform` target property doesn't enable true cross-compiling on its own. Essentially, all it does right now is to select which one of the platform files we link against. The platform support files for Mac should work on Linux too, hence the test.

To enable true cross-compiling, we need to add additional missing pieces (marked with `???`):
```cmake
# the name of the target operating system
# set(CMAKE_SYSTEM_NAME Windows) # This is set by the platform target property

# which compilers to use for C and C++
set(CMAKE_C_COMPILER   i586-mingw32msvc-gcc) # ???
set(CMAKE_CXX_COMPILER i586-mingw32msvc-g++) # ???

# where is the target environment located
set(CMAKE_FIND_ROOT_PATH  /usr/i586-mingw32msvc
    /home/alex/mingw-install) # ???

# adjust the default behavior of the FIND_XXX() commands:
# search programs in the host environment
set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER) # ???

# search headers and libraries in the target environment
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY) # ???
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY) # ???
```

lhstrh: Got it. What do we have to do to support a `CMAKE_SYSTEM_NAME` that is not a popular OS but something like Arduino or FlexPRET? I did find https://github.com/arduino-cmake/Arduino-CMake-NG.

Soroosh129: This is very promising. I think if we want to use this, from the looks of the example in the README of that repo, we need to generate a custom CMakeLists.txt entirely for Arduino since executables are created by calling a `add_arduino_executable` method, which is a custom CMake method.

lhstrh: Tagging @arengarajan99 for :point_up:

lhstrh:
Looks like this is a go?

Issue #87, opened 2020-01-09
Update Validator to ban the use of target-language keywords as identifiers for ports, actions, etc.


lhstrh:
Example: a port named `in` will cause problems in the generated code. Yes, this will be caught by the target compiler, but the error messages can be rather cryptic. It's an easy thing to handle on our end...

mew2ub:
Resolved.

Issue #62, opened 2019-11-21
Handling shutdown


lhstrh:
> \martin{To stop periodic timers (and maybe kill other pending events in the queue)
> we need a request-for-shutdown function.}
> \marten{I'm not convinced that we do. Can't the base reactor class simply implement
> a reaction that does this upon being triggered by a shutdown action?}
> \martin{How do you stop a periodic timer? Who decides that a shutdown happens?
> We should have an example application to explore this.}
> \edward{I suggest moving this discussion to the bug tracking system.}

In the CyPhy paper we assume that the container of a reactor schedules shutdown when appropriate. If this does not satisfactorily address the question, we could explore alternatives.

schoeberl:
> On 21 Nov, 2019, at 20:38, Marten Lohstroh <notifications@github.com> wrote:
> 
> \martin{To stop periodic timers (and maybe kill other pending events in the queue)
> we need a request-for-shutdown function.}
> \marten{I'm not convinced that we do. Can't the base reactor class simply implement
> a reaction that does this upon being triggered by a shutdown action?}
> \martin{How do you stop a periodic timer? Who decides that a shutdown happens?
> We should have an example application to explore this.}
> \edward{I suggest moving this discussion to the bug tracking system.}
> 
> In the CyPhy paper we assume that the container of a reactor schedules shutdown when appropriate. If this does not satisfactorily address the question, we could explore alternatives.
> 
> 

What does it mean “when appropriate”? The LF wiki talks about a stop() method that can be called.

Martin

lhstrh:
Execution can be requested to halt by calling a runtime function, as explained in the wiki. _When_ it is called, the shutdown sequence unfolds conform the CyPhy paper, by letting the top-level reactor schedule a shutdown action for all reactors contained by it, which then also schedule shutdown for the reactors that they contain, etc. This goes on until the entire reactor hierarchy has reacted to a `shutdown` action.

cmnrd:
I am a bit confused by this statement. What does 'which **then** also schedule shutdown' mean? Are all the shutdown reactions logically simultaneous or is there a microstep delay between the shutdown reactions of different hierarchy levels?

lhstrh:
All actions, including `shutdown` are always scheduled with a microstep
delay. The only exception is `startup`, which is scheduled without a
microstep delay.

cmnrd:
[...]

PR #1247, opened 2022-06-23
New keyword for history transitions

### LINKED ISSUE #1220: Minor issues with the syntax of modal reactors
It appears that a `reset` transition is the default, meaning that is no modifier is given in the transition effect, this is to mean that it is a reset transition. This could be confusing, because it may suggest that a "normal" transition is of a different kind than `reset` or `continue`. We had a similar problem with actions, and decided that it would be better to require a disambiguation from the programmer and thus only allow `logical action` or `physical action`, and not just `action`. I think we should take the same approach here, and enforce this in the validator so we can present a useful error message.

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1169#discussion_r894762670_

And:

I'm still a bit on the fence about the `continue` keyword. If refer to this a "history transition" and label it with an "H" in the diagram, then why don't we just use the `history` keyword and avoid all confusion?

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1169#discussion_r894763524_

a-sr:
Changes keyword for history transitions from `continue` to `history`.
The documentation is updated by [website PR #87](https://github.com/lf-lang/website-lingua-franca/pull/87).
Update of the syntax highlighting for VSCode in [vscode PR #62](https://github.com/lf-lang/vscode-lingua-franca/pull/62).

Also adds warning if the user does not specify a transition type but the types would actually have different effects on the target mode.

Closes #1220

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
@@ -2369,6 +2369,25 @@ public class LinguaFrancaValidationTest {
             "The state variable can not be automatically reset without an initial value.");
     }
 
+    @Test
+    public void testUnspecifiedTransitionType() throws Exception {
+        String testCase = """
+            target C;
+            main reactor {
+                initial mode IM {
+                    reaction(startup) -> M {==}
+                }
+                mode M {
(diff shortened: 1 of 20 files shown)


lhstrh (approved):
This looks great! I'll merge this and merge the accompanying documentation PR as well.

PR #1249, opened 2022-06-24
Preserve comments

petervdonovan:
This branch includes comments, wraps comments, and wraps code.

The diff is almost entirely formatting changes to the tests.

~~The modal models tests are not reformatted for reasons explained [here](https://github.com/lf-lang/lingua-franca/pull/1227#issuecomment-1173532853).~~ A "keep-format" magic comment is implemented in this branch to ensure that the annotated AST node will not be formatted. This allowed the modal models tests to be run through the formatter without problems.

Here are some potentially controversial issues that might be discussed:
- Semicolons are always omitted from LF code.
- All code blocks, including reaction bodies, are crammed onto one line when possible. ~~Target properties are crammed onto one line as well~~ -- to quote [the Black docs](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#pragmatism), "it fits!"
  - This may conserve vertical space, but it hurts maintainability. It is slightly annoying to turn an inlined code block into a multiline one when you want to add a line of code, for example.
- When possible, single-line comments are also crammed onto the same line as the AST nodes that they appear to annotate.
- Comments are line-wrapped automatically, and things that used to appear on two consecutive lines (without a space in between) are considered part of the same paragraph -- the line break is ignored. This is consistent with how Markdown behaves. Lines starting with special symbols such as `&` and `-` and `@` are not merged into previous lines, however. This protects bullet points and `@author` tags, for example.
- The user can opt out of formatting for a specific AST node using a single `keep-format` comment. For context, see [the other formatting PR](https://github.com/lf-lang/lingua-franca/pull/1227#issuecomment-1173532853).
- ~~inputs, outputs, state, etc. are all separated within a reactor. This means that even if the original textual representation of the program had some other organization -- for example, putting inputs next to outputs that are related to them -- that organization will be ignored.~~
- ~~These different groups of reactor elements are separated by blank lines, even if they contain only one element each. This can waste vertical space.~~
- ~~Alignment of single-line comments on consecutive lines is ignored. We could go the other way by making the formatter automatically do this alignment for you when possible, but that would make the implementation more complicated, and it would result in larger diffs (commenting one line could cause the formatter to change the alignment of a comment on an adjacent line)~~
- Commented-out code is handled badly by this formatter. Orphaned comments (comments that are not associated with any particular AST node) are also handled imperfectly. I think this is suboptimal but acceptable for now. It is usually bad practice to maintain commented-out code. Although there may be valid use-cases for orphaned comments, I believe that such use cases are rare.

In many cases, I made these choices not because of a strong personal opinion, but rather because I did not care very much and because they just happened to make the implementation of the formatter simpler, e.g. by reducing the number of special cases.


petervdonovan:
I think this branch finally has the functionality that I wanted it to have. It includes line wrapping because line wrapping and handling of comments are somewhat related features that benefited from being designed together.

I still need to refactor before this will be ready for review.

petervdonovan:
If anyone is curious about the state of this branch, I just pushed some commits that change the tests to show what the formatter does to them. In some cases I manually helped the formatter, but in any case the text of the edited tests are fixed points.

petervdonovan:
If we want to continue to support `// @label` annotations appearing in single-line comments, that could potentially hold up this PR even further. It is possible that such comments would have to be handled specially by the formatter to be extra sure that they are not put out of place.

This issue manifests itself in `test/Rust/src/ReactionLabels.lf`.

lhstrh:
I think those annotations have been deprecated, no? We no longer need to put them in comments.

petervdonovan:
OK, I think this PR really is ready for review now. It still probably has some bugs, but I fixed everything I found after formatting all 700+ tests.
* Semicolons are still removed whenever possible.
* Target properties (and key value pairs in general) are now expanded over multiple lines.
* Single-line comments are aligned with each other when possible.
* I did not add a nullity check in the formatter (in case of a method that is not implemented) because if a method returns `null`, then it will immediately raise an NPE, which is about the best runtime exception we could hope for.

cmnrd (commented):
One thing that I find odd when looking at the formatted code snippets is that empty lines are inserted between inputs, outputs, actions, state variables and so on. I did not notice this before. Also it appears that there is a fixed order enforced by the code formatter (i.e. first inputs, then outputs, then timers, ...).

Code formatters work on a thin line between being a helpful tool and being annoying. Programmers are typically annoyed if the formatter does not leave room for developing an own style and being in charge of how a class (or a reactor) is best presented. I see the logic behind grouping all inputs together, but maybe for some reactors it makes sens to pair an input with an action and a state variable because they are tightly coupled ion the reactors logic. I think it should be up to the programmer to decide.

petervdonovan:
Thanks for taking a look! Yes, I agree that this is a potentially annoying feature of the formatter. Currently it is absolutist (programmers craft ASTs and not text files!) but maybe the re-grouping and the empty lines are too extreme.

cmnrd (commented):
Yeah, I think its easy to fall in the trap of seeing source code only as something for machines. But in the end we also write it so that it is comprehensive for humans.

I really like the formatting results! I only have one small comment: I noticed that the formatting also changes the indentation of code within reaction bodies. Generally speaking this is fine, but are you sure that this cannot have any weird effects in languages like Python where indentation matters?

I did not do a full code review yet. Actually it appears there is something odd going on with the branch. Maye a rebase went wrong? It includes >330 commits and many of them are replicated commits from master.

petervdonovan:
> are you sure that this cannot have any weird effects in languages like Python where indentation matters?

Well, there is an edge case that I should probably address. The formatter is not supposed to be changing relative indentation within the code blocks.

> It includes >330 commits and many of them are replicated commits from master.

Oops, sorry, that is because the target branch is `pretty-printer` and not `master`, and those commits are not in `pretty-printer`. I guess I should just merge this one into `pretty-printer` without review and then open `pretty-printer` for review instead, because this PR has been open for so long.

PR #1252, opened 2022-06-24
Mutation

Wonseo-C:
I implemented the Mutation Generator part.
1.  Recognizing Sub-objects of Mutation: Fix [“LFScopePrividerImpl.java”](https://github.com/lf-lang/lingua-franca/compare/master...Wonseo-C:mutation?expand=1#diff-05d855e94b9c3a5006e670afb1fa9663a213981551c7f751ffddea7af1d1b71d)

There is some issue: Mutation's effect cannot be recognized by fixing "LFScopePrivederImpl.java," so I immediately fixed the ["LinguaFranca.xtext"](https://github.com/lf-lang/lingua-franca/blob/3a1f550176af967811b200f28f22b1748f17392e/org.lflang/src/org/lflang/LinguaFranca.xtext#L223) to solve it ...~~It Must be fixed!~~ (Deleting the bracket is the right way because VarRef has already the bracket)

2. Code Generator (Mutation): Fix ["TSReactionGenerator"](https://github.com/lf-lang/lingua-franca/compare/master...Wonseo-C:mutation?expand=1#diff-5fdb14bc11c18f34a23a51db49bd847b0bc8b386cc889dff06cb825d2ab20cf5)
* **Get Mutation**: fix the function of generateAllReactions to make it get Mutation before Reactions if it exists.
* **generateSingleMutation**: It has same logic to generateSingleReaction, except the part of generateMutationSignatureElementForPortEffect.
* **generateMutationSignatureElementForPortEffect**: change the effect's variable interface as ReadWrite to WritablePort.

So, by this branch, we can generate [Sieve.lf](https://github.com/lf-lang/benchmarks-lingua-franca/pull/20/files) , which needs to implement the Mutation part.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CPortGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CPortGenerator.java
@@ -111,7 +111,7 @@ public class CPortGenerator {
             if (AttributeUtils.isSparse(input.getDefinition())) {
                 return String.join("\n", result,
                     "if ("+input.getWidth()+" >= LF_SPARSE_WIDTH_THRESHOLD) {",
-                    "    "+portRefName+"__sparse = _lf_allocate(1,",
+                    "    "+portRefName+"__sparse = (lf_sparse_io_record_t*)_lf_allocate(1,",
                     "            sizeof(lf_sparse_io_record_t) + sizeof(size_t) * "+input.getWidth()+"/LF_SPARSE_CAPACITY_DIVIDER,",
                     "            &"+reactorSelfStruct+"->base.allocations);",
                     "    "+portRefName+"__sparse->capacity = "+input.getWidth()+"/LF_SPARSE_CAPACITY_DIVIDER;",


Wonseo-C:
I did PR about generating mutation. And I cannot set the reviewer because I'm not a member.
So, if you have time, then can you review it?
I think you already know all of this content.
@lhstrh @hokeun

lhstrh:
Added you to the repo, @Wonseo-C.

org.lflang/src/org/lflang/LinguaFranca.xtext
```diff
@@ -217,7 +217,7 @@ Mutation:
     ('mutation')
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')?
     (sources+=VarRef (',' sources+=VarRef)*)?
-    ('->' effects+=[VarRef] (',' effects+=[VarRef])*)?
+    ('->' effects+=VarRef (',' effects+=VarRef)*)?
```
hokeun, line 220: Just out of curiosity, what is the implication of this change?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -142,6 +143,49 @@ class TSReactionGenerator(
...
+            if (trigger is VarRef) {
+                mutationTriggers.add(trigger.generateVarRef())
+            } else if (trigger is BuiltinTriggerRef) {
+                when (trigger.type) {
+                    BuiltinTrigger.STARTUP  -> mutationTriggers.add("this.startup")
+                    BuiltinTrigger.SHUTDOWN -> mutationTriggers.add("this.shutdown")
+                    else -> {}
+                }
+            }
+        }
```
hokeun, line 165: This part of generateMutationString() seems to be exactly the same as that of generateReactionString(). How about we factor this out as a separate function and have it reused in both generateMutationString() and generateReactionString()?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -443,12 +502,222 @@ class TSReactionGenerator(
         )
     }
 
+    private fun generateSingleMutation(reactor : Reactor, mutation: Mutation): String {
```
hokeun, line 505: I think this is a great opportunity to refactor generateSingleReaction() function. Could you identify the common code lines between generateSingleReaction() and generateSingleMutation() separate the code lines out as functions, and have them called by generateSingleReaction() and generateSingleMutation()?

hokeun (commented):
Awesome progress! I left some suggestions to improve the modularity of the code.

org.lflang/src/org/lflang/LinguaFranca.xtext
Wonseo-C: Marten checked that it was just a mistake.
The VarRef already list variable.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: https://github.com/lf-lang/lingua-franca/pull/1252/commits/8dfcc2a223a5648cbed67ad6b615079036c3d4b3
I combined the generateMutationString() and generateReactionString().
It works successfully, but I'm not sure about my logic. @hokeun

lhstrh:
@Wonseo-C: why was this PR closed? Please always give a reason when closing a PR so that it's clear for others what is going on :-)

hokeun:
@lhstrh This is being tracked in https://github.com/lf-lang/lingua-franca/pull/1318 for now. :)

PR #1257, opened 2022-06-29
properly handle paths with space in lfc launch script

### LINKED ISSUE #1256: Unexpected Behavior on running tests when running on a path with a space.
When running a compiled LFC on a test file whose directory has a space (i.e. ~/Documents/Github/Lingua Franca/lingua-franca/tests/C/arduino/Blink.lf), the LFC cannot find the file and produces an error.

cmnrd:
Closes #1256


lhstrh (approved):
LGMT! I'm assuming the culprit were the back ticks?

cmnrd:
Yes, an in `include.sh` the problem was that the glob operator (`*`) does not work anymore when you simply put quotes around

PR #1265, opened 2022-07-01
Repair ability to execute commands with Bash in case command is not found on path

petervdonovan:
This repairs the "execute" command that for scripts that use Node. I believe that Node is unusual in the way it is added to `$PATH` -- it only is present in the shell environment, which might be why we have to invoke it in this special way.

Issue reported [here](https://github.com/lf-lang/vscode-lingua-franca/issues/64).


lhstrh (approved):
Thanks, @billy-bao and @petervdonovan!

Issue #1213, opened 2022-06-03
Cycle detection reports unrelated elements


a-sr:
The cycle detection reports elements that are not included in the actual cycle.
In this example, a cycle is formed only by sim.theta -> con.theta -> con.control -> sim.control. Yet, all ports and reactions are reported as part of the cycle.

![cycle_unrelated](https://user-images.githubusercontent.com/25612037/171852361-d6e19d80-d54c-4a4c-9458-74b1eaadc5bc.png)

I already tested if the issue is caused by the visualization but `ReactorInstance.getCycles()` reports the following list as causality loop.
```
[
reaction_0 of main.sim, 
reaction_0 of main.con, 
reaction_0 of main.sou, 
reaction_0 of main.out, 
PortInstance main.con.theta, 
PortInstance main.sim.theta, 
PortInstance main.sou.angle, 
PortInstance main.out.theta, 
PortInstance main.sim.control, 
PortInstance main.con.control, 
PortInstance main.out.control, 
PortInstance main.out.led, 
PortInstance main.con.led, 
PortInstance main.out.speaker, 
PortInstance main.sou.sound
]
```
Hence, the problem seems to be in the cycle detection itself.

edwardalee:
This is addressed by PR #1270.

a-sr:
While PR #1270 solves most of the cases, the cycle detection still reports ports as part of a cycle if they are downstream of a reaction on the cycle, even if they do not contribute to the cycle.
![CycleDetectionErrorPort](https://user-images.githubusercontent.com/25612037/184351861-c2bada15-19fb-4a35-9f33-5022b9e65083.png)
In this example the `game_over` port is incorrectly reported.

PR #1270, opened 2022-07-03
Do not include in cycles downstream reactions

edwardalee:
This PR addresses issue #1213. Before this PR, when a causality cycle was detected, reactions that were downstream of the cycle but not part of the cycle were reported to be part of the cycle.  This PR fixes it so that only the reactions in the cycle are highlighted.


lhstrh (approved):
Looks great! Thanks for fixing this!

PR #1271, opened 2022-07-03
Uclid5-based LF Verifier

lsk567:
This is a PR for an LF verifier built on top of [Uclid5](https://github.com/uclid-org/uclid), which analyzes an LF program with the C target and checks whether the program violates a user-specified property. Currently, this feature is enabled when the `@property` annotation is used.

A corresponding benchmark suite can be found here: https://github.com/lf-lang/lf-verifier-benchmarks.

Here is a list of remaining TODOs for this PR:
- [x] Implement an MTL parser that parses user-specified properties. Generate uclid axioms that correspond to the MTL properties.
- [x] Implement a C parser and generate uclid axioms (or smt-lib directly) from analyzable C fragments.
- [x] Automatically calculate the completeness threshold for an MTL property.
- [x] Support timers.
- [x] Speed up SMT file generation process.
- [x] Generate a `.dot` file for the state space diagram.
- [x] Support `lf_schedule_int` call (which takes 3 params).
- [x] [CEX] Parse negative numbers and a let binding from a counterexample.
- [x] A regression test suite and a CI workflow for the LF verifier, so that we know that adding a new feature would not break others (axioms are fairly delicate and might interfere with each other).
- [x] Trim the test cases so that they can be run on a machine with 7G RAM.
- [x] [State space] The event queue needs to sort events by name in addition to time.
- [x] Show a warning for "experimental feature".
- [x] Check for `uclid` and `z3`.

Major feature TODOs:
- Implement an algorithm for abstracting away in the uclid encoding components that cannot causally influence the truth value of a property.
- Support local variables and multiple updates to the same variable in C reaction bodies.
- [Important] Integrate with CBMC. Hopefully, let CBMC output reaction axioms.

Minor features / bugs TODOs:
- [LF] [!] Support physical actions (solver currently blocks).
- [LF] When checking the intervals of a temporal operator, do not take microsteps into account.
- [LF] If a reaction is triggered by startup, it cannot be triggered by another trigger yet.
- [LF] Unconnected input ports currently have nondeterministic behavior. It should not have any signals.
- [LF] Support reactor parameters.
- [LF] Support multiple reactions sensitive to the same trigger.
- [LF] Allow a reaction to be triggered by `startup` and other triggers. For example in `PingPong.lf`.
- [LF] Support reactor parameters.
- [LF] Support modal models.
- [LF & C] [!] Unused triggers (e.g. a logical action declared at the LF layer but not referenced in the reaction body) can block the model. Need to handle this either in the translation or in the encoding. Even if one branch of an if statement is not using it, the entire model is blocked.
```
reaction(startup, serve) -> send {=
    lf_set(send, self->pingsLeft);
    self->pingsLeft -= 1;
=}
```
- [C] Empty reaction bodies are not supported yet.
- [C] [50%] Support adding an additional delay when scheduling an action. The current implementation is almost there except that the state tuple needs to augment with a list of additional delay variables. The reaction axioms should set these variables, and the `Actions` axioms should reference these additional delay variables.
- [C] Support C time value macros (such as `MS(100)`).
- [C] In `TrafficLight.lf`, if an action is visible, not using it blocks the model. So if one branch of an if statement is not using it, the entire model is blocked.
- [State space] Detect Zeno behavior (zero delay cycle) during state space exploration. The current behavior seems to alternate between (0,0) and (0,1) (when running `Election.lf`).
- [State space] One-off timer declarations that look like `timer t(1 nsec)` currently send the explorer into an infinite loop.
- [Diagram] Render CEX diagrams.
- [MTL] In the ADAS model, the property, `ADASModel_l_reaction_0 ==> (F[0, 50 msec]( ADASModel_b_reaction_0 ))`, returns trivially true.
- [MTL] The parsing of `X()` is sometimes problematic.

For logical time-based semantics:
- Generate axioms for simultaneous reaction invocations while preserving reaction priorities.

Notes on maintaining this feature:
1. To develop the Uclid encoding, it is good to first develop a small example manually and then codify the axioms in the Uclid generator.

Debugging tips:
1. To debug the C AST, there is a handy function in `AstUtils.java`, `printStackTraceAndMatchedText()`, that can print the string matched by ANTLR as well as the stack trace of the visitor functions.
2. Unused triggers can block the model (producing trivially true).
3. Insufficient CT can block the model.


org.lflang/src/org/lflang/TimeValue.java
```diff
@@ -25,6 +25,8 @@
 
 package org.lflang;
 
+import java.sql.Time;
```
lsk567, line 28: This seems unnecessary. To remove.

lsk567:
I believe this PR is ready for review now. The Uclid test has passed in the CI, but some other tests have been failing mysteriously...

petervdonovan:
> some other tests have been failing mysteriously...

All of the tests that call `./gradlew build` are failing because some Java files are not formatted.

petervdonovan:
I just pushed a formatting commit, but it is not appearing here. It is visible in the "code" UI in GitHub, though, so I know the commits are in this branch on the remote.

core/src/main/java/org/lflang/analyses/statespace/Event.java
```diff
@@ -0,0 +1,44 @@
...
+  @Override
+  public int compareTo(Event e) {
+    // Compare tags first.
+    int ret = this.tag.compareTo(e.tag);
+    // If tags match, compare trigger names.
+    if (ret == 0) ret = this.trigger.getFullName().compareTo(e.trigger.getFullName());
+    return ret;
+  }
+
+  /** This equals() method does NOT compare tags, only compares triggers. */
```
petervdonovan, line 29: You might consider making this a method of its own that does not override `equals` for this reason. See the [`Comparable` contract](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html). Data structures in the Java standard library (probably including `PriorityQueue`) assume that the ordering induced by `compareTo` is total with respect to the equality defined by `equals`.

core/src/main/java/org/lflang/analyses/statespace/StateInfo.java
```diff
@@ -0,0 +1,24 @@
+/** A class that represents information in a step in a counterexample trace */
```
petervdonovan, line 1: Probably this comment belongs on line 6, above `public class StateInfo`?

> [A doc comment is written in HTML and must precede a class, field, constructor or method declaration](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html)

The same comment applies to `StateSpaceInfo.java`.

core/src/main/java/org/lflang/analyses/statespace/StateSpaceDiagram.java
```diff
@@ -0,0 +1,217 @@
...
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.lflang.TimeValue;
+import org.lflang.generator.CodeBuilder;
+import org.lflang.generator.ReactionInstance;
+import org.lflang.graph.DirectedGraph;
+
+// FIXME: Use a linkedlist instead.
```
petervdonovan, line 12: I think this is fine as-is because the Java standard `LinkedList` is extremely limited in that it only allows linear structures without looping nor branching. But this seems to be encoding structures that can be cyclic.

core/build.gradle
lhstrh, line 1: @cmnrd, could you please review the changes to this file?

core/src/main/java/org/lflang/analyses/statespace/StateSpaceExplorer.java
```diff
@@ -0,0 +1,329 @@
...
+   * Explore the state space and populate the state space diagram until the specified horizon (i.e.
+   * the end tag) is reached OR until the event queue is empty.
+   *
+   * <p>As an optimization, if findLoop is true, the algorithm tries to find a loop in the state
+   * space during exploration. If a loop is found (i.e. a previously encountered state is reached
+   * again) during exploration, the function returns early.
+   *
+   * <p>TODOs: 1. Handle action with 0 min delay. 2. Check if zero-delay connection works.
+   */
+  public void explore(Tag horizon, boolean findLoop) {
```
petervdonovan, line 75: The cyclomatic complexity of this method is probably around 24 based on the number of conditional branches and loops that it has, and 10 is considered a reasonable upper limit. Consider breaking this into smaller methods that each have documentation. Extremely long methods with high cyclomatic complexity and inline comments instead of method-level comments are an antipattern in our code base.

core/src/main/java/org/lflang/analyses/uclid/UclidGenerator.java
```diff
@@ -0,0 +1,1733 @@
...
+  //// Functions from generatorBase
+
+  @Override
+  public Target getTarget() {
+    return Target.C; // Works with a C subset.
+  }
+
+  @Override
+  public TargetTypes getTargetTypes() {
+    throw new UnsupportedOperationException("TODO: auto-generated method stub");
```
petervdonovan, line 1731: It might be wise to provide a short explanation of why it doesn't make sense to implement `getTargetTypes` instead of keeping the text `"TODO: auto-generated method stub"`.

petervdonovan (commented):
This is a large PR that looks like it represents a lot of work.

Today I only had the time to do a superficial pass over it. There were certain comments, such as
- FIXMEs that seem like they would be trivial to address before merging this PR
- chances to use pattern-matching syntax instead of casts
- usages of `Exceptions.sneakyThrow` which we usually do not need
- JavaDoc comments located at the top of files instead of the tops of classes
- occasional methods with cyclomatic complexity exceeding 10 or 15
- public or package-private mutable or non-final fields, protected fields that could be private, or private fields that could be final

where I may not have written the comment in every place where it would be applicable.

lsk567:
@petervdonovan @lhstrh Thanks very much for the detailed reviews! I will work on addressing these comments in the coming days. Since this PR is indeed fairly large, I would like to take the time to make sure that everything conforms to the standards before merging this in. :-)

.github/workflows/uclid-verifier-c-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+          repository: lsk567/lf-verifier-benchmarks
+          ref: main
+          path: lf-verifier-benchmarks
+      - name: Check out Uclid5 repository
+        uses: actions/checkout@v3
+        with:
+          repository: uclid-org/uclid
+          ref: 4fd5e566c5f87b052f92e9b23723a85e1c4d8c1c
+          path: uclid
+      - name: Set up JDK 17
```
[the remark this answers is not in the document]
lhstrh: I vaguely recall we stopped using that action due to flakiness, but that might be averted by the caching we do (except on Windows, where it might rear its head because don't have caching working on Windows.

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
```diff
@@ -79,6 +79,12 @@ public class Lfc extends CliBase {
       description = "Do not invoke target compiler.")
   private boolean noCompile;
 
+  @Option(
+      names = {"--no-verify"},
```
lhstrh, line 83: Should it be the default to run verification? What are the performance impacts? I'm thinking it might be better to have it the other way around...

core/src/main/java/org/lflang/generator/GeneratorBase.java
```diff
@@ -211,10 +192,6 @@ private void createMainInstantiation() {
    */
   public void doGenerate(Resource resource, LFGeneratorContext context) {
 
-    // FIXME: the signature can be reduced to only take context.
-    // The constructor also need not take a file config because this is tied to the context as well.
-    cleanIfNeeded(context);
```
lhstrh, line 216: What happened here?

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
petervdonovan: I think it only runs if there is a `@property` annotation (?)

lhstrh: Yes, that is true. But if you have code with properties, you don't necessarily want to check them every time you compile.

core/build.gradle
cmnrd: I pushed a change that uses the antlr plugin instead of the custom task. The grammar files are now located in the `src/main/antlr` source set. Applying the plugin revealed a couple of issues in our current configuration, which I also fixed.

cmnrd:
For some reason the CI is not triggering, not sure what is going on here.

lhstrh:
> For some reason the CI is not triggering, not sure what is going on here.

The branch is out of date and there are conflicts...

cmnrd:
Before we merge this, we should also consider how this can be integrated with the Maven build for epoch. We will also have to run antlr to generate the java files there. Perhaps @a-sr can comment on this?

core/src/main/java/org/lflang/analyses/uclid/UclidGenerator.java
```diff
@@ -0,0 +1,1733 @@
...
+                  + "("
+                  + "d"
+                  + "("
+                  + "i"
+                  + ")"
+                  + ")"
+                  + " == "
+                  + "false");
+        } else {
+          System.out.println("Unreachable!");
```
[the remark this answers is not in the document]
lsk567: Yes indeed, I will turn all of these println's into exceptions.

core/src/main/java/org/lflang/analyses/uclid/UclidGenerator.java
```diff
@@ -0,0 +1,1733 @@
...
+        li[i] = "true";
+        code.pr("|| rxn(i) == " + "{" + String.join(", ", li) + "}");
+      }
+      code.unindent();
+      code.pr("));");
+    }
+  }
+
+  /** Axioms for the trigger mechanism */
+  protected void generateTriggersAndReactions() {
```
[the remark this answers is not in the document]
lsk567: Great point...

core/build.gradle
lsk567: Thanks for fixing the antlr integration, @cmnrd !

.github/workflows/uclid-verifier-c-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Check out lingua-franca repository
+        uses: actions/checkout@v3
+        with:
+          path: lingua-franca
+          submodules: true
+          fetch-depth: 0
+      - name: Check out lf-verifier-benchmarks repository
+        uses: actions/checkout@v3
+        with:
+          repository: lsk567/lf-verifier-benchmarks
```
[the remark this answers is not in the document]
lsk567: Yes, indeed. The benchmark repo has been transferred to the lf-lang org. Will update this yml file.

core/build.gradle
```diff
@@ -44,6 +49,10 @@ dependencies {
         exclude group: 'org.eclipse.platform', module: 'org.eclipse.swt'
     }
 
+    antlr4 'org.antlr:antlr4:4.7.2'
+    implementation 'org.antlr:antlr4-runtime:4.7.2'
+    implementation 'org.json:json:20200518' // JSON
```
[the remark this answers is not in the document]
lsk567: Yes, you are right. We do not need the ANTLR code. Will change this to `compileOnly`.

core/src/main/java/org/lflang/analyses/cast/IfNormalFormAstVisitor.java
```diff
@@ -0,0 +1,92 @@
...
+ * is an ill-formed program.
+ *
+ * <p>In this program, visit() is the normalise() in the paper.
+ */
+public class IfNormalFormAstVisitor extends CBaseAstVisitor<Void> {
+
+  public CAst.StatementSequenceNode INF = new CAst.StatementSequenceNode();
+
+  @Override
+  public Void visitStatementSequenceNode(
```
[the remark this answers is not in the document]
lsk567: Here, `IfNormalFormAstVisitor` builds a new AST by traversing a C AST. Since this traversal only updates the new AST variable, the visit methods do not need to return anything, so we need to set the type argument of `CBaseAstVisitor` to `void`. However, `void` is a keyword in Java, not a class, and so `Void` needs to be used here instead. `Void` is a class in Java and can be used as a type argument in generics where you don't expect to use the result type. But since it has no usable constructors and its only instance is null, methods that have Void as a return type would have to return null. I admit it looks a bit strange, but I think it is a side effect of using Antlr's visitor pattern.

core/src/main/java/org/lflang/analyses/statespace/StateSpaceExplorer.java
```diff
@@ -0,0 +1,329 @@
...
+      else if (previousTag != null && currentTag.timestamp == previousTag.timestamp) {
+        // System.out.println("DEBUG: Case 3");
+        // Add reactions explored in the current loop iteration
+        // to the existing state space node.
+        currentNode.reactionsInvoked.addAll(reactionsTemp);
+        // Update the eventQ snapshot.
+        currentNode.eventQ = new ArrayList<Event>(eventQ);
+      } else {
+        // Unreachable
+        Exceptions.sneakyThrow(new Exception("Reached an unreachable part."));
```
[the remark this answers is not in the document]
lsk567: Yes, will take care of throwing exceptions properly...

core/src/main/java/org/lflang/analyses/cast/BuildAstParseTreeVisitor.java
```diff
@@ -0,0 +1,691 @@
...
+      CAst.VariableNode varNode = (CAst.VariableNode) primaryExprNode;
+      if (varNode.name.equals("self")) {
+        // return a state variable node.
+        return new CAst.StateVarNode(ctx.Identifier().get(0).getText());
+      } else if (ctx.Identifier().get(0).getText().equals("value")) {
+        // return a trigger present node.
+        return new CAst.TriggerValueNode(varNode.name);
+      } else if (ctx.Identifier().get(0).getText().equals("is_present")) {
+        // return a trigger value node.
+        return new CAst.TriggerIsPresentNode(varNode.name);
```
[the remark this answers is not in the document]
lsk567: Thanks. This is a great point.

core/src/main/java/org/lflang/analyses/cast/BuildAstParseTreeVisitor.java
```diff
@@ -0,0 +1,691 @@
...
+                  "Warning (line " + ctx.getStart().getLine() + "):",
+                  "lf_set must have 2 arguments. Detected " + ctx.argumentExpressionList().size(),
+                  "Marking the function call as opaque."));
+          return new CAst.OpaqueNode();
+        }
+        CAst.SetPortNode node = new CAst.SetPortNode();
+        node.left = visitAssignmentExpression(params.get(0));
+        node.right = visitAssignmentExpression(params.get(1));
+        return node;
+      } else if (varNode.name.equals("lf_schedule")) {
```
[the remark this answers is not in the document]
lsk567: This is a good point. I hope the verifier CI tests can help us detect such inconsistency in the future.

core/src/main/java/org/lflang/analyses/cast/BuildAstParseTreeVisitor.java
```diff
@@ -0,0 +1,691 @@
...
+      // Check if the declaration specifier is a type specifier: e.g. int or long.
+      DeclarationSpecifierContext declSpec = declSpecList.get(0);
+      if (declSpec.typeSpecifier() == null) {
+        System.out.println(
+            String.join(
+                " ",
+                "Warning (line " + ctx.getStart().getLine() + "):",
+                "only type specifiers are supported.",
+                "e.g. \"static const int\" is not analyzable.",
+                "Marking the declaration as opaque."));
```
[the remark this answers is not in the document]
lsk567: Great point. Thanks. :-)

core/src/main/java/org/lflang/generator/GeneratorBase.java
lsk567: The `cleanIfNeeded(context)` method call is relocated to `LFGenerator.java`, so that the `model-gen` folder can be cleaned too before verification is run.
https://github.com/lf-lang/lingua-franca/blob/baa5061c6297b989a6d26beb728cd368b20b36c0/core/src/main/java/org/lflang/generator/LFGenerator.java#L120-L121

core/src/main/java/org/lflang/analyses/uclid/MTLVisitor.java
```diff
@@ -0,0 +1,659 @@
...
+  // A custom dispatch function
+  public String _visitUnaryOp(
+      MTLParser.UnaryOpContext ctx,
+      String prefixIdx,
+      int QFIdx,
+      String prevPrefixIdx,
+      long horizon) {
+
+    // FIXME: Is there a more "antlr" way to do dispatch here?
+    if (ctx instanceof MTLParser.NoUnaryOpContext) {
```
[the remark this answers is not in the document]
lsk567: Oh wow, this is great. Thanks!

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
lhstrh: That makes sense!

core/src/main/java/org/lflang/analyses/cast/CToUclidVisitor.java
```diff
@@ -0,0 +1,432 @@
...
+
+  // The reactor that contains the reaction
+  protected ReactorInstance reactor;
+
+  // A list of all the named instances
+  protected List<NamedInstance> instances = new ArrayList<NamedInstance>();
+
+  // Quantified variable
+  protected String qv = "i";
+  protected String qv2 = "j";
```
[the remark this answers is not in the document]
lsk567: Yup, these should definitely be `final`... Thanks!

core/src/main/java/org/lflang/analyses/cast/CToUclidVisitor.java
```diff
@@ -0,0 +1,432 @@
...
+        if (((PortInstance) i).getDefinition().getName().equals(name)) {
+          return i;
+        }
+      } else if (i instanceof StateVariableInstance) {
+        if (((StateVariableInstance) i).getDefinition().getName().equals(name)) {
+          return i;
+        }
+      }
+    }
+    System.out.println("Named instance" + "not found.");
```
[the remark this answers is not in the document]
lsk567: Yes, this should indeed be an exception. In this context, I think it makes sense to throw an error if an instance is not found by name. We could relax this in the future if need be. Good catch!

core/src/main/java/org/lflang/analyses/cast/CToUclidVisitor.java
```diff
@@ -0,0 +1,432 @@
...
+    return str;
+  }
+
+  @Override
+  public String visitScheduleActionIntNode(ScheduleActionIntNode node) {
+    String name = ((VariableNode) node.children.get(0)).name;
+    NamedInstance instance = getInstanceByName(name);
+    ActionInstance action = (ActionInstance) instance;
+    String additionalDelay = visit(node.children.get(1));
+    String intValue = visit(node.children.get(2));
```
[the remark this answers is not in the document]
lsk567: Yes, I agree with your concern. The solution is to have a new `AstNodeTernary` class with exactly three children. Currently, `ScheduleActionNode` is extending `AstNodeDynamic`...

core/src/main/java/org/lflang/analyses/cast/IfNormalFormAstVisitor.java
```diff
@@ -0,0 +1,92 @@
...
+    this.INF.children.add(generateIfBlock(node, conditions));
+    return null;
+  }
+
+  @Override
+  public Void visitIfBlockNode(CAst.IfBlockNode node, List<CAst.AstNode> conditions) {
+    List<CAst.AstNode> leftConditions = new ArrayList<>(conditions);
+    leftConditions.add(node.left);
+    visit(((CAst.IfBodyNode) node.right).left, leftConditions);
+    if (((CAst.IfBodyNode) node.right).right != null) {
```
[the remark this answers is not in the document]
lsk567: Yes, this is a great point. I will address this in a future PR for readability.

a-sr:
> Before we merge this, we should also consider how this can be integrated with the Maven build for epoch. We will also have to run antlr to generate the java files there. Perhaps @a-sr can comment on this?

After a quick look, I am quite sure the [antlr4 maven plugin](https://www.antlr.org/api/maven-plugin/latest/usage.html) should be able to mimic the gradle build process for generating the parser code.

lsk567:
> > Before we merge this, we should also consider how this can be integrated with the Maven build for epoch. We will also have to run antlr to generate the java files there. Perhaps @a-sr can comment on this?
> 
> After a quick look, I am quite sure the [antlr4 maven plugin](https://www.antlr.org/api/maven-plugin/latest/usage.html) should be able to mimic the gradle build process for generating the parser code.

Thanks @a-sr for the feedback! I will give this plugin a try. :-)

core/src/main/java/org/lflang/analyses/statespace/Event.java
lsk567: Thanks for pointing this out. Yes, I will come up with another way to implement this without overriding `equals()` here because it is inconsistent with `compareTo()` above.

By the way, I recently found that it is a general rule that every time `equals()` is overridden, `hashCode()` also needs to be overridden. I found a few instances in the code base where a new `hashCode()` is not provided.

core/src/main/java/org/lflang/analyses/statespace/StateSpaceExplorer.java
lsk567: Yes, I see your concern @petervdonovan, and I agree this is not the best practice currently. Breaking this method apart could be non-trivial, however. If it is okay with you, can I address this issue in a future PR, maybe in #1830? PR #1830 reuses this infrastructure to do scheduling, and there will be a rather big refactoring to the `statespace` directory based on @oowekyala's comments. Perhaps it's better to do everything in one go, so there won't be too many conflicts.

lsk567:
I believe this PR is ready for another look now. :-)

core/src/main/java/org/lflang/analyses/cast/AbstractAstVisitor.java
```diff
@@ -0,0 +1,17 @@
+package org.lflang.analyses.cast;
+
+import java.util.List;
+
+/** Modeled after AbstractParseTreeVisitor.class */
```
lhstrh, line 5:
```suggestion
/** Modeled after {@link AbstractParseTreeVisitor}. */
```

core/src/main/java/org/lflang/analyses/cast/AbstractAstVisitor.java
lhstrh, line 1: I think the package name `cast` is unfortunate because that is a noun/verb that has nothing to do with the C AST that I assume you are referring to. I think `clang` or just `c` would be much better.

core/src/main/java/org/lflang/analyses/cast/IfNormalFormAstVisitor.java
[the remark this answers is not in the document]
lhstrh: @lsk567, please create an issue for this and then mark this conversation as resolved.

core/src/main/java/org/lflang/analyses/cast/VariablePrecedenceVisitor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.lflang.analyses.cast.CAst.*;
+
+/** This visitor marks certain variable node as "previous." */
+public class VariablePrecedenceVisitor extends CBaseAstVisitor<Void> {
+
+  // This is a temporary solution and cannot handle,
+  // e.g., self->s = (self->s + 1) - (2 * 2).
+  @Override
+  public Void visitAssignmentNode(AssignmentNode node) {
+    // System.out.println("******* In assignment!!!");
```
lhstrh, line 12: Stray debugging code.

core/src/main/java/org/lflang/analyses/cast/CToUclidVisitor.java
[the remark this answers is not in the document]
lhstrh: Once we have an issue, let's resolve this conversation.

core/src/main/java/org/lflang/analyses/statespace/StateSpaceExplorer.java
lhstrh: This part of the PR is most concerning, as it also has a bunch of FIXMEs and debugging code in it. If you plan to address this in #1830 soon, I'm OK with merging this as-is, but it need to be a priority in further development. Perhaps leave a comment in the class description stating that this code is somewhat experimental.

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -1957,7 +1957,10 @@ protected void setUpGeneralParameters() {
         "LOG_LEVEL", String.valueOf(targetConfig.logLevel.ordinal()));
     targetConfig.compileAdditionalSources.addAll(CCoreFilesUtils.getCTargetSrc());
     // Create the main reactor instance if there is a main reactor.
-    createMainReactorInstance();
+    // FIXME: is `hasDeadlines` here always false? That does not look right.
```
lhstrh, line 1960: This `FIXME` should not be left unaddressed. Please investigate.

lhstrh (commented):
This looks like it is very close to ready! I've left some comments.

core/src/main/java/org/lflang/generator/c/CGenerator.java
lsk567: I have determined that `hasDeadlines` is an unused global variable from legacy code. It is now removed.

lsk567:
Thanks for taking a closer look, @lhstrh!

.github/workflows/uclid-verifier-c-tests.yml
lhstrh, line 1: This workflow doesn't seem to execute any tests, only benchmarks, so we should probably rename this workflow. That brings me to another point though, which is that if we treat benchmarks as tests, code coverage remains zero: https://app.codecov.io/gh/lf-lang/lingua-franca/tree/verifier/core%2Fsrc%2Fmain%2Fjava%2Forg%2Flflang%2Fanalyses

lsk567: Just renamed the workflow to `uclid-verifier-c-benchmarks.yml`. These benchmarks currently function as the integration tests for `lfc` in the `test` directory.

Of course, it would be great to have unit tests, especially for the state space diagram generator. How about we add them in #1830?

lhstrh: I'm not quite sure what you mean. Are you referring to the tests [here](https://github.com/lf-lang/lf-verifier-benchmarks/tree/main/tests)?

lhstrh: I do think that it is a problem that this PR brings code coverage down by a whopping 7%:
```
Successful in 1s — 68.45% (-7.01%) compared to 4537cfb 
```
And I worry that it might take a very long time before this gets fixed. What would be against addressing it in this PR? Besides, #1830 is about static scheduling, not formal verification.

lsk567: Okay, I will address the coverage problem in this PR. #1830 also uses the state space diagram infrastructure, and I planned to refactor it significantly in #1830, so I was worrying that all the tests written for the state space diagram will no longer be useful after the refactoring in #1830.

lsk567: > 

I meant to say "these tests serve the same role as the LF integration tests do in the LF repo"

petervdonovan: > But for some reason, the coverage report is not seeing it. 

[Well, this is how you report results to CodeCov.](https://github.com/lf-lang/lingua-franca/blob/045c4294f593c2189ed7c5d92ad8524bcc2d0c16/.github/workflows/c-tests.yml#L67) But I am not sure that adding that will solve the problem. The coverage measuring tool has some integration with JUnit I think which means the tests have to be run using JUnit in order to measure coverage. @lhstrh will know more about this.

.github/workflows/uclid-verifier-c-tests.yml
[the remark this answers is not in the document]
petervdonovan: OK, but if we start seeing flakiness in this step, the action item is to change back to using the method that we currently use in `prepare-build-env` to select the correct Java version.

cmnrd (changes requested):
If I run `lfc --verify` without having the proper dependencies installed I get an NPE:
```
java.lang.NullPointerException: Cannot invoke "org.lflang.util.LFCommand.run()" because "command" is null
        at org.lflang.analyses.uclid.UclidRunner.run(UclidRunner.java:175)
        at org.lflang.generator.LFGenerator.runVerifierIfPropertiesDetected(LFGenerator.java:198)
        at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:124)
        at org.lflang.cli.Lfc.invokeGenerator(Lfc.java:193)
        at org.lflang.cli.Lfc.doRun(Lfc.java:156)
        at org.lflang.cli.CliBase.run(CliBase.java:145)
        at org.lflang.cli.CliBase.doExecute(CliBase.java:117)
        at org.lflang.cli.CliBase.cliMain(CliBase.java:109)
        at org.lflang.cli.Lfc.main(Lfc.java:143)
        at org.lflang.cli.Lfc.main(Lfc.java:133)
```
There should be a proper error message explaining what is missing.

.github/workflows/uclid-verifier-c-tests.yml
cmnrd: I am resolving this conversation. I added a new verifier test category for our integration tests and updated CI accordingly. At the moment, there is an assertion error in the tests that needs to be fixed. Thus code coverage is no complete yet (as the tests abort early).

cmnrd:
The new setup for the verifier test works now. However, the tests revealed a problem in the generator:
```
 java.lang.AssertionError
    	at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:96)
    	at org.lflang.ast.ASTUtils.createMainReactorInstance(ASTUtils.java:604)
    	at org.lflang.analyses.uclid.UclidGenerator.doGenerate(UclidGenerator.java:197)
    	at org.lflang.generator.LFGenerator.runVerifierIfPropertiesDetected(LFGenerator.java:182)
    	at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:124)
```
This wasn't observed before, because assertions are only enabled during testing.

lsk567:
Coverage is back to [75%](https://app.codecov.io/gh/lf-lang/lingua-franca/tree/verifier/)!

lhstrh:
That's great news!

lhstrh:
> > If I run `lfc --verify` without having the proper dependencies installed I get an NPE:
> > ```
> > lfc: fatal error: Error running generator
> > java.lang.NullPointerException: Cannot invoke "org.lflang.util.LFCommand.run()" because "command" is null
> >         at org.lflang.analyses.uclid.UclidRunner.run(UclidRunner.java:175)
> >         at org.lflang.generator.LFGenerator.runVerifierIfPropertiesDetected(LFGenerator.java:198)
> >         at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:124)
> >         at org.eclipse.xtext.generator.GeneratorDelegate.doGenerate(GeneratorDelegate.java:44)
> >         at org.eclipse.xtext.generator.GeneratorDelegate.generate(GeneratorDelegate.java:35)
> >         at org.lflang.cli.Lfc.invokeGenerator(Lfc.java:193)
> >         at org.lflang.cli.Lfc.doRun(Lfc.java:156)
> >         at org.lflang.cli.CliBase.run(CliBase.java:145)
> >         at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
> >         at picocli.CommandLine.access$1500(CommandLine.java:148)
> >         at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
> >         at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
> >         at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
> >         at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
> >         at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
> >         at picocli.CommandLine.execute(CommandLine.java:2170)
> >         at org.lflang.cli.CliBase.doExecute(CliBase.java:117)
> >         at org.lflang.cli.CliBase.cliMain(CliBase.java:109)
> >         at org.lflang.cli.Lfc.main(Lfc.java:143)
> >         at org.lflang.cli.Lfc.main(Lfc.java:133)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There should be a proper error message explaining what is missing.
> 
> Yes, this is a bug. I will fix it now.

Has this been fixed, @lsk567?

core/src/main/java/org/lflang/generator/LFGenerator.java
```diff
@@ -126,4 +140,96 @@ public void doGenerate(Resource resource, IFileSystemAccess2 fsa, IGeneratorCont
   public boolean errorsOccurred() {
     return generatorErrorsOccurred;
   }
+
+  /**
+   * Check if a clean was requested from the standalone compiler and perform the clean step.
+   *
+   * <p>FIXME: the signature can be reduced to only take context.
```
lhstrh, line 147:
```suggestion
```

lhstrh (approved):
Looks good to me! Thanks for addressing the code coverage issue!

lsk567:
> > > If I run `lfc --verify` without having the proper dependencies installed I get an NPE:
> > > ```
> > > lfc: fatal error: Error running generator
> > > java.lang.NullPointerException: Cannot invoke "org.lflang.util.LFCommand.run()" because "command" is null
> > >         at org.lflang.analyses.uclid.UclidRunner.run(UclidRunner.java:175)
> > >         at org.lflang.generator.LFGenerator.runVerifierIfPropertiesDetected(LFGenerator.java:198)
> > >         at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:124)
> > >         at org.eclipse.xtext.generator.GeneratorDelegate.doGenerate(GeneratorDelegate.java:44)
> > >         at org.eclipse.xtext.generator.GeneratorDelegate.generate(GeneratorDelegate.java:35)
> > >         at org.lflang.cli.Lfc.invokeGenerator(Lfc.java:193)
> > >         at org.lflang.cli.Lfc.doRun(Lfc.java:156)
> > >         at org.lflang.cli.CliBase.run(CliBase.java:145)
> > >         at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
> > >         at picocli.CommandLine.access$1500(CommandLine.java:148)
> > >         at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
> > >         at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
> > >         at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
> > >         at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
> > >         at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
> > >         at picocli.CommandLine.execute(CommandLine.java:2170)
> > >         at org.lflang.cli.CliBase.doExecute(CliBase.java:117)
> > >         at org.lflang.cli.CliBase.cliMain(CliBase.java:109)
> > >         at org.lflang.cli.Lfc.main(Lfc.java:143)
> > >         at org.lflang.cli.Lfc.main(Lfc.java:133)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > There should be a proper error message explaining what is missing.
> > 
> > 
> > Yes, this is a bug. I will fix it now.
> 
> Has this been fixed, @lsk567?

Yes, this has been fixed. Now, if any of the dependencies is not found when running `lfc` with the `--verify` flag, an error is printed.
```
lfc: info: Generating code for: file:/Users/shaokai/git/lingua-franca/test/C/src/verifier/ADASModel.lf
lfc: info: Generation mode: STANDALONE
lfc: info: Generating sources into: /Users/shaokai/git/lingua-franca/test/C/src-gen/verifier/ADASModel
lfc: error: Fail to check the generated verification models because Uclid5 or Z3 is not installed.

> Task :cli:lfc:run FAILED

```

lhstrh:
The only thing preventing this from getting merged is https://github.com/lf-lang/epoch/issues/19.

lsk567:
> The only thing preventing this from getting merged is [lf-lang/epoch#19](https://github.com/lf-lang/epoch/issues/19).

Yes, I am [working on it](https://github.com/lf-lang/epoch/tree/verifier). I also consulted @a-sr for help on this. It will probably take some extra time.

lsk567:
https://github.com/lf-lang/epoch/issues/19 is now ready to go. All credits go to @a-sr! I think we should merge this PR (#1271 ) first, since epoch depends on it?

core/src/main/java/org/lflang/generator/LFGenerator.java
```diff
@@ -109,6 +117,12 @@ public void doGenerate(Resource resource, IFileSystemAccess2 fsa, IGeneratorCont
 
     } else {
 
+      // If "-c" or "--clean" is specified, delete any existing generated directories.
+      cleanIfNeeded(lfContext);
+
+      // If @property annotations are used, run the LF verifier.
+      runVerifierIfPropertiesDetected(resource, lfContext);
```
cmnrd, line 124: Is there a reason for calling the verifier here instead of in GeneratorBase?

I spend some time today wondering why the verifier tests pass on my machine, although I don't have uclid installed. Turns out that the errors reported by the verifier are ignored, because the message reporter is reset at the beginning of `doGenerate()` in GeneratorBase. So it seems like the verifier should be called in GeneratorBase, after cleaning and resetting the message reporter.

lsk567: At the time when I implemented this,  it felt a bit more natural to call the verifier outside of GeneratorBase, because it focuses on generating target source code.

But I am certainly not against moving it into `GeneratorBase`. :-)

PR #1275, opened 2022-07-06
Work-in-progress: Handle physical action-triggered outputs in federated execution TS (previously handled by TAN - Time Advance Notice)

hokeun:
Related PR: https://github.com/lf-lang/reactor-ts/pull/106


org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -245,6 +242,48 @@ class TSGenerator(
...
+    private fun createMainReactorInstance() {
+        if (mainDef != null) {
+            if (main == null) {
+                // Recursively build instances. This is done once because
+                // it is the same for all federates.
+                main = ReactorInstance(
+                    ASTUtils.toDefinition(mainDef.reactorClass), errorReporter,
+                    unorderedReactions
+                )
+                val reactionInstanceGraph = main.assignLevels()
```
Soroosh129, line 261: Does this level assignment do anything here for reactor-ts?

hokeun: Not right now. I don't think we really use reactor instance in reactor-ts.

I think I just tried to do the same as the CGenerator.

Would you have any concern or advice?

Soroosh129: The old FIXME still applies here. For the purposes of correct level assignment (or correct topological sorting), we add phantom connections from source federates to destination federates that could cause compile errors or runtime errors if the source federate tries to write on those phantom connections. `removeRemoteFederateConnectionPorts(main)` just removes these phantom connections to avoid errors. This removal could lead to an incorrect topological sort in reactor-ts. We have a better potential solution for this, I think, which is to generate a [causality interface](https://github.com/lf-lang/lingua-franca/pull/1221#issue-1268041718) for the source federate in the destination federate instead of creating phantom connections.

hokeun: > The old FIXME still applies here.

Sorry, I don't think I have any idea about the old FIXME you're referring to...

> removeRemoteFederateConnectionPorts(main) just removes these phantom connections to avoid errors. This removal could lead to an incorrect topological sort in reactor-ts. We have a better potential solution for this, I think, which is to generate a https://github.com/lf-lang/lingua-franca/pull/1221#issue-1268041718 for the source federate in the destination federate instead of creating phantom connections.

I see. Thank you for your suggestion! But I'm not sure if we want to solve this problem here altogether. Would it be ok if I leave a TODO or FIXME here? Do you think getting away with `removeRemoteFederateConnectionPorts(main)` for now is the right approach for the moment?

Soroosh129: > Do you think getting away with removeRemoteFederateConnectionPorts(main) for now is the right approach for the moment?

I think so. The causality interface is set to be part of the FedGenerator effort anyway.

Soroosh129: > Sorry, I don't think I have any idea about the old FIXME you're referring to...

Oh, sorry, I was referring to [this](https://github.com/lf-lang/lingua-franca/blob/2011ccacb472451651d7e0d391127c8060fc9e54/org.lflang/src/org/lflang/generator/ts/TSGenerator.kt#L152).

hokeun: I tried deleting `removeRemoteFederateConnectionPorts(main)`, and I got an error like this.

```
lfc: error: Mixed triggers and effects from different federates. This is not permitted
 --> test/TypeScript/src/federated/DistributedLoopedPhysicalAction.lf:70:1
   |
69 | 
   | >>>>>>>>>>>>>>
70 | federated reactor DistributedLoopedPhysicalAction {
71 |     sender = new Sender();
   | ...
74 |     sender.out -> receiver.inp;
75 | }
   | < Mixed triggers and effects from different federates. This is not permitted
76 | 
```

Could we keep this for now? Would you have any suggestions?

Soroosh129: Oh, sorry, I misquoted your comment earlier. I meant to agree with the following:

> Would it be ok if I leave a TODO or FIXME here? 

I think we should keep `removeRemoteFederateConnectionPorts` for now, but leave a FIXME.

hokeun: Thanks @Soroosh129 ! I left a FIXME.

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -245,6 +242,50 @@ class TSGenerator(
...
+     * If a main or federated reactor has been declared, create a ReactorInstance of it.
+     * This will assign levels to reactions; then, if the program is federated,
+     * an AST transformation is performed to disconnect connections between federates.
+     */
+    private fun createMainReactorInstance() {
+        if (mainDef != null) {
+            if (main == null) {
+                // Recursively build instances. This is done once because
+                // it is the same for all federates.
+                main = ReactorInstance(
```
Soroosh129, line 256: Why is ReactorInstance instantiated here? This could be very expensive.

Soroosh129: I see that it's needed for `removeRemoteFederateConnectionPorts`. Would it work to limit the depth to 1, like it is done [here](https://github.com/lf-lang/lingua-franca/blob/e3c7fb439ec67b721aba6fdd468d36f58035836b/org.lflang/src/org/lflang/generator/GeneratorBase.java#L1121)?

hokeun: This is just following the implementation in the CGenerator.java here: https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/generator/c/CGenerator.java#L2628

Would you have any suggestion on where the ReactorInstance is instantiated instead?

hokeun: The comment above was a reply to the first comment, by the way. 

> I see that it's needed for removeRemoteFederateConnectionPorts. Would it work to limit the depth to 1, like it is done [here](https://github.com/lf-lang/lingua-franca/blob/e3c7fb439ec67b721aba6fdd468d36f58035836b/org.lflang/src/org/lflang/generator/GeneratorBase.java#L1121)?

Sure, I can try that. Do you think we should update the CGenerator's logic as well?

hokeun: I added createMainReactorInstance() to make `findOutputsConnectedToPhysicalActions()` work. So now TypeScript starts using ReactorInstance.

Soroosh129: Specifically, by phantom connections, I'm referring to the bold lines in the example below that is part of #608:

hokeun: > Ah, I see. Okay. That makes sense. I got my answer then. There is no need to modify anything then. Thanks for explaining this!

Oh, I see. Thanks for the clarification and review! I'll leave this as is now then. 👍

Soroosh129 (approved):
Looks great!

hokeun:
Thank you @Soroosh129 !

PR #1277, opened 2022-07-06
Getting the minimum delay

Wonseo-C:
I implemented getting a minimum delay value. By adding "this.LET" member in the ReactionInstance, I can get the delays at Ports and Actions by checking effects. Then, calculate the minimum value of those delays.


org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -189,6 +192,38 @@ public ReactionInstance(
         }
     }
 
+    private static TimeValue LET;
+    private TimeValue getLET() {
```
Soroosh129, line 196: JavaDoc is missing here.

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -189,6 +192,38 @@ public ReactionInstance(
...
+        TimeValue result = TimeValue.MAX_VALUE;
+        TimeValue temp = TimeValue.ZERO;
+        for (VarRef effect : definition.getEffects()) {
+            Variable variable = effect.getVariable();
+            if (variable instanceof Port) {
+                var portInstance = parent.lookupPortInstance(effect);
+                if (portInstance != null) {
+                    this.effects.add(portInstance);
```
Soroosh129, line 204: Why is `portInstance` added to `this.effects` here?

Soroosh129 (approved):
I only had a few minor comments. Other than that, this looks great to me!

Soroosh129:
We should also wait for all the checks to pass before merging in.

lhstrh (changes requested):
Looks good to me! I'm only requesting stylistic changes. See https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

Wonseo-C:
> Looks good to me! I'm only requesting stylistic changes. See https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
I reflected your request in https://github.com/lf-lang/lingua-franca/pull/1277/commits/8361e3a2c29b1f04309a6ec3baf6968681ddc23d!

lhstrh:
> @lhstrh What would be a good place to put a test for the `getLogicalExecutionTime()` method? I don't believe we have ReactionInstance tests.

Let's add a new class called `LetInferenceTests` or something like that in the following package:
https://github.com/lf-lang/lingua-franca/tree/master/org.lflang.tests/src/org/lflang/tests/compiler

@Wonseo-C: could you add some tests?

Wonseo-C:
> > @lhstrh What would be a good place to put a test for the `getLogicalExecutionTime()` method? I don't believe we have ReactionInstance tests.
> 
> Let's add a new class called `LetInferenceTests` or something like that in the following package: https://github.com/lf-lang/lingua-franca/tree/master/org.lflang.tests/src/org/lflang/tests/compiler
> 
> @Wonseo-C: could you add some tests?

OK! I'll try it and then share my work.

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -78,6 +80,7 @@ public ReactionInstance(
         super(definition, parent);
         this.index = index;
         this.isUnordered = isUnordered;
+        this.LET = getLET();
```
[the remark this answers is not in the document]
lhstrh: @Wonseo-C: note that in the PR conversation view, you can hit a button saying `Commit suggestion` to automatically commit the suggestion :-)

lhstrh (changes requested):
This looks good, but we still need a test (or two).

lhstrh:
Closing this PR since as it has been subsumed by #1292, which now also contains tests.

PR #1279, opened 2022-07-06
Fix banks of modal reactors

### LINKED ISSUE #1264: Banks of modal rectors do not receive inputs
In the following program, the Receiver is a bank of (trivial) modal reactors with just one mode.  Although all bank members receive an input (and print it), only the last bank member receives the input **inside** its one and only mode.
```
target C {timeout: 1 sec}

reactor Receiver(bank_index:int(0)){
    input msg:int;
    reaction(msg) {=
        lf_print("[At %lld] Receiver %d received message %d outside mode", 
            lf_time_logical_elapsed(), self->bank_index, msg->value
        );
    =}
    initial mode A {
        reaction(msg) {=
            lf_print("[At %lld] Receiver %d received message %d inside mode", 
                lf_time_logical_elapsed(), self->bank_index, msg->value
            );
        =}
    }
}

reactor Sender(bank_index:int(0), arrival_interval:time(5000 msec), charge_time:time(5000 msec)) {
    output msg:int;
    timer t(0, 1 sec);
    reaction(t) -> msg {=
        lf_set(msg, self->bank_index);
        lf_print("Sender %d sending its ID", self->bank_index);
    =}
}

main reactor{
    vc = new[3] Sender();
    cs = new[3] Receiver();
    vc.msg -> cs.msg;
}
```
image

The output is:
```
Sender 2 sending its ID
Sender 0 sending its ID
Sender 1 sending its ID
[At 0] Receiver 1 received message 1 outside mode
[At 0] Receiver 2 received message 2 outside mode
[At 0] Receiver 0 received message 0 outside mode
[At 0] Receiver 2 received message 2 inside mode
Sender 1 sending its ID
Sender 2 sending its ID
Sender 0 sending its ID
[At 1000000000] Receiver 2 received message 2 outside mode
[At 1000000000] Receiver 0 received message 0 outside mode
[At 1000000000] Receiver 1 received message 1 outside mode
[At 1000000000] Receiver 2 received message 2 inside mode
```
The output should be:
```
Sender 2 sending its ID
Sender 0 sending its ID
Sender 1 sending its ID
[At 0] Receiver 1 received message 1 outside mode
[At 0] Receiver 1 received message 1 inside mode
[At 0] Receiver 2 received message 2 outside mode
[At 0] Receiver 0 received message 0 outside mode
[At 0] Receiver 0 received message 0 inside mode
[At 0] Receiver 2 received message 2 inside mode
...
```

a-sr:
Fixes code generation for modal reactors to correctly account for banks.

Closes #1264


org.lflang/src/org/lflang/generator/python/PythonPortGenerator.java
```diff
@@ -195,7 +195,7 @@ public static String generatePythonListForContainedBank(String reactorName, Port
             "for (int i = 0; i < "+generateWidthVariable(reactorName)+"; i++) {",
             "    if (PyList_SetItem("+reactorName+"_py_list,",
             "            i,",
-            "            "+generateConvertCPortToPy("self->_lf_"+reactorName+"[i]."+port.getName(), widthSpec),
+            "            "+generateConvertCPortToPy((port.getWidthSpec() == null ? "&" : "") + "self->_lf_"+reactorName+"[i]."+port.getName(), widthSpec),
```
a-sr, line 198: @Soroosh129 I had to make this change in order to get my test models to compile. Otherwise, I got the following error from gcc:
```
error: incompatible type for argument 1 of ‘convert_C_port_to_py’
expected ‘void *’ but argument is of type ‘countercycle_next_t’ {aka ‘struct <anonymous>’}
  418 | PyObject* convert_C_port_to_py(void* port, int width) {
```

This occurred for every reaction that had an effect on a bank of reactors but only if the port was not a multiport.

Can you confirm that this fix is correct or is the actual problems somewhere else?

Soroosh129: I think a simpler potential solution would be to use `reactorName + "[i]." + port.getName()` instead of `"self->_lf_"+reactorName+"[i]."+port.getName()`.

Soroosh129, line 198:
```suggestion
            "            "+generateConvertCPortToPy(reactorName + "[i]." + port.getName(), widthSpec),
```

a-sr: That looks like a better solution. Thanks!

org.lflang/src/org/lflang/generator/c/CModesGenerator.java
```diff
@@ -2,8 +2,11 @@
 
 import java.util.List;
 
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
```
lhstrh, line 5: We've been avoiding unnecessary dependencies on Xtext/Xbase in our code generators. Could you using something from the standard Java library instead?

lhstrh (changes requested):
Looks good to me, modulo that Xbase import.

org.lflang/src/org/lflang/generator/c/CModesGenerator.java
a-sr: Sure. I removed the usage of that extension.

lhstrh (approved):
:rocket: :rocket: :rocket:

PR #1280, opened 2022-07-06
Silence spurious error messages from Pylint.

### LINKED ISSUE #1163: Python target does not work with Epoch
To reproduce:
```
cd lingua-franca/test/Python
rm -rf src-gen bin
```
Then compile any LF program, e.g., `src/ActionDelay.lf`.  Try to run it:
```
python3 src-gen/ActionDelay/ActionDelay.py
  File "/Users/eal/lingua-franca-master/git/lingua-franca/test/Python/src-gen/ActionDelay/ActionDelay.py", line 6, in <module>
    from LinguaFrancaActionDelay import (  # pylint: disable=no-name-in-module, import-error
ModuleNotFoundError: No module named 'LinguaFrancaActionDelay'
```

`lfc` works, however.

petervdonovan:
Fixes  #1163.


org.lflang/src/org/lflang/generator/python/PythonPortGenerator.java
```diff
@@ -228,23 +227,23 @@ private static String generateConvertCPortToPy(String port, String widthSpec) {
...
      */
     public static String generatePythonPortVariableInReaction(VarRef port) {
         String containerName = port.getContainer().getName();
         String variableName = port.getVariable().getName();
+        String tryStatement = "try: "+containerName+"  # pylint: disable=used-before-assignment";
         if (port.getContainer().getWidthSpec() != null) {
             // It's a bank
-            return String.join("\n",
-                "try: "+containerName,
+            return String.join(System.lineSeparator(),
```
lhstrh, line 237: I think this would cause a compilation performed on Windows to yield generated code with `\r\n` (CRLF) line endings, which I'm guessing we don't actually want?

petervdonovan: Why would we not want that? Windows users use those silly line endings all the time, and their programs may be goofy, but they do compile. I think our goal should be to generate code that compiles and that is consistent with the standard followed by the user's OS. [Until recently](https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/), Notepad did not recognize `\n` as an EOL sequence, so if a Windows user opens generated Python code with `\n` characters in Notepad, they might be confronted with a wall of jumbled text.

People using older Macs may have a similar problem.

If we don't care about those users, then that's OK with me -- we can just make it a standard in our code base to always just use `\n`.

(That said, I vaguely remember that there was some place where we actually wanted to use the `\n` character unconditionally, but I don't remember where.)

lhstrh: Notepad! :rofl: I was assuming that anyone who codes would use "normal" line endings, but I'm probably wrong about that. If we want to be platform specific, then are are a ton more of such instances throughout the code base that we'd have to touch to be consistent. I'm fine with making a choice to accommodate Notepad coders, but then we'd have to adjust _all_ `String.join` statements, not just these ones.

Soroosh129: I think if `docker: true` is set in the target properties, then the generated Python code will end up running in a Linux environment, even if lfc is used on Windows. I would vote to use `\n` to avoid a mishap with our Docker support.

lhstrh: Good point, @Soroosh129, I'm with you on this one.

petervdonovan: Okay, I'll revert this change. I searched around a little bit, and I guess that most modern editors should handle non-standard line feeds acceptably.

The formatter also has many uses of `System.lineSeparator()`. Should I change all of those to `\n` too? If we are using it in federated code generation where the generated code might be copied to other machines, then maybe we would want that.

petervdonovan, line 237:
```suggestion
            return String.join("\n",
```

org.lflang/src/org/lflang/generator/python/PythonPortGenerator.java
```diff
@@ -228,23 +227,23 @@ private static String generateConvertCPortToPy(String port, String widthSpec) {
...
+                tryStatement,
                 "except NameError: "+containerName+" = [None] * len("+containerName+"_"+variableName+")",
                 "for i in range(len("+containerName+"_"+variableName+")):",
                 "    if "+containerName+"[i] is None: "+containerName+"[i] = Make()",
                 "    "+containerName+"[i]."+variableName+" = "+containerName+"_"+variableName+"[i]"
             );
         } else {
-            return String.join("\n",
-                "try: "+containerName,
+            return String.join(System.lineSeparator(),
```
petervdonovan, line 245:
```suggestion
            return String.join("\n",
```

lhstrh (approved):

org.lflang/src/org/lflang/generator/python/PythonPortGenerator.java
lhstrh: > The formatter also has many uses of `System.lineSeparator()`. Should I change all of those to `\n` too? If we are using it in federated code generation where the generated code might be copied to other machines, then maybe we would want that.

I think that makes sense, yes.

PR #1282, opened 2022-07-07
Improve mode diagram layout

a-sr:
* Added correct spacing configuration for modes (they now follow the same base configuration as reactors)
* Added synthesis option to manually adjust global spacing (helpful to reduce diagram size for papers)
* Fixed problems in visualization of mode-crossing connections


edwardalee (approved):
LGTM!

PR #1283, opened 2022-07-07
C++: check if subdirectory was generated from LF-Cpp before including it in cmake

### LINKED ISSUE #1171: CMake fails with mixed projects
If I create a project that has a C target program (I tried this with this [federated program](https://github.com/lf-lang/examples-lingua-franca/blob/main/C/src/DistributedHelloWorld/HelloWorld.lf)) and then create a Cpp program in the same src directory, then if I compile first the C target program and then the Cpp target program, CMake fails with the following error:

```
Generating code for: platform:/resource/CMake%20Bug/src/RoomTemp.lf
******** mode: EPOCH
******** generated sources: /Users/eal/Tmp/Bug/src-gen/RoomTemp
Path: /Users/eal/Tmp/Bug/src-gen/RoomTemp /Users/eal/Tmp/Bug/src-gen/RoomTemp
--- Current working directory: /Users/eal/Tmp/Bug/build
--- Executing command: cmake --version
cmake version 3.22.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
--- Current working directory: /Users/eal/Tmp/Bug/build
--- Executing command: cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/Users/eal/Tmp/Bug -DCMAKE_INSTALL_BINDIR=bin -DREACTOR_CPP_VALIDATE=ON -DREACTOR_CPP_TRACE=OFF -DREACTOR_CPP_LOG_LEVEL=3 -DLF_SRC_PKG_PATH=/Users/eal/Tmp/Bug /Users/eal/Tmp/Bug/src-gen
-- The CXX compiler identification is AppleClang 13.1.6.13160021
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
See also "/Users/eal/Tmp/Bug/build/CMakeFiles/CMakeOutput.log".
See also "/Users/eal/Tmp/Bug/build/CMakeFiles/CMakeError.log".
CMake Error at CMakeLists.txt:44 (add_subdirectory):
  The source directory

    /Users/eal/Tmp/Bug/src-gen/FederatedHelloWorld

  does not contain a CMakeLists.txt file.


ERROR: cmake failed with error code 1
```

Notice that it seems to be demanding a CMakeLists.txt file in a directory unrelated to the program I am trying to compile. This seems to happen with any combination of C and Cpp programs in the same directory.

cmnrd:
This places a file called `.lf-cpp-marker` in each generated subdirectory. Cmake then uses this to determine whether it should include a found subdirectory or not.

Closes #1171


petervdonovan:
I think I once had build issues with the C++ target because of directories that were generated by an older version LFC (but I don't remember the details). Should this marker be more specific in some way? For example, should it have some kind of hash that lets us detect that the content of the directories is no longer valid?

cmnrd:
The C++ build system got a lot more robust, so the problem that you faced in the past should not occur anymore. The point that you raise is nonetheless a good one, I would treat it as a separate issue/feature. This will become more important once we are concerned with compatibility and maybe also have binary packages.

edwardalee (approved):
LGTM, although I'm not generally a fan of creating hidden files. Is there any reason to not just name the file "lf-cpp-marker" rather than ".lf-cpp-marker"?

cmnrd:
No, there is no reason other than personal preference. I prefer the hidden file, but I am also happy to change it. Maybe someone else has an opinion?

edwardalee:
OK, I'm convinced. Let's go with the hidden file.

PR #1285, opened 2022-07-07
Do not minimize the shadow jar

### LINKED ISSUE #1284: Gradle does not include Kotlin classes when compiling again
Compiling with gradle works fine when compiling for the first time. However, if I make changes to a kotlin file and invoke `.\gradlew build` again, the resulting jar does not include the Kotlin classes anymore. I have to run `.\gradlew clean` first to make it work again.

Any ideas what is going wrong here? I think we had a similar issue in the past, but I was not able to find something in our old issues.

cmnrd:
Issue #1284 was caused by the mimization step of the shadowJar gradle plugin. The main problem is, that we don't import kotlin classes directly, but load them manually in order to work around the build problems in Eclipse... Hence, the minimizer thinks it can exclude the Kotlin classes. While in principle it is possible to add particular excludes, this does not work properly when we compile a second time. The reason for this is unclear, but we might as well disable the minimization. This slightly increases the jar size from 23 MB to 28 MB.

Closes #1284

### DIFF
--- a/org.lflang.lfc/build.gradle
+++ b/org.lflang.lfc/build.gradle
@@ -35,11 +35,6 @@ task buildLfc() {
 
     shadowJar {
         exclude 'test/*'
-        minimize() {
-            exclude(dependency('log4j:log4j:.*'))
-            exclude(dependency('com.google.inject:guice:.*'))
-            exclude(dependency('org.lflang:org.lflang:.*'))
-        }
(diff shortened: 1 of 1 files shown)


lhstrh (changes requested):
I think this is a fine solution, but we should leave a note in the Gradle config explaining why we aren't doing minimization. We could just refer to this PR.

PR #1286, opened 2022-07-08
Cpp: fix triggering of reactions in multiple nested reactors

cmnrd:
The reactor-cpp runtime did not correctly handle multiple reactions in different nested reactors that are triggered by the same upstream port. This change pulls in a fix in reactor-cpp and adds an LF test.

Related reactor-cpp PR: https://github.com/lf-lang/reactor-cpp/pull/16


cmnrd:
@edwardalee, @Soroosh129 Do you know if we have a similar test for C?

edwardalee:
> @edwardalee, @Soroosh129 Do you know if we have a similar test for C?

I'm not sure, but I just added a C version of this test to the branch.

Soroosh129:
I couldn’t find a test that exactly matches what’s here (where you have two reactions at two different levels of the hierarchy be triggered by the same upstream port).

I did, however, find an apparent mistake in [Hierarchy.lf](https://github.com/lf-lang/lingua-franca/blob/c29a30ef750aee43c5ca3a62fac8d2063c96dd95/test/C/src/Hierarchy.lf#L44), where in the main reactor we have:
```
container.out -> print.in;
container.out -> print2.in;
```
I think this should be:
```
container.out -> print.in;
container.out2 -> print2.in;
```

edwardalee:
I just pushed a fix to this mistake to this branch.

> I did, however, find an apparent mistake in [Hierarchy.lf](https://github.com/lf-lang/lingua-franca/blob/c29a30ef750aee43c5ca3a62fac8d2063c96dd95/test/C/src/Hierarchy.lf#L44), where in the main reactor we have:
> 
> ```
> container.out -> print.in;
> container.out -> print2.in;
> ```
> 
> I think this should be:
> 
> ```
> container.out -> print.in;
> container.out2 -> print2.in;
> ```

cmnrd:
Thanks for providing the C test! I will merge it then.

PR #1287, opened 2022-07-08
Fixed issue with ASTUtils width() returnning -1 even when the width can be inferred from connections

hokeun:
This is to fix an error found while working on https://github.com/lf-lang/lingua-franca/pull/1275/.

Weirdly, adding `createMainReactorInstance()` to TypeScript code generator breaks following existing tests.

- [test/TypeScriptsrc/multiport/BankToBankMultiportAfter.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/BankToBankMultiportAfter.lf)
- test/TypeScriptsrc/multiport/BroadcastAfter.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/BroadcastAfter.lf)
- test/TypeScriptsrc/multiport/BroadcastMultipleAfter.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/BroadcastMultipleAfter.lf)
- test/TypeScriptsrc/multiport/MultiportFromBankHierarchyAfter.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/MultiportFromBankHierarchyAfter.lf)
- test/TypeScriptsrc/multiport/MultiportToBankAfter.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/MultiportToBankAfter.lf)
- test/TypeScriptsrc/multiport/MultiportToMultiport2After.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/MultiportToMultiport2After.lf)
- test/TypeScriptsrc/multiport/PipelineAfter.lf(https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/multiport/PipelineAfter.lf)

Below is an example raw error message from the broken tests' runLfc:

```
java.lang.NullPointerException: Cannot read field "_interleaved" because "src" is null
	at org.lflang.generator.ReactorInstance.connectPortInstances(ReactorInstance.java:863)
	at org.lflang.generator.ReactorInstance.establishPortConnections(ReactorInstance.java:925)
	at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:824)
	at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:112)
	at org.lflang.generator.ts.TSGenerator.createMainReactorInstance(TSGenerator.kt:256)
	at org.lflang.generator.ts.TSGenerator.doGenerate(TSGenerator.kt:154)
	at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:164)
	at org.lflang.lfc.Main.runGenerator(Main.java:293)
	at org.lflang.lfc.Main.main(Main.java:222)
```
 
Specifically, when the `width` becomes 0, `RuntimeRange<PortInstance> src` becomes null.


edwardalee (commented):
I'm not sure about this change. The failure in Cpp is worrisome, but I can't read the CI output so I don't know the failure actually is. The Cpp tests log output has an absurd number of printf outputs... Are these all needed?  Anyway, I search the logs for "failed" and "error" and found nothing, but the CI reports that the test failed.

I think that width == 0 in delays means that the width need to be inferred from the context. So I'm not sure that simply ignoring it is the right thing to do.  But again, I can't find anywhere that this field maxWidth is used (Eclipse tells me there are no references to it), so don't know what the implications are.

hokeun:
Thank you for your comments, @edwardalee!

> I'm not sure about this change. The failure in Cpp is worrisome, but I can't read the CI output so I don't know the failure actually is. The Cpp tests log output has an absurd number of printf outputs... Are these all needed? Anyway, I search the logs for "failed" and "error" and found nothing, but the CI reports that the test failed.

I think you meant this error in cpp-benchmark-tests/run:
Screen Shot 2022-07-09 at 3 31 14 PM

Here is the error message I tracked down:

```
Error executing job with overrides: ['test_mode=True', 'iterations=1', 'benchmark=savina_micro_threadring', 'target=lf-cpp', 'iterations=1', 'target.params.extra_args=[--external-runtime-path,/home/runner/work/lingua-franca/lingua-franca/reactor-cpp/install]']
An error occurred during Hydra's exception formatting:
TypeError("print_exception() got an unexpected keyword argument 'etype'")
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/utils.py", line 283, in run_and_report
    print_exception(etype=None, value=ex, tb=final_tb)  # type: ignore
TypeError: print_exception() got an unexpected keyword argument 'etype'
During handling of the above exception, another exception occurred:
  File "/home/runner/work/lingua-franca/lingua-franca/runner/run_benchmark.py", line 194, in <module>
    main()
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/main.py", line 48, in decorated_main
    _run_hydra(
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/utils.py", line 385, in _run_hydra
    run_and_report(
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/utils.py", line 294, in run_and_report
    raise ex
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/utils.py", line 211, in run_and_report
    return func()
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/utils.py", line 386, in <lambda>
    lambda: hydra.multirun(
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/hydra.py", line 140, in multirun
    ret = sweeper.sweep(arguments=task_overrides)
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/_internal/core_plugins/basic_sweeper.py", line 161, in sweep
    _ = r.return_value
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/core/utils.py", line 233, in return_value
    raise self._return_value
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/hydra/core/utils.py", line 160, in run_job
    ret.return_value = task_function(task_cfg)
  File "/home/runner/work/lingua-franca/lingua-franca/runner/run_benchmark.py", line 75, in main
    check_return_code(code, continue_on_error)
  File "/home/runner/work/lingua-franca/lingua-franca/runner/run_benchmark.py", line 107, in check_return_code
    raise RuntimeError(
RuntimeError: Command returned with non-zero exit code (1)
Error: Process completed with exit code 1.
```

To me, it looks like the root cause of the error above might not be related to this change, but I'm not 100% sure yet. Do you think you could recommend a point of contact to reach out to regarding this error? Maybe @cmnrd?

> I think that width == 0 in delays means that the width need to be inferred from the context. So I'm not sure that simply ignoring it is the right thing to do.

I'm not sure either... But without this change, we get an error in existing tests when `createMainReactorInstance()` is added. I updated the description of this PR to explain the error I'm trying to fix.

> But again, I can't find anywhere that this field maxWidth is used (Eclipse tells me there are no references to it), so don't know what the implications are.

You're right. `maxWidth` is only used within in the constructor of `RuntimeRange` to determine `width`. And it becomes problematic when `maxWidth`

hokeun:
@edwardalee, I fixed this issue in a different way, by inferring the width from connections when it's possible. I believe this fixes the fundamental issue. Do you think you could take another look when you get a chance? Thanks!

edwardalee (approved):
Ah, OK, this looks like a better way to solve the problem, and the output @cmnrd found indicates that indeed the test failure was related to the change. The relevant line was this one:
```
2022-07-11T21:20:54.1375165Z [2022-07-11 21:20:54,136][/home/runner/work/lingua-franca/lingua-franca/lf/bin/lfc][INFO] - org.lflang.generator.InvalidSourceException: Multiple ports with variable width on a connection.
```
I need someone to show me how to read the "Checks" tab for a PR. I was unable to find any meaningful output at all in that tab. Clearly I am looking in the wrong place.

cmnrd:
In the meantime I pushed a fix to the benchmarks repo which should reduce the noise in the benchmark logs (in particular the C++ logs). (See https://github.com/lf-lang/benchmarks-lingua-franca/pull/27).

In this particular case, the log for the c++ benchmarks [here](https://github.com/lf-lang/lingua-franca/runs/7290894303?check_suite_focus=true) did not contain the problem. So probably this is why you did not find it. Note the last line stating: 
```
This step has been truncated due to its large size. Download the full logs from the menu once the workflow run has completed.
```

I clicked in the gear in the top-right and then on "View raw logs" to see the full log. Then I searched the raw log. With the fix in https://github.com/lf-lang/benchmarks-lingua-franca/pull/27, the log should get much smaller and you should be able to search directly in GitHub's log view. Note that searching only works properly with the built-in search tool and not with you browser's search.

PR #1288, opened 2022-07-11
[cpp] CLI parser errors are caught and a print help message is printed

### LINKED ISSUE #1255: C++ prorgams should print help when using wrong CLI arguments
Currently, C++ binaries exit with a cryptic exception when an invalid CLI argument is used. Instead, we should print a nicely formatted error and print the help text.

cmnrd:
Instead of just failing with an exception on CLI parse errors, the LF binary will print a more informative error message and the help text. Also this PR improves the formatting of parse exceptions thrown by our custom parser for time values.

Closes #1255


lhstrh (approved):
Cool!

Issue #1278, opened 2022-07-06
Reactions do not consistently trigger banks


a-sr:
If two reactor banks are triggered by the same reaction, the bank that is listed as first effect, will only get a single bank member triggered.

This is an example:
```
target C;

reactor Hello {
    input I:int;
    
    reaction(I) {=
        printf("Hello\n");
    =}
}
reactor World {
    input I:int;
    
    reaction(I) {=
        printf("World\n");
    =}
}

main reactor {
    hellos = new[3] Hello()
    worlds = new[3] World()

    reaction(startup) -> hellos.I, worlds.I {=
        for(int i = 0; i < hellos_width; i++) {
            lf_set(hellos[i].I, true);
        }
        for(int i = 0; i < worlds_width; i++) {
            lf_set(worlds[i].I, true);
        }
    =}
}
```
The output is:
```
World
World
World
Hello
```

Swapping the order of the effects will output 3 x `Hello`, 1 x `World`.
Splitting the reaction into two, one for each bank, yields the correct behavior.

edwardalee:
PR #1289 fixes this.  The problem was a simple indexing issue that failed to take into account the width of the first bank.

edwardalee:
Unfortunately, my fix does not work. Reopening this issue and converting the PR #1289 to a draft.

PR #1289, opened 2022-07-11
Dual bank bugfix

edwardalee:
This fixes Reactions do not consistently trigger banks #1278.


edwardalee:
Right. I'm on it. Testing a better fix now.

cmnrd:
> Need to convert this PR to a draft, but there appears to be no way to do that.

The "convert to draft" button is hidden in the top sidebar, right under the list of reviewers and above "Assignees".

PR #1292, opened 2022-07-12
Implementation of LET inference in ReactorInstance

lhstrh:
This is a WIP with @CloverCho and @Wonseo-C to infer logical execution times in `ReactorInstance`.

FIXMEs:
 - [x] calculate LET for reaction (not reactor)
 - [x] clean up and comments
 - [x] write tests


org.lflang.tests/src/org/lflang/tests/compiler/LetInferenceTests.java
```diff
@@ -0,0 +1,137 @@
+package org.lflang.tests.compiler;/* Parsing unit tests. */
+
+/*************
+ Copyright (c) 2019, The University of California at Berkeley.
```
Soroosh129, line 4: The date might need to be update. Also, I think you might want to add Hanyang University as well.

org.lflang.tests/src/org/lflang/tests/compiler/LetInferenceTests.java
```diff
@@ -0,0 +1,137 @@
...
+import org.lflang.tests.LFInjectorProvider;
+import static org.lflang.ASTUtils.*;
+
+@ExtendWith(InjectionExtension.class)
+@InjectWith(LFInjectorProvider.class)
+
+/**
+ * Test for getting minimum delay in reactions.
+ * Checking the actions and port's delay,then get the minimum reaction delay.
+ */
```
Soroosh129, line 61: Author information is missing.

org.lflang/src/org/lflang/generator/ReactorInstance.java
```diff
@@ -1114,4 +1114,53 @@ private void setInitialWidth() {
      * Cached reaction graph containing reactions that form a causality loop.
      */
     private ReactionInstanceGraph cachedReactionLoopGraph = null;
+
+    private TimeValue let = null;
+    
+    public TimeValue getLogicalExecutionTime() {
```
Soroosh129, line 1120: This is confusing to me. Why do you calculate LET values for reactors?

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -448,6 +448,57 @@ public String toString() {
...
+        }
+
+        TimeValue let = null;
+
+        // Iterate over effect and find minimum delay.
+        for (TriggerInstance<? extends Variable> effect : effects) {
+            if (effect instanceof PortInstance) {
+                var afters = this.parent.getParent().children.stream().filter(c -> {
+                    if (c.isGeneratedDelay()) {
+                        return c.inputs.iterator().next().getDependsOnPorts().iterator().next().instance
```
Soroosh129, line 472: This line is quite hard to understand.

Why are you making calls to `iterator().next()` and not use something like `forEach`?

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -448,6 +448,57 @@ public String toString() {
...
+                                       .equals((PortInstance) effect);
+                    }
+                    return false;
+                }).map(c -> c.actions.iterator().next().getMinDelay())
```
Soroosh129, line 476: I might just be unfamiliar with the `.iterator().next()` pattern. Could you clarify what it does?

Soroosh129 (changes requested):
I think there is a residual piece of code in ReactorInstance that might need to be deleted.

I also had some comments about the usage of `.iterator().next()`, which were unclear for me.

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -448,6 +448,57 @@ public String toString() {
...
+     * Determine logical execution time for each reaction during compile
+     * time based on immediate downstream logical delays (after delays and actions)
+     * and label each reaction with the minimum of all such delays.
+     */
+    public TimeValue getLogicalExecutionTime() {
+        if (this.let != null) {
+            return this.let;
+        }
+
+        if (this.parent.isGeneratedDelay()) {
```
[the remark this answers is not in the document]
Wonseo-C: Because when we get the action delay by `after` lf syntax, the delay exists in the 'Delay' reactor, which is the additional reactor for only delay. So, if the parent is in the generated delay condition, we thought it should return the `TimeValue.ZERO`.

org.lflang/src/org/lflang/generator/ReactionInstance.java
Wonseo-C: In the 'Delay' reactor, there is only one delay reaction, so we simply used the first iteration.
Is it the right explanation? @CloverCho

org.lflang.tests/src/org/lflang/tests/compiler/LetInferenceTests.java
Wonseo-C: I reflected on your comment in https://github.com/lf-lang/lingua-franca/pull/1292/commits/ed0e00e44a54d108d45a0b269986c36333481ba2.

org.lflang.tests/src/org/lflang/tests/compiler/LetInferenceTests.java
Wonseo-C: I reflected on your comment in https://github.com/lf-lang/lingua-franca/pull/1292/commits/ed0e00e44a54d108d45a0b269986c36333481ba2

org.lflang/src/org/lflang/generator/ReactionInstance.java
Soroosh129: Ah, okay. I think `.get(0)` is what we've used most often.

org.lflang/src/org/lflang/generator/ReactorInstance.java
Wonseo-C: Sorry, this is my mistake forgot to delete this part.
I deleted it in https://github.com/lf-lang/lingua-franca/pull/1292/commits/ed0e00e44a54d108d45a0b269986c36333481ba2.

org.lflang/src/org/lflang/generator/ReactionInstance.java
Soroosh129: I think I understand this better now based on your previous comment. I would suggest using `.get(0)` if possible to make this clearer.

org.lflang/src/org/lflang/generator/ReactionInstance.java
Wonseo-C: Thank you for your nice programming advice!
I modified it in https://github.com/lf-lang/lingua-franca/pull/1292/commits/ff0782633922481312131e2321730ef447577194.

org.lflang/src/org/lflang/generator/ReactionInstance.java
Wonseo-C: I modified it in https://github.com/lf-lang/lingua-franca/commit/ff0782633922481312131e2321730ef447577194

Soroosh129 (approved):
LGTM!

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1622,6 +1622,12 @@ private void recordBuiltinTriggers(ReactorInstance instance) {
         }
     }
 
+    private void assignLetToReaction(ReactorInstance instance) {
+        for (ReactionInstance reaction : instance.reactions) {
```
hokeun, line 1626: Where does the return value of `getLogicalExecutionTime ` go? If we are not going to use the return value, shouldn't the return type void?

hokeun (approved):
Overall, this looks good!

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -448,6 +448,57 @@ public String toString() {
         return getName() + " of " + parent.getFullName();
     }
 
+    /**
+     * Determine logical execution time for each reaction during compile
+     * time based on immediate downstream logical delays (after delays and actions)
+     * and label each reaction with the minimum of all such delays.
+     */
+    public TimeValue getLogicalExecutionTime() {
```
hokeun, line 456: I think this should be renamed to `assignLogicalExecutionTime()` as it assigns LET rather than getting it.

hokeun (approved):

org.lflang/src/org/lflang/generator/c/CGenerator.java
Wonseo-C: Yes, this function was written for testing. So, I think it should be deleted.

org.lflang/src/org/lflang/generator/ReactionInstance.java
Wonseo-C: OK. I renamed the function name to `assignLogicalExecutionTime()`. https://github.com/lf-lang/lingua-franca/pull/1292/commits/3ca6084fa42d61554184d13bf47e9d754a153eb7

Wonseo-C:
@lhstrh I think it is ready to merge.

PR #1294, opened 2022-07-13
Fixed LET inference and added test

Wonseo-C:
1. Set delay reactor's let is zero.
2. Checking if delay reaction's ```DependsOnPort``` is same as effect or not.
3. In``` isGeneratedDelay()```, the function checks if the reactor's name match "delay."
4. Add ```assignLetToReaction``` function in the Generator after ```recordBuiltinTriggers```.


lhstrh (approved):
Looks great! :rocket:

PR #1295, opened 2022-07-14
removed +nightly command for cargo

revol-xut:
Tiny pull request so the rust code is built with:

```
cargo build --release
```

instead of

```
cargo +nightly build --release
```

The problem is that the +nightly argument only exists in cargo from rustup. So standard the "standard" cargo will fail.

```
       error: no such subcommand: `+nightly`
       --> (unknown file):null:1 - cargo failed with error code 101 and reported the following error
       (error: no such subcommand: `+nightly`
       
       
       lfc: fatal error: Aborting due to previous error
```

=> the cargo in my env is nightly


cmnrd (commented):
Note that we already have #1218. Maybe you can sync with @jhaye on this matter ;)

PR #1296, opened 2022-07-14
Include reactor-rs as a submodule

### LINKED ISSUE #1291: Add reactor-rust as a submodule
We plan to publish release notes for our runtime implementations as part of the Lingua Franca release notes. This will give the runtime implementations more visibility, and it compensates for a lack of versioning/releases of happening in their repositories. Currently, all our runtime implementations are hooked into the Lingua Franca repository as a submodule, _except for `reactor-rust`, for which we keep track of its refs in a separate file. To make it easier to automatically generate and include release notes for `reactor-rust`, we should also include it as a submodule.

Note the command `git submodule` can be used to find the ref a submodule points to.

Going forward, we should always include a runtime used by LF as a submodule, _unless_ the runtime maintains its own versioning/releases.

If we decide that using a submodule for reactor-rust is not a good fit, we can also resolve this by starting to do proper versioning for it and publish release notes separately.

cmnrd:
This PR adds reactor-rs as a submodule and makes sure that the library is copied to src-gen from the classpath. The generated Cargo.toml is adjusted, so that it reads the local version and only pulls from GitHub if another version is explicitly specified by the user.

Closes #1291


lhstrh (approved):
This looks great, but I think that instead of removing the `README.md`, we should update it and explain how a runtime other than what's in the submodule can be specified.

cmnrd:
I don't think that the README is the right place for this. We already explain the target properties `runtime-version` and `external-runtime-version` in the docs on the website (although they wrongly state that rust does not support `runtime-version`). For lfc, we already have the help text which explains the options, but we should probably also have a page in the docs dedicated to lfc.

PR #1297, opened 2022-07-15
Bump Klighd version to 2.2

a-sr:
Updates the target platform to the latest Klighd release.

**Important**:  With this change, the generated language/diagram server will only work with the upcoming (corresponding) klighd-vscode release.


org.lflang.diagram/src/org/lflang/diagram/synthesis/SynthesisRegistration.java
```diff
@@ -32,6 +36,17 @@ public void execute() {
         
         // Style Mod
         reg.registerStyleModifier(ReactionPortAdjustment.ID, new ReactionPortAdjustment());
+        
+        // Blacklist LF-specific properties that should be removed when a diagram is send from the diagram server to a client.
```
lhstrh, line 40:
```suggestion
        // Blacklist LF-specific properties that should be removed when a diagram is sent from the diagram server to a client.
```

org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaStyleExtensions.java
```diff
@@ -67,7 +67,10 @@
 @ViewSynthesisShared
 public class LinguaFrancaStyleExtensions extends AbstractSynthesisExtensions {
     
-    private static final Property<Colors> LABEL_PARENT_BACKGROUND = new Property<>(
+    /** 
+     * INTERNAL property to communicate a node's background color.
```
lhstrh, line 71: If internal then why `public`?

lhstrh (approved):
Looks good to me aside from the changes to the visibility modifiers, which, given the `INTERNAL` comments are probably necessary to accommodate some API...?

org.lflang.diagram/src/org/lflang/diagram/synthesis/styles/LinguaFrancaStyleExtensions.java
a-sr: I needed to change them to public in order to reference them in the blacklisting in SynthesisRegistration.java

PR #1298, opened 2022-07-18
Multiport iterator for more efficient access to sparse inputs

edwardalee:
This PR enables more efficient access to input multiports that are *sparse*, meaning that few of the channels (typically) have inputs present.  This is used as in the following example:
```
     @sparse
     input[100] in:int;
     reaction(in) {=
         struct lf_multiport_iterator_t i = lf_multiport_iterator(in);
         int channel = lf_multiport_next(&i);
         while(channel >= 0) {
             printf("Received %d on channel %d.\n", in[channel]->value, channel);
             channel = lf_multiport_next(&i);
         }
     =}
```
In my informal measurement, this yields a 15% improvement in the unthreaded execution of the Big benchmark (with the default setting of 120 Worker reactors).

This PR also includes fixes to the annotation scheme to enable annotations with no arguments, as in `@sparse` above. @lsk567: Please have a look at those changes. If we decide that the 15% improvement is not worth the additional complexity, then we should cherry pick those changes.

The following PRs should be merged with this:
* https://github.com/lf-lang/reactor-c/pull/95
* https://github.com/lf-lang/reactor-c-py/pull/15


org.lflang/src/org/lflang/AttributeUtils.java
```diff
@@ -86,7 +86,20 @@ public static String findLabelAttribute(EObject node) {
...
+    
+    /**
+     * Return true if the specified node is an Input and has an {@code @sparse}
+     * attribute.
+     * @param node An AST node.
+     */
+    public static boolean isSparse(EObject node) {
+        if (node instanceof Input) {
+            for (var attribute : getAttributes(node)) {
+                if (attribute.getAttrName().equals("sparse")) return true;
```
lhstrh, line 99: This check is case sensitive. Should it be?

org.lflang/src/org/lflang/generator/c/CPortGenerator.java
```diff
@@ -88,22 +96,39 @@ public static String initializeInputMultiport(
...
-                String.join("\n",
-                    "// width of -2 indicates that it is not a multiport.",
-                    portRefName+"_width = -2;"
                 );
+            }
+            return result;
+        } else {
+            return String.join("\n",
+                "// width of -2 indicates that it is not a multiport.",
+                portRefName+"_width = -2;"
```
lhstrh, line 129: Why not use an `enum` for this and ditch the comment?

lhstrh (approved):
It left some nitpicks...

org.lflang/src/org/lflang/AttributeUtils.java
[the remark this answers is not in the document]
edwardalee: Made this case insensitive so you can write @Sparse.

org.lflang/src/org/lflang/generator/c/CPortGenerator.java
edwardalee: The width can also be any nonnegative integer. I don't know how to use an enum for that.

lhstrh: Ah, right. Never mind this suggestion then.

PR #1299, opened 2022-07-20
Removal of "No CMake" build option

petervdonovan:
This branch deletes over 500 LOC and breaks the "no CMake" build option that directly invokes the C compiler. The corresponding PRs in `reactor-c` and `reactor-c-py` should be merged first.

It's a bit of a work-in-progress, but I wish to rush this into master to mitigate conflicts. I promise to help resolve any conflicts that result in the `fed-gen` branch.


test/C/src/serialization/include/CMakeListsExtension.txt
```diff
@@ -1,3 +1,4 @@
 find_package(std_msgs REQUIRED)
 
-ament_target_dependencies(${LF_MAIN_TARGET} PUBLIC std_msgs)
+# Tradeoff of not using ament described here: https://docs.ros.org/en/foxy/How-To-Guides/Ament-CMake-Documentation.html?highlight=ament_target_dependencies
+target_link_libraries(${LF_MAIN_TARGET} PUBLIC std_msgs)
```
Soroosh129, line 4: But this is a test for our ROS integration, not our build system.

Abandoning Ament would be altering our migration procedure.
There are also existing migrated ROS nodes that use `ament_target_dependencies` in the order of a couple of dozen nodes.

petervdonovan: Thanks for pointing that out. I'll look for another solution.

.github/workflows/c-tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   run:
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
+        platform: [windows-latest]
```
petervdonovan, line 24:
```suggestion
        platform: [ubuntu-latest, macos-latest, windows-latest]
```

.idea/codeStyles/Project.xml
```diff
@@ -47,6 +46,8 @@
           <emptyLine />
         </value>
       </option>
+      <option name="ALIGN_MULTILINE_RECORDS" value="false" />
```
petervdonovan, line 49: I took the liberty to set the margin lines and to change alignment according to my personal preferences since we have no style guide... I'm happy to undo this though.

org.lflang/src/org/lflang/generator/c/Actions.java
```diff
@@ -25,7 +25,7 @@
  * @author{Alexander Schulz-Rosengarten <als@informatik.uni-kiel.de>}
  * @author{Hou Seng Wong <housengw@berkeley.edu>}
  */
-public class CActionGenerator {
+public class Actions {
```
petervdonovan, line 28: **These renamings are totally a stylistic choice which I am happy to revert.**

Reasoning: Having all the files in `org.lflang.generator.c` start with "C" and end with "Generator" is a waste of characters that makes it unnecessarily difficult to find the desired file. It also doesn't really make sense since some of these files can only loosely be considered as "generators" -- a suffix that serves more to comply with Java's tyrannical nounification conventions than to convey useful information.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -73,77 +78,212 @@ public class CCompiler {
...
-        String file,
-        boolean noBinary,
         GeneratorBase generator,
         LFGeneratorContext context
     ) throws IOException {
-        if (noBinary && context.getMode() == LFGeneratorContext.Mode.STANDALONE) {
-            errorReporter.reportError("Did not output executable; no main reactor found.");
-        }
-        LFCommand compile = compileCCommand(file, noBinary);
+        // Set the build directory to be "build"
```
petervdonovan, line 108: Much of the code here is not new; it is just copied from `CCmakeCompiler`, which is now gone.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1433,7 +1387,7 @@ private void generateInteractingContainedReactors(
             // Generate one struct for each contained reactor that interacts.
             body.pr("struct {");
             body.indent();
-            for (Port port : contained.portsOfInstance(containedReactor)) {
+            for (org.lflang.lf.Port port : contained.portsOfInstance(containedReactor)) {
```
petervdonovan, line 1390:
```suggestion
            for (Port port : contained.portsOfInstance(containedReactor)) {
```

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1879,24 +1834,6 @@ public static String variableStructType(TriggerInstance<?> portOrAction) {
         return portOrAction.getParent().reactorDeclaration.getName().toLowerCase()+"_"+portOrAction.getName()+"_t";
     }
 
-    /**
```
petervdonovan, line 1882: These two methods were unused, maybe because they perform trivial operations. I deleted them so that my IDE wouldn't mark them with warnings.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -2164,60 +2101,13 @@ private void initializeOutputMultiports(ReactorInstance reactor) {
...
-     * parameter references on the specified self struct.
-     * @param port The port.
-     * @param contained If the port belongs to a contained reactor, then
-     *  the contained reactor's instantiation. Otherwise, null.
-     * @param reactorInstance The reactor referring to this port. If null, "self" will be used
-     *  to reference the reactor.
-     * @return The width expression for a multiport or an empty string if it is
-     *  not a multiport.
-     */
-    protected String multiportWidthSpecInC(Port port, Instantiation contained, ReactorInstance reactorInstance) {
```
petervdonovan, line 2186: This method was also unused.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -2250,10 +2139,19 @@ protected void setUpGeneralParameters() {
         }
         if (hasModalReactors) {
             // So that each separate compile knows about modal reactors, do this:
-            targetConfig.compileDefinitions.put("MODAL_REACTORS", "");
+            targetConfig.compileDefinitions.put("MODAL_REACTORS", "TRUE");
         }
-        if (targetConfig.threading) {
+        if (targetConfig.threading) {  // FIXME: This logic is duplicated in CMake
```
petervdonovan, line 2144: In light of my decreased availability, I would like to expedite the process of merging this, even if we have to turn this comment into an issue.

org.lflang/src/org/lflang/generator/CodeBuilder.java
```diff
@@ -134,7 +134,8 @@ public void prSourceLineNumber(EObject eObject) {
             // Extract the filename from eResource, an astonishingly difficult thing to do.
             URI resolvedURI = CommonPlugin.resolve(eObject.eResource().getURI());
             // pr(output, "#line " + (node.getStartLine() + offset) + ' "' + FileConfig.toFileURI(fileConfig.srcFile) + '"')
-            pr("#line " + (node.getStartLine() + offset) + " \"" + resolvedURI + "\"");
+            // FIXME: Re-enable line directives before merging
+            // pr("#line " + (node.getStartLine() + offset) + " \"" + resolvedURI + "\"");
```
petervdonovan, line 138:
```suggestion
            pr("#line " + (node.getStartLine() + offset) + " \"" + resolvedURI + "\"");
```

org.lflang/src/org/lflang/generator/c/Preambles.java
```diff
@@ -19,40 +19,47 @@
...
  *
- * @author{Edward A. Lee <eal@berkeley.edu>}
- * @author{Marten Lohstroh <marten@berkeley.edu>}
- * @author{Mehrdad Niknami <mniknami@berkeley.edu>}
- * @author{Christian Menard <christian.menard@tu-dresden.de>}
- * @author{Matt Weber <matt.weber@berkeley.edu>}
- * @author{Soroush Bateni <soroush@utdallas.edu>
- * @author{Alexander Schulz-Rosengarten <als@informatik.uni-kiel.de>}
- * @author{Hou Seng Wong <housengw@berkeley.edu>}
+ * @author Edward A. Lee <eal@berkeley.edu>
```
petervdonovan, line 22: This change silences a warning that is very common in our code base

org.lflang/src/org/lflang/generator/python/Ports.java
```diff
@@ -13,15 +13,12 @@
...
 import static org.lflang.generator.c.CUtil.generateWidthVariable;
 
-public class PythonPortGenerator {
+public class Ports {
     public static final String NONMULTIPORT_WIDTHSPEC = "-2";
 
     /**
-     * Generate code to convert C actions to Python action capsules
-     * @see pythontarget.h.
-     * @param pyObjectDescriptor A string representing a list of Python format types (e.g., "O") that
```
petervdonovan, line 22: The changes in this file are due to stale comments.

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -1,5 +1,3 @@
-/* Generator for the Python target. */
```
petervdonovan, line 1: This change silences a warning. Perhaps comments of this sort should be discouraged (their content is obvious given the class name)

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -211,10 +223,7 @@ public String generatePythonCode(FederateInstance federate, String pyModuleName)
             "# so that these names will be assumed present without having to compile and install.",
             "# pylint: disable=no-name-in-module, import-error",
             "from "+pyModuleName+" import (",
-            "    Tag, action_capsule_t, compare_tags, get_current_tag, get_elapsed_logical_time,",
-            "    get_elapsed_physical_time, get_logical_time, get_microstep, get_physical_time,",
-            "    get_start_time, port_capsule, request_stop, schedule_copy,",
-            "    start",
```
petervdonovan, line 217: The deleted functions were already deprecated

org.lflang/src/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -235,57 +244,7 @@ public String generatePythonCode(FederateInstance federate, String pyModuleName)
...
-
-    /**
-     * Generate the setup.py required to compile and install the module.
-     * Currently, the package name is based on filename which does not support sharing the setup.py for multiple .lf files.
-     * TODO: use an alternative package name (possibly based on folder name)
-     *
-     * If the LF program itself is threaded or if tracing is enabled, NUMBER_OF_WORKERS is added as a macro
-     * so that platform-specific C files will contain the appropriate functions.
-     */
-    public String generatePythonSetupFile(String lfModuleName, String pyModuleName) {
```
petervdonovan, line 250: The Python build now uses C's CMake infrastructure, not `setuptools`.

test/C/src/serialization/include/CMakeListsExtension.txt
petervdonovan: The "solution" that I chose was to enable a deprecated feature of CMake. Since there isn't really any case where it makes sense to use the syntax that ament uses, I think it's fine to live with the deprecation warnings for as long as ament continues to use that syntax, even if the process of transitioning takes years.

.github/workflows/c-tests.yml
petervdonovan, line 24:
```suggestion
        platform: [ubuntu-latest, macos-latest, windows-latest]
```

petervdonovan:
I do not know why the Python tests segfaulted in macOS in the CI run for [64e80ab](https://github.com/lf-lang/lingua-franca/pull/1299/commits/64e80ab8e6153aefaaf175e8ec4a8ddbc4c875df) but not for [81a7cb5](https://github.com/lf-lang/lingua-franca/pull/1299/commits/81a7cb52465b614b0f40aaa81e362b3fe3110e5e). I suspect that both have a memory error, but that whether it causes a crash depends on the runner instance that we get or is somehow determined by the state of the memory in the instance. Certainly we have a host of memcheck errors both here and in master...

petervdonovan:
Based on the output (after setting the output to be sent directly to stdout instead of being collected and maybe thrown away), I think that the `Minimal.lf` is not even able to start on macOS -- it seems to just immediately segfault before any debug prints (from setting logging to DEBUG) can appear. This makes it seem like maybe the issue could be something simple, like a missing DLL.

org.lflang/src/org/lflang/generator/c/Actions.java
lhstrh: I agree with you, but I think it might be best not to bundle such refactoring in this PR. One advantage of the current naming scheme is that it's clear from the filename or class name immediately which target it is for, so there is no need to disambiguate. That said, I think that the naming you suggest here is better. Perhaps @edwardalee has an opinion about this?

org.lflang/src/org/lflang/generator/c/CCompiler.java
lhstrh: :+1:

org.lflang/src/org/lflang/generator/c/Actions.java
lhstrh: How about `CGenerator` and `PythonGenerator` though? Should we drop the prefix on those as well? Or move them out of the `c` and `python` package, respectively?

lhstrh:
Perhaps @lsk567 would be able to help you debug the macOS issue?

org.lflang/src/org/lflang/generator/c/Actions.java
cmnrd: While I agree that this seems like a waste of characters, it is a common style in Java projects. Also I second @edwardalee's comment about distinguishing tabs/buffers and also I think it is easier to find the specific file you are looking for when using a fuzzy search. So I lean towards keeping things as they are, but I don't have a strong opinion here.

However, I don't think this PR is the place to make such a style change. If we want to change the style, we should do so consistently in the entire code base (including the C++ and  Rust generators for instance), and also record it in our style guide (which we don't have yet, but we should add it to the draft).

petervdonovan: It seems like this change is both controversial and unimportant, so I'll just revert it.

petervdonovan:
Here is a catalog of things that did not work... Just writing them down so that I don't forget and waste time trying them again.
1. In CI, the error is indeed a "Segmentation fault: 11." It appears in **all** Python tests on macOS.
2. ASan makes no mention of illegal memory accesses when run locally on Ubuntu.
3. When run with the correct debug version of Python, Valgrind makes no mention of illegal memory accesses on Ubuntu.
4. When run by SSHing into the GitHub Actions macOS machine, the Python tests pass and ASan makes no mention of illegal memory accesses.
5. When run on a 2013 MacBook Air with macOS 11, `Minimal.lf` and I think some other Python tests pass; to my knowledge, they probably all pass since this segfault seems like an "all or nothing" phenomenon.
6. When run in CI with debug prints enabled, no debug prints appear.
7. When run in CI in LLDB, echoing the string "run\nrun\nbt\n" into stdin, no output appears.
8. It seems infeasible to run tests in CI on macOS with GDB due to security issues.
9. It is infeasible to do a binary search through the version history to figure out where the error first appeared because in this PR, there was a long period during which the tests were not compiling. Either we use `setup.py` or we use CMake; there is very little in-between here.

I just got something that might be meaningful out of ASan, however:
```
Running Minimal.py
==13401==AddressSanitizer: libc interceptors initialized
|| `[0x200000000000, 0x7fffffffffff]` || HighMem    ||
|| `[0x140000000000, 0x1fffffffffff]` || HighShadow ||
|| `[0x120000000000, 0x13ffffffffff]` || ShadowGap  ||
|| `[0x100000000000, 0x11ffffffffff]` || LowShadow  ||
|| `[0x000000000000, 0x0fffffffffff]` || LowMem     ||
MemToShadow(shadow): 0x120000000000 0x123fffffffff 0x128000000000 0x13ffffffffff
redzone=16
max_redzone=2048
quarantine_size_mb=256M
thread_local_quarantine_size_kb=1024K
malloc_context_size=30
SHADOW_SCALE: 3
SHADOW_GRANULARITY: 8
SHADOW_OFFSET: 0x100000000000
==13401==Installed the sigaction for signal 11
==13401==Installed the sigaction for signal 10
==13401==Installed the sigaction for signal 8
==13401==T0: stack [0x7ffee907f000,0x7ffeea07f000) size 0x1000000; local=0x7ffeea07b2a0
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==Checking file existence is not allowed under sandbox.
==13401==AddressSanitizer Init done
AddressSanitizer:DEADLYSIGNAL
=================================================================
==13401==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x000109dafe0e bp 0x7ffeea07b710 sp 0x7ffeea07b700 T0)
==13401==The signal is caused by a READ memory access.
==13401==Hint: address points to the zero page.
    #0 0x109dafe0e in PyDict_New+0x15 (Python:x86_64+0xafe0e)
    #1 0x109dde616 in PyType_Ready+0x69f (Python:x86_64+0xde616)
    #2 0x1096b8633 in PyInit_LinguaFrancaMinimal pythontarget.c:328
    #3 0x1067973e4 in _imp_create_dynamic+0x5a4 (libpython3.10.dylib:x86_64+0x2043e4)
    #4 0x106685144 in cfunction_vectorcall_FASTCALL+0x54 (libpython3.10.dylib:x86_64+0xf2144)
    #5 0x10674e5b5 in _PyEval_EvalFrameDefault+0x36e5 (libpython3.10.dylib:x86_64+0x1bb5b5)
    #6 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #7 0x10675a69e in call_function+0xae (libpython3.10.dylib:x86_64+0x1c769e)
    #8 0x10675122e in _PyEval_EvalFrameDefault+0x635e (libpython3.10.dylib:x86_64+0x1be22e)
    #9 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #10 0x10675a69e in call_function+0xae (libpython3.10.dylib:x86_64+0x1c769e)
    #11 0x106750377 in _PyEval_EvalFrameDefault+0x54a7 (libpython3.10.dylib:x86_64+0x1bd377)
    #12 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #13 0x10675a69e in call_function+0xae (libpython3.10.dylib:x86_64+0x1c769e)
    #14 0x10674dc16 in _PyEval_EvalFrameDefault+0x2d46 (libpython3.10.dylib:x86_64+0x1bac16)
    #15 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #16 0x10675a69e in call_function+0xae (libpython3.10.dylib:x86_64+0x1c769e)
    #17 0x10674dc16 in _PyEval_EvalFrameDefault+0x2d46 (libpython3.10.dylib:x86_64+0x1bac16)
    #18 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #19 0x10675a69e in call_function+0xae (libpython3.10.dylib:x86_64+0x1c769e)
    #20 0x10674dc16 in _PyEval_EvalFrameDefault+0x2d46 (libpython3.10.dylib:x86_64+0x1bac16)
    #21 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #22 0x106620b19 in object_vacall+0x1d9 (libpython3.10.dylib:x86_64+0x8db19)
    #23 0x106620dc8 in _PyObject_CallMethodIdObjArgs+0xf8 (libpython3.10.dylib:x86_64+0x8ddc8)
    #24 0x1067927e7 in PyImport_ImportModuleLevelObject+0x427 (libpython3.10.dylib:x86_64+0x1ff7e7)
    #25 0x106750941 in _PyEval_EvalFrameDefault+0x5a71 (libpython3.10.dylib:x86_64+0x1bd941)
    #26 0x106749903 in _PyEval_Vector+0x1b3 (libpython3.10.dylib:x86_64+0x1b6903)
    #27 0x1067cb2aa in pyrun_file+0x14a (libpython3.10.dylib:x86_64+0x2382aa)
    #28 0x1067caac7 in _PyRun_SimpleFileObject+0x147 (libpython3.10.dylib:x86_64+0x237ac7)
    #29 0x1067c9fcd in _PyRun_AnyFileObject+0x7d (libpython3.10.dylib:x86_64+0x236fcd)
    #30 0x1067f1129 in pymain_run_file_obj+0xe9 (libpython3.10.dylib:x86_64+0x25e129)
    #31 0x1067f0a04 in pymain_run_file+0x54 (libpython3.10.dylib:x86_64+0x25da04)
    #32 0x1067f02f4 in Py_RunMain+0x974 (libpython3.10.dylib:x86_64+0x25d2f4)
    #33 0x1067f129a in pymain_main+0x2a (libpython3.10.dylib:x86_64+0x25e29a)
    #34 0x1067f1e89 in Py_BytesMain+0x29 (libpython3.10.dylib:x86_64+0x25ee89)
    #35 0x7fff203cbf3c in start+0x0 (libdyld.dylib:x86_64+0x15f3c)

==13401==Register values:
rax = 0x0000000000000000  rbx = 0x0000000000000000  rcx = 0x0000000000002100  rdx = 0x00000001096c56f8  
rdi = 0x00000001096c5650  rsi = 0x0000000000000000  rbp = 0x00007ffeea07b710  rsp = 0x00007ffeea07b700  
 r8 = 0x0000000000000000   r9 = 0x0000000000000000  r10 = 0x0000000000000000  r11 = 0x000000010693ffa8  
r12 = 0x0000000109c12160  r13 = 0x0000000109c08670  r14 = 0x0000000109bc4df0  r15 = 0x00000001096b8620  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (Python:x86_64+0xafe0e) in PyDict_New+0x15
==13401==ABORTING
/Users/runner/work/_temp/eb6be2ad-c0f9-49a0-8ff3-653fe594d20e.sh: line 6: 13401 Abort trap: 6           python3 /Users/runner/work/lingua-franca/lingua-franca/test/Python/src-gen/Minimal/Minimal.py
```

petervdonovan:
I said that our code didn't appear in the stack trace, but I was wrong. It's right there, when we call `PyType_Ready(&py_port_capsule_t)`. So I am optimistic that we can fix this very soon.

cmnrd:
> `==13401==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x000109dafe0e bp 0x7ffeea07b710 sp 0x7ffeea07b700 T0)
==13401==The signal is caused by a READ memory access.`

That address looks really weird. Looks like an incremented `NULL` pointer...

Since there is no debug output showing, I would a assume that the segfault happens sometime during initialization. Have you tried to place a `printf` directly at the start of the main function? If this does not show, the problem might lie in the loading of some dynamically linked libraries. Let me know if I can be of any help with this.

petervdonovan:
Thanks Christian for taking a look. I guess the debug print would have to be entered into a fork/local version of the CPython repo in order for it to work, but I guess that's the next thing I should try.

petervdonovan:
Status update from core dump (the problem is that `%rax` is null; note also that the function `PyDict_New` does not take arguments, which makes it difficult to figure out what caused `%rax` to be null):
```
(lldb) register read
General Purpose Registers:
       rax = 0x0000000000000000
       rbx = 0x0000000100694b78  python3`_Py_PackageContext
       rcx = 0x0000000000002100
       rdx = 0x00000001009a46f8  py_port_capsule_t + 168
       rdi = 0x00000001009a4650  py_port_capsule_t
       rsi = 0x0000000010000000
       rbp = 0x00007ffeef9f6850
       rsp = 0x00007ffeef9f6840
        r8 = 0x00000000000024d3
        r9 = 0x0000000000000008
       r10 = 0x00007fc596700000
       r11 = 0x0000000000000000
       r12 = 0x00007fc596720ba0
       r13 = 0x00000001004d1699  "PyInit"
       r14 = 0x00007fc596720580
       r15 = 0x0000000000000000
       rip = 0x0000000100a6326a  Python`PyDict_New + 21
    rflags = 0x0000000000010202
        cs = 0x000000000000002b
        fs = 0x0000000000000000
        gs = 0x0000000000000000

(lldb) fr s 0
frame #0: 0x0000000100a6326a Python`PyDict_New + 21
Python`PyDict_New:
->  0x100a6326a <+21>: movq   0x10(%rax), %rax
    0x100a6326e <+25>: movl   0x2c40(%rax), %ecx
    0x100a63274 <+31>: testl  %ecx, %ecx
    0x100a63276 <+33>: je     0x100a632da               ; <+133>
(lldb) l
   346 
   347      m = PyModule_Create(&MODULE_NAME);
   348 
   349      if (m == NULL) {
   350          return NULL;
   351      }
   352 
(lldb)
```

Assembly (manually truncated):
```
Python`PyDict_New:
    0x100a63255 <+0>:   pushq  %rbp
    0x100a63256 <+1>:   movq   %rsp, %rbp
    0x100a63259 <+4>:   pushq  %r14
    0x100a6325b <+6>:   pushq  %rbx
    0x100a6325c <+7>:   incq   0x24277d(%rip)            ; empty_keys_struct
    0x100a63263 <+14>:  movq   0x261506(%rip), %rax      ; _PyRuntime + 616
->  0x100a6326a <+21>:  movq   0x10(%rax), %rax
    0x100a6326e <+25>:  movl   0x2c40(%rax), %ecx
    0x100a63274 <+31>:  testl  %ecx, %ecx
    0x100a63276 <+33>:  je     0x100a632da               ; <+133>
    0x100a63278 <+35>:  decl   %ecx
    0x100a6327a <+37>:  movl   %ecx, 0x2c40(%rax)
    0x100a63280 <+43>:  movslq %ecx, %rcx
```
I'm inexperienced with x86 and don't understand why we are referring to the return value register `%rax` at the beginning of this function. All we've done so far AFAICT is to push saved registers onto the stack and increment a global counter -- no function calls -- so why would we expect to know anything about the value stored in `%rax`?

petervdonovan:
Here is the situation in the parent frame (`PyDict_Ready`):
```
(lldb) x/20i $pc-32
    0x100a91abf: 6d                    insl   %dx, %es:(%rdi)
    0x100a91ac0: d0 0f                 rorb   (%rdi)
    0x100a91ac2: 84 a8 03 00 00 48     testb  %ch, 0x48000003(%rax)
    0x100a91ac8: 8b 53 28              movl   0x28(%rbx), %edx
    0x100a91acb: 48 83 c3 28           addq   $0x28, %rbx
    0x100a91acf: 48 85 d2              testq  %rdx, %rdx
    0x100a91ad2: 0f 84 4f fc ff ff     je     0x100a91727               ; <+744>
    0x100a91ad8: eb a2                 jmp    0x100a91a7c               ; <+1597>
    0x100a91ada: e8 76 17 fd ff        callq  0x100a63255               ; PyDict_New
->  0x100a91adf: 48 85 c0              testq  %rax, %rax
    0x100a91ae2: 0f 84 1d 10 00 00     je     0x100a92b05               ; <+5830>
    0x100a91ae8: 48 8b 7d d0           movq   -0x30(%rbp), %rdi
    0x100a91aec: 48 89 87 08 01 00 00  movq   %rax, 0x108(%rdi)
    0x100a91af3: 48 8b 55 c0           movq   -0x40(%rbp), %rdx
    0x100a91af7: e9 ab f9 ff ff        jmp    0x100a914a7               ; <+104>
    0x100a91afc: 48 8b b7 00 01 00 00  movq   0x100(%rdi), %rsi
    0x100a91b03: 48 85 f6              testq  %rsi, %rsi
    0x100a91b06: 0f 84 80 04 00 00     je     0x100a91f8c               ; <+2893>
    0x100a91b0c: bf 01 00 00 00        movl   $0x1, %edi
    0x100a91b11: 31 c0                 xorl   %eax, %eax
```
```
(lldb) register read
General Purpose Registers:
       rbx = 0x0000000100694b78  python3`_Py_PackageContext
       rbp = 0x00007ffeef9f68d0
       rsp = 0x00007ffeef9f6860
       r12 = 0x00007fc596720ba0
       r13 = 0x00000001004d1699  "PyInit"
       r14 = 0x00007fc596720580
       r15 = 0x0000000000000000
       rip = 0x0000000100a91adf  Python`PyType_Ready + 1696
13 registers were unavailable.

```

cmnrd:
> Assembly (manually truncated):
> 
> ```
> Python`PyDict_New:
>     0x100a63255 <+0>:   pushq  %rbp
>     0x100a63256 <+1>:   movq   %rsp, %rbp
>     0x100a63259 <+4>:   pushq  %r14
>     0x100a6325b <+6>:   pushq  %rbx                             ; Until here it should be setting up the stack
>     0x100a6325c <+7>:   incq   0x24277d(%rip)               ; empty_keys_struct ; 0x24277d(%rip) refers to some global variable and this variable is incremented
>     0x100a63263 <+14>:  movq   0x261506(%rip), %rax      ; _PyRuntime + 616 ; This is another global variable that is loaded into rax
> ->  0x100a6326a <+21>:  movq   0x10(%rax), %rax           ; This loads the memory value add  address rax + 0x10 into rax
>     0x100a6326e <+25>:  movl   0x2c40(%rax), %ecx
>     0x100a63274 <+31>:  testl  %ecx, %ecx
>     0x100a63276 <+33>:  je     0x100a632da               ; <+133>
>     0x100a63278 <+35>:  decl   %ecx
>     0x100a6327a <+37>:  movl   %ecx, 0x2c40(%rax)
>     0x100a63280 <+43>:  movslq %ecx, %rcx
> ```
> 
> I'm inexperienced with x86 and don't understand why we are referring to the return value register `%rax` at the beginning of this function. All we've done so far in `PyDict_New` AFAICT is to push saved registers onto the stack and increment a global counter -- no function calls -- so why would we expect to know anything about the value stored in `%rax`?

I am not sure what you mean `rax` is a general purpose register, not a return value register. I added a few comments to the code above. The problem is that `movq   0x261506(%rip), %rax` apparently loads 0. This must be some global variable. The comment on the right says `_PyRuntime + 616 ` but I am not sure what this refers to.

petervdonovan:
Okay, I guess I just got confused between Intel and AT&T syntax. What Christian said clears that up now. Here is another confusion that I have:
```
    0x108f6f25c: 48 ff 05 7d 27 24 00     incq   0x24277d(%rip)            ; empty_keys_struct
    0x108f6f263: 48 8b 05 06 15 26 00     movq   0x261506(%rip), %rax      ; _PyRuntime + 616
->  0x108f6f26a: 48 8b 40 10              movq   0x10(%rax), %rax
    0x108f6f26e: 8b 88 40 2c 00 00        movl   0x2c40(%rax), %ecx
    0x108f6f274: 85 c9                    testl  %ecx, %ecx
    0x108f6f276: 74 62                    je     0x108f6f2da               ; <+133>
    0x108f6f278: ff c9                    decl   %ecx
    0x108f6f27a: 89 88 40 2c 00 00        movl   %ecx, 0x2c40(%rax)
    0x108f6f280: 48 63 c9                 movslq %ecx, %rcx
    0x108f6f283: 48 8b 9c c8 c0 29 00 00  movq   0x29c0(%rax,%rcx,8), %rbx
    0x108f6f28b: 83 3d 12 64 24 00 00     cmpl   $0x0, 0x246412(%rip)      ; _Py_tracemalloc_config + 3
    0x108f6f292: 0f 85 90 00 00 00        jne    0x108f6f328               ; <+211>
    0x108f6f298: 48 c7 03 01 00 00 00     movq   $0x1, (%rbx)
    0x108f6f29f: 48 8d 05 3a 27 24 00     leaq   0x24273a(%rip), %rax      ; empty_keys_struct
    0x108f6f2a6: 48 89 43 20              movq   %rax, 0x20(%rbx)
    0x108f6f2aa: 48 8d 05 77 1d 26 00     leaq   0x261d77(%rip), %rax      ; empty_values
    0x108f6f2b1: 48 89 43 28              movq   %rax, 0x28(%rbx)
    0x108f6f2b5: 48 c7 43 10 00 00 00 00  movq   $0x0, 0x10(%rbx)
(lldb) p &empty_keys_struct
(PyDictKeysObject *) $6 = 0x0000000108a927b0
```
We have that 0x108f6f263 + 0x261506 != 0x0000000108a927b0. Furthermore, I thought the data segment was supposed to be above the text segment, but the address of `empty_keys_struct` (which is a static initialized variable) is lower than program counter. Sorry if this is a dumb question, but could this be related to my [switching on position independent code](https://github.com/lf-lang/lingua-franca/blob/88731c882d86fc51d090cff85e8bf2b151c080c1/org.lflang/src/org/lflang/generator/python/PythonGenerator.java#L797) for our C extension? I don't understand why this would have any effect on where `empty_keys_struct` is, but I do not know how else to interpret this address mismatch.

petervdonovan:
Following a conversation with @edwardalee, I tried to reproduce the segfault in an SSH session in the GitHub Actions runner and succeeded, but only when using the same debug version of Python that I had compiled in the workflow. In the same SSH session, the same program ran successfully in the default Python interpreter that came with the runner.

This makes it look as if the segfault I observed was not the same segfault that was showing up in all the tests (which used the regular default Python that comes with the runner) and that I had just compiled Python incorrectly somehow. But if I do the same process, without a debug version of Python, I still get a segfault in the GitHub Actions environment, but then I cannot reproduce the segfault upon SSHing in, even if I do not re-compile anything at all. Of course, I also failed to reproduce after just running `make` in the build directory.

However, it does indeed seem like the specific comments I had about the address of `empty_keys_struct` were a red herring because in the regular default Python, the address is quite right (i.e., 0x24277d + 0x10d53d263 = 0x10D77F9E0)
```
(lldb) p &empty_keys_struct
warning: could not execute support code to read Objective-C class data in the process. This may reduce the quality of type information available.
(void **) $0 = 0x000000010d77f9e0
(lldb) x/20i $pc-16
    0x10d53d25a: 56                       pushq  %rsi
    0x10d53d25b: 53                       pushq  %rbx
    0x10d53d25c: 48 ff 05 7d 27 24 00     incq   0x24277d(%rip)            ; empty_keys_struct
    0x10d53d263: 48 8b 05 06 15 26 00     movq   0x261506(%rip), %rax      ; _PyRuntime + 616
->  0x10d53d26a: 48 8b 40 10              movq   0x10(%rax), %rax
    0x10d53d26e: 8b 88 40 2c 00 00        movl   0x2c40(%rax), %ecx
```
So that particular mystery is solved, I guess. The questions that remain are
- What global variable is accessed in the line `movq   0x261506(%rip), %rax      ; _PyRuntime + 616`? It should be possible to figure this out by examining the source code for `PyDict_New`. Currently my best guess is that it is that this is the location of the thread state. The assembly seems to be inlined, so I had to follow some procedure calls to get to this conclusion, but in particular I suspect that this line corresponds to line 70 of `pycore_pystate.h`, which reads: `    return (PyThreadState*)_Py_atomic_load_relaxed(&runtime->gilstate.tstate_current);
`. On line 38 of `pycore_runtime.h`, it reads:
```
/* Assuming the current thread holds the GIL, this is the
       PyThreadState for the current thread. */
    _Py_atomic_address tstate_current;
```
- Why is that global variable `NULL` when we run the Python interpreter in the GH actions runner but not when we SSH in? Could it just be because we forgot to acquire the GIL? Or is it something to do with the thread state? The relevant typedef might be the one on line 82 of `pystate.h`.

cmnrd:
> We have that 0x108f6f263 + 0x261506 != 0x0000000108a927b0. Furthermore, I thought the data segment was supposed to be above the text segment, but the address of `empty_keys_struct` (which is a static initialized variable) is lower than program counter. Sorry if this is a dumb question, but could this be related to my [switching on position independent code](https://github.com/lf-lang/lingua-franca/blob/88731c882d86fc51d090cff85e8bf2b151c080c1/org.lflang/src/org/lflang/generator/python/PythonGenerator.java#L797) for our C extension? I don't understand why this would have any effect on where `empty_keys_struct` is, but I do not know how else to interpret this address mismatch.

I think this is a very good question actually. I don't have a definite answer though. This stack overflow question might be relevant: https://stackoverflow.com/questions/29813643/pointer-deferencing-in-x86-assembly-code. Do we compile all the code with `-fPIC`? You could also check the compiler commands that cmake invokes by setting the environment `VERBOSE=1`. Also I am curious what `empty_keys_struct` represents. Is this something defined by us or by the Python libs?

petervdonovan:
We just compile the C extension with `-fPIC`, I believe, and `empty_keys_struct` is in the `cpython` repository, with the documentation, "This immutable, empty PyDictKeysObject is used for PyDict_Clear() (which cannot fail and thus can do no allocation)." I didn't think `-fPIC` could be the reason for the address mismatch because I didn't think it applied to this `cpython` code.

In any case, the address mismatch is gone when we use the precompiled Python that comes with the GitHub Actions runner, but the segfault is still triggered by the same instruction with the same (wrong) value.

petervdonovan:
If I try to do something completely different than before (acquire the GIL), I still get a segfault from accessing the address `NULL + 0x10`. I think the Python interpreter just isn't initialized. I've heard that [this can happen due to linking against two different versions of Python](https://stackoverflow.com/questions/4985933/is-the-pythreadstate-of-the-main-python-thread-expected-to-be-null). The error also seems to go away in a virtual environment. For these reasons, I suspect that CMake is simply not discovering the correct Python interpreter.

cmnrd:
In this case it should be helpful to get the verbose output of cmake and compare the executed commands in CI to a local run.

petervdonovan:
I did as Christian suggested and can confirm that CMake is discovers a different version of Python than what is discovered by `which python`. To be specific, invocations of `gcc` include the option `-isystem /usr/local/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/include/python3.10` whereas when I SSH into the runner I see the following:
```
Mac-1666038353816:lingua-franca runner$ which python3
/usr/local/bin/python3
Mac-1666038353816:lingua-franca runner$ ls -al /usr/local/bin/python3
lrwxr-xr-x  1 runner  admin  40 Oct 11 00:31 /usr/local/bin/python3 -> ../Cellar/python@3.10/3.10.7/bin/python
```

Basically, one Python is in the Cellar, and the other is not.

There are ways to fix this, e.g., [by explicitly specifying the path to Python when we invoke CMake.](https://stackoverflow.com/questions/7660001/cmake-finds-wrong-python-libs); furthermore, I know where in the code base this change would have to be made. However, there are a few reasons to stop here:
- Every build environment is different, and an end user can always find ways to break their build by setting up their environment in special ways. Arguably, whether this is a bug in the build system or in the environment is a matter of perspective.
- It is possible to work around this issue by using a virtual environment (which most Python programmers should do anyway).
- This bug has never been observed except in CI, and it could be very rare.
- It is not clear to me that other ways of integrating Python with CMake are recommended over the one I used; therefore, it is not obvious to me that fixing this bug would reduce the total number of bugs that we have.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -820,6 +821,14 @@ private static <T extends EObject> void insertModeElementsAtTextualPosition(List
         list.addAll(idx, elements);
     }
 
+    public static <T extends EObject> Iterable<T> allElementsOfClass(
+        Resource resource,
+        Class<T> clazz
```
lhstrh, line 826:
```suggestion
        Class<T> elementClass
```

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -820,6 +821,14 @@ private static <T extends EObject> void insertModeElementsAtTextualPosition(List
...
+    ) {
+        //noinspection StaticPseudoFunctionalStyleMethod
+        return Iterables.filter(IteratorExtensions.toIterable(resource.getAllContents()), clazz);
```
lhstrh, line 829:
```suggestion
        return Iterables.filter(IteratorExtensions.toIterable(resource.getAllContents()), elementClass);
```

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -227,15 +217,15 @@ public enum TargetProperty {
...
+
     /**
      * Flags to be passed on to the target compiler.
      */
     FLAGS("flags", UnionType.STRING_OR_STRING_ARRAY,
-            Arrays.asList(Target.C, Target.CCPP), (config, value, err) -> {
-                config.compilerFlags = ASTUtils.elementToListOfStrings(value);
-            }),
-    
+        Arrays.asList(Target.C, Target.CCPP), (config, value, err) -> {
```
lhstrh, line 225: If we use Cmake, we don't know which compiler gets used. Then how do we know which flags to use? Shouldn't we also get rid of this option then?

org.lflang/src/org/lflang/federated/launcher/FedCLauncher.java
```diff
@@ -75,28 +75,28 @@ String compileCommandForFederate(FederateInstance federate) {
             errorReporter.reportError("Failed to create file config for federate "+federate.name);
             return "";
         }
-        
+
         String commandToReturn = "";
-        // FIXME: Hack to add platform support only for linux systems. 
+        // FIXME: Hack to add platform support only for linux systems.
```
lhstrh, line 80: Could you please create an issue for this and reference it in the code?

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -73,77 +79,217 @@ public class CCompiler {
...
+        FileConfig fileConfig,
+        ErrorReporter errorReporter,
+        boolean cppMode
+    ) {
         this.fileConfig = fileConfig;
         this.targetConfig = targetConfig;
         this.errorReporter = errorReporter;
         this.commandFactory = new GeneratorCommandFactory(errorReporter, fileConfig);
-        this.CppMode = false;
+        this.cppMode = cppMode;
```
lhstrh, line 95: :rocket:

org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -258,4 +259,35 @@ CodeBuilder generateCMakeCode(
 
         return cMakeCode;
     }
+
```
lhstrh, line 262: JavaDoc missing for these two new methods.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -73,77 +79,217 @@ public class CCompiler {
...
-        if (noBinary && context.getMode() == LFGeneratorContext.Mode.STANDALONE) {
-            errorReporter.reportError("Did not output executable; no main reactor found.");
-        }
-        LFCommand compile = compileCCommand(file, noBinary);
+        // Set the build directory to be "build"
+        Path buildPath = fileConfig.getSrcGenPath().resolve("build");
+        // Remove the previous build directory if it exists to
+        // avoid any error residue that can occur in CMake from
+        // a previous build.
+        // FIXME: This is slow and only needed if an error
```
lhstrh, line 114: Good point. Please create and issue and reference it here.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -26,28 +26,34 @@
...
  *
  * @author Soroush Bateni <soroush@utdallas.edu>
- * @author{Edward A. Lee <eal@berkeley.edu>}
- * @author{Marten Lohstroh <marten@berkeley.edu>}
- * @author{Christian Menard <christian.menard@tu-dresden.de}
- * @author{Matt Weber <matt.weber@berkeley.edu>}
+ * @author Edward A. Lee <eal@berkeley.edu>
+ * @author Marten Lohstroh <marten@berkeley.edu>
+ * @author Christian Menard <christian.menard@tu-dresden.de
+ * @author Matt Weber <matt.weber@berkeley.edu>
```
lhstrh, line 56: Please add yourself :-) That's also appropriate for a bunch of other files here that you've touched with a lot more than just minor refactoring work.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -210,7 +345,7 @@ public LFCommand compileCCommand(
 
         // If there is no main reactor, then use the -c flag to prevent linking from occurring.
         // FIXME: we could add a `-c` flag to `lfc` to make this explicit in stand-alone mode.
-        // Then again, I think this only makes sense when we can do linking.
+        //  Then again, I think this only makes sense when we can do linking.
```
lhstrh, line 348: Does this need a second look as suggested by the old FIXME? Or else, can it be removed?

org.lflang/src/org/lflang/generator/c/CGenerator.java
lhstrh: Let's turn it into an issue.

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -378,19 +378,6 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
...
     }
 
-    /**
-     * Copy a list of files from a given source directory to a given destination directory.
-     * @param srcDir The directory to copy files from.
-     * @param dstDir The directory to copy files to.
-     * @param files The list of files to copy.
-     * @throws IOException If any of the given files cannot be copied.
-     */
-    public static void copyFilesFromClassPath(String srcDir, Path dstDir, List<String> files) throws IOException {
```
lhstrh, line 388: I guess we no longer copy lists of individual files from the class path?

lhstrh (approved):
This looks great! Congrats on finally getting this PR to pass the tests!! :fireworks: I left some comments/requests. I wonder how much trouble these changes will give us in `fed-gen`... Perhaps you could assist with the merge? I haven't reviewed the changes in `reactor-c` yet. I will do that next.

lhstrh:
Looks like there is no accompanying PR for `reactor-c-py`?

org.lflang/src/org/lflang/TargetProperty.java
petervdonovan: Yes, I think it is always dangerous and discouraged to use compiler flags with CMake. However, they can be a handy workaround when platform-independent strategies fail, so I am inclined to keep the option here.

petervdonovan: Currently we issue a warning, not an error, when you use compiler flags with CMake. It seems reasonable to stick with that policy.

org.lflang/src/org/lflang/util/FileUtil.java
petervdonovan: Yes, that was an objective of this PR

org.lflang/src/org/lflang/generator/c/CCompiler.java
petervdonovan: As far as I know, the "produce an object file instead of an executable" option does not even work -- I have not seen tests that use it. Therefore, my guess is that this FIXME is still unresolved. I personally do not see why we would want to examine questions related to that compilation option in the near future; given the way connections are resolved at compile time, it is unclear to me whether it ever will be feasible to make it work.

org.lflang/src/org/lflang/federated/launcher/FedCLauncher.java
petervdonovan: Done: #1418

PR #1300, opened 2022-07-20
Custom icons for reactors in diagrams

a-sr:
Adds support for using the new attribute notation for annotating reactors with an icon.


lhstrh (approved):
This looks good, but it makes me wonder allowable image formats/extensions. We should probably instate some checks to ensure that images are viable and can be displayed correctly.

a-sr:
Ok, I improved the error reporting and added validation rule to issue warnings for bad icon paths.

lhstrh:
Very cool!

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -104,6 +104,59 @@ public static IPath toIPath(URI uri) throws IOException {
...
+        File file = new File(path);
+        if (file.exists()) {
+            try {
+                return file.toURI();
+            } catch (Exception e) {
+                // nothing
+            }
+        }
+        // Check if path is relative to LF file
+        if (resource != null) {
```
lhstrh, line 136: Since the introduction of `FileConfig`, we have been interpreting paths relative to the project root, which is understood to be the parent of the `src` directory. We should probably include this possibility in the search. Would it make sense to pass in a `FileConfig` instance? We could also take the resource and find the `src` in its path, but that would be duplicating logic we already have in `FileConfig`...

a-sr: I moved this code from the diagram package to that utility class in order to use the same code base for the validator, however, in the diagram synthesis and the validator, there is only the model. How do I obtain the a `FileConfig` for a model in that context? I looks like `FileConfig` is only available in the generator.

lhstrh: You're right. I think this solution is fine, but it should also search relative to the root of the project (not just relative to the file itself). The root of the project is just the parent of the nearest containing `src` directory.

a-sr: Ah, yes, I can do that.

PR #1308, opened 2022-07-25
Use HTTPS for Rust reactor runtime submodule

jhaye:
Use HTTPS for consistency. Also for practical purposes, because Github denies access via SSH for those who don't have access rights.


cmnrd (approved):
Good catch!

PR #1310, opened 2022-07-29
Removal of error reporting markers from parameter types

Wonseo-C:
The string returned by <code>getTargetType</code> included error reporting markers, which it shouldn't.

Before this change:
image

After this change:
image

### DIFF
--- a/org.lflang/src/org/lflang/generator/TargetTypes.java
+++ b/org.lflang/src/org/lflang/generator/TargetTypes.java
@@ -163,7 +163,7 @@ public interface TargetTypes {
         } else if (type.isVariableSizeList) {
             return getTargetVariableSizeListType(type.baseType());
         }
-        return type.toText();
+        return type.toOriginalText();
     }
 
     /**
--- a/org.lflang/src/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
+++ b/org.lflang/src/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
@@ -126,7 +126,7 @@ class TSParameterPreambleGenerator(
             var customArgType: String? = null
             var customTypeLabel: String? = null
 
-            val paramType = getTargetType(parameter)
+            val paramType = parameter.type.id
             if (paramType == "string") {
                 mainParameters.add(parameter)
                 customArgType = "String";
--- a/org.lflang/src/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
+++ b/org.lflang/src/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
@@ -126,7 +126,7 @@ class TSParameterPreambleGenerator(
             var customArgType: String? = null
             var customTypeLabel: String? = null
 
-            val paramType = parameter.type.id
+            val paramType = getTargetType(parameter)
             if (paramType == "string") {
                 mainParameters.add(parameter)
                 customArgType = "String";


cmnrd (commented):
I don't know the specifics of the TS code generation, but simply using the type id does not seem like the right solution to me. What if the type is `time`? Or `time[]`? It seems that `getTargetType()` does not do exactly what is needed for TS here.

Wonseo-C:
@cmnrd @hokeun 
I tried debugging <code>getTargetType</code> and found that there is some bug.
The function returns <code>parameter.inferredType</code>'s <code>toText()</code>.
So, there is some bug with <code>toText()</code>'s return value.
image
Also, I found that when I used the <code>toOriginalText()</code>, I got the expected return value "number".
image
How can I fix this problem?

cmnrd:
@petervdonovan could you have a look at this? Is it save to just use `toOriginalText()` here? It looks like the code tagging is breaking things in TypeScript.

Wonseo-C:
> @petervdonovan could you have a look at this? Is it save to just use `toOriginalText()` here? It looks like the code tagging is breaking things in TypeScript.

I changed `getTargetType()`'s return value `toOriginalText()`

petervdonovan:
Sorry for missing @cmnrd's comment.

Since I don't know the full motivation behind this change, it is difficult for me to comment on whether it is necessary. It can cause slight regressions in the error reporting functionality of the language server, which needs these tags in order to report errors on the correct lines in VS Code.

A possible alternative would be to strip the `CodeMap` tags out downstream, as is done [here](https://github.com/lf-lang/lingua-franca/blob/be71d83d94b91d05304dbee864f04af033b9d953/org.lflang/src/org/lflang/generator/rust/RustModel.kt#L671) using `CodeMap.fromGeneratedCode(someToTextOutputWithTags).generatedCode`. This might not be the most elegant solution, but I think it would be the right choice if the result of `getTargetType` is only wrong for a few specific cases.

lhstrh:
@petervdonovan: the problem is that the type of the parameter instead of returning something like `string` or `number` would return a much longer string with additional markers in it used for error reporting. I think this is a legitimate bug, and the text referenced in the method should be the original one, not the modified one. I'm not quite sure how this fix would affect error reporting... 

Are you suggesting that we replace the line that says `type.toOriginalText()` with `CodeMap.fromGeneratedCode(type).generatedCode`? If you could make a concrete code suggestion in the `Files changed`, that would be great.

Wonseo-C:
> Are you suggesting that we replace the line that says `type.toOriginalText()` with `CodeMap.fromGeneratedCode(type).generatedCode`? If you could make a concrete code suggestion in the `Files changed`, that would be great.

@lhstrh Does this mean that we'll use `CodeMap.fromGeneratedCode(type).generatedCode` like under the screenshot?
I did debug like this, and then I got the value of `string\n`, not `string`.
스크린샷 2022-08-11 오후 2 15 53

petervdonovan:
Sorry for the late response...

> Are you suggesting that we replace the line that says type.toOriginalText() with CodeMap.fromGeneratedCode(type).generatedCode? If you could make a concrete code suggestion in the Files changed, that would be great.

That is not quite what I meant. If we did that, it would be equivalent to just using `toOriginalText`, except that it is more complicated because it puts the `CodeMap.Correspondence` tags in and then takes them out again.

I cannot suggest a change in the `Files changed` because my suggestion applied only to the code that originally motivated this PR. I do not know why you wanted to change `getTargetType`, but I guessed that it was because somewhere else in the code base, the `CodeMap.Correspondence` tags were causing problems. If the tags are causing problems elsewhere, then the change I suggested would be a workaround.

> I think this is a legitimate bug, and the text referenced in the method should be the original one, not the modified one.

I'm not sure why this would be the case. The philosophy behind my liberal use of these tags is that the primary purpose of the functions that convert `EObject`s to text is for code generation, and that all generated code copied verbatim from the source should have a mapping. To access the contents of the `EObject`, we can just directly call its getters, right? It strikes me as slightly hacky to use the functions related to code generation to stringify these objects, and then use regexes and string functions to analyze them -- sure, we do it, and I probably have done it too, but it isn't something we should aspire to because it is a heuristic-based approach that creates weird bugs and corner cases.

That said, this has been a lot of discussion over a relatively harmless change. I am happy to see this PR get merged if that's necessary in order to make progress.

lhstrh:
I guess the real issue is that the markers should associate not with a parameter type but with a declaration that uses it... I agree that it would be reasonable to merge this in order to make progress... This can always be revised at a later time.

lhstrh (approved):

PR #1312, opened 2022-08-01
Optimize access to sparse multiports in the C++ target

revol-xut:
# Multiport Optimization

Ports contained inside a multiport are now smart and will help the multiport to keep track of all the present ports.
This changes will increase performance for sparse multiports e.g. Big Benchmark. Previously a `std::vector` was used 
to store ports this is now replaced by `reactor::Multiport` class.

This changes are related to:
- https://github.com/lf-lang/reactor-cpp/pull/24


test/Cpp/src/multiport/BankSelfBroadcast.lf
```diff
@@ -16,10 +16,14 @@ reactor A(bank_index:size_t(0)) {
         out.set(bank_index);
     =}
     reaction(in) {=
+        auto activated_ports = in.active_ports_indices();
+
+        auto contained = [&](std::size_t value) {
```
cmnrd, line 21: What is the purpose of the changes in this file?

cmnrd (commented):
The changes here look good to me (I only had one question). But we should wait with merging this until https://github.com/lf-lang/reactor-cpp/pull/24 is merged.

lib/scripts/launch.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/lfc
```
cmnrd, line 1: that doesn't look right ;)

test/Cpp/src/multiport/BankSelfBroadcast.lf
```diff
@@ -18,8 +18,6 @@ reactor A(bank_index: size_t(0)) {
     reaction(in) {=
         for (size_t i = 0; i < in.size(); i++) {
             if (in[i].is_present()) {
-                std::cout << "Reactor " << bank_index << " received "
```
cmnrd, line 21: Is there a reason for deleting this?

test/Cpp/src/multiport/CallbackMultiport.lf
```diff
@@ -0,0 +1,61 @@
+// Test bank of reactors to multiport input with id parameter in the bank.
```
cmnrd, line 1: The name of the test seems outdated. Maybe PresentPorts.lf? Also the description isn't very clear

cmnrd (commented):
This looks good! I added a few nitpicks. Also it seems like you changed the refs of other submodules then reactor-cpp, which should not be needed.

cmnrd:
Could you also update the PR description?

org.lflang/src/org/lflang/generator/cpp/CppPortGenerator.kt
```diff
@@ -66,7 +66,7 @@ class CppPortGenerator(private val reactor: Reactor) {
             ${name}.reserve($width);
             for (size_t __lf_idx = 0; __lf_idx < $width; __lf_idx++) {
               std::string __lf_port_name = "${name}_" + std::to_string(__lf_idx);
-              ${name}.emplace_back(__lf_port_name, this);
+              ${name}.emplace_back(__lf_port_name, this, ${name}.get_active_ports(), __lf_idx);
```
cmnrd, line 69: Looks like this causes compile errors. Shouldn't this simply pass a pointer to the Multiport?

cmnrd (approved):
Glad that we can finally merge this!

PR #1318, opened 2022-08-03
Simplification of grammar that encodes mutations with the same productions as reactions

Wonseo-C:
This PR:
* Updates the Lingua Franca grammar to delete the Mutation class, and merge it with the Reaction class. (They have the similar properties `Trigger`, `Source`, `Effect`, `Code`)
* Updates the TypeScript code generator to uses a `WritablePort` for mutations


hokeun:
@lhstrh, I talked to @Wonseo-C and we found that [`Reaction`](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/LinguaFranca.xtext#L197) and [`Mutation`](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/LinguaFranca.xtext#L219) have a lot in common. Do you know if we can define a parent of these two objects in [`LinguaFranca.xtext`](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/LinguaFranca.xtext)? If we have a sort of "parent class" of [`Reaction`](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/LinguaFranca.xtext#L197) and [`Mutation`](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/LinguaFranca.xtext#L219), we will be able to write code a lot cleaner without repeating the same code. :) For example, we can define common functions used by procedures creating reactions and mutations.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -103,43 +107,66 @@ class TSReactionGenerator(
...
+            if (reaction.deadline != null) {
+                reactDeadline = generateDeadlineHandler(reaction, reactPrologue, reactEpilogue, reactSignature)
+            }
         }
+
         return with(PrependOperator) {
             """
             |
-            |this.addReaction(
+            |this.add${if (reaction.isMutation()) "mutation" else "reaction"}(
```
lhstrh, line 152:
```suggestion
            |this.add${if (reaction.isMutation()) "Mutation" else "Reaction"}(
```

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -180,6 +207,21 @@ class TSReactionGenerator(
         }
     }
 
+    private fun generateMutationSignatureElementForPortEffect(effect: VarRef): String {
```
lhstrh, line 210: Comments missing

org.lflang/src/org/lflang/scoping/LFScopeProviderImpl.java
```diff
@@ -257,7 +259,7 @@ protected IScope getScopeForVarRef(VarRef variable, EReference reference) {
     private RefType getRefType(VarRef variable) {
         if (variable.eContainer() instanceof Deadline) {
             return RefType.DEADLINE;
-        } else if (variable.eContainer() instanceof Reaction) {
+        } else if (variable.eContainer().getClass() == ReactionImpl.class) {
```
lhstrh, line 262:
```suggestion
        } else if (variable.eContainer() instanceof Reaction) {
```

org.lflang/src/org/lflang/scoping/LFScopeProviderImpl.java
```diff
@@ -266,7 +268,7 @@ private RefType getRefType(VarRef variable) {
             } else if (reaction.getEffects().contains(variable)) {
                 return RefType.EFFECT;
             }
-        } else if (variable.eContainer() instanceof Connection) {
+        } else if (variable.eContainer().getClass() == ConnectionImpl.class) {
```
lhstrh, line 271:
```suggestion
        } else if (variable.eContainer() instanceof Connection) {
```

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: I wrote the comments for this function in https://github.com/lf-lang/lingua-franca/pull/1318/commits/dbec99308314d3e9cb63d9f47ace6c83530d1bfc

Wonseo-C:
* I deleted the unnecessary code lines (https://github.com/lf-lang/lingua-franca/pull/1318/commits/fbec70a2776aa3fa7bbd7a4bef12533a689de94e, https://github.com/lf-lang/lingua-franca/pull/1318/commits/4d49227638e48405bb262cebf498baaf6081f0b8)
* Combined the function of generating Mutation and Reaction (https://github.com/lf-lang/lingua-franca/pull/1318/commits/e444eb44f94360cbf4b78e90570d2f0d0ce0ae84)
* Added comment for `generateMutationSignatureElementForPortEffect` (https://github.com/lf-lang/lingua-franca/pull/1318/commits/dbec99308314d3e9cb63d9f47ace6c83530d1bfc)

[Sieve.lf](https://github.com/lf-lang/benchmarks-lingua-franca/blob/main/TS/Savina/src/parallelism/Sieve.lf), the simple example file using the `mutation`, is built well. @lhstrh

org.lflang/src/org/lflang/generator/ts/TSImportPreambleGenerator.kt
```diff
@@ -55,7 +55,7 @@ class TSImportPreambleGenerator(
             |import {Action as __Action, Startup as __Startup, FederatePortAction as __FederatePortAction} from './core/action'
             |import {Bank as __Bank} from './core/bank'
             |import {FederatedApp as __FederatedApp} from './core/federation'
-            |import {InPort as __InPort, OutPort as __OutPort, Port as __Port} from './core/port'
+            |import {InPort as __InPort, OutPort as __OutPort, Port as __Port, WritablePort} from './core/port'
```
lhstrh, line 58:
```suggestion
            |import {InPort as __InPort, OutPort as __OutPort, Port as __Port, WritablePort as __WritablePort} from './core/port'
```

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -1,5 +1,6 @@
 package org.lflang.generator.ts
 
+import org.eclipse.emf.ecore.EObject
```
lhstrh, line 3:
```suggestion
```

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -103,43 +107,66 @@ class TSReactionGenerator(
...
-        for (trigger in reaction.triggers) {
-            if (trigger is VarRef) {
-                reactionTriggers.add(trigger.generateVarRef())
-            } else if (trigger is BuiltinTriggerRef) {
-                when (trigger.type) {
-                    BuiltinTrigger.STARTUP  -> reactionTriggers.add("this.startup")
-                    BuiltinTrigger.SHUTDOWN -> reactionTriggers.add("this.shutdown")
-                    else -> {}
+
+        if (reaction.isMutation()) {
```
lhstrh, line 118: Why are these cases handled separately? They look the same to me...

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -103,43 +107,66 @@ class TSReactionGenerator(
         reaction: Reaction,
         reactPrologue: String,
         reactEpilogue: String,
-        reactFunctArgs: StringJoiner,
+        reactFuncArgs: StringJoiner,
         reactSignature: StringJoiner
     ): String {
         // Assemble reaction triggers
+        var reactCode = ""
+        var reactDeadline = "}"
```
lhstrh, line 115: Why is the variable initialized with a closing curly bracket?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -333,8 +379,13 @@ class TSReactionGenerator(
                 reactSignatureElement += ": Sched<" + getActionType(effect.variable as Action) + ">"
                 schedActionSet.add(effect.variable as Action)
             } else if (effect.variable is Port){
-                reactSignatureElement += ": ${generateReactionSignatureElementForPortEffect(effect)}"
-                reactEpilogue.add(generateReactionEpilogueForPortEffect(effect))
+                if (reaction.isMutation()) {
```
lhstrh, line 382: Duplicate code...

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -449,6 +500,7 @@ class TSReactionGenerator(
         // If the app is federated, only generate
         // reactions that are contained by that federate
         val generatedReactions: List<Reaction>
+        val generatedMutations: List<Reaction>
```
lhstrh, line 503: Why is the variable called `generatedMutations`? I think just `mutations` is more accurate.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -464,11 +516,20 @@ class TSReactionGenerator(
...
         }
+        // get mutations
+        generatedMutations = reactor.reactions.stream().filter{it -> it.isMutation()}.collect(Collectors.toList())
 
         ///////////////////// Reaction generation begins /////////////////////
         // TODO(hokeun): Consider separating this out as a new class.
+        if (generatedMutations.isNotEmpty()) {
+            // Write the mutation itself
+            for (mutation in generatedMutations) {
+                reactionCodes.add(generateSingleReaction(reactor, mutation))
```
lhstrh, line 530: Does this method account for the fact that the given reaction is a mutation? If so, then why are we handling mutations and reactions separately at all?

lhstrh (changes requested):

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: I changed the variable name! https://github.com/lf-lang/lingua-franca/commit/b349e6010a667132c40f78a71128e268fecd00dc

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: I defined the `reactDeadline` as `"}"`, because when the deadline is null, then the deadline should be empty.
So, at [this line](https://github.com/lf-lang/lingua-franca/pull/1318/files#diff-5fdb14bc11c18f34a23a51db49bd847b0bc8b386cc889dff06cb825d2ab20cf5R119), I replaced the value of `reactDeadline` if the reaction's deadline isn't null.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: Yes, I deleted the duplicate code. https://github.com/lf-lang/lingua-franca/pull/1318/commits/0d49d6e37ef948cd22628394e533ffab613e524b

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: I deleted the duplicate code. https://github.com/lf-lang/lingua-franca/pull/1318/commits/d58c06655ed025e9537ca8b9dbf5c5af04316984

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
lhstrh: I think the old line `139` is actually much more self-explanatory. I.e.: 
```
${" |    "..if (reaction.deadline != null) generateDeadlineHandler(reaction, reactPrologue, reactEpilogue, reactSignature) else "}"}
```

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -103,11 +107,20 @@ class TSReactionGenerator(
...
-        reactFunctArgs: StringJoiner,
+        reactFuncArgs: StringJoiner,
         reactSignature: StringJoiner
     ): String {
         // Assemble reaction triggers
+        var reactCode = ""
+        var reactDeadline = "}"
         val reactionTriggers = StringJoiner(",\n")
+
+        if (!reaction.isMutation()) {
```
lhstrh, line 118: Can mutations not have deadlines? I think this can all go away and we can just reinstate the original line `139`.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -119,27 +132,29 @@ class TSReactionGenerator(
                 }
             }
         }
+        reactCode = reaction.code.toText()
```
lhstrh, line 135: This variable also seems unnecessary...

Wonseo-C: Thank you. I deleted the unnecessary variable! https://github.com/lf-lang/lingua-franca/pull/1318/commits/f327feef4c499641d83833eb1fd3591f20c35834

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: You're right. I deleted this part. https://github.com/lf-lang/lingua-franca/pull/1318/commits/f327feef4c499641d83833eb1fd3591f20c35834

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: Thank you for your nice code suggestion! I reflected your comment in [f327fee](https://github.com/lf-lang/lingua-franca/commit/f327feef4c499641d83833eb1fd3591f20c35834).

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -119,27 +124,28 @@ class TSReactionGenerator(
...
             |
-            |this.addReaction(
+            |this.add${if (reaction.isMutation()) "Mutation" else "Reaction"}(
             |    new __Triggers($reactionTriggers),
-            |    new __Args($reactFunctArgs),
+            |    new __Args($reactFuncArgs),
             |    function ($reactSignature) {
             |        // =============== START react prologue
-        ${" |        "..reactPrologue}
+            ${" |        "..reactPrologue}
```
lhstrh, line 136: The indentation here is weird. Same on line 139, 142, 145, etc.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
```diff
@@ -180,6 +186,25 @@ class TSReactionGenerator(
         }
     }
 
+    /**
+     * Generate the mutation's effect type as "WritablePort,"
+     * which differs from the reaction's effect type.
+     */
+    private fun generateMutationSignatureElementForPortEffect(effect: VarRef): String {
```
lhstrh, line 193: I realize that you had to make this distinction because we set the data type up this way in reactor-ts, but it wasn't a very well-though-out decision on my part to implement it that way. It doesn't hurt reactions to be able to inspect the port, so I suggest that instead augment `ReadWrite` with `getPort()` and `MultiReadWrite` with `getPorts` and remove the distinction between reactions and mutations. WDYT?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
lhstrh: My question is: why do we gather mutations separately from reactions if for both we end up calling `generateSingleReaction`, anyway?

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: @lhstrh I think it is an excellent idea!
But I have a question; is the `WritablePort` not necessary? If `WritablePort` is deleted, then there will be many problems with implementation. The reason is the WritablePort is class, and the ReadWrite is type.

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: Yes, you're right. I was mistaken that the `Reaction` and `Mutation`  were separated.
I deleted this part. https://github.com/lf-lang/lingua-franca/pull/1318/commits/61bb0286a4433f3d8c16efa3ed4ba017c7d307b0

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: https://github.com/lf-lang/lingua-franca/pull/1318/commits/e07dc3972dac5adfd4a7e864607f6f746cb6f049

org.lflang/src/org/lflang/generator/ts/TSReactionGenerator.kt
Wonseo-C: I tried to fix it, but there are some problems.
If we add the `getPort()` in the `ReadWrite`, we should get the `IOPort`. So, the `ReadWrite`'s type should be `Present`. It caused a type error. (The `ReadWrite` is no class)
I think that we can use the `asWritable()`, which is the `IOPort`'s function of returning the `writablePort`.

https://github.com/Wonseo-C/reactor-ts/tree/getport : This is the branch where I worked.

Wonseo-C: @lhstrh :
By discussing with @hokeun, we guessed your purpose of this review is to make the simple code. And we decided to implement it like this -> https://github.com/lf-lang/lingua-franca/pull/1318/commits/74ad4094bc837200e4667eb3864f78b8b9aa326b.
We thought your suggestion could cause another problem because of the type error (detail explained above), so I just modified the `generateReactionSignatureElementForPortEffect` to get different strings if it is a mutation or not.

lhstrh (approved):
Looks good! Thanks for these fixes!

hokeun:
Great work @Wonseo-C ! Thanks for the review @lhstrh !

PR #1319, opened 2022-08-04
Added 'TagString' test

patilatharva:
This test ensures tag objects have the string representation: Tag(time=$time, microstep=$microstep)

### DIFF
--- a/test/Python/src/TagString.lf
+++ b/test/Python/src/TagString.lf
@@ -1,0 +1,35 @@
+// Tests the get_microstep() function in the python target.
+target Python {
+    fast: false
+};
+main reactor TagString {
+    preamble {=
+        import sys
+    =}
+    state s(1);
+
+    // timer t(0, 1 msec);
+    logical action l;
+    reaction(startup) -> l {=
+        l.schedule(0);
+    =}
+
+    reaction(l) -> l {=
+        tag_str = str(lf.tag())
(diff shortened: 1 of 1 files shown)


Soroosh129:
Hi @patilatharva 

It looks like the test you've added is filing with the following error:
```Python
expected tag string representation: Tag(time=$time, microstep=$microstep), instead got: {tag_str}
```
Do you know why that might be?

patilatharva:
This seems to be because the python target submodule wasn't updated. The git history is a bit messy now but I just updated so the test should pass. Also, thanks to your comment I noticed a formatting error in the error message, which I fixed as well.

Soroosh129:
> This seems to be because the python target submodule wasn't updated. 

Ah I see.

> The git history is a bit messy now.

No worries. We an squash the commits. Thanks for adding this feature!

lhstrh:
@patilatharva: I looks like this PR has been under the radar for a while. Shall we bring it back into a state to where we can merge it? One issue with this PR is that I don't quite understand what it attempts to do, so if you could update the description to clarify, that would be a great first step. Thanks, and sorry for the delayed handling of this!

patilatharva:
Yes, thanks for bringing this back to my attention! I'll make it ready for review.

petervdonovan (approved):

PR #1323, opened 2022-08-11
Reduced verbosity in reported output

petervdonovan:
These messages should be kept out of the "output" that appears in VS Code.


lhstrh (approved):
LGTM!

PR #1324, opened 2022-08-12
Fix layout for error messsage node

a-sr:
Due to a change in the default layout direction in ELK, the error message node for detected cycles was place left of the reactor instead of top. This PR fixes that by making the assumed layout options explicit.


edwardalee (approved):
LGTM!

PR #1338, opened 2022-08-30
Use Node version 18 in CI

petervdonovan:
We should probably decide exactly what Node version(s) we support (e.g., [here](https://github.com/lf-lang/vscode-lingua-franca/blob/051579e8ea4752c1c095e779e78ce02099406722/src/config.ts#L34)), but that can be done later.

Probably the only lines of code that _actually_ caused the problem (of not recognizing optional chaining syntax) were the ones in reactor-ts that explicitly pinned our Node version to version 10.


lhstrh (approved):
:rocket:

PR #1341, opened 2022-08-31
Improve model order layout configuration

soerendomroes:
Full Control mode sometimes produces bad drawings that cannot be fully controlled since edges are not explicit in LF and many dummy nodes are created by fixed port sides.

LF needs a model order strategy that is strict during crossing minimization, which requires to set a model order for all elements (also actions), and less strict during crossing minimization.

fullControlNeedsGreedySwitch
Here it would be better to activate the greedy switch heuristic to get a better drawing.
fullControlNeedsGreedySwitch-OYES
Here it is also not possible to route the edge below the nodes, but at least no crossings are created.

Example model:
```
target C

main reactor {
    reaction (startup) -> a {= =}
    logical action a:char*;
    reaction (a) -> b {= =}
    physical action b:char*;
    reaction (b) -> a {= =}
    
}
```

### DIFF
--- a/org.lflang.diagram/src/org/lflang/diagram/synthesis/util/LayoutPostProcessing.java
+++ b/org.lflang.diagram/src/org/lflang/diagram/synthesis/util/LayoutPostProcessing.java
@@ -65,6 +65,10 @@ public class LayoutPostProcessing extends AbstractSynthesisExtensions {
     /** Reactions and reactors are ordered by their model order if no additional crossing are created. */
     public static final String TIE_BREAKER = "Optimize Crossings";
     /**
+     * As FULL_CONTROL but does one sided greedy switch to eliminate crossings.
+     */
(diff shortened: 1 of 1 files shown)


soerendomroes:
This might have some problem, I will leave as a draft for now.

lhstrh:
This PR as been marked as draft, citing a potential problem. Are you sure this isn't ready for review?

soerendomroes:
I did not have time yet to address the feedback from the last meeting I attended and the added option does not work for all models.

lhstrh:
OK, no problem. Just checking in :-) This part of the code base is less likely to accumulate conflicts, so it's OK if this needs to wait a bit.

lhstrh:
Hi @soerendomroes, should we keep this PR open?

soerendomroes:
Ill reopen it once I have time.

PR #1348, opened 2022-09-03
Safe handling of physical actions in single-threaded C runtime

lhstrh:
Prior to these changes, it was not possible to asynchronously access the event queue from an interrupt service routine on bare-iron platforms. This PR -- currently still a work in progress -- should enable the use of physical actions on embedded platforms.
Known issues:
 - [ ] the compiler currently forces the the use of the threaded runtime if there are physical actions; this behavior needs to be changed, but it's not clear how exactly. Not all embedded platforms are single-threaded.
 - ...


petervdonovan:
The tracing tests are failing on Windows for both Python and C with ERROR_ACCESS_DENIED, which seems to just be the same as a segmentation fault, and I cannot reproduce the error locally, even on Windows.

lhstrh:
That is rather bizarre... 😔

lhstrh:
It sounds like a race condition. I believe that if tracing is enabled the threaded runtime is always used...

edwardalee:
See https://github.com/lf-lang/reactor-c/pull/106#pullrequestreview-1095541919

erlingrj:
@This is ready for merge. Arduino is still supported as a target platform. I think you removed the support in reactor-c, so should we remove it here as well?

lhstrh:
Not sure about Arduino. It kind of depends on what @arengarajan99 is up to, but I haven't heard from him in a while.

lhstrh (commented):
Looks great, @erlingrj! Thanks for much for whipping this into shape.

Issue #1345, opened 2022-09-02
Modal validation is incorrect


edwardalee:
With apologies for not reducing to a minimal example, the attached `src/Robot_Solution_724.lf` has an incorrect validation error. It gives an error message "This reactor contains state variables that are not reset upon  mode entry." But the highlighted reactor has no state variables at all. More oddly, removing the *next* instantiated reactor eliminates the error. The next instantiated reactor does have state variables, but those are *all* marked reset.
[bug.tar.gz](https://github.com/lf-lang/lingua-franca/files/9474810/bug.tar.gz)

a-sr:
I am not able to get the model from the archive. After unpacking the `bug` archive the `Robot_Solution_724.lf` is not a text file but an archive itself that seems to be empty.

edwardalee:
Oops, apologies.  Here is a new archive.
[bug.tar.gz](https://github.com/lf-lang/lingua-franca/files/9476248/bug.tar.gz)

a-sr:
Thanks. I was able to take a look at the model. Unfortunately, I am unable to reproduce the error.
The only validation error that I get is in line 76 (`aconverter.angle -> robot.angle;`) where the types do not match.
The modes do not raise any errors but contrary to the initial description none has inner reactor instantiations that could cause the bug. I was also unable to find the _next_ instantiation that might be involved in the problem.
Could it be possible that this version has _next_ removed to eliminate the error? If that the case, which reactor and where do I have to instantiate to cause the error?

a-sr:
Independent from this specific problem, I decided to improve the error messaging for this validation rule in PR #1349. Since we should not have to speculate which state variables the validator is referring to, it will now list the involved state variables.

edwardalee:
Strange. I'm now unable to reproduce the error also... Closing the issue.

PR #1349, opened 2022-09-03
Improve error messaging for mode validation

a-sr:
The validation error for state variables without reset now exposes the problematic state variables and their reactor.

This should also help narrowing down the problem in #1345


edwardalee (approved):
Nice!  Thanks for dealing with this.

PR #1355, opened 2022-09-06
Add target property for specifying additional ROS dependencies

cmnrd:


Rodario (approved):
Tested it with a pub/sub setup that uses a custom ROS2 message package and it works. 
It should be noted that LF does not build your additional packages for you. That needs to be done prior to compiling with lfc.

lhstrh:
What's the goal of this PR? A description would be nice :-)

PR #1356, opened 2022-09-06
Reduction of output in formatter tests

### LINKED ISSUE #1354: Formatter test logs are too verbose
While looking for a test failure in the CI log I found tens of thousands of lines of code printed by the formatter: https://github.com/lf-lang/lingua-franca/runs/8206312377?check_suite_focus=true#step:4:24332

Looks like the CI logs contain all the test sources of the repo... This is too much for a browser tab, and the build failure is hidden

It starts with the following lines:
```
RoundTripTests > roundTripTest() STANDARD_OUT
    Indexing...
```
I don't see a RoundTripTests file in the repo though. Do you know where this is coming from and how to stop it?

petervdonovan:
Fixes #1354.

This silences the formatter test except when there is an assertion failure, in which case the relevant information is printed.

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/compiler/RoundTripTests.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/RoundTripTests.java
@@ -46,31 +46,30 @@ public class RoundTripTests {
 
     private void run(Path file) throws Exception {
         Model originalModel = parse(file);
-        System.out.printf("Running formatter on %s...%n", file);
         Assertions.assertTrue(originalModel.eResource().getErrors().isEmpty());
         // TODO: Check that the output is a fixed point
         final int smallLineLength = 20;
-        String squishedTestCase = FormattingUtils.render(originalModel, smallLineLength);
-        System.out.printf(
-            "The following is what %s looks like after applying formatting with the preferred line "
-                + "length set to %d columns:%n%s%n%n",
-            file.getFileName(),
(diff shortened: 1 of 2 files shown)


org.lflang.tests/src/org/lflang/tests/compiler/RoundTripTests.java
```diff
@@ -46,31 +46,30 @@ public void roundTripTest() throws Exception {
...
+            System.out.printf(
+                "The following is what %s looks like after applying formatting with the preferred line "
+                    + "length set to %d columns:%n%s%n%n",
+                file.getFileName(),
+                smallLineLength,
+                squishedTestCase
+            );
+            throw new junit.framework.AssertionFailedError(String.format(
+                "The reformatted version of %s was not equivalent to the original file.",
+                file.getFileName()
```
lhstrh, line 69: I noticed that the error reporting tests do not report the path to the offending file, like this line does. We should add it there, too.

lhstrh (approved):
Thanks for the fix!

org.lflang.tests/src/org/lflang/tests/compiler/RoundTripTests.java
petervdonovan: You are right. Done!

oowekyala:
Thanks Peter!

PR #1357, opened 2022-09-07
Bump Rust runtime

jhaye:
https://github.com/lf-lang/reactor-rs/pull/28 was a major fix that gets rid of most of our problems with the runtime.


lhstrh (approved):
:rocket: :rocket: :rocket:

PR #1358, opened 2022-09-07
Only record tests in test/*/src

oowekyala:
The current implementation also walked into src-gen. This causes problems with my setup because the Rust runtime repo has some LF files, and since it is copied into src-gen, these were registered as tests and caused failures eg of the formatter tests.


cmnrd (approved):
LGTM! But I am also tagging Marten since he knows the tests better

lhstrh (approved):
LGTM! Thanks for the fix :-)

Issue #1321, opened 2022-08-07
Surprising behavior involving a reactor that splits a multiport into individual ports


petervdonovan:
I'm labeling this as a question because it's possible this stems from a misunderstanding on my part of the semantics of multiports. I had difficulty getting any programs similar to these to work.

### IndexIntoMultiportOutput

This program compiles successfully and executes with the C target, but its behavior is inconsistent with my expectations. I expect it to print "1 = 1" and "2 = 2", but instead, it prints "1 = 0" and "2 = 1". This is because instead of connecting the `i`th reaction to the `i`th multiport channel, it fills reactions into the multiport channels starting at channel 0. I find this kind of behavior strange and hard to reason about. The way messages are routed inside `MultiportSplitter` should not be affected by how things are connected outside of it.

![Screenshot from 2022-08-06 18-39-01](https://user-images.githubusercontent.com/33707478/183271394-ea9586f2-2487-4b90-87ba-60f05e114c4e.png)

This is the output:
```
---- Start execution at time Sat Aug  6 18:37:04 2022
---- plus 231711826 nanoseconds.
---- Using 1 workers.
1 = 0
2 = 1
---- Elapsed logical time (in nsec): 0
---- Elapsed physical time (in nsec): 208,243
```

This is the program:
```
target C

reactor ReactorWithMultiport(width: int(3)) {
    output[width] out: int

    reaction(startup) -> out {=
        for (int i = 0; i < self->width; i++) {
            lf_set(out[i], i);
        }
    =}
}

reactor MultiportSplitter(width: int(3)) {
    input[width] in: int

    output out0: int
    output out1: int
    output out2: int

    in -> out0, out1, out2
}

main reactor IndexIntoMultiportOutput {
    source = new ReactorWithMultiport()
    splitter = new MultiportSplitter()

    source.out -> splitter.in

    reaction(splitter.out1) {=
        printf("1 = %d\n", splitter.out1->value);
    =}
    reaction(splitter.out2) {=
        printf("2 = %d\n", splitter.out2->value);
    =}
}
```

---

### IndexIntoMultiportInput

This program fails with the C++ target, I think during the assemble phase of execution, and I do not understand why.
![index-into-multiport-input](https://user-images.githubusercontent.com/33707478/183271234-a5b7f9e6-ca82-4bdb-b4a3-768160721b97.png)
This is the output of the program:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  unexpected case
[...]

PR #1361, opened 2022-09-08
Support passthrough connections in the C++ target

cmnrd:
This pulls in https://github.com/lf-lang/reactor-cpp/pull/29 and addresses the problem in the C++ target described in https://github.com/lf-lang/lingua-franca/issues/1321. It also adds the two programs written by @petervdonovan as tests.

### DIFF
--- a/test/Cpp/src/multiport/IndexIntoMultiportInput.lf
+++ b/test/Cpp/src/multiport/IndexIntoMultiportInput.lf
@@ -38,15 +38,9 @@ main reactor IndexIntoMultiportInput {
 
     splitter.out -> receiver.in
 
-    reaction(startup) -> splitter.in0 {=
-        splitter.in0.set(0);
-    =}
+    reaction(startup) -> splitter.in0 {= splitter.in0.set(0); =}
 
-    reaction(startup) -> splitter.in1 {=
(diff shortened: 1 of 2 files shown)


petervdonovan (approved):
This looks right to me. Thanks!

test/Cpp/src/multiport/IndexIntoMultiportInput.lf
```diff
@@ -0,0 +1,46 @@
+target Cpp
+
```
lhstrh, line 2: I think it's a good practice to add a description to the test so that it's clear to others what the point of the test is.

test/Cpp/src/multiport/IndexIntoMultiportOutput.lf
```diff
@@ -0,0 +1,64 @@
...
+reactor ReactorWithMultiport(width: size_t(3)) {
+    output[width] out: int
+
+    reaction(startup) -> out {=
+        for (size_t i{0}; i < width; i++) {
+            out[i].set(i);
+        }
+    =}
+}
+
```
lhstrh, line 12: Idem

lhstrh (approved):
Looks great, modulo comments.

Issue #1363, opened 2022-09-14
java.lang.SecurityException: class "org.eclipse.core.runtime.OperationCanceledException"'s signer information does not match signer information of other classes in the same package


petervdonovan:
[This PR run](https://github.com/lf-lang/reactor-c/actions/runs/3048968494) is unrelated to our Java sources or build system. It adds a Markdown file in `reactor-c`. It fails with a `java.lang.SecurityException`:

```
java.lang.SecurityException: class "org.eclipse.core.runtime.OperationCanceledException"'s signer information does not match signer information of other classes in the same package
                                    at java.base/java.lang.ClassLoader.checkCerts(ClassLoader.java:1158)
                                    at java.base/java.lang.ClassLoader.preDefineClass(ClassLoader.java:902)
                                    at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1010)
                                    at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
                                    at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
                                    at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
                                    at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
                                    at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
                                    at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
                                    at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
                                    at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
                                    at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3373)
                                    at java.base/java.lang.Class.getDeclaredConstructors(Class.java:2555)
                                    at com.google.inject.spi.InjectionPoint.forConstructorOf(InjectionPoint.java:292)
                                    at com.google.inject.internal.ConstructorBindingImpl.create(ConstructorBindingImpl.java:117)
                                    at com.google.inject.internal.InjectorImpl.createUninitializedBinding(InjectorImpl.java:716)
                                    at com.google.inject.internal.InjectorImpl.createJustInTimeBinding(InjectorImpl.java:941)
[...]

PR #1365, opened 2022-09-15
Removal of unnecessary build dependency to address security exceptions

cmnrd:
This fixes https://github.com/lf-lang/lingua-franca/issues/1363 (at least locally on my machine, let's wait for CI). I think the issue is that `org.eclipse.core.runtime` had a conflicting version. Interestingly, it seems to work fine if we just remove our own version specification and let gradle resolve the dependency from xtext.

I am not sure if we would also need to make a similar change in the Maven configuration.

This fixes https://github.com/lf-lang/lingua-franca/issues/1363

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,6 @@ lsp4jVersion=0.14.0
 mwe2LaunchVersion=2.12.2
 openTest4jVersion=1.2.0
 resourcesVersion=3.16.0
-runtimeVersion=3.25.0
 shadowJarVersion=7.1.2
 xtextGradleVersion=3.0.0
 xtextVersion=2.27.0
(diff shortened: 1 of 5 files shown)


petervdonovan (approved):
`TimeLimit.lf` timed out for Python, but I'm inclined to ignore it. It is possible that we do not need that test anymore.

`Count3Modes.lf` timed out for C. That test has some delays in it. I suspect that this timeout can also be safely ignored.

I do not understand why this works, but I'm thankful that it does. It fixes things on my machine as well.

cmnrd:
I also don't fully understand why this works. It was more like a lucky shot :) I read [hear](https://www.eclipse.org/eclipse/platform-core/) that `org.eclipse.core.runtime` is supplanted by Equinox and just tried to remove it.

lhstrh:
:man_shrugging:

PR #1366, opened 2022-09-15
Add test for #763

### LINKED ISSUE #763: Rust main parameters need to be literals for CLI to compile
If you build a reactor with CLI generation enabled, the main parameter default values need to be literals. This is because the default values are used by the CLI generator and are actually directly copy-pasted as a string as the default value for the help text. This string is then parsed by the CLI parser at runtime, and of course not interpreted as a Rust expression. We use this CLI generation framework that translates attributes to calls to the programmatic API to build the CLI parser. I don't see a way to fix this without switching to using the programmatic API of clap directly.

I found this out because I couldn't write `1u64 << 60`.
```rust
target Rust {
  cargo-features: [ "cli" ],
};
main reactor (param: u32(1u64 << 60)) { }
```

oowekyala:
We can now close #763 

Note that this contains #1365 as that fixes my build

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,6 @@ lsp4jVersion=0.14.0
 mwe2LaunchVersion=2.12.2
 openTest4jVersion=1.2.0
 resourcesVersion=3.16.0
-runtimeVersion=3.25.0
 shadowJarVersion=7.1.2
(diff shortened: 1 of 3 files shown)


lhstrh (approved):
:rocket:

PR #1367, opened 2022-09-15
Support scheduling physical actions synchronously

oowekyala:
This is needed for a benchmark (Philosophers I think?)


lhstrh (approved):
:rocket: :rocket: :rocket:

lhstrh:
Looks like there is still a problem with https://github.com/lf-lang/reactor-rs/pull/30 and the submodule is also not updated?

jhaye (approved):
I [fixed Sleeping Barber](https://github.com/lf-lang/benchmarks-lingua-franca/pull/46) with these changes and it works well.

PR #1370, opened 2022-09-16
Adjustment of code generator to address #1368

### LINKED ISSUE #1368: Memory error with reactions writing to multiple ports
The following program fails with either `FATAL ERROR: Out of memory!` or a segmentation fault (depending on insertion of printfs):
```
target C;
reactor C {
    output x:int;
    output z:int;
    timer t(0, 100 msec);
    reaction(t) -> x, z {=
        lf_set(x, 42);
        lf_set(z, 44);
    =}
}
main reactor {
    c = new C();
    reaction(c.z) {=
        lf_print("c.z = %d", c.z->value);
    =}
}
```
image

lsk567:
Fixes #1368.


test/C/src/MultipleOutputs.lf
```diff
@@ -0,0 +1,22 @@
...
+        lf_set(x, 42);
+        lf_set(z, 44);
+    =}
+}
+main reactor {
+    c = new C();
+    reaction(c.z) {=
+        lf_print("c.z = %d", c.z->value);
+    =}
+}
```
lhstrh, line 22:
```suggestion
}

```

lhstrh (approved):
Awesome, thanks for tracking this down! You might want to run the formatter on the test to make sure it is formatted correctly and avoid a test failure.

lhstrh:
Note: I ran `./gradlew runLff --args "test/C/src/MultipleOutputs.lf"`

edwardalee (approved):
Great catch!  To be a test, the MultiportOutputs.lf needs to call `lf_print_error_and_exit` if the reaction to to `c.z` is not invoked. That can be handled in a shutdown reaction that checks a state variable.

Issue #1173, opened 2022-05-16
Epoch IDE fails to launch on MacOS


Shrajan:
I followed the instructions mentioned on the official page (https://www.lf-lang.org/docs/handbook/epoch-ide) to install Epoch IDE on a MacBook Pro M1. I get the following error when I try to launch the application with version 0.2.0 as well as the nightly build.

```
The JVM shared library "/Library/Java/JavaVirtualMachines/jdk-17.0.3.1.jdk/Contents/Home/bin/../lib/server/libjvm.dylib"
does not contain the JNI_CreateJavaVM symbol.
```

I followed the steps mentioend here https://stackoverflow.com/a/33477302/2919326, and modified the `Info.plist` file to set the correct Java path. Unfortunately, the problem still persists. 

Please let me know how to solve this. Thank you very much.

Shrajan:
@lhstrh Thank you for the quick reply.

$JAVA_HOME:
```
echo $JAVA_HOME
/Library/Java/JavaVirtualMachines/jdk-17.0.3.1.jdk/Contents/Home
```

`Info.plist` Java Version matches `$JAVA_HOME`: `jdk-17.0.3.1.jdk`
```
<string>-vm</string><string>/Library/Java/JavaVirtualMachines/jdk-17.0.3.1.jdk/Contents/Home/bin/java</string>
```

Output of `/usr/libexec/java_home -V`:
```
Matching Java Virtual Machines (2):
    18.0.1 (arm64) "Azul Systems, Inc." - "Zulu 18.30.11" /Library/Java/JavaVirtualMachines/zulu-18.jdk/Contents/Home
    17.0.3.1 (arm64) "Oracle Corporation" - "Java SE 17.0.3.1" /Library/Java/JavaVirtualMachines/jdk-17.0.3.1.jdk/Contents/Home
/Library/Java/JavaVirtualMachines/zulu-18.jdk/Contents/Home
```

lhstrh:
I don't see any red flags here, but there's also no way for me to replicate because I don't have a MacBook. Tagging @Soroosh129 who is also running on Apple silicon.

lhstrh:
Just bumping this thread to let everyone know this is on our radar and we're working on it.

Note to self: this is the process we need to automate in CI: https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution#3087722

lhstrh:
[...]

Issue #1276, opened 2022-07-06
Epic: Simplify the structure of the core repo and factor out Epoch in its own repo


cmnrd:
This epic specifies the new envisioned and simplified project structure as well as concrete incremental steps we can take towards this goal.

### Overview

Our current is very bloated and it is hard for newcomers to find their way through the forest of packages and directories. This structure has been been mostly dictated by Xtext and Eclipse. However, since we are moving away from Eclipse as our primary development IDE and Epoch is only one of our supported LF IDEs, we should aim at freeing our core repository from the constraints and overhead that Eclipse imposes. The best way to do this, is to separate our "core" repository from Epoch. Ideally, we strip the lingua-franca repository from all unnecessary Eclipse dependencies and Eclipse related code. This should instead be hosted in a separte Epoch repository where lingua-franca is a submodule.

### The Goal

- The lingua-franca repository should look something like this: 
  ```Text
  lingua-franca/
  ├── bin/
  ├── lib/
  ├── src/
  ├── test/
  ```
  Of course we will need some additional files and directories for gradle and other meta files, but the idea is to have a single point of entry for the source code. All code that is needed for our command line tools and the language server, should be located within `src/`. This includes the contents of `org.lflang`, `org.lflang.lfc`, `org.lflang.diagrams` and `org.lflang.lds`. The primary build tool for this repository would be gradle and any IDE could be used for development
- The Epoch repository would look similar to the current lingua franca repository. Most notably, `org.lflang` would become a submodule and point to the lingua-franca repository. The primary build tool for this repository would be maven and Eclipse the main IDE for development.
- Since the restructuring will be a breaking change, we also want to use the opportunity to fix #1078 and #423. The goal is to clean our git history and push the cleaned history to a new main branch. This way we can keep the old master branch for legacy, but start developing on main in our freshly cleaned up repo.

### Action Items
- [x] Bundle the core functionality of LF in `org.lflang`
  - [x] #1421
  - [x] Move contents of `org.lflang.diagram` to a subpackage `diagram` within `org.lflang`
  - [x] Build  language server with Gradle which depends on publication of KlighD package
  - [x] Remove `org.lflang.lds`because no longer needed
  - [x] Remove "Eclipse switch" from `mwe2` file in `lingua-franca` repo
  - [x] #1800
- [x] https://github.com/lf-lang/lingua-franca/issues/1491 
[...]

PR #1371, opened 2022-09-17
Removal of PyDev and Kotlin plugin dependencies from Epoch

lhstrh:
Attempt to address code signing issues discussed in https://github.com/lf-lang/lingua-franca/issues/1173


lhstrh:
@a-sr: having removed the Kotlin plugin is causing compile errors. Could we just introduce the dependency through Maven rather than Tycho?

lhstrh:
Also tagging @oowekyala for help on this.

a-sr:
> @a-sr: removal of the Kotlin plugin is causing compile errors in the Maven build. Could we just introduce the dependency through Maven rather than Tycho?

I am not sure if that is possible because Kotlin is a runtime dependency and Eclipse only accepts plugins in its runtime. We could try to pull the dependency via maven central and then put it on the plugin's classpath but I have the feeling that we already tried that in the past and it was problematic. One thing I am certain of is that it will complicate the build process in the Eclipse IDE because Eclipse does not like to mix maven and eclipse dependencies in the same project.
I hope the restructuring in #1276 will resolve the barriers in the future. However, the structure described there is not yet possible. I have to change the build process for the diagrams and the LDS first, and also have to find a workaround to get it working in an Eclipse plugin without problems on the classpath. It is definitely on my short-term Todo list but I currently do not have the time to tackle this.

lhstrh:
OK, thanks for looking into this, @a-sr. It sounds like there is no way to properly support Epoch on macOS then, because code signing won't work if the Kotlin plugin is included, and building Epoch won't work without the Kotlin plugin.

lhstrh:
Confirmed that this doesn't impair the code generators that are written in Kotlin.

Issue #1052, opened 2022-03-17
Add style check to build process


housengw:
Seems like we could add spotless (https://github.com/diffplug/spotless) to our build process to perform automatic style checks.
Works with Java, Kotlin and many other languages,

cmnrd:
I think this is a great good idea! However, more important than style checking in CI, is that we agree on a common workflow and style. Currently, it seems everyone uses different tooling and we have constant refactoring of code going back and force between different styles. Looks like spotless could also help solving this issue, as it integrates with various IDEs (but apparently not Eclipse...).

petervdonovan:
I just thought I would take the opportunity to share something neat that I found. [This method in ASTUtils](https://github.com/lf-lang/lingua-franca/blob/1ae23fee64ac401d5e72eacccdbde267f09e2ec4/org.lflang/src/org/lflang/ASTUtils.java#L1519) has 88 lines with a cyclomatic complexity of close to 20 (I think), which is about double [the score that seems to be conventionally used as an upper limit](https://docs.microsoft.com/en-us/visualstudio/code-quality/code-metrics-cyclomatic-complexity?view=vs-2022#the-magic-number).

If LF semantics were inherently that complicated, we would probably need to have simpler semantics. However, it seems more likely that we just need CI to automatically reject methods with a cyclomatic complexity greater than 10.

petervdonovan:
Here is the output of a few formatters that integrate with Spotless. There are several other options, but these are perhaps the most similar to our current style. This is relevant to #478 because if we choose an automatic formatter, it might not be compatible with a customized set of project-specific formatting rules. For example, the only significant customization that is possible with `googleJavaFormat` is choosing between 2 and 4 spaces of indentation.

It would be great to be able to just pick an automatic formatter (or else close this and #478 and let everyone format files however they want) so that we can liberate humans from drudgery, reduce bikeshedding during code reviews etc., and avoid turning style into a years-long project

- [googleJavaFormat-aosp.txt](https://github.com/lf-lang/lingua-franca/files/9610959/googleJavaFormat-aosp.txt)
- [clangFormat-LLVM.txt](https://github.com/lf-lang/lingua-franca/files/9610956/clangFormat-LLVM.txt)
- [clangFormat-Chromium.txt](https://github.com/lf-lang/lingua-franca/files/9610958/clangFormat-Chromium.txt)

The Gradle configuration used to try out this formatting appears in #1374.

cmnrd:
[...]

PR #1374, opened 2022-09-20
Add Spotless Gradle plugin

petervdonovan:
I have gotten feedback from multiple people that the requirement that LF tests be formatted has the potential to be annoying if it is not documented properly (and perhaps automated or hooked into some established workflow).

Unfortunately, using client-side Git hooks for the formatter seems like it would require each person to manually add the Git hook (for some reason, client-side Git hooks apparently cannot be tracked). If we were to use a server-side Git hook, then people would find out about formatting problems after they have already pushed their work, which can be annoying.

Because there has already been plenty of talk about adding Spotless to our regular workflow (#1052), it seems to me like the overhead of dealing with formatting tools can be minimized by just using the Spotless Gradle plugin for both LF formatting and Java formatting.

If we cannot quickly reach consensus about Java formatting, I propose to comment Java formatting out of this PR so that we at least have an established way of invoking `lff`. Otherwise, people would have to get used to directly invoking the `runLff` task, and then when we have a Java formatter or linter, they would have to get used to that again. Additionally, Spotless has advantages over direct invocation of the formatter. For example, it only runs the formatter on files that have changed.

Development of a comprehensive `CONTRIBUTING.md` file and building of consensus around coding standards have already been started in #478; therefore, they are both non-goals of this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,6 @@
+import com.diffplug.gradle.spotless.SpotlessTask
+import lfformat.LfFormatStep
+
 buildscript {
     repositories {
         mavenCentral()
@@ -12,13 +15,14 @@ plugins {
     id "com.github.johnrengelman.shadow" version "${shadowJarVersion}"
     id 'java'
     id 'jacoco'
+    id "com.diffplug.spotless" version "6.11.0"
 }
 
 subprojects {
     repositories {
         mavenCentral()
     }
-    
+
     apply plugin: 'kotlin'
     compileKotlin {
         destinationDir = compileJava.destinationDir
@@ -74,3 +78,38 @@ gradle.projectsEvaluated {
         }
     }
 }
+
+
+spotless {
+    repositories {
+        mavenCentral()
+    }
+    // optional: limit format enforcement to just the files changed by this feature branch
+    ratchetFrom 'origin/master'
+
+    format 'misc', {
+        // define the files to apply `misc` to
+        target '*.gradle', '*.md', '.gitignore'
+
+        // define the steps to apply to those files
+        trimTrailingWhitespace()
+        indentWithSpaces() // or spaces. Takes an integer argument if you don't like 4
+        endWithNewline()
+    }
+
+    format 'linguaFranca', {
+        addStep(LfFormatStep.create())
+        target 'test/*/src/**/*.lf' // you have to set the target manually
+        targetExclude 'test/**/failing/**'
+    }
+
+    java {
+        target 'org.lflang*/src/**/*.java', 'buildSrc/**/*.java'
+        // The following is quoted from https://github.com/google/google-java-format
+        // "Note: There is no configurability as to the formatter's algorithm for formatting.
+        // This is a deliberate design decision to unify our code formatting on a single format."
+        googleJavaFormat('1.15.0').aosp().reflowLongStrings()
+        formatAnnotations()
+    }
+}
+tasks.withType(SpotlessTask) { it.dependsOn(":org.lflang.cli:jarLff") }
--- a/build.gradle
+++ b/build.gradle
@@ -98,7 +98,7 @@ spotless {
     }
 
     format 'linguaFranca', {
-        addStep(LfFormatStep.create())
+        addStep(LfFormatStep.create(project.projectDir))
         target 'test/*/src/**/*.lf' // you have to set the target manually
         targetExclude 'test/**/failing/**'
     }
--- a/build.gradle
+++ b/build.gradle
@@ -108,7 +108,7 @@ spotless {
         // The following is quoted from https://github.com/google/google-java-format
         // "Note: There is no configurability as to the formatter's algorithm for formatting.
         // This is a deliberate design decision to unify our code formatting on a single format."
-        googleJavaFormat('1.15.0').aosp().reflowLongStrings()
+        googleJavaFormat('1.15.0').reflowLongStrings()
         formatAnnotations()
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ plugins {
     id "com.github.johnrengelman.shadow" version "${shadowJarVersion}"
     id 'java'
     id 'jacoco'
-    id "com.diffplug.spotless" version "6.11.0"
+    id "com.diffplug.spotless" version "${spotlessVersion}"
 }
 
 subprojects {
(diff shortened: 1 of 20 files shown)


buildSrc/src/main/java/lfformat/LfFormatStep.java
```diff
@@ -0,0 +1,58 @@
...
+            Process p =
+                    new ProcessBuilder(
+                                    List.of(
+                                            "java",
+                                            "-jar",
+                                            Path.of(
+                                                            "org.lflang.cli",
+                                                            "build",
+                                                            "libs",
+                                                            "org.lflang.cli-0.3.1-SNAPSHOT-lff.jar")
```
lhstrh, line 34: This should not be hard coded but read from here: https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/StringsBundle.properties

lhstrh (changes requested):
Looks great! Left some comments.

buildSrc/src/main/java/lfformat/LfFormatStep.java
```diff
@@ -0,0 +1,58 @@
...
+                                            "--dry-run",
+                                            file.getAbsoluteFile().toString()))
+                            .start();
+            StringBuilder output = new StringBuilder();
+            BufferedReader in = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            String line;
+            while ((line = in.readLine()) != null) {
+                if (!output.isEmpty()) output.append("\n");
+                // Filtering by "lff: " is yet another string-processing hack that is not airtight!
+                if (!line.startsWith("lff: ")) output.append(line);
```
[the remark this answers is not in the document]
petervdonovan: `lff` prints a message saying that it has finished formatting. An alternative would be to modify `lff` so that this message is suppressed when `lff` is invoked with the `--dry-run` option.

build.gradle
```diff
@@ -12,13 +15,14 @@ plugins {
     id "com.github.johnrengelman.shadow" version "${shadowJarVersion}"
     id 'java'
     id 'jacoco'
+    id "com.diffplug.spotless" version "6.11.0"
```
lhstrh, line 18: Let's define the version in `gradle.properties`

build.gradle
```diff
@@ -74,3 +78,38 @@ gradle.projectsEvaluated {
...
+        target 'test/*/src/**/*.lf' // you have to set the target manually
+        targetExclude 'test/**/failing/**'
+    }
+
+    java {
+        target 'org.lflang*/src/**/*.java', 'buildSrc/**/*.java'
+        // The following is quoted from https://github.com/google/google-java-format
+        // "Note: There is no configurability as to the formatter's algorithm for formatting.
+        // This is a deliberate design decision to unify our code formatting on a single format."
+        googleJavaFormat('1.15.0').reflowLongStrings()
```
lhstrh, line 111: Idem

buildSrc/src/main/java/lfformat/LfFormatStep.java
```diff
@@ -0,0 +1,87 @@
...
+          ResourceBundle.getBundle(
+              "StringsBundle",
+              Locale.getDefault(),
+              new URLClassLoader(new URL[] {resourcePath.toUri().toURL()}));
+      final Path lffPath =
+          Path.of(
+              "org.lflang.cli",
+              "build",
+              "libs",
+              String.format("org.lflang.cli-%s-lff.jar", properties.getString("VERSION")));
```
lhstrh, line 67: :rocket:

buildSrc/build.gradle
```diff
@@ -0,0 +1,9 @@
+repositories {
+    mavenCentral()
+    dependencies {
+        // https://mvnrepository.com/artifact/com.diffplug.spotless/spotless-lib
+        implementation group: 'com.diffplug.spotless', name: 'spotless-lib', version: '2.30.0'
```
lhstrh, line 5: Use variable for version number

gradle.properties
```diff
@@ -4,6 +4,7 @@ version=0.3.1-SNAPSHOT
 
 [versions]
 commonsCliVersion=1.4
+googleJavaFormatVersion=1.8
```
lhstrh, line 7: Inconsistent with `build.gradle`?

lhstrh (approved):
Only minor nitpicks

lhstrh (approved):

cmnrd:
I noticed that this PR created a new top-level directory `buildSrc`. What is this directory used for? We are currently trying to simplify the repository structure (see https://github.com/lf-lang/lingua-franca/issues/1276) and adding a new directory seems counter productive. Is there a way to move the contents to other locations and integrate with existing files?

petervdonovan:
I do not think it would be a good idea to get rid of `buildSrc`. Here is an excerpt from a conversation I had with Marten:
> Gradle files are like a DSL that is embedded in Groovy. This DSL is pretty limited. Consequently, you are allowed to write custom logic in Java instead of in Gradle files, and then import that (into your Gradle build scripts). This custom imperative code is not part of the core of the project; instead, it is part of the build system that builds the project. **Therefore, it belongs in a buildSrc directory, and not an src directory**

Also, regarding this particular `buildSrc` directory:
> It's totally a hack... because my goal was to invoke Lff, which is indeed what we are trying to build.
> But it works just fine because it invokes the jar of Lff instead of depending directly on the Java code. The jar is built without invoking the formatting step.

So in short, the my general approach to the problem at hand was a little weird -- I wrote Java code that executes a JAR that has to be built separately in order to avoid a circular dependency. However, unless/until we decide to use an entirely different strategy, it is indeed appropriate to keep the code in `buildSrc`.

cmnrd:
Ok, I understand now what buildSrc means. Thanks for explaining. So ideally we would move all the `org.lflang.*` packages into a top-level src/ directory or out of this repo. Then we would have a structure where we have bin/ build/ buildSrc/ lib/ rsc/ in the top level.

I guess invoking the jar is part of the reason why formatting can be slow. I don't quite see the circular dependency though. Could you explain a bit? Since we refactor the repo and the build anyway, we should aim at resolving this problem in the process.

petervdonovan:
> I don't quite see the circular dependency though. Could you explain a bit?

Well, there is no circular dependency, but if `buildSrc` invoked the Java code directly instead of the JAR then there would be a circular dependency because usually (maybe not for us but in general), `buildSrc` has to be compiled before compiling anything else -- this is just the way Gradle sets it up.

cmnrd:
Ok, I see now. Thanks!

Issue #1375, opened 2022-09-24
Collect code coverage information also for the runtimes


cmnrd:
Most of our testing is done in form of integration tests given as LF programs. For the compiler code-base we have a setup that provides us with code coverage information, but we don't have such a setup yet for the runtimes. I am not sure if this is possible for all the targets, but it seems doable to also collect coverage for the C and C++ runtimes. This would be based on [gcovr](https://gcovr.com/en/stable/getting-started.html) and only require to compile all tests with the `--coverage` flag of gcc enabled. This would likely only work on Linux though.

cmnrd:
https://github.com/lf-lang/lingua-franca/pull/1376 implements the generation of files containing coverage data also for C. However, this data currently cannot be used for the following reasons:
1. The `#line` directives in the generated c code confuse lcov because it cannot map object code to its C source code.
2. For each test program the reactor-c files are compiled separately. This way, lcov cannot tell that they are the same and merge coverage information from multiple test runs.

Issue #1314, opened 2022-08-02
Test for invalid states in CI


petervdonovan:
For C, Python, C++, and Rust, it might be a good idea to
- run tests with assertions enabled (perhaps in debug mode)
- run at least some tests through `memcheck`

Related: #1313, https://github.com/lf-lang/reactor-c/issues/24.

cmnrd:
Yes! Somehow I thought I already forced on the Debug mode for our C++ tests, but I also recently noticed that this isn't the case. Note that we can also use `RelWithDebInfo` as it also included the assertions.

I also like the idea of using memcheck. It slows down programs considerably, but our tests are also quite small, so it might  not make much of a difference.

PR #1376, opened 2022-09-25
Generate code coverage information for the C and C++ runtime

cmnrd:
This PR introduces a new build type `Test` which is similar to `Debug` but also generates code coverage information (if possible for the given target and platform). This build type is automatically used for all of our integration tests.  This PR further implements code coverage generation for the C and C++ on Linux. The PR also adjusts the C++ CI so that a code coverage report is generated as a downloadable artifact of the CI job.

While this PR also ensures that coverage data is generated for C tests, we cannot yet leverage this information. This has two reasons:
1. The `#line` directives in the generated c code confuse lcov because it cannot map object code to its C source code.
2. For each test program the reactor-c files are compiled separately. This way, lcov cannot tell that they are the same and merge coverage information from multiple test runs.

This address the first item in #1314 (Run all tests in debug mode with assertions enabled) and also addresses #1375 for C++.


lhstrh (approved):
Where does the coverage get reported?

cmnrd:
I don't know yet. I configured CI to publish an artifact, that should be downloadable, but I need to wait for CI to finish to test how this works ;)

lhstrh:
It looks like `lcov` is also supported by CodeCov: https://about.codecov.io/tool/lcov/

cmnrd:
In principle this is ready to be merged now. But for some reason the serialization tests fail now with linker errors. I don't understand how my changes can lead to broken linking... @petervdonovan maybe you have an idea what could cause this?

petervdonovan:
The Windows Python tests are also failing because no debug version of Python is installed. I think that my CMake PR caused the Python target to be responsive to the `Build-Type` parameter, which I do not think was true before. I suspect that this is why CMake is trying to use a debug version of Python. We should decide whether to install a debug version of Python in CI (which could be tricky on Windows), suppress the `Build-Type` parameter in the Python generator, or suppress the `Build-Type` parameter in CI.

cmnrd:
We could also run all the Python tests using the Release mode. But I don't understand why the build would fail in the Debug mode. The build type should only affect the source files that we actually build, but have no impact on the dependencies. Or am I missing something here about how the Python build works?

cmnrd:
Ok, apparently python is special in this regard. I found a solution here: https://stackoverflow.com/questions/56798472/force-linking-against-python-release-library-in-debug-mode-in-windows-visual-stu

cmnrd:
I finally merged this. Note that I changed the PR name back to "Generate code coverage information for the C and C++ runtime". This is in line with our guidelines as 'generate' is not the action conducted by me but the actual contribution of the PR.

PR #1381, opened 2022-09-28
Use debug logging for all LF tests run in CI

cmnrd:
Essentially, this change ensures that all LF tests are run with the `logging: Debug` target property set. Since for some tests this produces huge amounts of output, this PR also fixes those tests. Furthermore, it fixes some corner cases in the error reporting mechanism, so that we should now see more informative error messages when something unexepected is happening (One example is an out of memory error in the threads that we use for processing stdout and stderr).

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/LFTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/LFTest.java
@@ -193,7 +193,7 @@ public class LFTest implements Comparable<LFTest> {
          * String buffer used to record the standard output and error
          * streams from the input process.
          */
-        final StringBuffer buffer = new StringBuffer();
+        StringBuffer buffer = new StringBuffer();
 
         /**
          * Return a thread responsible for recording the standard output stream
@@ -238,6 +238,10 @@ public class LFTest implements Comparable<LFTest> {
         @Override
         public String toString() {
             return buffer.toString();
+        }
+
+        public void clear() {
+            buffer = null;
         }
     }
 }
--- a/org.lflang.tests/src/org/lflang/tests/LFTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/LFTest.java
@@ -126,9 +126,14 @@ public class LFTest implements Comparable<LFTest> {
     public void reportErrors() {
         if (this.hasFailed()) {
             System.out.println("+---------------------------------------------------------------------------+");
-            System.out.println("Failed: ");
+            System.out.println("Failed: " + this);
             System.out.println("-----------------------------------------------------------------------------");
             System.out.println("Reason: " + this.result.message + " Exit code: " + this.exitValue);
+            if (this.exitValue.equals("139")) {
+                // The java ProcessBuiler and Process interface does not allow us to reliably retrieve stderr and stdout
+                // from a process that segfaults. We can only print a message indicating that the putput is incomplete.
+                System.out.println("This exit code typically indicates a segfault. In this case, the execution output is likely missing or incomplete.");
+            }
             printIfNotEmpty("Reported issues", this.issues.toString());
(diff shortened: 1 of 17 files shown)


org.lflang.tests/src/org/lflang/tests/LFTest.java
```diff
@@ -121,34 +123,29 @@ public boolean hasFailed() {
...
-            sb.append("Reason: ").append(this.result.message).append(" Exit code: ").append(this.exitValue).append(System.lineSeparator());
-            appendIfNotEmpty("Reported issues", this.issues.toString(), sb);
-            appendIfNotEmpty("Compilation output", this.compilationLog.toString(), sb);
-            appendIfNotEmpty("Execution output", this.execLog.toString(), sb);
-            sb.append("+---------------------------------------------------------------------------+\n");
-        return sb.toString();
-        } else {
-            return "";
+            System.out.println("+---------------------------------------------------------------------------+");
+            System.out.println("Failed: ");
```
petervdonovan, line 129:
```suggestion
            System.out.println("Failed: " + this.name);
```

test/C/src/Deadline.lf
```diff
@@ -14,11 +14,7 @@ reactor Source(period: time(3 sec)) {
         if (2 * (self->count / 2) != self->count) {
             // The count variable is odd.
             // Take time to cause a deadline violation.
-            // Do not use nanosleep() here because if this is threaded,
-            // it gets awakened right away.
```
petervdonovan, line 18: I checked the documentation of the sleep functions that our platform abstraction uses, and I did not find any support for this comment. I'm not sure why it was here for so long, but thanks for removing it.

lhstrh: I'm fine with the removal of the comment, but I think that `nanosleep` returns prematurely if the thread receives a signal.

test/C/src/TimeLimit.lf
```diff
@@ -39,15 +39,15 @@ reactor Destination {
...
+        if (self->s != 12) {
+            fprintf(stderr, "ERROR: Expected 12 but got %d.\n", self->s);
             exit(1);
         }
         lf_print("Approx. time per reaction: %lldns", lf_time_physical_elapsed()/(self->s+1));
     =}
 }
 
-main reactor TimeLimit(period: time(1 usec)) {
+main reactor TimeLimit(period: time(1 sec)) {
```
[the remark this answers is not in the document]
lhstrh: :point_up:

test/Python/src/TimeLimit.lf
```diff
@@ -35,14 +35,14 @@ reactor Destination {
 
     reaction(shutdown) {=
         print("**** shutdown reaction invoked.")
-        if self.s != 10000002:
+        if self.s != 12:
```
lhstrh, line 38: Line 39 below is no longer correct with this change.

test/C/src/Deadline.lf
petervdonovan: Yes, that's true. Is that likely to happen in the typical running of our tests?

lhstrh: I think it may have been the reason that the busy-wait approach was opted for.

petervdonovan: If the only signals we expect are those that signify critical errors or requests to terminate the program, then I would not worry about them. However, I do not have a strong opinion on this issue.

cmnrd: Since the deadline tests pass it means that the sleep is long enough. I made some local tests and indeed `lf_nanosleep` does not return prematurely (in my tests). While this does not give a guarantee, it shows at least that the existing comment is wrong (it does not get awakened right away). We could put the sleep in a loop, but it would do much more sense to do this in the implementation of `lf_nanosleep` itself.

PR #1384, opened 2022-09-29
Augmented Support for Arduino

arengarajan99:
This PR will address the following issues dealing with running Arduino on LF
- Autogeneration of custom Board Options file via tweaking of Platform parameter to support overloaded definitions (Complete)
- Copy over of Arduino-Toolchain files to allow self-sufficient compilation of LF-Arduino programs without reliance on building LF from source (should support Nightly Build and release versions) (Complete)
-  Support of external Arduino-supported libraries (Complete)


erlingrj:
@lhstrh Anirudh would like to finish this line of work before applying the changes related to 64bit and nanoseconds to the Arduino port. I think this is reasonable, but I also think it is possible to apply the other changes first. I am mentioning this because there have very recently been discussions on what is really the job of LFC and how should we extend support to multiple embedded targets. This PR sort of represents one possible path where we tightly integrate support for embedded targets into LFC and try to have "end-to-end" build and flash through LFC.

I do think that we can merge this in even before making a decision on how to support embedded targets. In the worst case we would have to do refactoring later. I am pushing a bit to merge lock_time branch into reactor-c main so that we can start working on supporting more embedded targets

lhstrh (approved):
This looks good to me! I am afraid that the approach of hard coding all platforms in an enum will require too much maintenance, but I'm not sure we have a viable alternative. These are certainly improvements over what's currently in `master`, so let's merge.

PR #1387, opened 2022-09-30
Switch to own provider plugin for external dependencies

### LINKED ISSUE #1311: Sync Maven dependencies with Gradle dependencies
It appears that Maven uses a Kotlin version below 1.5 while in gradle we use Kotlin 1.6.20. This surfaced because #1218 made use of the `lowercase` function introduced in Kotlin 1.5. But in the corresponding CI run (https://github.com/lf-lang/lingua-franca/runs/7590796430?check_suite_focus=true) the Maven build fails. The kotlin versions should be synchronized between both build systems. While we are at it, we should also double check all other dependencies.

### LINKED ISSUE #1377: Publish up-to-date Kotlin dependencies as plugin
Create a repository that contains an Eclipse plugin that mirrors the up-to-date libraries of Kotlin from maven central and provides them via an update site for the Eclipse runtime.

a-sr:
Configures the maven+tycho build infrastructure to use the new [*org.lflang.epoch.external.dependencies* plugin](https://github.com/lf-lang/epoch-external-dependencies) that currently only supplies up-tp-date Kotiln dependencies but also can  be used for other dependencies that are not available via Eclipse update sites.

Closes #1377
Closes #1311


org.lflang/src/org/lflang/generator/rust/RustGenerator.kt
```diff
@@ -135,7 +135,7 @@ class RustGenerator(
 
         if (cargoReturnCode == 0) {
             // We still have to copy the compiled binary to the destination folder.
-            val isWindows = System.getProperty("os.name").toLowerCase().contains("win")
+            val isWindows = System.getProperty("os.name").lowercase().contains("win")
```
a-sr, line 138: @cmnrd I tested the Kotlin support by using the `lowercase` function as in #1311. It compiles with maven and works in the Epoch product. I hope this is sufficient.

a-sr:
@lhstrh I suspect you will ask why we need a separate repository and update site for this and don't do this directly in the org.lflang plugin. Even if there is an additional maintenance overhead, it is a cleaner solution. Otherwise, it would obstruct the build process in the Eclipse IDE.
We should be able to get rid of this workaround (and the repository) when we finish the project restructuring in the future (#1276).

By using our own plugin and a more recent version of Kotlin, it might be possible to have signed jars for #1173.

lhstrh (approved):
This looks great! I take it that we have to change the `pom.xml` in `epoch-external-dependencies` to specify a new `kotlin-version`? One downside of having this in a separate repo is that this our version update script won't be able to do this, but if it's only one dependency then this seems manageable.

a-sr:
Sadly this is not enough, since the [manifest](https://github.com/lf-lang/epoch-external-dependencies/blob/master/org.lflang.epoch.external.dependencies/META-INF/MANIFEST.MF) and [build.properties](https://github.com/lf-lang/epoch-external-dependencies/blob/master/org.lflang.epoch.external.dependencies/build.properties) refer to the library jar files including version number.

lhstrh:
Would it be OK to have a little Python script in `epoch-external-dependencies` to do version updates? Or could this potentially break things?

a-sr:
No, I can't think of cases where it would break anything. Unless, the Kotlin guys change the names of the provided jars or the packages provided by them. But that would take a manual effort in any case.

cmnrd:
Thanks a lot @a-sr. Your test using `lowercase()` already shows that the new Kotlin version is used. I double checked by adding a print statement displaying the Kotlin version during code generation and it indeed shows `1.6.20`. So this also fixed https://github.com/lf-lang/lingua-franca/issues/1311 (I added this to the PR description).

I am wondering though if we still need the kotlin plugins listed in the `pom.xml` files. It seems like the Kotlin configuration their is redundant and was never actually used. If that is the case, should we remove all mentions of the kotlin plugin from the `pom.xml` files?

cmnrd (approved):

cmnrd:
One more question: Does this mean that we can now compile the Kotlin classes when building Epoch from the developer Eclipse setup?

cmnrd:
> Thanks a lot @a-sr. Your test using `lowercase()` already shows that the new Kotlin version is used. I double checked by adding a print statement displaying the Kotlin version during code generation and it indeed shows `1.6.20`. So this also fixed #1311 (I added this to the PR description).
> 
> I am wondering though if we still need the kotlin plugins listed in the `pom.xml` files. It seems like the Kotlin configuration their is redundant and was never actually used. If that is the case, should we remove all mentions of the kotlin plugin from the `pom.xml` files?

I tried this and it seems that we still need to use the kotlin plugins in the `pom.xml` files. I tried to remove them locally. Compilation than still succeeds, but the resulting Epoch cannot load the Kotlin classes.

a-sr:
> One more question: Does this mean that we can now compile the Kotlin classes when building Epoch from the developer Eclipse setup?

No, the new dependency plugin only provides the necessary *runtime* libraries, it still needs to be combined with the maven build for compilation. That is also the reason why you need the dependency in the poms.

PR #1389, opened 2022-10-01
Default line length set to 100 for LF files

petervdonovan:
This is done in the name of consistency because we use a line length of 100 in Java files now.


lhstrh:
Should we also change to 2 tabs instead of 4?

petervdonovan:
Okay, the tests are now formatted with 2 spaces indentation and 100 columns width. I also reformatted some Markdown and build.gradle to match.

CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Changelog
- 
+
```
lhstrh, line 2: How come these `*.md` files are affected? It seems to me that they shouldn't be.

CHANGELOG.md
```diff
@@ -388,7 +388,7 @@ This is the first stable release of Lingua Franca. Aside from numerous bugfixes,
 - Multiports and banks are now iterable in the Python target ([#713](https://github.com/lf-lang/lingua-franca/pull/713))
 - Fixed an issue where top-level custom Python classes were being serialized incorrectly
 - `bank_index` (useful for banks of reactors) is now a proper parameter ([#424](https://github.com/lf-lang/lingua-franca/pull/424))
-  that can be passed down the reactor hierarchy via parameter assignment.
+that can be passed down the reactor hierarchy via parameter assignment.
```
lhstrh, line 391: This change isn't right... Looks like it's part of the sentence that starts with the previous list item.

lhstrh (approved):
Look good to me, modulo the thing I commented on (and the remaining conflicts).

CHANGELOG.md
petervdonovan: This change was made automatically. I think that it is semantically equivalent Markdown and that the removal of this indentation should not separate it from the list. ([Here is how it looks on GitHub.](https://github.com/lf-lang/lingua-franca/blob/update-lf-code-width/CHANGELOG.md)) To separate it from the list, I think you would have to add a newline.

lhstrh: That may be, but that just means it wasn't right in the first place :-) Let's correct it...

petervdonovan: I'm confused because it looks right to me as rendered on GitHub. This last line appears as part of the list item that starts with `bank_index`.

lhstrh: To be clear, something was off about that list item, but it wasn't due to the formatting changes in this PR.

lhstrh:
@petervdonovan: shall we bite the bullet on this and get it merged? Modulo conflicts, this looks ready to me...

petervdonovan:
> shall we bite the bullet on this and get it merged? Modulo conflicts, this looks ready to me...

Yep, I'm on board with this, it seems straightforward and harmless

petervdonovan:
Sorry, it looks like I forgot to follow through with this one. I just got it back up to date with `master` and it should be ready to merge. It will cause lots of problems but so be it I guess.

lhstrh:
Let's hold off for a bit until we've bitten the bullet on the formatting of the Java code that this is supposed to achieve consistency with...

CHANGELOG.md
petervdonovan: Marking this as resolved by the latest commit on this branch.

cmnrd (approved):
Great! I would be fine with biting smaller bullets instead of one huge one. I don't see good reasons to hold this off and many conflicts should be resolvable by applying lff in the PRs.

lhstrh:
@petervdonovan: let's merge this one too?

petervdonovan:
This branch has had some issues with the Python LSP tests. I will merge it as soon as this is resolved.

petervdonovan:
It looks like two `CCpp` federated tests failed. The first one that failed looks like it could be the issues that we have had with the global variables used in name mangling, but the second one is less clear (there is only a tiny snippet of output shown for some reason).

The debug output is quite short, so I don't think this is related to running out of heap space...

PR #1398, opened 2022-10-06
Zero-width multiports

petervdonovan:
This PR fixes zero-width multiports for the C target.

### DIFF
--- a/org.lflang/src/org/lflang/generator/PortInstance.java
+++ b/org.lflang/src/org/lflang/generator/PortInstance.java
@@ -38,19 +38,19 @@ import org.lflang.lf.Port;
 import org.lflang.lf.WidthSpec;
 import org.lflang.lf.WidthTerm;
 
-/** 
+/**
  * Representation of a compile-time instance of a port.
  * Like {@link ReactorInstance}, if one or more parents of this port
  * is a bank of reactors, then there will be more than one runtime instance
  * corresponding to this compile-time instance.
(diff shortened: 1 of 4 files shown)


edwardalee (approved):
LGTM!  Thanks for doing this.

cmnrd:
What is the motivation for using zero width multiports? And what exactly does this PR fix?

petervdonovan:
> What is the motivation for using zero width multiports? And what exactly does this PR fix?

In our conversation about mutations, I started making unsubstantiated claims that we could have a lot of expressive power using other means. This prompted me to write some weird programs in the experimental repo. Among these was [this one](https://github.com/lf-lang/experimental-lingua-franca/blob/composition/C/src/Composition/Hypercube.lf), which had issues with connections not being made and with wrong error messages that there were more sources than destinations in a connection and vice versa due to its zero-width multiports. I wanted zero-width multiports because they are necessary in order to continue a pattern across a set of similar reactors. I wanted to continue a pattern across a set of similar reactors so that I could argue that those similar reactors should all be generated from the same source code.

petervdonovan:
I'm not seeing the light at the end of the tunnel on this one. Let's shelve this until we know the future of `initialize_trigger_objects` and the other complex code that operates on the exploded reactor instance graph.

lhstrh:
Closing this based on the last comment by @petervdonovan in this thread.

PR #1401, opened 2022-10-10
Removal of obsolete Eclipse installation files and formatter

lhstrh:
Now that Eclipse installation is governed by Oomph and we have a Spotless configuration in Gradle for formatting, these files should no longer be necessary.


a-sr (approved):
LGTM

Issue #1403, opened 2022-10-10
LSP test failures


lhstrh:
LSP tests are randomized, and tempting to "fix" by just rerunning them. To ensure that these problems get fixed someday, here's a list of them. 

- [ ] https://github.com/lf-lang/lingua-franca/actions/runs/3222520208/jobs/5271773981
```
LspTests > pythonValidationTestSyntaxOnly() STANDARD_OUT
    Expecting an error to be reported at line 61... Success.
    Validating...
    Code validation complete.
    Expecting an error to be reported at line 29... but the expected error could not be found.
    ManualDelayedReaction.lf failed. Content of altered version of ManualDelayedReaction.lf:
    ------------------------------------------------------------------------------

        0 target Python {
        1     # Set keepalive to false since this is a test and schedule is called on a
        2     # physical action from within a reaction. This is one of the special rare
        3     # cases where the user might want to manually override keepalive.
        4     keepalive: false
        5 }
        6 
        7 # That's the stuff that shall be generated for the after
        8 reactor GeneratedDelay {
        9     input y_in
       10     output y_out
       11     state y_state(0)
       12 
       13     physical action act(0 msec)  # TODO: delay in act or the schedule call?
       14 
       15     reaction(y_in) -> act {=
       16         self.y_state = y_in.value
       17         act.schedule(MSEC(100))
       18     =}
       19 
       20     reaction(act) -> y_out {= y_out.set(self.y_state) =}
       21 }
       22 
       23 reactor Source {
       24     output out
       25     timer t
       26 
       27     reaction(t) -> out {= out.set(1) =}  # reaction(t) -> out after 100 msec {=
       28 }
    -> 29         +++++;
       30 
       31 reactor Sink {
       32     input s_in
       33 
       34     reaction(s_in) {=
       35         elapsed_logical = lf.time.logical_elapsed()
       36         logical = lf.time.logical()
       37         physical = lf.time.physical()
       38         print("Nanoseconds since start: {:d} {:d} {:d}.\n".format(logical, physical, elapsed_logical))
       39         if elapsed_logical < MSEC(100):
       40             sys.stderr.write("Expected {:d} but got {:d}.\n".format(MSEC(100), elapsed_logical))
       41             exit(1)
       42     =} deadline(200 msec) {=  =}
       43 }
       44 
       45 main reactor ManualDelayedReaction {
       46     source = new Source()
       47     sink = new Sink()
       48     g = new GeneratedDelay()
       49 
[...]

PR #1404, opened 2022-10-11
Handling for commented-out reactions in integration tests

petervdonovan:
These confuse the error inserter that is used in the randomized LSP tests. The intent was never to make the error inserter work perfectly, but this bad behavior has caused enough confusion for other people to justify taking a few minutes to fix.

Related: https://github.com/lf-lang/lingua-franca/issues/1403#issue-1403810430


lhstrh:
Does this fix the particular bug listed in #1403?

petervdonovan:
> Does this fix the particular bug listed in https://github.com/lf-lang/lingua-franca/issues/1403?

That was the intention, yes.

PR #1409, opened 2022-10-11
Draft: Arduino Support for Arduino IDE v2.0

arengarajan99:
With the release of Arduino IDE v2.0 comes a whole new infrastructure that renders our original CMake setup obsolete. 

Though we still maintain support up to v1.18.18, v2.0 no longer has an avr-gcc compiler and instead uses Arduino-CLI's API model to build out its programs instead. As such, this PR seeks to remove the CMake-Toolchain library completely, remove the build steps, and instead leverage the arduino-cli. 

The Arduino-CLI is limited due to the lack of CMake support, and will need to follow specifications according to the Sketch build process listed here: https://arduino.github.io/arduino-cli/0.27/sketch-build-process/

This PR will also include a transition back to 64 bit support for consistency with the LF timing model.

- No more CMake/CCpp target required since we return control to build-dependency of arduino-cli (Would then support windows)


org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -241,12 +241,7 @@ CodeBuilder generateCMakeCode(
         cMakeCode.unindent();
         cMakeCode.pr(")");
         cMakeCode.newLine();
-
-        if (this.targetConfig.platformOptions.platform == Platform.ARDUINO) {
```
erlingrj, line 245: A question is, could we achieve this outside LFC by passing CMake files to "cmake-include" target property? I guess it is fine to have plug-and-play support for a platform like Arduino, but for other embedded boards I think we cant have all this specific logic inside LFC.

petervdonovan:
I noticed that I got invited to help with this task. My personal motivation for working on the build system has been to make it more maintainable so that I do not have to work on the build system as much later :grin:.  In particular, my goal since August has been to keep the core logic of the build system alongside the C code itself so that
- the structure of the program, and therefore its meaning, is unambiguous (for example, you can see from looking at the repository what code is included, and when)
- changes to the C code do not require corresponding changes in another repository
- build system logic is not duplicated
- the code generator does not have to do too many different things/does not have too much code churn

To this end, maybe there is more I should do after #1299 that would help with Arduino support? For example, if it seems like CMake is not general enough for a CMake build system to target the platforms we want -- Maybe it would help if I shift logic from CMake to the preprocessor so that the build system logic is more "portable" in the sense that it does not rely as much on CMake? Just looking for a way to avoid having to permanently maintain scripts that reach in and copy individual files.

lhstrh:
Sounds interesting, @petervdonovan. Let's discuss in person? I, too, am weary of the scripts that copy files, but I'm hoping that these may be a stop-gap measure until we have our package manager...

petervdonovan:
I am also wondering if, as an alternative to shifting as much logic as possible to preprocessor directives, we could instead precompile the runtime or maybe even the entire program as a static library (like, a `.a` file) and then just use it in a small `.ino` wrapper. [This page](https://arduino.github.io/arduino-cli/0.28/library-specification/) makes it seem like even libraries are supposed to be compiled using `arduino-cli`, in which case this idea is pointless, but I just thought I would check since I don't fully understand the problem at hand.

arengarajan99:
@petervdonovan Arduino libraries are never normally pre-compiled. They typically are made up of a C/C++ class wrapped inside a header file that is then included in your sketch file to be used. The arduino-cli does lazy compilation by compiling an entire library file that is included within your sketch (with the only exception being any file located under a directory named src). Pre-compiling will not work for our use case since arduino-cli simply won't recognize them and it would break the mold of how they're normally defined. The script, though it isn't ideal, seems like the only stop-gap solution that can let us keep the normal reactor-c structure (will be hard to convert this into an Arduino style library).

petervdonovan:
> @petervdonovan Arduino libraries are never normally pre-compiled.

This used to be true, but it has been false since release 1.8.6 according to [the link that I included in my comment above](https://arduino.github.io/arduino-cli/0.28/library-specification/).

PR #1411, opened 2022-10-12
Differentiation between multi-threaded and single-threaded preprocessor directives in C

erlingrj:
This PR adds two compiler definitions to the C target: `LF_MULTI_THREADED` and `LF_SINGLE_THREADED` which reflect the target property "threading". I have previously used NUMBER_OF_WORKERS to try to get the same info, however, NUMBER_OF_WORKERS is defined even if `threading: false`, if `tracing: true`. 

Thus I propose to make the fact that we are single-threaded or multi-threaded globally available. It would suffice with only a single flag, but I think it is clearer with a "positive definition" of something than a negative.


edwardalee (commented):
Apologies for the delay.  I'm not sure this change makes sense. I suspect that tracing won't actually work in an unthreaded runtime, otherwise I don't know why it would be setting NUMBER_OF_WORKERS.

erlingrj:
`tracing: true` sets NUMER_OF_WORKERS=0. But it does not include the threaded runtime. I.e. reactor.c is used, not reactor_threaded.c. I suspect the flag is set because some functionality needed for tracing is hidden behind #ifdef NUMBER_OF_WORKERS. But it does not use the threaded runtime unless threading is specified.

For LET scheduling we need to know whether we need to know, in reactor_common.c whether we are in a single-threaded or multi-threaded runtime. This cannot be checked by `#ifdef NUMBER_OF_WORKERS`. It could also be checked with

```
#ifdef NUMBER_OF_WORKERS
#if NUMBER_OF_WORKERS == 0
// single-threaded with tracing enabled
#else
// multi-threaded
#endif
#else
// Single-threaded without tracing
#endif
```

I.e. I don't really need a new flag, but it seems a bit clunky

edwardalee:
Ah, I see.  I think the reason is simply that tracing needs a to report a thread number.  There may be a better solution, but your solution seems fine, let's merge it.

edwardalee (approved):
Let's merge this.

erlingrj:
Who can merge this for me? @lhstrh ?

lhstrh:
@erlingrj: Just resolved a conflict and waiting for CI. We can merge soon as we're green!

PR #1412, opened 2022-10-13
Use path, not URI in line-directives (#1407)

erlingrj:
Addressing #1407


edwardalee (approved):
LGTM.  Does this correctly report errors on the right line in Epoch?

cmnrd:
Also I am wondering how this affects error reporting via the LSP and on the command line in lfc. Tagging @petervdonovan

erlingrj:
I am away for a week. I'm not sure how this affects Epoch or LSP.

petervdonovan (approved):
I think we can merge this now. Thanks Erling!

Issue #1380, opened 2022-09-28
Disallow `--fast` if there are physical actions


lhstrh:
Or else, the program is bound to experience runtime errors due to attempts to schedule an event earlier than current time by.

axmmisaka:
Should be resolved, see https://github.com/lf-lang/lingua-franca/issues/1480#issuecomment-1615841669 .

PR #1413, opened 2022-10-13
Disallow fast if physical action is present

axmmisaka:
This hopefully resolves https://github.com/lf-lang/lingua-franca/issues/1380

I'm still onboarding (sadly) and can't test the result properly; this should be working but I'm marking as WIP and won't roll CICD until things are figured out on my end.

- [x] Better error message (current one sounds very confusing)
- [x] Test this with TS targets
- [x] Test with C++ targets
- [x] Test with Rust targets
- [ ] Add tests (unit tests) for this functionality (example in https://github.com/lf-lang/lingua-franca/pull/1413#issuecomment-1299091828)


org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -421,6 +421,9 @@ public void accommodatePhysicalActionsIfPresent() {
                             "Using the threaded C runtime to allow for asynchronous handling of physical action " +
                             action.getName()
                         );
+                        if (targetConfig.fastMode) {
```
edwardalee, line 424: This doesn't look quite right. This check is performed only when threading is turned off. Shouldn't it always be performed?

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -156,6 +157,12 @@ public class TargetConfig {
      */
     public boolean fastMode = false;
 
+    /**
+     * If fastMode is true, this field records the AST Node that made 
+     * fastMode to be true.
+     */
+    public Element fastModeASTNode = null;
```
edwardalee, line 164: It looks like this field is set but never read. Is this needed?

edwardalee (changes requested):
Thanks for dealing with this. I think it needs small changes, as detailed in the comments.

org.lflang/src/org/lflang/TargetConfig.java
axmmisaka: It's needed in https://github.com/lf-lang/lingua-franca/pull/1413/files#diff-36305ea0c57443b1c6234992d94f5217bb222b410cb31b339b0582d9e96178bcR425 (sole purpose is to provide error message)

axmmisaka:
FYI, after some discussion yesterday the conclusion is:
1. Refactor my commit in `CGenerator` to `GeneratorBase` so that all targets can support this feature; and
2. Make action finding an AST utility (there might be one now actually)

axmmisaka:
Appears to be working...... with this C example. I will add a roadmap to what else to do before merging.
image

```
target C {
    fast: true
};
reactor Physical {
    input x:int;
    physical action a;
    reaction(x) -> a {=
        lf_schedule(a, 0);
    =}
    reaction(a) {=
        interval_t elapsed_time = lf_time_logical_elapsed();
        printf("Action triggered at logical time %lld nsec after start.\n", elapsed_time);
    =}
}

```

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -632,6 +635,25 @@ public static String getUniqueIdentifier(Reactor reactor, String name) {
...
         return ASTUtils.collectElements(definition, featurePackage.getReactor_Actions());
     }
+
+
+    /**
+     * Given a generator class, return a list of all its physical actions.
+     * @param generator Generator class definition.
+     */
+    public static List<Action> allPhysicalActions(GeneratorBase generator) {
+        // TODO: Merge these into one line of filter
```
axmmisaka, line 645:
```suggestion
```

This is probably fine (with `ArrayList`)

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -156,6 +157,12 @@ public class TargetConfig {
      */
     public boolean fastMode = false;
```
axmmisaka, line 158: Check where `fastMode` is set to true, and if only at places where `fastModeASTNode` can be specified, maybe remove this.

org.lflang/src/org/lflang/generator/GeneratorBase.java
```diff
@@ -140,6 +140,7 @@ public abstract class GeneratorBase extends AbstractLFValidator {
      * using a command like `foo = new Foo();`
      */
     protected List<Reactor> reactors = new ArrayList<>();
+    public List<Reactor> getReactors() { return this.reactors; }
```
axmmisaka, line 143: Remove this, and pass in reactors for `allPhysicalActions1

org.lflang/src/org/lflang/TargetConfig.java
axmmisaka: Maybe make a separate issue/PR, this property has at least 8 usages.

axmmisaka:
I can't find a way to add unit test to this whole functionality sadly (if I add, it would be to test `allPhysicalActions`; I could check if that is necessary but instantiation would take some time honestly) and system test doesn't seem to support expected failed result.

org.lflang/src/org/lflang/generator/GeneratorBase.java
```diff
@@ -296,6 +296,11 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
             targetConfig,
             errorReporter
         );
+
+        // If fastMode is enabled, report error on the AST Node 
+        if (!ASTUtils.allPhysicalActions(this.reactors).isEmpty() && targetConfig.fastMode) {
+            errorReporter.reportError(targetConfig.fastModeASTNode, "When physical action is present `fast` must be false.");
```
oowekyala, line 302: Should this really be an error? The user could have included a reactor that contains physical actions internally without knowing. That would break their build, even though `fast` is not supposed to change the semantics of the program but just be an optimization AFAIU. I think a warning would be more appropriate

I also prefer to avoid giving commands in messages to the user ("fast must be false"). I find a warning message like the following feels nicer: "Fast mode cannot be enabled in a program with physical actions, this will be ignored."

axmmisaka: I’ll make changes accordingly and discuss this with @lhstrh in our next meeting

lhstrh: I like @oowekyala's suggestion to ignore the `fast` flag and showing a warning that explains why.

lhstrh:
> I can't find a way to add unit test to this whole functionality sadly (if I add, it would be to test `allPhysicalActions`; I could check if that is necessary but instantiation would take some time honestly) and system test doesn't seem to support expected failed result.

In the `org.lflang.test.compiler` package, you'll find a number of classes (e.g., `LinguaFrancaValidationTest`) in which unit tests are defined. Both the `allPhysicalActions` method and the new error/warning could be tested like that.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -633,6 +636,24 @@ public static String getUniqueIdentifier(Reactor reactor, String name) {
     public static List<Action> allActions(Reactor definition) {
         return ASTUtils.collectElements(definition, featurePackage.getReactor_Actions());
     }
+
+
+    /**
+     * Given a list of reactors, return a list of all physical actions.
+     * @param reactors Generator class definition.
+     */
+    public static List<Action> allPhysicalActions(Iterable<Reactor> reactors) {
```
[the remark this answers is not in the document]
axmmisaka: Just on the second point: `allPhysicalActions()` is also used in here: https://github.com/lf-lang/lingua-franca/pull/1413/files#diff-36305ea0c57443b1c6234992d94f5217bb222b410cb31b339b0582d9e96178bcR409 utilising the actual content, I can add two methods for efficiency but I'm unsure if that would be preferred.

org.lflang/src/org/lflang/generator/GeneratorBase.java
axmmisaka: If we have consensus I will open an issue potentially changing error here to warning?
https://github.com/lf-lang/lingua-franca/blob/d9604195fb1be5abefd715c8a064b8489dd92500/org.lflang/src/org/lflang/validation/LFValidator.java#L1093-L1097

lhstrh: I don't think it's worth opening a new issue for that. It's fine to amend #1380.

axmmisaka: This is for `LFValidator` tho, they already have such functionality but it's an error not warning.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -406,7 +406,7 @@ public CGenerator(FileConfig fileConfig, ErrorReporter errorReporter) {
     public void accommodatePhysicalActionsIfPresent() {
         // If there are any physical actions, ensure the threaded engine is used and that
         // keepalive is set to true, unless the user has explicitly set it to false.
-        List<Action> physicalActions = ASTUtils.allPhysicalActions(this.reactors);
+        List<Action> physicalActions = this.getPhysicalActions();
         if (!physicalActions.isEmpty()) {
```
oowekyala, line 410: Here the list is built also only to fetch at most the first element... Consider using stream methods, eg
```java
Optional<Action> firstPhysicalAction = reactors.stream()
                 .flatMap(r -> allActions(r).stream())
                 .filter(act-> act.getOrigin() == ActionOrigin.PHYSICAL)
                 .findFirst()
```
These can often replace loops and read much less complex.

axmmisaka: Alternatively, I can just revert this change or make `getPhysicalActions` return an iterator - unsure if that would be satisfactory under Java

axmmisaka:
Unsure if this makes the evaluation lazy, but if it does I'd say this is the most abstract way to do so

oowekyala:
> Unsure if this makes the evaluation lazy, but if it does I'd say this is the most abstract way to do so

Yes, the evaluation of streams is always lazy up to the final terminal operation (`findAny` or `findFirst` here). Thanks for the change, it looks nice

axmmisaka:
Updates on test:
It appears that errors reported in `GeneratorBase` does not get tested anywhere (at least not in `org.lflang.test.compiler`). `LFValidator`, on the other hand, does not seem to error if both are present (which it will with VSC plugin). I might be misunderstanding what I was doing, I suppose......

axmmisaka:
See #1480 for update; I might need to later close this PR or revert some of the changes once we figure out how that issue is to be solved.

axmmisaka:
Closing this as it is stale and resolved, see https://github.com/lf-lang/lingua-franca/issues/1480#issuecomment-1615841669 .

PR #1420, opened 2022-10-21
Disallow reaction code to modify the multiport container

revol-xut:
Currently, reaction bodies that are triggered by a multiport get full access to the multiport data structure. Apart from reading from or writing to individual ports, this also includes adding or removing port instances from the mutiport container. This PR pulls in the changes from https://github.com/lf-lang/reactor-cpp/pull/33 and adjusts the cods generator and tests accordingly.


cmnrd (commented):
The changes look good to me. You forgot to update reactor-cpp though, but instead reactor-c and reactor-c-py changed for some reason. I would recommend to run ` git submodule update --init --recursive` and be cautious about which changes to stage.

revol-xut:
I think nothing really is blocking this pr. I would merge this and then fix the benchmark repo

lhstrh:
But then master will fail in the interim, which we try to avoid. Why not fix the benchmark tests first?

cmnrd (approved):
This looks great! Thanks for your fixes.

Issue #1359, opened 2022-09-07
TypeScript tests are way too slow in CI


cmnrd:
At the moment the TypeScript Test for MacOS and Windows run for more than 4 hours. This is way too long. Do we know why it takes so long (for Linux it is only one hour)? Is there a way to cut down the number of tests run? Or can we maybe split the TypeScript tests into multiple jobs that can run in parallel?

lhstrh:
They shouldn't be so slow, but it hasn't been an issue to address it. Perhaps @byeong-gil has any ideas?

byeong-gil:
@lhstrh Actually, I have run the TypeScript test only in Linux. So I don't have any idea about this problem yet... 
I'll try to examine what happens in the Windows TypeScript test.

lhstrh:
One thing we could do is disable the tests on Windows and Mac, but that, of course, isn't really addressing the problem...

petervdonovan:
The only suggestion I have ever put forward to fix this was to bundle the dependencies of the reactor-ts runtime into it. Everything that it would need would be stuffed into one big JS file, which we could minify if we want. A downside is that if the end user wants to use the same library as is used by the runtime, then they would have to import it again, and there would be code duplication in the generated JS. I almost implemented this already, actually, in https://github.com/lf-lang/lingua-franca/pull/1322, but I gave up because I was inexperienced with the NPM/TS ecosystem, and just getting anything to work was a painful process for me.

I personally would be happy with considerably scaling back the tests on Windows and Mac, at least as a stopgap measure. Node is supposed be mostly platform-independent (right?), and although I remember addressing platform compatibility issues to get the TS target to work on Windows, they were pretty superficial bugs that would have been caught by a simple smoke test.

petervdonovan:
By "superficial bugs," I mean things like using Bash commands in the scripts in `package.json`, not misbehaving JavaScript code. It seems like `microtime`, which we depend on, might be doing some low-level magic, but AFAIK we have not have issues with that dependency.

PR #1426, opened 2022-10-22
Usage of a published version of reactor-ts

petervdonovan:
Fixes https://github.com/lf-lang/lingua-franca/issues/1359
Tasks completed:
- [x] Remove `reactor-ts` submodule
- [x] Facilitate local testing using a command line `./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="file:<some-path-to-a-local-reactor-ts>"`
- [x] Setup `reactor-ts` CI to run integration tests against a particular commit
- [x] Provide documentation about new development workflow


org.lflang/src/lib/ts/tsconfig.json
```diff
@@ -13,7 +13,7 @@
         "strictBindCallApply": true,
         "strictNullChecks": true,
         "strictFunctionTypes": true,
-        "typeRoots": ["./node_modules/@types/", "./node_modules/reactor-ts/src/core/@types/"]
+        "typeRoots": ["./node_modules/@types/", "./node_modules/@lf-lang/reactor-ts/src/core/@types/"]
```
lhstrh, line 16: Note: the second element of the list should no longer be necessary once we remove our dependency on `ulog` as I suggested.

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -384,8 +384,21 @@ private static void checkAndReportFailures(Set<LFTest> tests) {
      * @throws IOException if there is any file access problem
      */
     private LFGeneratorContext configure(LFTest test, Configurator configurator, TestLevel level) throws IOException {
+        var props = new Properties();
+        var sysProps = System.getProperties();
+        // Set the external-runtime-path property if it was specified.
+        if (sysProps.containsKey("runtime")) {
```
lhstrh, line 390: Note to self: this code is outdated and won't work.

org.lflang.cli/src/org/lflang/cli/Lfc.java
```diff
@@ -95,20 +96,18 @@ enum CLIOption {
         /**
          * Construct a new CLIOption.
          *
-         * @param opt         The short option name. E.g.: "f" denotes a flag
-         *                    "-f".
-         * @param longOpt     The long option name. E.g.: "foo" denotes a flag
-         *                    "--foo".
-         * @param hasArg      Whether or not this option has an argument. E.g.:
+         * @param parameter   teThe build parameter that this CLI parameter corresponds to.
```
petervdonovan, line 99:
```suggestion
         * @param parameter   The build parameter that this CLI parameter corresponds to.
```

petervdonovan (commented):
I can't approve this because of how GitHub behaves, but it looks right to me. I agree that it is cleaner to introduce the `BuildParm` enum.

test/TypeScript/README.md
```diff
@@ -1,2 +1,19 @@
...
+```
+./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="~/lf-lang/reactor-ts"
+```
+- To point to a particular ref (e.g., `feature-branch`) of the `reactor-ts` repository:
+```
+./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#feature-branch"
+```
+- Note that `lfc` can be pointed to an alternative runtime as well, using the `external-runtime-path` switch:
+```
+lfc test/TypeScript/src/Minimal.lf --external-runtime-path ~/lf-lang/reactor-ts
```
cmnrd, line 18: Note that for lfc we also have the `--runtime-version` argument which is intended for specifying a particular ref to be fetched from the main runtime repository. Is this also supported for TS now? If not, I think we should add it.

cmnrd (approved):
This looks good to me! The improvements for Linux and Windows are significant, but oddly on MacOS the tests are still running after 2 hours. Please also consider my comment regarding `--runtime-version`

test/TypeScript/README.md
lhstrh: OK, I've added this (and removed the overloading of the `-Druntime` switch as it 1) isn't strictly necessary and 2) doesn't seem to allow `pnpm` to reuse prior installations of `@lf-lang/reactor-ts`. Instead, in CI, I'm just checking out the repo and pointing to the location in the filesystem for tests invoked as part of `lf-lang/reactor-ts` CI runs.

lhstrh (approved):

PR #1427, opened 2022-10-24
Draft: Modal let

erlingrj:
Companion PR to https://github.com/lf-lang/reactor-c/pull/118


org.lflang/src/org/lflang/generator/c/CModesGenerator.java
```diff
@@ -199,6 +200,25 @@ public static String generateLfHandleModeChanges(
         );
     }
     
+    /**
+     * Generate function for getting reactor_mode_state_t array and its length
+     *
+     * @param hasModalReactors True if there is modal model reactors, false otherwise
```
edwardalee, line 206:
```suggestion
     * @param hasModalReactors True if there are modal model reactors, false otherwise
```

edwardalee (approved):
LGTM!

erlingrj:
Update:

As suggested by @a-sr, we code-generate the function `_lf_mode_get_transitioning_reactors` which will take a `vector_t` as argument and fill it with those reactors that are supposed to have a mode transition. This is accompanied by a `_lf_mode_collect_transitioning_reactors` in `reactor-c` that actually searches through all the modal reactors and fills up the vector

PR #1422, opened 2022-10-21
New lff script and `org.lflang.cli` moved into `org.lflang` as subpackage 

cmnrd:
This PR refactors the org.lflang.cli package.

1. It includes the `org.lflang.cli` package as a subpackage in `org.lflang`. This closes https://github.com/lf-lang/lingua-franca/issues/1421.
2. It simplifies the gradle build such that a single jar (`org.lflang-0.3.1-SNAPSHOT.jar`) is created. This jar includes both the lfc and lff main classes. The main build task is now called `buildAll`. `buildLfc` is still available but shows a deprecation warning.
3. It introduces an `lff` script to invoke the lingua franca formatter. This closes https://github.com/lf-lang/lingua-franca/issues/1369.
4. It renames the build script to `build-lf-cli` and drops the `--run` option. Note that the old build script is still available, but prints a deprecation warning. The `--run` option was dropped because it would be unclear which command to run. Also the run tasks that we provide for gradle already implement this functionality. 

There are still a couple of unresolved issues though:
- [x] Fix and rename the `build-lfc` script. Should we name it `build-all`?
- [x] Add `lff.ps1` script for Windows. (@axmmisaka)
- [x] There is a test class [LfcIssueReportingTest.kt](https://github.com/lf-lang/lingua-franca/blob/f1c4e65912cfdd103ca4b3fdf7f9c1aa509e54b0/org.lflang.cli/test/kotlin/org/lflang/cli/tests/LfcIssueReportingTest.kt) in `org.lflang.cli.test`, but I was not able to move it to the new location and make it work with gradle. @oowekyala can you help with this?
- [x] We should update the developer documentation and use the `buildAll` task instead of `buildLfc`. Also `build-lf-cli` should be used instead of `build-lfc`. See https://github.com/lf-lang/website-lingua-franca/issues/106
- [x] Open an issue to keep track of the deprecated tools and remove them for next release. See https://github.com/lf-lang/lingua-franca/issues/1431
- [x] Update the `.github/scripts/package_lfc.sh` script (and rename to `package.sh`?)
- [x] Document changes required to release workflow https://github.com/lf-lang/release-tools/issues/16

PR #1430, opened 2022-10-27
PowerShell scripts for lfc and lff

axmmisaka:
As mentioned in the commit message, this current implementation is very *NIX-y/bashy and not powershell-y. While I don‘t doubt its correctness this adds complications; maybe we should just pass the thing in as argument.


axmmisaka:
There appears to be some misunderstanding until I saw this: https://github.com/lf-lang/lingua-franca/pull/1422#issuecomment-1293139385

cmnrd (changes requested):
This looks great! While testing this I noticed one problem though. It seems like we cannot pass any additional arguments to lfc and lff anymore. For instance `.\bin\lfc.ps1 -c test\C\src\Minimal.lf` produces an error complaining that `test\C\src\Minimal.lf` does not exist. Do you have an idea what is going on here? It works perfectly fine without any additional arguments.

axmmisaka:
Ah I see, my hypothesis is that PS takes `./bin` as its PWD… because running `.\bin\lfc.ps1 -c .\..\test\C\src\Minimal.lf`works.

axmmisaka:
Unsure if using `.` or `Invoke-Command` would be better

lhstrh (approved):
Looks good to me! Thanks for fixing this :-)

cmnrd:
Thanks @axmmisaka!

Issue #1146, opened 2022-05-08
Authentication and communication security in federated execution


hokeun:
## Motivation
The current implementation of federated execution lacks authentication and encryption mechanisms to prevent adversarial federates from joining the federation or sending malicious messages (e.g., bad tags or bad sensor readings) to RTI or other federates.
- Use case:
  - Simple threat model: an adversarial federate on the same network. 
- Non-goal:
  - We're not targeting applications running on the cloud (e.g., containerized execution environment) where we can use existing security infrastructure.
  - Inventing our own protocol or crypto - we will use standard protocols and cryptos already available.

## Goals
1. **Authentication:** To authenticate a federate properly when the federate joins a federation. 
2. **Communication security:** To ensure confidentiality and integrity (+ message authenticity) of the network messages between RTI and federates as well as among federates.

### Target property for configuring authentication

- We can use the target properties to configure authentication mechanism, which will include the necessary libraries (e.g., OpenSSL for C federates) and generate code for performing authentication.

```
target C {
    auth: Hmac
};
```

- To begin with, we can have following target property options:
  - **No auth property**: This is default, using hex string matching for authentication - current approach.
  - **auth: Hmac**: This option uses the Key-hashed MAC ([HMAC](https://en.wikipedia.org/wiki/HMAC)) using the shared secret = federation ID = hex string. Instead of sending this federation ID in plain text, we will use this for a challenge-response using randomly generated nonce for each federate. We may need a longer string, though, for example, 32 bytes for SHA 256.

- In the future, we will be able to use full-fledged authentication, for example, using different keys for each federate or using public-key cryptography for authentication. 
  - **auth: SST**: This option uses [SST](https://github.com/iotauth/iotauth), an open-source toolkit providing authentication and authorization services for the IoT. An API in C lanagues is under development at @hokeun's research group.

### Target property for configuring communication security for network messages
- Target properties can also be used to configure communication security

```
target C {
    communication-security: {
        key-distribution: FederationID,
        encryption: AES-128-CBC,
        message-authentication: HmacSHA256
    }
};
```

- Here are target property options to begin with
  - **No communication-security property**: By default, we do not provide any encryption or message authentication 
[...]

PR #1432, opened 2022-10-30
Option of using OpenSSL for authentication between RTI and federates

Jakio815:
NOTE: Please note that there's an issue with CI tests, mentioned at the bottom of this description.

## Overview
This PR is the first step to add security to LF, by including OpenSSL in the C target.

This PR is related to the PR lf-lang/reactor-c#105 and resolves lf-lang/lingua-franca#1146. Both PRs should be merged to work.

The main idea is in lf-lang/lingua-franca#1146, and this PR is an implementation of 
  - **auth: Hmac**: This option uses the Key-hashed MAC ([HMAC](https://en.wikipedia.org/wiki/HMAC)) using the shared secret = federation ID = hex string. Instead of sending this federation ID in plain text, we will use this for a challenge-response using randomly generated nonce for each federate. We may need a longer string, though, for example, 32 bytes for SHA 256.

## Approach
```
target C {
   auth: true 
}
```
This option includes OpenSSL libraries, and use HMAC authentications between the RTI and the federates. It creates a random nonce and creates a HMAC tag, and they process a 3 way handshake starting from the RTI. The details of the 3-way handshake is in lf-lang/reactor-c#105. The option is currently implemented as a boolean option.

The main protocol is like the following.
**1. RTI_HELLO (RTI -> Federate)
2. FED_RESPONSE  (Federate -> RTI) 
3. RTI_RESPONSE  (RTI -> Federate)**

For test, build the RTI.
```
// org.lflang\src\lib\c\reactor-c\core\federated\RTI\
mkdir build && cd build
cmake -DAUTH=ON ../
make
sudo make install
```

There is a simple test .lf file.
```
./gradlew runlfc --args test/C/src/federated/SimpleFederatedAuthenticated.lf
./test/C/bin/SimpleFederatedAuthenticated
```

RTI does not include OpenSSL libraries when -DAUTH=ON is not commanded. It is set OFF on default.
`runlfc` command also does not include OpenSSL libraries if `target C { auth: true }` is not coded.

For clarification between lf-lang/reactor-c#105 and this PR, this PR is about,
1. Including OpenSSL libraries optionally in CMake.
2. Code generation part of adding ```auth``` options.
3. Example code of a simple example of authentication in ```test/C/src/federated/SimpleFederatedAuth.lf```

PR lf-lang/reactor-c#105 will be the details of the 3-way handshake. 

## CI Tests
cpp-ros2 tests and serialization tests are not passing. I sent an issue #1483.


Jakio815 (commented):
.

.github/workflows/c-tests.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       - name: Install dependencies OS X
         run: |
           brew install coreutils
+          brew install openssl
+          brew link openssl --force
```
Jakio815, line 47: Only adding export OPENSSL_ROOT_DIR="/usr/local/opt/openssl" does not work, because it seems that every run opens a new bash.

.github/workflows/ci.yml
```diff
@@ -54,12 +54,12 @@ jobs:
...
   # Run the C integration tests.
   c-tests:
-    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@master
+    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@auth
     needs: cancel
 
   # Run the CCpp integration tests.
   ccpp-tests:
-    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@master
+    uses: lf-lang/lingua-franca/.github/workflows/c-tests.yml@auth
```
Jakio815, line 62: Changed `install-rti/install.sh` and `c-tests.yml` so changed this temporarily. Need to change back to master after merged.

Jakio815:
@lhstrh @edwardalee Would you please check out this PR when available? Thanks!

lhstrh:
Thanks for the reminder, @Jakio815, will put in a review soon.

.github/workflows/ci.yml
lhstrh: The way to do  it is to keep it like this; merge (not yet remove the branch!); and push a change to `master` to update these lines. After that, the branch can be safely deleted. I usually take care of this, so you don't have to worry about it.

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -58,7 +58,14 @@
  * @author{Marten Lohstroh <marten@berkeley.edu>}
  */
 public enum TargetProperty {
-
+    /**
+     * Directive to allow including OpenSSL libraries and process HMAC authentication.
+     * FIXME: The option will be updated from boolean to string for support of various options.
```
lhstrh, line 63: Let's not leave this `FIXME` here. I think we can just remove it.

test/C/src/federated/SimpleFederatedAuthenticated.lf
```diff
@@ -0,0 +1,23 @@
+/**
+ * This simple test checks if federate authentication works by adding `auth`
+ * target property.
+ */
+target C {
+    timeout: 2 secs,
+    build-type: RelWithDebInfo,
+    auth: true,  // FIXME: This will be updated to string such as HMAC.
```
lhstrh, line 8: Same here. The policy is to not leave dangling `FIXME`s in `master`.

lhstrh (approved):
This looks good to me!

org.lflang/src/org/lflang/TargetProperty.java
Jakio815: Fixed it.

test/C/src/federated/SimpleFederatedAuthenticated.lf
Jakio815: Fixed it. I won't leave them next time!

PR #1439, opened 2022-11-01
Empty bracket pairs properly tokenized

oowekyala:
This allows spaces in empty bracket pairs in types, like `int[  ]`. Although it doesn't look good, I think it's surprising to make the language space-sensitive in this case only. It also allows commenting things out, eg `int[/*4*/]`.

The change removes one token from the grammar: `[]` is now tokenized as `[` then `]`, like in C and all other target languages that use these tokens.

This change is extracted from #544


lhstrh (approved):
If only all PRs were this clean an concise! :laughing:

PR #1441, opened 2022-11-01
Cleanup initializer grammar

### LINKED ISSUE #657: Stop special casing `{= 0 =}`
It's very useful to allow specifying a zero duration as just "0":
```
a: time(0)
```

But in #544 I found out we go out of our way to interpret `{= 0 =}` in the same way. I think that's not useful (why would you write `{=0=}` if you can write `0`), and also introduces a dangerous precedent for trying to interpret the contents of `{=...=}`. Using `0 ms` or `0` is the *LF* syntax to express time, and `{= 0 =}` is the target language syntax, which we're supposed not to make any assumptions about. Also our implementation could be simpler if we didn't support it.

I think we should just remove support for that.

oowekyala:
Changes extracted from #544.

This does not change the LF language, it just introduces the node `Initializer` to fold some duplicated code in the grammar file. Later this can be changed to support the new `=` syntax described in #986 

Since the code generation in the C++ target needed to be updated to adjust to the grammar changes, we also used the opportunity to improve the handling of parameters in C++ and fix a couple of related issues along the way.

Fixes https://github.com/lf-lang/lingua-franca/issues/419
Fixes #657
Fixes https://github.com/lf-lang/lingua-franca/issues/605


org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
```diff
@@ -55,28 +49,29 @@ class CppInstanceGenerator(
...
-            CppTypes.getCppArgumentForParameter(param, this)
+    private fun Instantiation.getParameterStruct(): String {
+        // If this is a bank instantiation and the instantiated reactor defines a "bank_index" parameter, we have to set
+        // bank_index here explicitly.
+        var prefix = "${cppType}::Parameters{"
+        if (isBank && reactor.hasBankIndexParameter())
+            prefix += ".bank_index = __lf_idx, "
+        return parameters.joinToString(", ", prefix, "}") {
+            val expr = it.rhs.exprs[0]
+            assert(!it.rhs.isBraces && !it.rhs.isParens && it.rhs.exprs.size == 0)
```
oowekyala, line 60: I'm not sure this works... You're extracting an element at the line above so probably the assert should check for size == 1 (not 0) and be on the line above

org.lflang/src/org/lflang/generator/cpp/CppExtensions.kt
```diff
@@ -83,20 +80,13 @@ fun Expression.toCppCode(outerContext: Boolean = false, inferredType: InferredTy
  * @param outerContext A flag indicating whether to generate code for the scope of the outer reactor class.
```
oowekyala, line 80: This part of the comment needs to be removed as well

org.lflang/src/org/lflang/generator/cpp/CppTypes.kt
```diff
@@ -52,67 +49,36 @@ object CppTypes : TargetTypes {
...
+        TimeUnit.WEEK   -> "d*7"
+        else            -> ""
     }
 
-    return buildString {
-        append(getTargetType(inferredType)) // treat as ctor call
+/**
+ * Returns a C++ variable initializer.
+ */
+fun CppTypes.getCppInitializer(init: Initializer?, inferredType: InferredType): String {
```
oowekyala, line 74: Will this work out of the box with = initializers? Will we just have to make this return `= init.exprs[0]` basically?

org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
cmnrd: Oh, right. Also the assertion should come before the variable access. I think we never really use the weird syntax with multiple expressions, so the assertion war my way of checking.

I am seeing strange errors when I run all the tests (also in CI) that look like this:
```
Failed: src/concurrent/DeadlineHandledAboveThreaded.lf
    -----------------------------------------------------------------------------
    Reason: No information available. Exit code: ?
    Compilation output:
    Generating code for: file:/home/runner/work/lingua-franca/lingua-franca/test/Cpp/src/concurrent/DeadlineHandledAboveThreaded.lf
    ******** mode: STANDALONE
    ******** generated sources: /home/runner/work/lingua-franca/lingua-franca/test/Cpp/src-gen/concurrent/DeadlineHandledAboveThreaded
```
It dosen't provide a clear reason and I cannot reproduce the error when running lfc manually. So perhaps it is because the assertion is only executed during testing? But if so, then why is the error not properly reported in our tests?

org.lflang/src/org/lflang/generator/cpp/CppTypes.kt
cmnrd: Yes, for most cases this should suffice.

cmnrd:
While working on this, I noticed that we need to reformat most of touched java files to make spotless happy. I think it is best to only do this shortly before merging to keep the diff clean for now.

org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
oowekyala: Assertions are only executed if the JVM is passed the flag `-ea`. Possibly the lfc launcher doesn't use this flag

lhstrh: It doesn't..

cmnrd (commented):
I think this PR is taking shape now and I will mark it as ready for review. We still have to fix a couple of tests though.

I fixed the C++ tests (for the most parts) and we now have passing tests for C, C++, and Python. There is still a couple of Windows tests failing which I will look into.

@oowekyala There appears to be a problem with multiports in Rust. Cold you take a look? This could be related to the fact that the rust generator uses `toCppCode` on the width.

@petervdonovan could I ask you to help with the failing unit tests (round trip) and the failing LSP tests? I don't understand why the unit test fails. It seems to insert some erroneous code into one of our test files but then expects the file to parse. For the LSP test, I was not able to find the actual error in the log. I am not sure what I need to search for her, but 'error' or 'fail' does not work. It would be great if we could further improve the outputs of theses tests so that the log clearly states where the error happened and also points to how it could be fixed.

@lhstrh The TS tests fail and likely the TS generators still need to be updated. Who could help with this?

lhstrh:
If they have a chance, perhaps @hokeun or @byeong-gil could have a look at the TypeScript tests?

cmnrd:
Thanks for your fix @petervdonovan! So the round trip tests apply the formatting to our tests and check if the result is still valid LF?

hokeun:
> If they have a chance, perhaps @hokeun or @byeong-gil could have a look at the TypeScript tests?

Sorry @lhstrh, I'm tied up with some other things right now, but I should be able to take a look at it this weekend!

cmnrd:
Actually I just fixed the TS tests. It turned out to be a simple fix. Sorry for the noise!

cmnrd (approved):
Looks like all the tests pass now. As far as I am concerned, this is ready to be merged.

org.lflang/src/org/lflang/ASTUtils.java
```diff
@@ -432,15 +432,17 @@ private static Instantiation getDelayInstance(Reactor delayClass,
...
             } else {
                 widthSpec.setOfVariableLength(true);
             }
             delayInstance.setWidthSpec(widthSpec);
         }
         Assignment assignment = factory.createAssignment();
         assignment.setLhs(delayClass.getParameters().get(0));
-        assignment.getRhs().add(delay);
+        Initializer init = factory.createInitializer();
+        init.getExprs().add(connection.getDelay());
```
oowekyala, line 444: Maybe we should assert the following
```suggestion
        init.getExprs().add(Objects.requireNonNull(connection.getDelay(), "null delay"));
```

org.lflang/src/org/lflang/generator/c/CTypes.java
```diff
@@ -6,6 +6,7 @@
 import org.lflang.ErrorReporter;
 import org.lflang.InferredType;
 import org.lflang.generator.TargetTypes;
+import org.lflang.lf.ParameterReference;
```
oowekyala, line 9: looks superfluous

org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
```diff
@@ -51,64 +50,45 @@ class CppInstanceGenerator(
...
-                    "${param.targetType}{${assignment.rhs.joinToString(", ") { it.toCppCode() }}}"
-                } else {
-                    "${param.targetType}(${assignment.rhs.joinToString(", ") { it.toCppCode() }})"
-                }
+    private fun Instantiation.getParameterStruct(): String {
+        val assignments = parameters.mapNotNull {
+            if (it.rhs.isParens || it.rhs.isBraces) {
+                errorReporter.reportError(it, "Parenthesis based initialization is not allowed here!")
+                return@mapNotNull null
             }
```
oowekyala, line 58: Nit: annotated returns are usually not necessary, eg here using a when statement here would be cleaner IMO

org.lflang/src/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
```diff
@@ -63,8 +63,8 @@ class CppStandaloneCmakeGenerator(private val targetConfig: TargetConfig, privat
             |  endif()
             |endif()
             |
-            |# require C++ 17
-            |set(CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard is cached for visibility in external tools." FORCE)
+            |# require C++ 20
```
oowekyala, line 66: Maybe this "20" should be in a constant somewhere (eg `const val CPP_STANDARD` or so)? Would be easier to update later

oowekyala (commented):
Thanks for pushing this forward @cmnrd!

petervdonovan (approved):
Thanks! This substantially cleans up and adds tests for some code that I left and the cleanups in the grammar look good to me. I was not able to find mistakes

Issue #1448, opened 2022-11-02
Index calculation does not use deadline


erlingrj:
When calculating the index of a reaction in `setReactionPriorities` in `CTriggerObjectsGenerator` the code-generator uses the `deadline`  member of `ReactionInstance` this seems to be initialized to FOREVER and never updated.

lhstrh:
@erlingrj wasn't this fixed by #1451?

PR #1451, opened 2022-11-02
Fix for broken deadline propagation mechanism

erlingrj:
This PR fixes two things:

1. The ReactionInstance.deadline field was not initialized which made the index field of the `reaction_t` always use FOREVER as the deadline. I.e. the EDF scheduling of reactions didn't work on single-threaded
2. Add getInheretedDeadline() function to ReacitonInstance where it searches for the minimum deadline of its downstream reactions. This is then used to calculate the priority of a reaction.

This solves the problem of AfterDelay reactors blocking deadline reactions because they have low priority.

I am using this fix on an outdated branch it seems to work.

TODO:
- [x] Fix getInheritedDeadline() it should search recursively through the downstream.


org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -499,6 +500,22 @@ public TimeValue assignLogicalExecutionTime() {
         return this.let = let;
     }
 
+    /**
+     * Find the inherited deadline which is the least deadline of all
+     * downstream reactions. This inherited deadline is used to assign
+     * priority to the reaction
+     */
+    // FIXME: Needs to be recursive to support chains.
+    public TimeValue getInheritedDeadline() {
```
edwardalee, line 509: Shouldn't this also check against this.deadline?

erlingrj: It is checked somewhere else. But I will revise the implementation and use:
ReactionInstance.declaredDeadling: Deadline specified by user.
ReactionInstance.deadine for the "inferred" deadline which is either MIN(downstream deadlines) or declaredDeadline.

erlingrj:
I have added 3 test programs that fails on master which verifies that the deadline reactions are indeed prioritized

erlingrj:
@edwardalee Update:

1. Fix function `ReactionInstance.getInferredDeadline()` it recursively calls into all direct downstream reactions and returns the minimum inferredDeadline. This function could be optimized because the results could be stored within each ReactionInstance, e.g. in `.deadline`, to avoid recomputing it multiple times. 
2. When generating reaction initialization, getInferredDeadline is called and we just update `.deadline` to make the fewest changes to existing code.
3. Had to specifically cast "index" value into an unsigned representation. Not entirely sure how it worked previously. Since MAX_LONG left-shifted by any amount will yield a negative long...

erlingrj:
Actually, ecplise-tests fails due to stack overflow in getInferredReactions. Can there be cycles when you follow the dependentReactions?

erlingrj:
@edwardalee I pushed a fix, but sure if it is satisfactory. The stack-overflow happens because a reactionInstance can have itself in the set of `dependetReactions`. This happens if there is a spiralling bank of Reactors. This puzzles me, but a fix is to detect this and stop the recursion. I have done so in `setInferredDeadline`. A possible improvement would be to invoke `setInferredDeadline` at a different place, but I am not sure where

erlingrj:
Actually, that fix is not enough as the spiralling can be through other reactions and thus not be caught. There has to be some way of differentiating between reactions in different reactors in a bank?

edwardalee:
Oh, I think I know what the problem is. Each instance of ReactionInstance can stand for more than one "runtime instance." I.e., there is only one instance of ReactionInstance for the whole bank. The docs for ReactionInstance say this:
```
 * Like {@link ReactorInstance}, if one or more parents of this reaction
 * is a bank of reactors, then there will be more than one runtime instance
 * corresponding to this compile-time instance.  The {@link #getRuntimeInstances()}
 * method returns a list of these runtime instances, each an instance of the
 * inner class {@link ReactionInstance.Runtime}.  Each runtime instance has a "level", which is
 * its depth an acyclic precedence graph representing the dependencies between
 * reactions at a tag.
```
I think the deadline needs to be calculated and stored in `ReactionInstance.Runtime`. In fact, that inner class already has a field `deadline`. I think that the algorithm for assigning deadlines should be just like the `assignLevels` method of `ReactionInstanceGraph`, except it should start from the leaf nodes rather than the root nodes and should work upstream rather than downstream.  This could perhaps be done at the same time as assignLevels.

It looks like there was a previous attempt to propagate deadlines. In `ReactionInstanceGraph.addDownstreamReactions` there are these lines:
```
                        // Propagate the deadlines, if any.
                        if (srcRuntime.deadline.compareTo(dstRuntime.deadline) > 0) {
                            srcRuntime.deadline = dstRuntime.deadline;
                        }
```
However, this won't do the job, I think.  Deadlines, like levels, propagate more than one level.  I think the implementation needs to be like `assignLevels`, which uses Kahn's algorithm.

erlingrj:
It is a bit suspicious that only the macOS tests are failing. But I have now a design which I believe works. It followed Edwards proposal and mimics assignLevels, only doing it backwards. I have added a call to `assignInferredDeadlines` from the `rebuild` method. This means that any time a ReactionInstanceGraph is constructed, both levels and deadlines are assigned. This means that when we call `ReactorInstance.assignLevels` we are also assigning deadlines.

edwardalee:
The MacOS CI runs seem to be very vulnerable to unexpected deadline violations. It seems that deadlines of hundreds of milliseconds are not enough.  Maybe these tests are being run on a mechanical Turing machine and hence are very slow?

erlingrj:
> The MacOS CI runs seem to be very vulnerable to unexpected deadline violations. It seems that deadlines of hundreds of milliseconds are not enough. Maybe these tests are being run on a mechanical Turing machine and hence are very slow?

Aha. I increased the deadlines now, their magnitude doesn't matter since I just wanna check that priorities are assigned correctly. I guess technically a Turing machine couldn't execute a LF program since we depend on constant "interaction" with an external clock and control flow is dictated by this interaction (deadlines)?

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -362,6 +357,22 @@ public Set<Integer> getLevels() {
...
+     * a bank and its dependencies on other reactions pass through multiports.
+     */
+    public Set<TimeValue> getInferredDeadlines() {
+        Set<TimeValue> result = new LinkedHashSet<>();
+        // Force calculation of levels if it has not been done.
+        parent.assignLevels();
+        for (Runtime runtime : runtimeInstances) {
+            result.add(runtime.deadline);
+        }
+        return result;
```
erlingrj, line 373: We dont really need this. We could just search through the list.

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -92,6 +92,14 @@ public void rebuild() {
...
         }
+        // FIXME: Is it OK to also assign deadlines here?
+        // Also do the traversal in the opposite order to calculate the deadlines
+        this.clear();
+        addNodesAndEdges(main);
+        // Assign a level to each reaction. 
+        // If there are cycles present in the graph, it will be detected here.
+        assignInferredDeadlines();
+
     }
```
erlingrj, line 103: Can we check for no deadlines in the program.

erlingrj: And not assign deadlines if so

erlingrj: @edwardalee can I assume that the ReactionInstanceGraph always is connected to the top-level ReactorInstance? If that is true I can check whether there are any declaredDeadlines within the ReactorInstance. If it is not necessarily the top-level I would have to check somewhere else. If there is any deadline in the  program then we should assign inferred deadlines

erlingrj: I think the answer to that is no? We make a ReactionInstanceGraph for each ReactorInstance. So we need a global flag saying this program has deadlines. Any idea on where to check it and where to store it?

org.lflang/src/org/lflang/generator/c/CTriggerObjectsGenerator.java
```diff
@@ -344,20 +345,28 @@ private static boolean setReactionPriorities(
...
         var prolog = new CodeBuilder();
         var epilog = new CodeBuilder();
         for (ReactionInstance r : reactor.reactions) {
             if (currentFederate.contains(r.getDefinition())) {
-                var levels = r.getLevels();
-                if (levels.size() != 1) {
+                var levelSet = r.getLevels();
+                var deadlineSet = r.getInferredDeadlines();
+
+                if (levelSet.size() != 1 || deadlineSet.size() != 1) {
```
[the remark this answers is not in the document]
erlingrj: To clarify. This is for banked reactors so the Reactions within have the same declared deadline, if it exists. We have 4 scenarios
1. There is only 1 level (which means they are not connected to each other) But they have different inferredDeadlines due to being connected.
- Then we need an array of deadlines but only a single level
2. There are multiple levels, i.e. they have dependencies on each other. But all the inferred deadlines are the same. 
- Then we only need array of levels. 
3. There is multiple levels and multiple deadlines
4. One level and all have the same deadline

In fact, we should just store the index directly in the array, not both level and deadline

org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
edwardalee: There is always a top-level ReactorInstance. 

There is code somewhere that searches globally to see whether there is a physical action. Perhaps that could be a model for code that checks for a deadline.

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -351,6 +351,7 @@ public Set<ReactionInstance> dependsOnReactions() {
     public Set<Integer> getLevels() {
         Set<Integer> result = new LinkedHashSet<>();
         // Force calculation of levels if it has not been done.
+        // FIXME: Is it necessary to repeat this everywhere?
```
erlingrj, line 354: @edwardalee Are there any valid reason for all the time checking if the levels have been assigned? This check is performed for each reaction in the program. I am not performing similar test on the deadline assignments, I assume that it is done for the top-level once before code-generating any priorities/indices for the reactions

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -2571,6 +2571,9 @@ private void createMainReactorInstance() {
                     errorReporter.reportError("Main reactor has causality cycles. Skipping code generation.");
                     return;
                 }
+                if (hasDeadlines) {
+                    this.main.assignDeadlines();
+                }
```
erlingrj, line 2576: This is where deadlines are assigned and propagated. Later levels are also assigned. Still "assignLevels" are invoked a lot later, "just in case" it seems

org.lflang/src/org/lflang/generator/c/CTriggerObjectsGenerator.java
```diff
@@ -343,30 +343,42 @@ private static boolean setReactionPriorities(
     ) {
         var foundOne = false;
         // Force calculation of levels if it has not been done.
+        // FIXME: Why is this repeated all over? Are there scenarios where we haven't assigned levels yet?
```
erlingrj, line 346: Again assignLevels are called, in case it hasn't been done. It is also invoked each time we call "getLevels"

erlingrj:
Sorry for the delay here, let`s get this fixed merge

org.lflang/src/org/lflang/generator/ReactionInstance.java
edwardalee: I'm not sure I understand the question, but it looks like getLevels() is called only once per ReactorInstance. The assignLevels() function caches the result, so it doesn't recompute it.  I guess we could cache the result of getLevels(), but I don't think it will change anything currently.

erlingrj: I am referring to calling `assignLevels()` all the time. Either it is redundant (because it is invoked on the top-level reactor at the beginning of code generation) or somehow there are scenarios where we are inside the code generator and for some reason levels have not been assigned yet. It is important because in the latter scenario, also the deadlines will not have been assigned. So I would have to replicate that caching logic for the deadlines also.

org.lflang/src/org/lflang/generator/ReactorInstance.java
```diff
@@ -182,13 +182,31 @@ public ReactorInstance(Reactor reactor, ReactorInstance parent, ErrorReporter re
...
+     * @return
+     */
+    public ReactionInstanceGraph assignDeadlines() {
+        if (depth != 0) return root().assignDeadlines();
+        if (cachedReactionLoopGraph == null) {
+            cachedReactionLoopGraph = new ReactionInstanceGraph(this);
+        }
+        cachedReactionLoopGraph.rebuildAndAssignDeadlines();
+        return cachedReactionLoopGraph;
+
```
edwardalee, line 208: Remove blank line?

edwardalee (approved):
Very nice!  Let's merge!

lhstrh (approved):
LGTM. Only minor nitpicks about comments.

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -362,6 +357,22 @@ public Set<Integer> getLevels() {
         }
         return result;
     }
+    
+    /**
+     * Return a set of deadlines that runtime instances of this reaction have.
+     * A ReactionInstance may have more than one level if it lies within
```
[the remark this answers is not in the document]
erlingrj: Yes you are right. Will fix. It was copied from `getLevels()`. I am, unfortunately a bit confused by why `assignLevels` are being invoked a lot it seems totally redundant. It should be enough to invoke it once for the top-level main reactor. And this is done. But it is also repeated for each reactor, and then for each reaction within each reactor. I am gonna stick with calling assignDeadlines a single time

org.lflang/src/org/lflang/generator/ReactorInstance.java
```diff
@@ -189,6 +189,22 @@ public ReactionInstanceGraph assignLevels() {
         }
         return cachedReactionLoopGraph;
     }
+
+    /**
+     * This function assigns/propagates deadlines through the Reaction Instance Graph
+     * it performs Kahn`s algorithm in reverse, starting from the leaf nodes and
+     * propagates deadlines upstream. To reduce cost, it should only be invoked when
```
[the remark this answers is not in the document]
erlingrj: Fixed

erlingrj:
I still have some doubts whether the deadline propagation is implemented correctly. How it works now:

When main reactor is created, if there is a deadline in the program, the deadlines are propagated on the ReactionInstanceGraph using Kahns Algorithm. The levels are also propagated. This is the only place where the deadline propagation algorithm is invoked. My doubt arises from this: assignLevels are called multiple other places also, which doesn't result in extra work because the results are cached. But we don't do this caching for the deadlines. So it suggests that there are scenarios where levels (and deadlines) are not propagated from the top-level main reactor. But I cannot think of such an example, but there must be a reason for calling assignLevels multiple places "just in case"

lhstrh:
I agee, @erlingrj, that it's sus. The only way to eliminate this doubt is to make sure `assignLevels` is only called once, like it used to be.

erlingrj:
@lhstrh @edwardalee. Can we merge this? My latest changes was to remove seemingly unnecessary calls to `assingLevels`, no failing tests tells me that they were indeed superfluous and that our implementation of deadline propagation is thus correct

lhstrh:
There is still an unaddressed `TODO` in the description?

erlingrj:
Its done

org.lflang/src/org/lflang/generator/ReactionInstance.java
```diff
@@ -356,12 +351,27 @@ public Set<ReactionInstance> dependsOnReactions() {
     public Set<Integer> getLevels() {
```
lhstrh, line 351: Unless `assignLevels` is invoked in the constructor, there is no assurance that this method will always return a sensible result. Where is `assignLevels` called?

edwardalee:
I'm all for merging this!

org.lflang/src/org/lflang/generator/ReactionInstance.java
erlingrj: I think the problem that was addressed by calling assignLevels, "just in case" is that the level of a reaction is not known until you have constructed the whole reaction graph. So you cannot just do it in the constructor. A simple fix would be to add a simple function to reactorInstance which checks if levels has been assigned yet. (hasLevelsBeenAssigned()) this would walk up the containment hierarchy to the main reactor and check if it has created a reactionInstanceGraph yet (and thus assigned levels). If this for some reason is not true, then I propose to throw an error because the developer might be requesting the levels at a point in time where they are not available (e.g. before the reaction graph has been fully constructed).

How does this sound?

lhstrh: OK, that sounds like a good stop-gap measure to me. Still not fond of the original design, but the error will help prevent confusion.

cmnrd:
Does this fix https://github.com/lf-lang/lingua-franca/issues/1448?

PR #1452, opened 2022-11-02
Fix TS CI on Windows

cmnrd:


lhstrh (approved):
Thanks for fixing this. Sorry for the sloppiness on my part...

PR #1456, opened 2022-11-03
Fix python tests

### LINKED ISSUE #1455: Python test failures
See: https://github.com/lf-lang/lingua-franca/actions/runs/3381421755/jobs/5615545726

cmnrd:
Fixes #1455.

### DIFF
--- a/.github/workflows/py-tests.yml
+++ b/.github/workflows/py-tests.yml
@@ -33,7 +33,6 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-        if: ${{ runner.os == 'Windows' }}
       - name: Install dependencies OS X
         run: |
           brew install coreutils
--- a/org.lflang/src/org/lflang/generator/c/CCompiler.java
+++ b/org.lflang/src/org/lflang/generator/c/CCompiler.java
@@ -195,8 +195,9 @@ public class CCompiler {
                 buildPath);
         if (command == null) {
             errorReporter.reportError(
-                    "The C/CCpp target requires CMAKE >= 3.5 to compile the generated code. " +
-                            "Auto-compiling can be disabled using the \"no-compile: true\" target property.");
+                "The C/CCpp target requires CMAKE >= " + CCmakeGenerator.MIN_CMAKE_VERSION
+                    + "3.5 to compile the generated code. " +
+                    "Auto-compiling can be disabled using the \"no-compile: true\" target property.");
         }
         return command;
     }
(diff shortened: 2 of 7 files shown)


.github/workflows/py-tests.yml
```diff
@@ -33,7 +33,6 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-        if: ${{ runner.os == 'Windows' }}
```
petervdonovan, line 36: Christian might know this already, but just in case: This line (deleted in this PR) was added by me specifically to make the macOS tests pass. The GH Actions runners have Python pre-installed, so the existence of the extra Python seemed to confuse CMake. The test failures we are seeing know are probably the same as what blocked me for more than a month.

cmnrd: You are probably right here. The strange thing is that cmake does not even use the version installed by the github action (3.10), but uses some other version that it finds (3.8). I don't know how to resolve this... The default Python version in the MacOS runner is now 3.11 which our Python target is not compatible to. I will continue trial and error and see where it gets me...

Peter, you know the Python target better than me: What do you think how much effort it would be to actually fix the errors that we see popping up in CI when Python 3.11 is used?

cmnrd: I just noticed that I made a mistake and excluded version 3.10, which explains why cmake selected version 3.8 (which seems to be preinstalled as well)

cmnrd:
It looks like it works fine when we specify a range of possible Python versions in the cmake script. @petervdonovan what do you think about this solution? Should we merge it?

cmnrd:
The LSP tests on Mac OS consistently fail. I tried 3 times. Run 1 and 3 have the same error, but in run 2 another error is reported. I am not sure if there is an actual problem lurking here, or if I just had bad luck with the LSP tests.

petervdonovan:
Yes, I think it's a real bug. Probably when we changed Python versions, the error messages coming out of Python changed slightly and failed to parse.

org.lflang/src/org/lflang/generator/c/CCompiler.java
```diff
@@ -195,8 +195,9 @@ public LFCommand compileCmakeCommand() {
                 buildPath);
         if (command == null) {
             errorReporter.reportError(
-                    "The C/CCpp target requires CMAKE >= 3.5 to compile the generated code. " +
-                            "Auto-compiling can be disabled using the \"no-compile: true\" target property.");
+                "The C/CCpp target requires CMAKE >= " + CCmakeGenerator.MIN_CMAKE_VERSION
+                    + "3.5 to compile the generated code. " +
```
petervdonovan, line 199:
```suggestion
                    + " to compile the generated code. " +
```

petervdonovan:
Not only is the LSP tests problem apparently platform-specific, but I am having trouble reproducing it even by running CI from a branch that is based on this branch. I'm worried that the two issues blocking this and #1457 will be difficult to find.

PR #1457, opened 2022-11-03
TypeScript generator cleanups

oowekyala:
Changes extracted from #1441

ExpressionGenerator is removed because it is redundant to TargetTypes. The rest should be a simple refactoring without functional change.


oowekyala:
The remaining failing tests are because of #1455, and there is also an LSP test failing randomly: https://github.com/lf-lang/lingua-franca/actions/runs/3386947686/jobs/5639171134#step:16:420

I don't understand why an error is expected here. Is the LSP checking for undefined names in a Python file? How accurate can that be?

cmnrd:
After running the workflow again, the LSP error also shows in master: https://github.com/lf-lang/lingua-franca/actions/runs/3384118963/jobs/5673489810
I suspect that in the last run the LSP test war run with a old version of the MacOS runner.

org.lflang/src/org/lflang/CommonExtensions.kt
```diff
@@ -158,6 +158,12 @@ fun <T> Iterable<T>.joinWithCommasLn(
     transform: (T) -> CharSequence = { it.toString() }
 ): String = joinWithCommas(prefix, postfix, skipLines = true, trailing, transform)
 
+fun <T> Iterable<T>.joinWithLn(
```
lhstrh, line 161: KDoc missing.

org.lflang/src/org/lflang/generator/GeneratorUtils.kt
```diff
@@ -42,3 +47,38 @@ fun EObject.locationInfo(): LocationInfo {
         lfText = toTextTokenBased() ?: ""
     )
 }
+
+
```
lhstrh, line 51: Newly added functions should also have KDoc.

org.lflang/src/org/lflang/generator/ts/TSImportPreambleGenerator.kt
```diff
@@ -67,36 +67,29 @@ class TSImportPreambleGenerator(
...
     private fun generateDefaultImports(): String {
-        return with(PrependOperator) {
-            """
-                |// Code generated by the Lingua Franca compiler from:
-                |// file:/${filePath.toString()}
-                $DEFAULT_IMPORTS
-                |
-            """.trimMargin()
-        }
+        return """
```
lhstrh, line 70: I'm not sure I understand why it is possible to do away with the `with` clause here?

lhstrh (approved):
LGTM! Thanks for these cleanups, @oowekyala :+1:

org.lflang/src/org/lflang/generator/ts/TSImportPreambleGenerator.kt
oowekyala: That's because the prepend operator is not used here (that's the `..` operator on String). DEFAULT_IMPORTS is written to take advantage of `trimMargin`. Looking at this, it might be cleaner to change that, because this code relies on the fact that DEFAULT_IMPORTS has the same indentation as the other text blocks that include it...

org.lflang/src/org/lflang/generator/GeneratorUtils.kt
```diff
@@ -42,3 +47,48 @@ fun EObject.locationInfo(): LocationInfo {
...
+ * Returns the target code for the [getActualValue] of the
+ * param for this instantiation.
+ */
+fun TargetTypes.getTargetInitializer(param: Parameter, inst: Instantiation): TargetCode {
+    val init = inst.getActualValue(param)
+    return getTargetInitializer(init, param.type, param.isInitWithBraces)
+}
+
+/**
+ * Returns the actual value of a parameter for the given instantiation.
```
lhstrh, line 74:
```suggestion
 * Return the actual value of a parameter for the given instantiation.
```

org.lflang/src/org/lflang/generator/GeneratorUtils.kt
```diff
@@ -42,3 +47,48 @@ fun EObject.locationInfo(): LocationInfo {
...
+ * source code is invalid (param is required)
+ */
+fun Instantiation.getActualValue(param: Parameter): List<Expression> =
+    parameters.firstOrNull { it.lhs == param }?.rhs
+        ?: param.init
+        ?: throw InvalidLfSourceException(this, "No value for parameter ${param.name}")
+
+
+/**
+ * Returns the target code for the given expression, given
```
lhstrh, line 86:
```suggestion
 * Return the target code for the given expression, given
```

org.lflang/src/org/lflang/generator/GeneratorUtils.kt
```diff
@@ -42,3 +47,48 @@ fun EObject.locationInfo(): LocationInfo {
...
+ * that it's a time expression.
+ */
+fun TargetTypes.getTargetTimeExpr(v: Expression): TargetCode =
+    this.getTargetExpr(v, InferredType.time())
+
+/** If this is null, returns the literal 0. */
```
lhstrh, line 92:
```suggestion
/** If this is null, return the literal 0. */
```

lhstrh (approved):

PR #1459, opened 2022-11-04
Update Rust runtime, add AsyncCallback

oowekyala:
This shows that lf-lang/reactor-rs#34 works properly. The test doesn't reproduce lf-lang/reactor-rs#33 though.


test/Rust/src/concurrent/AsyncCallback.lf
```diff
@@ -0,0 +1,61 @@
+// Test asynchronous callbacks that trigger a physical action.
```
lhstrh, line 1: Probably have to run this through our code formatter `lff` to let this pass CI.

lhstrh (approved):

lhstrh:
The latest Rust test failure looks a bit sus...

oowekyala:
There is definitely something weird going on: https://github.com/lf-lang/lingua-franca/actions/runs/3429702634/jobs/5727270072#step:8:164
>  Reason: Test did not pass. Exit code: 139
>  This exit code typically indicates a segfault. In this case, the execution output is likely missing or incomplete.

I'm not sure how a segfault is possible given the test is safe Rust. I also would not expect the race condition to manifest like this. I think it would have been caught by the scheduler and cause a panic. I'll look into it

oowekyala:
@lhstrh It looks like the new test is passing with the latest master. I still don't know why it used to segfault... maybe we can merge it anyway :shrug:

lhstrh (approved):

PR #1461, opened 2022-11-05
Debug macos python lsp in ci

petervdonovan:


petervdonovan:
I think this issue has been resolved.

cmnrd:
Is there a way to add more information to the log produced by the LSP tests? I am always puzzled by LSP test failures because the log only states that something went wrong but we do not see which error message the test actually expected and which error message we got.

petervdonovan:
To make sure we are talking about the same thing, here is what the test output looks like:
```
    Expecting an error to be reported at line 24... but the expected error could not be found.
    Pipeline.lf failed. Content of altered version of Pipeline.lf:
    ------------------------------------------------------------------------------

        0 target Python {
        1     timeout: 2 sec
        2 }
        3 
        4 reactor TakeTime {
        5     input _in
        6     output out
        7 
        8     reaction(_in) -> out {=
        9         offset = 0
       10         for i in range(10000):
       11             offset+=1
       12 
       13         out.set(_in.value + offset)
       14     =}
       15 }
       16 
       17 reactor Print {
       18     input _in
       19     state count(0)
       20     state received(0)
       21 
       22     reaction(_in) {=
       23         self.received += 1
    -> 24         undefined_name15291838("Received: {:d} at logical time {:d}".format(_in.value, lf.time.logical_elapsed()))
       25         if _in.value != (self.count + 40000):
       26             sys.stderr.write("ERROR: Expected {:d}.\n".format(self.count + 40000))
       27             exit(1)
       28         self.count+=1
       29     =}
       30 
       31     reaction(shutdown) {=
       32         if self.received == 0:
       33             sys.stderr.write("ERROR: Final reactor received no data.\n")
       34             exit(3)
       35     =}
       36 }
       37 
       38 main reactor Pipeline {
       39     timer t(0, 200 msec)
       40     state count(0)
       41 
       42     c1 = new TakeTime()
       43     c2 = new TakeTime()
       44     c3 = new TakeTime()
       45     c4 = new TakeTime()
       46     p = new Print()
       47 
       48     c1.out -> c2._in after 200 msec
       49     c2.out -> c3._in after 200 msec
       50     c3.out -> c4._in after 200 msec
       51     c4.out -> p._in
       52 
       53     reaction(t) -> c1._in {=
       54         c1._in.set(self.count)
       55         self.count += 1
       56     =}
       57 }

    ------------------------------------------------------------------------------
```

The test with the syntax error or undefined name in it is printed, with an arrow (`->`) indicating the erroneous line (line 24 in this case). So it usually is possible to see what error was supposed to have been reported. Is the problem just that this output is hard to find because too much other output is printed? Or should the test print out another message describing the nature of the error that was inserted?

cmnrd:
I think it is clear from this output where the error is expected to be reported, but it is unclear what exactly the expected error message is. This makes it hard to understand why the test fails if for instance the message reported by the target compiler changed. Or do we only test for the location and not for the concrete error message?

If the test fails, this typically means that some other error occurred. In order to understand what is going on, it would be helpful to see all error messages that where actually reported.

petervdonovan:
Yes, we only test for the location and not the concrete error message. The error messages that are actually reported do appear in CI I believe, because [`MockLanguageClient` prints them](https://github.com/lf-lang/lingua-franca/blob/bc699cc9e5fe2649f532d11710e467ae70ee0f7d/org.lflang.tests/src/org/lflang/tests/lsp/MockLanguageClient.java#L38). They do appear in CI [here](https://github.com/lf-lang/lingua-franca/actions/runs/3415848917/jobs/5685403525#step:18:125), for example.

cmnrd:
Oh, I see... Thanks!

I guess this still leaves the question open what happens if the parsing fails. Is it possible somehow to print the raw output of the commands run in the background somehow?

petervdonovan:
Yes. The stdout is forwarded in the LSP tests to the same extent as it is forwarded in `lfc`, but I guess the output does sometimes get swallowed. Part of this is because it isn't human-readable and therefore wouldn't normally be printed by `lfc`, but we can add a `verbose` or `debug` option so that it does get printed.

PR #1467, opened 2022-11-10
Add CLI unit tests

### LINKED ISSUE #1429: CLI arg validation
Currently, CLI arguments are passed on to the code generator as properties almost blindly. It should validate arguments in `Lfc.java` (and add unit tests that cover that code).

oowekyala:
I found a bug whereby wrong CLI options were reported as "no such file or directory", and found that there are no tests for this, so here they are.

Ref #1429 

TODO before ready: add more tests


cmnrd (approved):
This looks great! Thanks for taking care of the tests!

PR #1470, opened 2022-11-11
Cleanup AttributeUtils

cmnrd:
While trying to use the Attribute syntax I found that the `AttributeUtils` could use some cleanups. In particular, this change removes the grammar rule `AttrParmValue` and instead simply uses `Literal`. Since some of the type checking done required new functions for checking literals, I also decided to move all functions regarding expressions to a new class `ExpressionUtils`. And while doing so I also realized that AstUtils, ExpressionUtils, and AttributeUtils woulb probably better located in the `org.lflang.ast` subpackage. I realize that this last change might be causing conflicts with other large outstanding PRs. Feel free to revert the last commit if it seems too intrusive.

### DIFF
--- a/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
+++ b/org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
@@ -38,9 +38,6 @@ import java.util.stream.Stream;
 
 import javax.inject.Inject;
 
-import org.eclipse.elk.alg.layered.options.EdgeStraighteningStrategy;
-import org.eclipse.elk.alg.layered.options.FixedAlignment;
-import org.eclipse.elk.alg.layered.options.GreedySwitchType;
 import org.eclipse.elk.alg.layered.options.LayerConstraint;
 import org.eclipse.elk.alg.layered.options.LayeredOptions;
 import org.eclipse.elk.alg.layered.options.NodePlacementStrategy;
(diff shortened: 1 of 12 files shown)


oowekyala:
I understand where this is coming from, but maybe we should wait to move these classes. The AST nodes themselves don't live in the `ast` package -maybe they should be moved there as well. I'm partly saying this because #1441 is still in flight as you know, and I want it to avoid the fate of #544.

I suggest we should have a ticket about planned package restructurings, and do everything in bulk as a planned effort when that achieves a critical mass.

cmnrd:
OK, wie could add it as an item to #1276.

It looks like the tests do not pass anyway, I will probably only get to fix them and of next week.

lhstrh:
This is cool, but I'd prefer to defer merging this into `master` until `fed-gen` is merged. We're getting pummeled with conflicts otherwise, making it even harder to get it ready to merge.

org.lflang.diagram/src/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
```diff
@@ -1390,7 +1387,7 @@ private KNode addErrorComment(KNode node, String message) {
     
     private Iterable<KNode> createUserComments(EObject element, KNode targetNode) {
         if (getBooleanValue(SHOW_USER_LABELS)) {
-            String commentText = AttributeUtils.label(element);
+            String commentText = AttributeUtils.getAttributeValue(element, "label");
```
oowekyala, line 1390:
```suggestion
            String commentText = AttributeUtils.getLabel(element);
```

oowekyala (approved):

cmnrd:
I force pushed a more minimalist version today, that only cleans up the attributes but avoids additional refactoring.

PR #1474, opened 2022-11-17
Update spotlessLinguaFranca task.

petervdonovan:
This fell out of sync with the structure of the repository during Christian's recent refactoring because there were no regression tests for it. Thanks to @Jakio815 for reporting this problem.

### DIFF
--- a/buildSrc/src/main/java/lfformat/LfFormatStep.java
+++ b/buildSrc/src/main/java/lfformat/LfFormatStep.java
@@ -61,18 +61,19 @@ public final class LfFormatStep {
               new URLClassLoader(new URL[] {resourcePath.toUri().toURL()}));
       final Path lffPath =
           Path.of(
-              "org.lflang.cli",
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
Thanks for the fix!

Jakio815:
@petervdonovan  Would you please check if `./gradlew spotlessApply` works regularly on `org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java` ? Even if I add only a word, the whole file gets reformatted.

petervdonovan:
> Even if I add only a word, the whole file gets reformatted.

I think that is the intended behavior. It's OK if entire files get completely reformatted. It will cause a lot of frustration and unnecessarily complex merge conflicts for people who have open PRs, but I think that is an unavoidable consequence of transitioning to automatic formatting.

cmnrd:
Can we add a a CI test for this? Would be great if we can avoid breaking it again

PR #1476, opened 2022-11-17
Stop unpacking Rust runtime

### LINKED ISSUE #1471: Stop unpacking Rust runtime into src-gen
This is not useful and can actually be harmful. Recently, the benchmark runner also filled nearly 600 GiB of disk space with copies of the runtime, because it's copied for every benchmark program:
Pasted image 20221114120225

AFAICT, there is no use to it because Cargo has to download other dependencies from the internet anyway, so it's not like having a local copy allows LFC to run offline. Copies also prevent reusing build artifacts with a cache system like sccache.

oowekyala:
Fix #1471

This has the following changes (compared to master):
- The Rust submodule stays in the repo currently. It can be practical for development, and it also allows generating the release notes automatically.
- However, the Rust runtime is not copied into the LF jar anymore. It is also not unpacked by LFC manually anywhere, instead we leave it to Cargo to find a place.
- There is a new version-controlled file `org.lflang/src/lib/rs/runtime-version.properties` that contains the version of the rust runtime. This is necessary to have the version available at runtime, so that LFC can ask Cargo to fetch a "known good" runtime. The file should be kept in sync with the submodule.
- There are two new gradle tasks 
  - `gradle checkRuntimeVersionFileUpToDate` checks that the file is in sync with the submodule. That's meant for CI.
  - `gradle updateRustRuntime` updates the file to match the current state of the submodule. This is for developer convenience and needs to be called manually (it's not part of buildAll or so).


cmnrd (commented):
I see one problem with this: This does not work with the maven build, right? Hence the generated file would not be available when generating code from Epoch.

oowekyala:
> I see one problem with this: This does not work with the maven build, right? Hence the generated file would not be available when generating code from Epoch.

Right, I had forgotten that... Maybe one solution would be to put the logic to get the revisions into a bash script and then invoke it from maven with ant... I'm not sure whether tycho would allow that...

oowekyala:
Ok this now has the following changes (compared to master):
- The Rust submodule stays in the repo currently. It can be practical for development, and it also allows generating the release notes automatically.
- However, the Rust runtime is not copied into the LF jar anymore. It is also not unpacked by LFC manually anywhere, instead we leave it to Cargo to find a place.
- There is a new version-controlled file `org.lflang/src/lib/rs/runtime-version.properties` that contains the runtime version of the rust runtime. This is necessary to have the version available at runtime, so that LFC can ask Cargo to fetch a "known good" runtime. The file should be kept in sync with the submodule.
- There are two new gradle tasks 
  - `gradle checkRuntimeVersionMatches` checks that the file is in sync with the submodule. That's meant for CI.
  - `gradle updateRustRuntime` updates the file to match the current state of the submodule. This is for developer convenience and needs to be called manually (it's not part of buildAll or so).

lhstrh:
For `reactor-ts`, I've actually removed as a submodule, the reason being that we started doing releases on our own, with separate release notes. Leaving `reactor-ts` as a submodule means that we will publish release notes of it in the `lingua-franca` repo. Is that the intended behavior?

oowekyala:
> For reactor-ts, I've actually removed as a submodule, the reason being that we started doing releases on our own, with separate release notes. Leaving reactor-ts as a submodule means that we will publish release notes of it in the lingua-franca repo. Is that the intended behavior?

Tbh the rust runtime crate is not particularly stable right now, so I'm a bit uncomfortable with publishing semantically versioned releases just yet. We would be bumping the major release number every few weeks. I think we should reconsider this later, I'll open a ticket to track this

cmnrd (commented):
This looks good to me! I noticed that the new gradle task for checking is called `checkRuntimeVersionFileUpToDate`, not `checkRuntimeVersionMatches`.

We also need to updates the CI scripts to call the new gradle checks. I had a quick look at this, but not sure yet where to add it. Probably to the rust tests? While looking at the Rust tests, I also noticed that we need to update the mechanism that is used to enforce a specific runtime version when the workflow is called from the reactor-rs repo.

Could you also update the PR description?

oowekyala:
Thanks for bringing this up with the reactor-rs CI. I don't think the checkRuntimeVersionFileUpToDate task will work when the workflow is called with a custom runtime-ref (like the CI of reactor-rs does)...

The two builds check different things
- The build of lingua-franca uses the submodule version that is checked in in this repo. It should run tests with the runtime version we will write in the lfc jar (so it needs the submodule version to be in sync with the runtime version file). In that case it makes sense to run the check.
- The build of reactor-rs checks that the Rusts tests of the lingua-franca repo pass when run with a newer runtime version. The runtime version is then out of sync, but that's on purpose. We should be able to build a newer version of reactor-rs, then later update the version that is used by LFC.

   I think we should disable the check that the versions match in this case. But that's not enough to make lfc link to the newer runtime version. Possibly, the build should set the environment variable `LOCAL_RUST_REACTOR_RT` to point to a submodule and ignore the runtime version file.

oowekyala:
See also lf-lang/reactor-rs#38

PR #1485, opened 2022-11-22
Fix tracing

lsk567:
This PR fixes the Makefile for building the trace tools.

### DIFF
--- a/util/tracing/makefile
+++ b/util/tracing/makefile
@@ -3,7 +3,11 @@
 # @author: Edward A. Lee
 
 CC=gcc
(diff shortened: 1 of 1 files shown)


cmnrd (commented):
Shouldn't the trace tools for C rather be located in the reactor-c repository? Or maybe in its own repo? They are only useful for the C target and also only really interesting for us developers (at least at the moment). Also it seems pretty weird to me, to have include paths pointing into our LF source tree.

Note that the tools for C++ tracing are currently also located in the reactor-cpp repo. Also

edwardalee (approved):
LGTM.  I agree that this should be in reactor-c, but I suggest that be done as a separate PR.

Issue #1460, opened 2022-11-05
Weird build errors occur when `src-gen` is not deleted prior to recompiling


petervdonovan:
Example: https://github.com/lf-lang/reactor-c/issues/121

This does not occur every time. Perhaps it only happens when you switch to a different submodule version or when the previous build failed, but definitely there are various build errors that happen only when the `src-gen` directory is not deleted before compiling.

This issue was probably introduced by #1299 and its companion PRs. Originally I wanted to pretend that everything was fine and say that having to always delete the `src-gen` directory is normal, but whenever something goes wrong this bug is just one more thing that everyone has to check, so it can be a time sink.

My best guess is that the fix will involve adding a clean step in `CCompiler.java`. This should be done carefully because deleting things unnecessarily can make builds slower by making it impossible to reuse old results (although this effect should be measured -- I don't understand everything that CMake is doing).

axmmisaka:
I discussed with professor @edwardalee today and it appears that the issue typically happens when recompiling after `lfc` is recompiled, which is, imo, expected (I don't know but I guess if you `make` a project partly with`gcc` and the rest with a fork it's not guaranteed to work); 
My plan is to compare the jar file timestamp with the generated file timestamp and recompile if the jar is newer, but your input will be appreciated.

petervdonovan:
> the issue typically happens when recompiling after lfc is recompiled, which is, imo, expected 

Oh, great! I hope that's all it is. Yes, if that plan is not hard to implement, then it sounds like a good idea.

cmnrd:
Can you elaborate on why it is expected that the build breaks after lfc is rebuild? Unless the bundled reactor-c or the way we generate cmake files changed, rebuilding with cmake should not be a problem. And even then most changes should not cause problems. Only if for instance a file is deleted or moved in reactor-c, I would expect problems, as then the source tree in src-gen differs from the expected tree. Note that the C++ code  is build in a very similar way and I haven't seen such problems in quite a while.

Is there a way to reproduce this error? And have you analyzed why the build fails? It seems to me that there is another problem lurking here and deleting src-gen probably just masks it instead of solving it.

petervdonovan:
[...]

PR #1490, opened 2022-11-22
Avoid reading the same input stream twice

cmnrd:
This fixes a nasty bug in FileUtil, where an input stream was read twice while copying. The first read was to check if the destination file needs to be modified, and the second read was to actually perform the copy. Since input streams can only be read once, the second read would then copy an empty byte array and hence write an empty file.

This likely fixes https://github.com/lf-lang/lingua-franca/issues/1460


edwardalee (approved):
Good catch!  Thanks!

petervdonovan (approved):
Thanks Christian!

PR #1493, opened 2022-11-24
Make sure federated RTI Porperties given in the command line override…

ChadliaJerad:
… those in the .lf file, even when 'localhost' is passed. This fix does not use 'null' as default value. Therefore, the previously implemented logic does not break.

### DIFF
--- a/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
+++ b/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
@@ -94,10 +94,14 @@ public class FedGenerator {
     private final Map<Integer, FederateInstance> federateByID = new LinkedHashMap<>();
     /**
      * The federation RTI properties, which defaults to 'localhost: 15045'.
+     * The "overridden" Pair guarantees following user directives, without setting
+     * the default values to be "null". A particular use case is when the lf file
+     * sets RTI properties, and the command line overrides these.
      */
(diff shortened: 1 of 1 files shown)


org.lflang/src/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -94,10 +94,14 @@ public synchronized void report(int id, int x, Procedure1<Integer> callback) {
...
      * The federation RTI properties, which defaults to 'localhost: 15045'.
+     * The "overridden" Pair guarantees following user directives, without setting
+     * the default values to be "null". A particular use case is when the lf file
+     * sets RTI properties, and the command line overrides these.
      */
     final LinkedHashMap<String, Object> federationRTIProperties = CollectionLiterals.newLinkedHashMap(
         Pair.of("host", "localhost"),
-        Pair.of("port", 0) // Indicator to use the default port, typically 15045.
+        Pair.of("port", 0), // Indicator to use the default port, typically 15045.
+        Pair.of("overridden", false)
```
edwardalee, line 104: Perhaps this should be named "host_overridden"? It isn't clear which property it is referring to.

edwardalee (approved):
Looks OK to me, modulo the comment.

ChadliaJerad:
No longer useful.

PR #1497, opened 2022-11-30
Spotless check in CI

petervdonovan:
This serves as a regression test for the automatic code formatting workflow. The CI step should also fail if modified files have code formatting violations.


lhstrh (approved):
LGTM!

cmnrd (approved):
Great! If we want to enforce code formatting in the future, we could also add it to this CI job.

PR #1498, opened 2022-12-06
Correctly report assertion errors and exceptions during code generation

cmnrd:
While working on https://github.com/lf-lang/lingua-franca/pull/1441, I noticed that assertion errors during code generation are not correctly reported by our test framework. This is because the TestBase used AssertionError to indicate that a test should be aborted  without additional reporting. This change instead introduces a custom exception type for aborting tests and makes sure that assertion errors and other exceptions are properly reported.


org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -467,19 +468,28 @@ protected void addExtraLfcArgs(Properties args) {
...
             generator.doGenerate(test.fileConfig.resource, fileAccess, test.context);
-            result = test.context.getResult();
-            if (generator.errorsOccurred()) {
-                test.result = Result.CODE_GEN_FAIL;
-                throw new AssertionError("Code generation unsuccessful.");
-            }
+        } catch (AssertionError | Exception e) {
+            test.result = Result.CODE_GEN_FAIL;
+            // Add the stack trace to the test output
+            e.printStackTrace(new PrintWriter(test.compilationLog, true));
```
oowekyala, line 484: It would be cleaner to add a method `handleCodeGenException(Throwable)` to LFTest that does this, instead of making the compilation log public.

Also this logic would probably better be moved to the parent method, and here a TestExecutionException be thrown that has the exception `e` as its cause

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -435,7 +436,7 @@ private LFGeneratorContext configure(LFTest test, Configurator configurator, Tes
     /**
      * Validate the given test. Throw an AssertionError if validation failed.
```
oowekyala, line 437:
```suggestion
     * Validate the given test. Throw an TestExecutionException if validation failed.
```

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -447,11 +448,11 @@ private void validate(LFTest test, IGeneratorContext context) {
                     test.result = Result.VALIDATE_FAIL;
                 }
             }
-        } catch (Exception e) {
+        } catch (Exception | AssertionError e) {
```
oowekyala, line 451: I don't think this works properly. We expect the exception to be thrown by line 442 here (`validator.validate`). If that happens, then the test log will remain empty because the exception is ignored here

I think this catch block should throw a TestExecutionException with the `e` exception here as a cause, and the printing be handled in the parent block

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -696,11 +706,9 @@ private void validateAndRun(Set<LFTest> tests, Configurator configurator, TestLe
                     test.result = Result.TEST_PASS;
                 }
 
-            } catch (AssertionError e) {
-                // Do not report assertion errors. They are pretty printed
-                // during reporting.
-            } catch (Exception e) {
-                test.issues.append(e.getMessage());
+            } catch (TestExecutionException e) {
```
oowekyala, line 709: Nit: the `execute` phase of the test is the only one that will not throw a TestExecutionException, which seems to contradict the naming.

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -696,11 +706,9 @@ private void validateAndRun(Set<LFTest> tests, Configurator configurator, TestLe
...
+                // This exception is thrown by any of the test steps above. We mainly use the exception
+                // to abort test execution. There is no need to report the error here.
```
oowekyala, line 711: This is where the exception should be handled though, so that failures in any of the substeps are handled uniformly.

org.lflang.tests/src/org/lflang/tests/TestBase.java
cmnrd: Do you have a name proposal? I chose the name to indicate an error during the execution of one of our tests, where executing a test is comprised of validating the code, generating target code, compiling and running the binary.

org.lflang.tests/src/org/lflang/tests/TestBase.java
petervdonovan: I think that I have run into issues with the test framework swallowing exceptions before. If a message were printed here, then there would be less risk that the code will change slightly and then start swallowing exceptions again.

petervdonovan (approved):
I left a minor comment, but mostly this looks good to me.

lhstrh:
This looks good to me too, but the error in ccp-tests should be looked at. Unrelated to the changes in the PR, I'm sure, but worrisome nonetheless.

cmnrd:
I reworked this completely. The exception `TestError` is now the one and only mechanism for reporting problems during test execution. This also allowed to improve the general design of LFTest and make all its internal state private.

Thanks @oowekyala for your review comments! I was a bit annoyed at first because it means more work, but I think the new design is much more robust and there should be an error message printed regardless of why the test failed.

lhstrh:
This will have some significant conflicts with what I'm doing in `fed-gen`. Specifically, I removed `fileConfig` from `LFTest` and made it accessible through the `context`. I'll put in a review when I can.

org.lflang.tests/src/org/lflang/tests/TestRegistry.java
```diff
@@ -386,32 +386,22 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) {
...
+                // Try to parse the file.
+                Resource r = rs.getResource(URI.createFileURI(path.toFile().getAbsolutePath()),true);
                 // FIXME: issue warning if target doesn't match!
                 LFTest test = new LFTest(target, path);
-                EList<Diagnostic> errors = r.getErrors();
-                if (!errors.isEmpty()) {
-                    for (Diagnostic d : errors) {
-                        test.issues.append(d.toString());
-                    }
-                    test.result = Result.PARSE_FAIL;
```
petervdonovan, line 401: It looks like this functionality was removed without being implemented anywhere else. Any particular reason for this?

petervdonovan (changes requested):
I left a minor comment. If I was mistaken or if it gets fixed, then I think this is good to merge.

petervdonovan:
Looks like yet another segfault in `AsyncCallback`. That shouldn't stop this from being merged since the issue was not created by this PR.

org.lflang.tests/src/org/lflang/tests/TestRegistry.java
cmnrd: The parser check was duplicated before. It still exists in `configure()` in `TestBase`. The reason the check is removed here, is that we cannot throw anything except from `IOException` from `visitFile`.

org.lflang.tests/src/org/lflang/tests/LFTest.java
```diff
@@ -128,19 +142,39 @@ public void reportErrors() {
...
+        result = e.getResult();
+        if (e.getMessage() != null) {
+            issues.append(e.getMessage());
+        }
+        if (e.getException() != null) {
+            issues.append(System.lineSeparator());
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
+            e.getException().printStackTrace(pw);
+            issues.append(sw);
```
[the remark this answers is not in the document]
cmnrd: Good point. Thanks! I just pushed a fix

cmnrd:
> This will have some significant conflicts with what I'm doing in `fed-gen`. Specifically, I removed `fileConfig` from `LFTest` and made it accessible through the `context`. I'll put in a review when I can.

I am happy to help resolve any merge conflicts. I feel like this refactoring was needed and it should actually make it easier to implement changes in the future.

petervdonovan (approved):

cmnrd:
@lhstrh Is it Ok if I merge this? I can help to merge these changes into fed gen.

lhstrh (approved):

PR #1499, opened 2022-12-08
Extension cpp lsp

Aldenysq:


petervdonovan:
Maven Tycho is not happy:
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.6.0:compile (default-compile) on project org.lflang.diagram: Compilation failure: Compilation failure: 
Error:  /home/runner/work/lingua-franca/lingua-franca/org.lflang.diagram/src/org/lflang/diagram/lsp/CppLanguageServer.java:[14] 
Error:  	import com.fasterxml.jackson.databind.ObjectMapper;
Error:  	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error:  Access restriction: The type 'ObjectMapper' is not API (restriction on classpath entry '/home/runner/.m2/repository/p2/osgi/bundle/com.fasterxml.jackson.core.jackson-databind/2.13.2.v20220426-1653/com.fasterxml.jackson.core.jackson-databind-2.13.2.v20220426-1653.jar')
```

I think this makes no sense. I'm pretty sure `ObjectMapper` is API.

Is there a way we can exclude this code from the Tycho build? Maybe in the same way as the way we exclude Kotlin? Epoch does not use this code.

org.lflang.diagram/src/org/lflang/diagram/lsp/LFLanguageServer.java
```diff
@@ -18,17 +20,19 @@ public class LFLanguageServer extends KGraphLanguageServerExtension {
...
-        try {
-            return super.hover(params, cancelIndicator);
-        } catch (IndexOutOfBoundsException e) {
-            return IHoverService.EMPTY_HOVER;  // Fail silently
+        if (!initialized) {
+            CppLanguageServer.init();
+            initialized = true;
         }
+        return CppLanguageServer.hoverRequest(params);
+        // return IHoverService.EMPTY_HOVER;  // Fail silently
```
petervdonovan, line 36:
```suggestion
        try {
            if (!initialized) {
                CppLanguageServer.init();
                initialized = true;
            }
            return CppLanguageServer.hoverRequest(params);
        } catch (IndexOutOfBoundsException e) {
            return IHoverService.EMPTY_HOVER;  // Fail silently
        }
```
The old code which got deleted in this PR does "Fail silently," as it says, which is of course very bad. However, I want to keep it around until we move away from Xtext. See https://github.com/lf-lang/vscode-lingua-franca/issues/63 (and the comment about the "hacky little patch") for background

petervdonovan: Oh, wait, never mind. If we don't use the default hover method, we don't have the `IndexOutOfBoundsException` from the framework, so that comment was not exactly right. But we do use the default hover method if we are not hovering over C++ code. So actually, the right thing to do is to use the old hacky thing only if we are not hovering over C++.

Aldenysq:
Commit mostly works except it highly depends on a user using "lf build" all the time, but not pressing Command+S which saves a file. If a user presses Command+S, the current file that a user is working on does not produce any output anymore (I assume the issue is somewhere where the file is closed and reopened).
Also, if some src-gen directory is deleted, feature stops working.
Steps to reproduce:
- need to install clang
- compile the project
- Open .lf file, it should start building once the file is open
- Then the feature should work, 
- 1. try deleting src-gen files and rebuilding again
- 2. try pressing command+s, in both cases it stops working.
There is also a bug that if a user has backslash in their code, the output is not consistent, but it is due to not proper serialization when sending json request and will be fixed soon. I should also think about other special characters that might appear in code that would not properly interact with json request.

Issue #1342, opened 2022-08-31
`LFFormatter` is slow


petervdonovan:
This is an issue extracted from the review of #1227.

Code formatting should be fast. This is feasible because formatting should be linear in the length of a file, and file lengths should always be bounded in practice (e.g., no more than a few thousand lines). It is necessary because it is common for programmers to set VS Code to "format on save" so that the code on their screen is never ugly for long, even if they are too lazy to format anything by hand.

However, the algorithm used in `LFFormatter` is not fast, nor is it even linear in the length of a file. It has undergone zero performance optimization because it is good enough to format the toy examples in our test suite and because when I wrote it, I was primarily concerned with rapidly setting an example of how LF files should be formatted. It should be easy to optimize the algorithm and at least give it an acceptable time complexity.

If we cannot make the algorithm faster than a typical human reaction time for all non-pathological files, then we should provide a way to cancel the formatting process in the IDE. This is easy to do using the cancel indicator object.

petervdonovan:
While working on #1389, I noticed that the `spotlessLinguaFrancaApply` task can take quite a while. I found myself waiting many minutes before any changes became visible in the file system. This can be frustrating if you don't know what is going on. This is probably either because I invoke the LFF jar and start up the JVM for each test or because I did not worry about about time complexity in the formatter. The person-hours saved by taking these shortcuts were worth it for me.

petervdonovan:
Update -- I think the real problem is that the Spotless format task restarts the jar for every file, which is bad. This should be an easy fix.

petervdonovan:
A fix for this is in progress. Until it is merged I recommend to
* use `./gradlew spotlessJavaApply` to format only the Java files using Spotless
* use `lff` to format the LF files that were modified. IIRC `lff` accepts globs.

cmnrd:
Can we configure spotless such that it does not format the LF files when running the gradle task? I mean, we do have lff and if it is the preferred method for formatting files is lff, then why have a redundant gradle task? We can quite as well run `./gradlew runLff --args .`

petervdonovan:
Spotless provides some features, such as ensuring that the formatter is idempotent and checking that files are formatted, without actually formatting them. In order for these to work, we have to plug the LF formatter into Spotless.

PR #1500, opened 2022-12-08
Fix display of MalleableStrings.

petervdonovan:
Fixes an issue reported by @edwardalee today which results in the following diagram:
image

### DIFF
--- a/org.lflang/src/org/lflang/ast/MalleableString.java
+++ b/org.lflang/src/org/lflang/ast/MalleableString.java
@@ -99,6 +99,15 @@ public abstract class MalleableString {
         return ret;
     }
 
+    @Override
+    public String toString() {
+        List<String> temp = comments;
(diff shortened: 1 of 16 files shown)


lhstrh (approved):
Now that's what I call a quick fix :rocket:

Formatting tests are wreaking havoc though...

petervdonovan:
> Formatting tests are wreaking havoc though...

This change uncovered a subtle bug in the formatting logic. It is not very good that the formatter has so many subtle bugs. A fix is in progress but will involve reformatting several tests, which takes time (#1342).

petervdonovan:
The test failure in the last commit was probably nondeterministic given that it a) did not appear in the commit before it and b) seems to be unrelated to any changes anywhere in this branch. I think it is pragmatic to merge this branch despite the test failure.

PR #1501, opened 2022-12-08
CI formatting check disabled

petervdonovan:
See discussion on Zulip. People might have different opinions about this, but any decision is fine with me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,9 +26,10 @@ jobs:
     needs: cancel
 
   # Check that automatic code formatting works.
-  format:
-    uses: lf-lang/lingua-franca/.github/workflows/format.yml@master
-    needs: cancel
+  # TODO: Uncomment after fed-gen is merged.
+  # format:
+  #   uses: lf-lang/lingua-franca/.github/workflows/format.yml@master
+  #   needs: cancel
 
   # Run the unit tests.
   unit-tests:


lhstrh:
I agree that failures are a bit harsh (and annoying). Maybe we should print warnings in the summary instead?

lhstrh:
Using `echo "::warning file={name},line={line},endLine={endLine},title={title}::{message}"`

petervdonovan:
Well, if they are warnings, then they will get ignored, and they will accumulate over time -- this is what happened in the C runtime. I don't think it's a problem to strictly enforce code formatting. If it is something that people just always do (it's just one command), then it will not take much time and we will finally be able to stop talking about it.

lhstrh:
I've just suggested an alternative approach in Zulip. It is to instate a bot that reacts to any pushes to master and applies formatting and files a PR with changes if necessary. That will get the job done and separates the formatting stuff clearly from the semantic changes.

petervdonovan:
> I've just suggested an alternative approach in Zulip. It is to instate a bot that reacts to any pushes to master and applies formatting and files a PR with changes if necessary.

If anyone has time to set that up, then I'm all for it -- my ultimate goal is to spend less time on formatting. But I think Christian's point still stands that we do not want to make big outstanding PRs like `fed-gen` (or maybe the 6000 line PR that Shaokai has) unmergeable.

lhstrh:
I can set it up. It going through an auxiliary PR gives us the ability to moderate and punt on formatting to accommodate outstanding PRs.

cmnrd (approved):
Thanks! should we merge this then?

Issue #1496, opened 2022-11-30
Precisely specify the behaviour of timeout and implement it correctly in all targets.


cmnrd:
While trying to fix https://github.com/lf-lang/lingua-franca/issues/725 for C++, I realized that we don't have a clear definition of timeout and targets implement this differently. The website says:
> The [target property timeout](https://www.lf-lang.org/docs/handbook/target-declaration#timeout) specifies the last logical time at which reactions should be triggered. The last invocation of reactions will be at tag (timeout, 0).

My interpretation of this sentence is that the shutdown reaction should be invoked precisely at tag (timeout, 0). 

In C++, currently the last tag is (timeout, 1). In C, (timeout, 0) is only the last tag if keepalive is enabled or there is another event scheduled at (timeout, 0). Otherwise, shutdown will be invoked at the last scheduled event with a tag smaller than (timeout, 0).

How is this implemented in the other targets and what should be the correct behavior?

edwardalee:
To me, the word "timeout" implies a bound, a maximum. I think that if keepalive == false, it is ok to end execution early and invoke shutdown reactions. It would seem strange to me to keep a program running while it can't do anything. 

As for the micostep, since the time is known in advance, I see no reason to postpone execution to microstep 1.

cmnrd:
The microstep 1 in C++ is an artifact of how timeout is currently implemented in C++. #725 already says that this should be changed.

> To me, the word "timeout" implies a bound, a maximum. I think that if keepalive == false, it is ok to end execution early and invoke shutdown reactions. It would seem strange to me to keep a program running while it can't do anything.

I agree to this for programs that do not schedule events beyond the specified timeout (For instance if there are only startup reactions or timers without a period). However, for programs that would run indefinitely without a timeout (like any program that uses a periodic timer), I would expect the shutdown reactions to be invoked precisely at the timeout. However, in C, a program with a timer t(500ms, 1s) and timeout=2s terminates at tag (1500ms, 0), which seems odd to me.

lhstrh:
@edwardalee, I agree with the intuition, but the wording in the documentation doesn't convey this meaning. How about we rephrase it as follows:
[...]

PR #1507, opened 2022-12-16
C++: use a native implementation of timeout

cmnrd:
This pulls in https://github.com/lf-lang/reactor-cpp/pull/36 which extends reactor-cpp with a native timeout handler. This PR adjusts the code generator accordingly and removes the old Timeout mechanism. The timeout policy implemented by this PR is to invoke shutdown precisely at tag  (timeout, 0).

Note that this PR does not provide a solution for https://github.com/lf-lang/lingua-franca/issues/1496 as further discussion on the semantics of timeout is required. Thus, PR is only the first step and provides a simple timeout implementation as this is needed to move forward with the enclave feature. We can refine the policy in future PRs.

Fixes https://github.com/lf-lang/lingua-franca/issues/725


org.lflang/src/lib/cpp/time_parser.hh
```diff
@@ -87,30 +107,30 @@ std::stringstream &operator>>(std::stringstream& in, reactor::Duration& dur) {
...
-               unit == "minute" || unit == "minutes") {
+               unit == "minute" || unit == "minutes" || unit == "m") {
       std::chrono::duration<double, std::ratio<60, 1>> tmp{value};
       dur = std::chrono::duration_cast<reactor::Duration>(tmp);
-    } else if (unit == "hour" || unit == "hours") {
+    } else if (unit == "hour" || unit == "hours" || unit == "h") {
       std::chrono::duration<double, std::ratio<3600, 1>> tmp{value};
       dur = std::chrono::duration_cast<reactor::Duration>(tmp);
-    } else if (unit == "day" || unit == "days") {
+    } else if (unit == "day" || unit == "days" || unit == "?") {
```
oowekyala, line 130:
```suggestion
    } else if (unit == "day" || unit == "days" || unit == "d") {
```

org.lflang/src/lib/cpp/time_parser.hh
```diff
@@ -131,6 +151,8 @@ std::string validate_time_string(const std::string& time) {
     return "The empty string is not a valid time!";
   } else if (trimmed[0] == '-') {
     return "Negative values are not a valid time!";
+  } else if (iequals("forever", time)) {
+    return ""; // "forever" is a valid value
```
oowekyala, line 155: Is the value "forever" defined by LF or by the C++ target only? It's the first time I see this.

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -123,14 +120,17 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
...
-    private fun assembleReaction(reaction: Reaction) = with(PrependOperator) {
-        """
+    private fun assembleReaction(reaction: Reaction): String {
+        val sources = reaction.sources.filter { it.variable is Port }
+        return with(PrependOperator) {
+            """
             |// ${reaction.name}
         ${" |"..reaction.triggers.joinToString(separator = "\n") { declareTrigger(reaction, it) }}
-        ${" |"..reaction.sources.joinToString(separator = "\n") { declareDependency(reaction, it) }}
+        ${" |"..sources.joinToString(separator = "\n") { declareDependency(reaction, it) }}
```
oowekyala, line 129: Can one not use an action as a source? Or is it handled somewhere else?

cmnrd: Yes, actions can be sources, but we don't need to tell this the runtime (the runtime only keeps track of the reaction that an action triggers). They are only brought into scope elsewhere.

org.lflang/src/lib/cpp/time_parser.hh
cmnrd: Forever is also used in the C target. I am not sure if it is specified anywhere to be honest.

org.lflang/src/org/lflang/generator/cpp/CppExtensions.kt
```diff
@@ -1,13 +1,7 @@
...
 
 import org.eclipse.emf.ecore.resource.Resource
-import org.lflang.InferredType
-import org.lflang.TargetProperty
-import org.lflang.TimeUnit
-import org.lflang.indexInContainer
-import org.lflang.isBank
-import org.lflang.isGeneric
-import org.lflang.isMultiport
+import org.lflang.*
```
lhstrh, line 4: I believe we don't allow wildcard imports in our style guide.

lhstrh (approved):
:+1:

org.lflang/src/org/lflang/generator/cpp/CppExtensions.kt
cmnrd: For Kotlin we refer to the [Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html) which don't specify this. Wild card imports are quite idiomatic in Kotlin and are applied automatically by the IntelliJ auto-formatter.

PR #1508, opened 2022-12-19
Provide an infrastructure for AST ransformations and factor out the after delay transformation

cmnrd:
The main motivation of this PR is to apply the after delay transformation only optionally when needed by the target.

Currently, the after delay transformation is handled in between some other important  passes in `GeneratorBase` and cannot easily be moved such that it can be called directly by our code generators before or after invoking `GeneratorBase`. Therefore, this PR introduces a `Transformation` interface which allows the target generators to register transformations which are then automatically applied.

This PR implements the `AfterDelayTransformation` and factors out the target specific code into a new interface `DelayBodyGenerator` which the targets need to provide.


marcusrossel (approved):

org.lflang/src/org/lflang/ast/AstTransformation.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Interface for AST Transfomations
+ */
+public interface AstTransformation {
+
+    /**
+     * Apply the AST transformation to all given reactors.
+     */
+    void applyTransformation(List<Reactor> reactors);
```
lhstrh, line 15: I would change this to take in the entire resource so that the interface is more general.

org.lflang/src/org/lflang/generator/cpp/CppDelayBodyGenerator.kt
```diff
@@ -0,0 +1,46 @@
...
+        // forward body for ${action.name}
+        lfutil::after_forward(&${action.name}, &${port.name});
+        """.trimIndent()
+    }
+
+    override fun generateDelayGeneric() = "T"
+
+    override fun generateAfterDelaysWithVariableWidth() = false
+
+}
```
lhstrh, line 46:
```suggestion
}

```

org.lflang/src/org/lflang/generator/ts/TSDelayBodyGenerator.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun generateForwardBody(action: Action, port: VarRef): String {
+        return "${ASTUtils.generateVarRef(port)} = ${action.name} as ${getActionType(action)};"
+    }
+
+    override fun generateDelayGeneric(): String {
+        return "T extends Present"
+    }
+
+    override fun generateAfterDelaysWithVariableWidth() = false
+}
```
lhstrh, line 37:
```suggestion
}

```

lhstrh (approved):
I think this looks great! Thanks for doing this! Aside from making the transformation optional, this code cleanup was much overdue. I've left some minor comments.

org.lflang/src/org/lflang/ast/AstTransformation.java
cmnrd: Yes, my first intuition was also to pass the model and/or the resource. However, it appears that there is some intelligence built into generating this list of reactors (this is handled by the instance graph). We would need to duplicate that code or refactor it to make it reusable, but I wanted to avoid spending too much time on this. Also I am not sure which information other transformations would need. I suggest to revisit this as part of fixing https://github.com/lf-lang/lingua-franca/issues/1510

PR #1517, opened 2022-12-21
Make Cmake output less verbose by suppressing install messages

cmnrd:
With this change, the output produced by cmake when installing the project is reduced to this:
```
Install the project...
-- Install configuration: "Release"
```
Before this change, it would also print a verbose list of all installed files (which in the C++ target can be quite long). This list does not provide much useful information though and we might as well suppress it.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
@@ -164,6 +164,10 @@ public class CCmakeGenerator {
         cMakeCode.pr("endif()\n");
         cMakeCode.newLine();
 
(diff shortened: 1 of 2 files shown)


petervdonovan (approved):
I do sometimes use the CMake output for debugging C build problems, so it is not necessarily what I would personally prefer. However, I'm OK with this change if people like it.

cmnrd:
This does not remove the build output, only the install output that is printed after building. For the C target, this will only remove this line:
```
-- Installing: ~/projects/lf/lingua-franca/test/C/bin/Minimal
```
In the C++ target the output is a lot more verbose, but it does not provide useful information in both targets.

PR #1518, opened 2022-12-22
Preserve time units when formatting LF code

cmnrd:
Currently, the formatter changes time units to the "canonical" name as defined in `TimeUnit`. For instance, it would change `2 s` to `2 sec`. I think we should leave the decision of which time units to choose to the user. Personally (maybe because I am European), I prefer SI units. Which unit works the best might also depend on the target language. For instance, the C++ chrono library allows writing `2s`, but not `2sec`. So it would be idiomatic to use the SI units in LF-C++ programs.


petervdonovan (approved):
LGTM!

PR #1520, opened 2022-12-24
Use of enums instead of strings

### LINKED ISSUE #1481: Make generator property check to use `LFGeneratorContext.BuildParm` enum rather than strings
We should not use a string here but refer to the enum.

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1475#discussion_r1026799805_

Right now this type of code is [all over the place](https://github.com/search?q=repo%3Alf-lang%2Flingua-franca+getArgs%28%29.containsKey&type=code), it works the same but maintainability is probably much better (also allows static checking).

axmmisaka:
This hopefully resolves #1481, yet I wonder if writing a wrapper for `LFGeneratorContext`'s instances would be a better idea......


axmmisaka:
https://github.com/lf-lang/lingua-franca/pull/1520/files#diff-973c82d851bc08b9498f46e484445a06b70a57fb72f307e22eae4ceb41e1c18fR111-R115

These are unchanged because they don't exist anymore......

lhstrh:
> This hopefully resolves #1481, yet I wonder if writing a wrapper for `LFGeneratorContext`'s instances would be a better idea......

Could you elaborate on this suggestion? The changes you propose in this PR look reasonable to me...

lhstrh (approved):

axmmisaka:
> > This hopefully resolves #1481, yet I wonder if writing a wrapper for `LFGeneratorContext`'s instances would be a better idea......
> 
> Could you elaborate on this suggestion? The changes you propose in this PR look reasonable to me...

Sorry for the late late reply, I was travelling in Europe!
I was thinking about adding something like `context.containsParam(NO_COMPILE)`

lhstrh:
@axmmisaka, I probably messed up the imports when merging.

What I don't like about this solution is the verbosity of using `getKey()` all the time. I know that I'm responsible for that part of the design, but I wonder if there is a better way...

lhstrh:
Your wrapper idea addresses it in part, but someone not aware of the API might just use the enum straight up (without calling `getKey`) and find out that in some cases this doesn't work.

axmmisaka:
I'll check what happened; yet without calling `getKey()` I think it should error? I'll double check

lhstrh:
Yes, without getKey currently does not work.

PR #1521, opened 2022-12-29
Refactoring + fix Docker support

lhstrh:
This PR contains changes that have to do with the communication between code generators. The goal is to fix the Docker support, but the necessary structural changes are quite significant.


org.lflang/src/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -235,28 +242,39 @@ private Map<Path, CodeMap> compileFederates(LFGeneratorContext context, Map<Path
...
+                    }
+
                 };
-                Resource res = rs.getResource(URI.createFileURI(
-                    fileConfig.getFedSrcPath().resolve(fed.name + ".lf").toAbsolutePath().toString()
-                ), true);
+
                 gen.doGenerate(res, fsa, cont);
                 codeMapMap.putAll(cont.getResult().getCodeMaps());
+                // FIXME
```
lhstrh, line 275: Note: This is the hook where we are supposed to do the docker file generation.

petervdonovan:
I cannot locally reproduce the failing Python validation tests -- not by running them with the same random seed, nor by rebuilding the VS Code extension and manually introducing the same errors into the file. The errors are reported correctly, even in the federated test where the LSP tests have been failing in CI. I'm going to move on to other tests that are currently failing.

lhstrh:
I don't think it's super important to get everything to pass before we merge this into `fed-gen`. Let's merge this soon as we're convinced that what's here makes sense. We'll probably have quite a lot of merge conflicts to solve...

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -606,7 +607,7 @@ private List<ProcessBuilder> getFederatedDockerExecCommand(LFTest test) {
             System.out.println(Message.MISSING_DOCKER);
             return List.of(new ProcessBuilder("exit", "1"));
         }
-        var srcGenPath = test.context.getFileConfig().getSrcGenPath();
+        var srcGenPath = new FedFileConfig.FedFiles(test.context.getFileConfig()).getFedSrcGenPath();
```
lhstrh, line 610: I think the right fix is to remove `getFedSrcGenPath` and put its contents in an override of `getSrcGenPath`. WDYT?

petervdonovan: Yes, the way it is done now is a little weird. I am certainly open to refactoring, but I just wanted to first get to something that works.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -97,6 +99,8 @@ public abstract class FileConfig {
      */
     public final boolean useHierarchicalBin;
 
+    public final CppFileConfig.CppFiles cpp;
```
cmnrd, line 104: I don't think that the file config should contain target specific extensions. Why does this change from the previous design? I think the target specific file config classes should extend the base class. We can create the correct one when creating the Context, but only store a reference to `FileConfig`. Then we can still cast to the target specific version in the generator.

If that does not work seems like a bad design, we could alternatively make `FileConfig` and `CppFileConfig` completely separated classes. The C++ generator could then internally create an instance of `CppFileConfig`.

org.lflang/src/org/lflang/FileConfig.java
```diff
@@ -16,6 +16,8 @@
 
 
 import org.lflang.generator.GeneratorUtils;
+import org.lflang.generator.cpp.CppFileConfig;
```
cmnrd, line 21: This is not going to break the Eclipse build. Currently our java files cannot import from Kotlin without breaking the Eclipse build...

org.lflang/src/org/lflang/generator/cpp/CppFileConfig.kt
```diff
@@ -52,24 +52,26 @@ class CppFileConfig(resource: Resource, srcGenBasePath: Path, useHierarchicalBin
         outPath.resolve("share")
     )
 
-    /** Relative path to the directory where all source files for this resource should be generated in. */
-    private fun getGenDir(r: Resource): Path = getDirectory(r).resolve(r.name)
+    class CppFiles(val fileConfig: FileConfig) {
```
cmnrd, line 48: Why add a nested class?

org.lflang/src/org/lflang/generator/LFGenerator.java
```diff
@@ -54,50 +57,60 @@ public class LFGenerator extends AbstractGenerator {
...
+            if (FedASTUtils.findFederatedReactor(resource) != null) {
+                return new FedFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+            }
+            switch (target) {
+            case CCPP:
+            case C: return new CFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+            case Python: return new PyFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+            case CPP:
+            case Rust:
+            case TS:
```
cmnrd, line 80: It seems like we are still instantiating the target specific file configs here. So why are there all the changes to the target file configs?

cmnrd (commented):
I am a bit puzzled by the redesign of CppFileConfig and left a few comments. I want to highlight here that the changes break compilation from Eclipse (see my comment).

org.lflang/src/org/lflang/FileConfig.java
lhstrh: Yes, it will either have to be casting or creating a completely separate class. Do you have a preference for either?

petervdonovan:
This appeared in CI for the latest commit:
```
#### Bringing the RTI back to foreground so it can receive Control-C.
    RTI -i ${FEDERATION_ID} -n 2 -c init exchanges-per-interval 10
    Federate 1: ERROR: Received message too late. Already at stop tag.
    Current tag is (1000000, 0) and intended tag is (0, 0).
```
Unlike the other errors, which seem to be mostly build errors, this seems like a logical error in coordination with the RTI. At least one other test had an STP violation. Others had reactors receiving incorrect data. Just commenting about this because it seems like this bug could be a little deeper than the others.

org.lflang/src/org/lflang/FileConfig.java
lhstrh: Good catch. I've revised this in 1cc48459e.

org.lflang/src/org/lflang/FileConfig.java
lhstrh: Fixed in 1cc48459e.

org.lflang/src/org/lflang/generator/cpp/CppFileConfig.kt
lhstrh: Fixed in 1cc48459e.

org.lflang.tests/src/org/lflang/tests/TestBase.java
lhstrh: I've fixed this because I had to for another reason in 1cc48459e. I hope it still works.

lhstrh:
Let's merge this into `fed-gen` and continue there. We already have more tests passing here than in `fed-gen`.

org.lflang/src/org/lflang/generator/cpp/CppPreambleGenerator.kt
```diff
@@ -26,6 +26,7 @@ package org.lflang.generator.cpp
 
 import org.eclipse.emf.common.util.EList
 import org.eclipse.emf.ecore.resource.Resource
+import org.lflang.FileConfig
```
cmnrd, line 29:
```suggestion
```

org.lflang/src/org/lflang/generator/cpp/CppReactorGenerator.kt
```diff
@@ -24,6 +24,7 @@
 package org.lflang.generator.cpp
 
 import org.lflang.ErrorReporter
+import org.lflang.FileConfig
```
cmnrd, line 27:
```suggestion
```

org.lflang/src/org/lflang/generator/cpp/CppRos2NodeGenerator.kt
```diff
@@ -1,5 +1,6 @@
 package org.lflang.generator.cpp
 
+import org.lflang.FileConfig
```
cmnrd, line 3:
```suggestion
```

org.lflang/src/org/lflang/generator/cpp/CppStandaloneMainGenerator.kt
```diff
@@ -1,5 +1,6 @@
 package org.lflang.generator.cpp
 
+import org.lflang.FileConfig
```
cmnrd, line 3:
```suggestion
```

org.lflang/src/org/lflang/generator/cpp/CppValidator.kt
```diff
@@ -1,6 +1,7 @@
 package org.lflang.generator.cpp
 
 import org.lflang.ErrorReporter
+import org.lflang.FileConfig
```
cmnrd, line 4:
```suggestion
```

PR #1527, opened 2023-01-05
Refactor error reporter

oowekyala:
This has turned into a refactoring. The original line number "bugfix" is gone. I checked that the line numbers are correct now on the CLI, and I trust the tests to do so for the LSP.

The main change is to change the ErrorReporter interface to become a staged builder. It used to have a lot of different overloads with different ways to pass a position (either a file, or a file with a line, or two positions, or a node), multiplied by the different report levels (error, warning, info). These overloads were incomplete as not all combinations existed. The parameter order was sometimes inconsistent too, and implementations needed to implement a lot of methods, often by duplicating code.

The solution I'm proposing is to split the interface into two stages: the first stage specifies a position, the second specifies the message and report level. This makes it so that the number of methods to implement is drastically reduced, and at the same time all combinations of position type and report level are covered.

Concretely, instead of calling `errorReporter.reportError(node, "message")`, you now call 
```java
errorReporter.at(node).error("message")
```
There are multiple overloads of `at` for the different kinds of positions. The object returned by `at` has methods `error`, `warning`, `info`, and also a generic `report(DiagnosticSeverity, String)`.

The report methods do not return a string anymore, as this was barely used (I remember only a single usage in non-dead code).

If we don't have a position, we can call `nowhere().error("...")`. Some errors are legitimately not bound to a file, but other `nowhere` calls should be inspected to find out whether there is a node in scope on which we can sensibly report.

Nearly all `System.out.println` calls are replaced to calls to the `info` method for consistency. There are a couple which I couldn't port yet, eg in LFCommand. ErrorReporter is renamed MessageReporter to emphasize the fact that this is not only for errors, which may be why those System.out.println were there in the first place.


org.lflang/src/org/lflang/generator/rust/RustValidator.kt
```diff
@@ -103,39 +105,43 @@ class RustValidator(
         @JsonProperty("expansion") val expansion: RustSpanExpansion?
     ) {
         val start: Position
-            get() = Position.fromOneBased(lineStart, columnStart)
+            get() = Position.fromZeroBased(lineStart, columnStart)
```
oowekyala, line 108: This is a bugfix, previously Cargo messages were not associated with the correct line

oowekyala: Thanks for looking into this! I noticed this fixes things in the lfc cli, but maybe this was not the correct fix. I'll investigate tomorrow

petervdonovan: Interesting -- for the specific program that failed in CI, after reverting this change it looked like the line reported by `lfc` CLI was correct

lhstrh:
@oowekyala and @petervdonovan, what should we do with this PR? Looks like there is useful stuff in it, but it has fallen behind...

oowekyala:
I tried to update this, but I don't really know what this is about anymore... Formatting the files obscures the diff enormously. What's the deal with that? Should I open a separate PR with just formatting changes so that we see clearer? How do you deal with that in other PRs?

lhstrh:
Once the formatter is applied in the branch, it should be easier to do the merge... It looks like you've done the opposite, which may have been harder, but the resulting diff should be OK, is it not?

lhstrh:
@oowekyala, what shall we do with this PR?

core/src/main/java/org/lflang/generator/HumanReadableReportingStrategy.java
```diff
@@ -85,75 +84,66 @@ public void report(String validationOutput, ErrorReporter errorReporter, Map<Pat
...
           Position.fromOneBased(
               Integer.parseInt(matcher.group("line")),
-              Integer.parseInt(
-                  matcher.group("column") != null
-                      ? matcher.group("column")
-                      : "0") // FIXME: Unreliable heuristic
-              );
+              column == null
+                  ? 1 // FIXME: Unreliable heuristic
+                  : Integer.parseInt(column));
```
petervdonovan, line 103: Thanks for this bugfix.

petervdonovan (approved):
Thanks Clement! I agree that this is a cleaner design and that it is better to do this rather than to print directly to stdout.

PR #1482, opened 2022-11-19
Migration of Epoch into separate repository

a-sr:
This PR factors out all Eclipse (Epoch) specific projects and restructures the remaining source files such that they can be used as a sub module. The new repository is here: https://github.com/lf-lang/epoch
Implements parts of #1276.

Involved tasks (LF repo):
- [x] Move contents of `org.lflang.diagram` and  `org.lflang.ide` to `org.lflang`
- [x] Remove all Epoch related projects
- [x] Move `EclipseErrorRepoter` to Epoch repo
- [x] Adjust mwe build configuration
- [x] Document mwe build configuration implications (@a-sr)
- [x] Adjust Gradle build
- [x] Adjust GH build actions
- [x] Fix Gradle tests

Involved tasks (Epoch repo):
- [x] Include lingua-franca repository as submodule
- [x] Adjust Maven build
- [x] Adjust mwe build configuration
- [x] Setup GH build actions
- [x] Adjust Oomph setup
- [x] (not possible) Reduce redundancy in mwe configurations
- [x] Fix eclipse test (template check)
- [x] Implement Epoch in-tool compilation test
- [ ] Set repo to public 

Remaining tasks:
- https://github.com/lf-lang/vscode-lingua-franca/issues/100
- https://github.com/lf-lang/lingua-franca/issues/1613
- https://github.com/lf-lang/lingua-franca/issues/1614
- https://github.com/lf-lang/epoch/issues/3
- https://github.com/lf-lang/lf-lang.github.io/issues/128

PR #1530, opened 2023-01-07
Enable Epoch build for M1 Mac

hokeun:
I propose adding ARM 64 Mac OS X as an Epoch build environment to support M1 Mac.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,6 +248,11 @@
                             <ws>cocoa</ws>
                             <arch>x86_64</arch>
                         </environment>
+                        <environment>
+                            <os>macosx</os>
+                            <ws>cocoa</ws>
+                            <arch>aarch64</arch>
+                        </environment>
                     </environments>
                 </configuration>
             </plugin>


lhstrh:
Good idea! @a-sr has a major refactoring of the build system that is basically ready to go. Let's have him decide which should go first?

edwardalee (approved):
Looks good to me!

a-sr:
Since the refactoring is already done, PR #1482 will removed the changed file anyway. Hence, I would suggest to drop this PR to prevent a conflict with #1482, unless someone needs an Epoch IDE for an M1 Apple before we merge #1482.
In any case, I created an PR in the new epoch repo (https://github.com/lf-lang/epoch/pull/1) that will add M1 support for future Epoch versions.

a-sr (approved):

hokeun:
Thanks, everyone! I'm not sure about the timeline for merging https://github.com/lf-lang/lingua-franca/pull/1482, but I would like to suggest merging this for now as some people (including myself and some of my students/TAs who are preparing for the embedded systems class at Hanyang) are using M1 Mac. I checked https://github.com/lf-lang/lingua-franca/pull/1482, and it seems to simply remove this pom.xml file. Thus, I think this change shouldn't block https://github.com/lf-lang/lingua-franca/pull/1482. Please let me know if this causes any conflicts, I will be happy to help and revert this.

PR #1532, opened 2023-01-09
Support for Arduino CLI

arengarajan99:
With the new Arduino 2.0 IDE release, the functionality that allowed us to use CMake (the avr compiler) has been abstracted away, prompting us to move to Arduino-CLI, the backbone of the Arduino build system. Arduino-CLI has limitation with regards to how to compile multilayered projects, but allows for lazy compilation in a specified src folder. 

This PR uses Arduino specifications in the following workflow: LFC generated code is treated like an Arduino Sketch and the corresponding reactor-c files are placed in a src folder to be lazy compiled. To do this, I removed files that aren't used by Arduino that when compiled will cause redefinition errors and changed all includes using Regex to be relative links inside the src folder since Arduino doesn't support adding additional include paths. 

To try this:
 - Install Arduino CLI
 - Plug in your Arduino board
 - Run `arduino-cli board list` and record the Fully Qualified Board Name (FQBN) and port
 - Compile your Lingua Franca program with the `platform: {name: arduino, board: <FQBN>}` target property
 - Change directory to the generated sources, and run `arduino-cli upload -b <FQBN> -p <port>`

Features implemented:
 - [x] LF programs that specify `arduino` as the platform without specifying a `board` or `port` will only generate C code (as it works now), but print a message instructing the user how to obtain the information and provide it to LF
 - [x] For LF programs that do specify a `board` and `port`, `arduino-cli` is invoked automatically (unless the `no-compile` flag is used)
 - [x] As of yet, Arduino boards that run an MBED RTOS cannot handle physical actions. The compiler needs to reject programs with physical actions that target such boards.


.github/workflows/c-tests.yml
```diff
@@ -49,16 +49,13 @@ jobs:
...
+          brew install arduino-cli
         if: ${{ runner.os == 'macOS' }}
-      - name: Install Arduino Platform for Linux
+      - name: Install Arduino CLI for Linux
         run: |
-          wget "https://downloads.arduino.cc/arduino-1.8.19-linux32.tar.xz"
-          tar -xvf arduino-1.8.19-linux32.tar.xz
-          cd arduino-1.8.19/
-          sudo ./install.sh
+          curl -fsSL https://raw.githubusercontent.com/arduino/arduino-cli/master/install.sh | sh
```
lhstrh, line 58: It looks like this installs the latest `arduino-cli` from `master`, which could cause unexpected problems when something in their `master` branch changes that breaks compatibility with us. It might be better to pin `arduino-cli` to a specific version number.

org.lflang/src/org/lflang/generator/c/CPreambleGenerator.java
```diff
@@ -39,32 +39,37 @@ public static String generateIncludeStatements(
     ) {
         var tracing = targetConfig.tracing;
         CodeBuilder code = new CodeBuilder();
-        if (cppMode) {
+        if (cppMode || targetConfig.platformOptions.platform == Platform.ARDUINO) {
```
lhstrh, line 42: Shouldn't we just set `cppMode` accordingly if platform is `Arduino`?

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -383,6 +387,88 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
...
+    public static void relativeIncludeHelper(Path dir) throws IOException {
+        System.out.println("Relativizing all includes in " + dir.toString());
+        List<Path> allPaths = Files.walk(dir)
+            .filter(Files::isRegularFile)
+            .sorted(Comparator.reverseOrder())
+            .collect(Collectors.toList());
+        Map<String, Path> fileStringToFilePath = new HashMap<String, Path>();
+        for (Path path : allPaths) {
+            String fileName = path.getFileName().toString();
+            fileStringToFilePath.put(fileName, path);
```
lhstrh, line 450: Maybe check here whether the key already exists, in which case we've hit an error condition.

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -383,6 +387,88 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
...
+
+    private static String fileNameMatchConverter(String fileName, Path currPath, Map<String, Path> fileStringToFilePath) 
+        throws NullPointerException {
+        // First get the child file
+        int lastPath = fileName.lastIndexOf(File.separator);
+        if (lastPath != -1){
+            fileName = fileName.substring(lastPath+1);
+        }
+        Path p = fileStringToFilePath.get(fileName);
+        if(p == null){
```
lhstrh, line 428:
```suggestion
        i f(p == null) {
```

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -383,6 +387,88 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
...
+        for (Path path : allPaths) {
+            String toCheck = path.toString().toLowerCase();
+            if (toCheck.contains("cmake") || toCheck.contains("semaphore")
+            || path.endsWith("lib/util.c") || (toCheck.contains("core/platform/") && !toCheck.contains("lf_arduino"))) {
+                Files.delete(path);
+            } 
+        }
+    }
+
+    private static String fileNameMatchConverter(String fileName, Path currPath, Map<String, Path> fileStringToFilePath)
```
lhstrh, line 420: JavaDoc missing.

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -383,6 +387,88 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
...
+        Path p = fileStringToFilePath.get(fileName);
+        if(p == null){
+            return "#include \"" + fileName + "\"";
+        }
+        String relativePath = currPath.getParent().relativize(p).toString();
+        return "#include \"" + relativePath + "\"";
+    }
+
+    /**
+     * Converts all includes recursively inside files within a specified folder to relative links
```
lhstrh, line 436:
```suggestion
     * Convert all includes recursively inside files within a specified folder to relative links
```

org.lflang/src/org/lflang/util/FileUtil.java
```diff
@@ -383,6 +387,88 @@ private static void copyDirectoryFromJar(JarURLConnection connection, final Path
         }
     }
 
+    /**
+     * Deletes Unused Files from Arduino-CLI based compilation.
```
lhstrh, line 391:
```suggestion
     * Delete unused Files from Arduino-CLI based compilation.
```

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -609,21 +585,32 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
...
-                mainDef != null,
-                cMakeExtras,
-                targetConfig
-            );
-            try {
-                cmakeCode.writeToFile(cmakeFile);
-            } catch (IOException e) {
-                //noinspection ThrowableNotThrown,ResultOfMethodCallIgnored
-                Exceptions.sneakyThrow(e);
+            if (targetConfig.platformOptions.platform != Platform.ARDUINO){
```
lhstrh, line 588:
```suggestion
            if (targetConfig.platformOptions.platform != Platform.ARDUINO) {
```

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -874,7 +861,7 @@ private void generateCodeForCurrentFederate(
             // is set to decentralized) or, if there are
             // downstream federates, will notify the RTI
             // that the specified logical time is complete.
-            if (CCppMode) code.pr("extern \"C\"");
+            if (CCppMode || targetConfig.platformOptions.platform == Platform.ARDUINO) code.pr("extern \"C\"");
```
lhstrh, line 864: Idem.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -2199,13 +2191,19 @@ protected void setUpGeneralParameters() {
...
         }
         if (targetConfig.threading && targetConfig.platformOptions.platform == Platform.ARDUINO) {
-
             //Add error message when user attempts to set threading=true for Arduino
             if (targetConfig.setByUser.contains(TargetProperty.THREADING)) {
                 errorReporter.reportWarning("Threading is incompatible on Arduino. Setting threading to false.");
             }
             targetConfig.threading = false;
         }
+        if (!targetConfig.noCompile && targetConfig.platformOptions.platform == Platform.ARDUINO) {
```
lhstrh, line 2200: Why only do this if `!targetConfig.noCompile`?

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -609,21 +585,32 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
...
+                );
+                try {
+                    cmakeCode.writeToFile(cmakeFile);
+                } catch (IOException e) {
+                    //noinspection ThrowableNotThrown,ResultOfMethodCallIgnored
+                    Exceptions.sneakyThrow(e);
+                }
+            } else {
+                try {
+                    Files.deleteIfExists(fileConfig.getSrcGenPath().resolve("src/lib/util.c"));
```
lhstrh, line 607: Why is this happening, and why is `src/lib/util.c` mentioned separately?

lhstrh (changes requested):
Great work, @arengarajan99! I've left some minor comments.

.github/workflows/c-tests.yml
arengarajan99: This was on their how to install arduino-cli guide for non-Mac systems. I'll try to point it to the current version that works.

org.lflang/src/org/lflang/generator/c/CPreambleGenerator.java
arengarajan99: Wasn't sure if we should move away from cppmode since we no longer need the CCpp target (C target works since we are not relying on any building or CMake through LFC). I can modify it to default to CppMode for the other things though!

org.lflang/src/org/lflang/util/FileUtil.java
arengarajan99: Will do.

org.lflang/src/org/lflang/util/FileUtil.java
arengarajan99: Will add.

org.lflang/src/org/lflang/generator/c/CGenerator.java
arengarajan99: I'm trying to send a warning saying that the LFC won't compile for the Arduino Platform and that you must use Arduino-CLI to complete compilation.

org.lflang/src/org/lflang/generator/c/CPreambleGenerator.java
lhstrh: Ah, good point. I hadn't thought of that. What "other things" are you referring to though?

org.lflang/src/org/lflang/generator/c/CGenerator.java
arengarajan99: util.c is a conflicting filename, I need to remove it here before doing relative path changing.

arengarajan99: This contains deprecated functions that are no longer in use as well.

lhstrh: What's the conflict?

arengarajan99: The way I change relative paths is to create a map between the filename and its Path. There is util.c inside of lib and util.c inside of core/utils, which will create a conflict. We don't use util.c afaik since the functions are out of date, so I opted to take it out before relativizing paths.

org.lflang/src/org/lflang/generator/c/CPreambleGenerator.java
arengarajan99: Disregard, looks like CppMode is otherwise not needed for Arduino.

arengarajan99: The Or Conditions can be kept as is or I can force it to enter CppMode (which seems like a waste since there's certain places where I need to override CppMode code with custom Arduino Code)

org.lflang/src/org/lflang/generator/c/CGenerator.java
lhstrh: Oh, I see. I wasn't aware that we had two `util.c` files... Your solution sounds reasonable but we probably shouldn't have these conflicting files to begin with. A related issue, https://github.com/lf-lang/reactor-c/issues/111, came up recently. Maybe @erlingrj or @petervdonovan want to weigh in on this.

org.lflang/src/org/lflang/generator/c/CGenerator.java
lhstrh: I guess I was thrown off by the comment on line `2201`, which mentions `threading=true` which seems to have nothing to do with this.

org.lflang/src/org/lflang/generator/c/CPreambleGenerator.java
lhstrh: OK, fine to leave "as is" then.

org.lflang/src/org/lflang/util/FileUtil.java
arengarajan99: Added.

org.lflang/src/org/lflang/generator/c/CGenerator.java
erlingrj: We should change that. The `util.c` that you are talking about implements deprecated API functions. We could rename it to `depreacted.c`? Or something similar.

lhstrh: Please update `reactor-c` to `2553fc84899128a8ba44f3173c287bb986578661` and remove the special case of `util.c`. After that, this conversation can be marked "resolved."

lhstrh (changes requested):
Looks ready to me, modulo the remaining `util.c` issue. Will approve and merge once fixed.

lhstrh:
Oh, there are also a bunch of conflicts with `master` that are probably a result of the work done in `fed-gen`. Can you handle these, or should I take a look, @arengarajan99?

lhstrh (approved):

test/C/src/arduino/Fade.lf
```diff
@@ -5,7 +5,7 @@
  * the PWM pins are identified with a "~" sign, like ~3, ~5, ~6, ~9, ~10 and
  * ~11.
  */
-target CCpp {
+target C {
```
lhstrh, line 8: Why is this no longer valid?

lhstrh:
The Zephyr tests also broke, so maybe something went wrong with the merge. Maybe @erlingrj can help?

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -476,65 +477,22 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
...
-                StringBuilder s = new StringBuilder();
-                s.append("set(ARDUINO_BOARD \"");
-                s.append("\")");
-                FileUtil.writeToFile(s.toString(), fileConfig.getSrcGenPath().resolve("toolchain/BoardOptions.cmake"));
-            }
-
-
-                // If Zephyr target then copy default config and board files 
-                //  for Zephyr support
-                if (targetConfig.platformOptions.platform == Platform.ZEPHYR) {
```
erlingrj, line 519: The removal of the following lines causes the CI failure for Zephyr.

lhstrh: Brought these lines back. Let's see what happens in CI :-)

org.lflang.tests/src/org/lflang/tests/TestRegistry.java
```diff
@@ -149,12 +145,12 @@ public enum TestCategory {
         DOCKER(true),
         DOCKER_FEDERATED(true, "docker" + File.separator + "federated"),
         SERIALIZATION(false),
-        ARDUINO(false, TestLevel.BUILD),
+        ARDUINO(false),
```
lhstrh, line 148:
```suggestion
        ARDUINO(false, TestLevel.BUILD),
```

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -572,6 +577,47 @@ static public String stackTraceToString(Throwable t) {
...
+    }
+    
+    /**
+     * Return a ProcessBuilder used to test Arduino in a non-federated, unthreaded environment.
+     * See the following for references on arduino-cli:
+     * https://arduino.github.io/arduino-cli/
+     *
+     * @param test The test to get the execution command for.
+     */
+    private ProcessBuilder getArduinoExecCommand(LFTest test) {
```
lhstrh, line 595: This does not belong in the test framework. Let's invoke `arduino-cli` as part of the build process (provided that the necessary target properties are provided). If they're missing, we should print a message telling the user how to obtain them. We can just set the test phase to `BUILD` as I had suggested, because, either way, the tests themselves will not be executed in CI.

lhstrh (changes requested):
Getting close! Minor changes requested.

org.lflang/src/org/lflang/generator/c/CGenerator.java
lhstrh: Looks like that worked. Thanks, @erlingrj!

org.lflang.tests/src/org/lflang/tests/TestBase.java
arengarajan99: Done.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1978,13 +1994,18 @@ protected void setUpGeneralParameters() {
...
             if (targetConfig.setByUser.contains(TargetProperty.THREADING)) {
-                errorReporter.reportWarning("Threading is incompatible on Arduino. Setting threading to false.");
+                System.out.println("Threading is incompatible on Arduino. Setting threading to false.");
             }
             targetConfig.threading = false;
         }
+
+        if (targetConfig.platformOptions.platform == Platform.ARDUINO && !targetConfig.noCompile
+            && targetConfig.platformOptions.board == null) {
+            System.out.println("To enable compilation for the arduino platform, you must specify the board name by its FQBN. Setting to no compile.");
```
lhstrh, line 2006:
```suggestion
            System.out.println("To enable compilation for the Arduino platform, you must specify the fully-qualified board name (FQBN) in the target property. For example, platform: {name: arduino, board: arduino:avr:leonardo}. Entering "no-compile" mode and generating target code only.);
```

org.lflang/src/org/lflang/util/ArduinoUtil.java
```diff
@@ -0,0 +1,101 @@
...
+    private ErrorReporter errorReporter;
+
+    public ArduinoUtil (LFGeneratorContext context, GeneratorCommandFactory commandFactory, ErrorReporter errorReporter) {
+        this.context = context;
+        this.commandFactory = commandFactory;
+        this.errorReporter = errorReporter;
+    }
+
+    /**
+     * Returns true if arduino-cli exists, false otherwise.
```
lhstrh, line 30:
```suggestion
     * Return true if arduino-cli exists, false otherwise.
```

org.lflang/src/org/lflang/util/ArduinoUtil.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.lflang.ErrorReporter;
+import org.lflang.FileConfig;
+import org.lflang.TargetConfig;
+import org.lflang.generator.GeneratorCommandFactory;
+import org.lflang.generator.LFGeneratorContext;
+
+import org.eclipse.xtext.xbase.lib.Exceptions;
+
+public class ArduinoUtil {
```
lhstrh, line 17: This file is missing some JavaDoc.

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -1189,7 +1188,6 @@ public enum UnionType implements TargetPropertyType {
         SCHEDULER_UNION(Arrays.asList(SchedulerOption.values()), SchedulerOption.getDefault()),
         LOGGING_UNION(Arrays.asList(LogLevel.values()), LogLevel.INFO),
         PLATFORM_UNION(Arrays.asList(Platform.values()), Platform.AUTO),
-        BOARD_UNION(Arrays.asList(Board.values()), Board.NONE),
```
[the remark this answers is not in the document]
arengarajan99: yes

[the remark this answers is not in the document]
arengarajan99: Board is now simply a string

org.lflang/src/org/lflang/TargetConfig.java
```diff
@@ -439,22 +377,25 @@ public static class PlatformOptions {
...
+         * when we want to flash to an Arduino Nano 33 BLE board, we can use the string arduino:mbed_nano:nano33ble
          */
         public String board = null;
 
+
         /**
          * The string value used to determine the port on which to flash the compiled program (i.e. /dev/cu.usbmodem21301)
          */
-        public String port = "";
+        public String port = null;
```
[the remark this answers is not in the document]
arengarajan99: yes, I run null checks on port.

org.lflang/src/org/lflang/util/ArduinoUtil.java
arengarajan99: Fixed.

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -671,6 +678,12 @@ private void validateAndRun(Set<LFTest> tests, Configurator configurator, TestLe
                 validate(test);
                 if (level.compareTo(TestLevel.CODE_GEN) >= 0) {
                     generateCode(test);
+                    var srcBasePath = test.getFileConfig().srcPkgPath.resolve("src");
```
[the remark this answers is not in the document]
arengarajan99: Fixed.

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -47,11 +48,14 @@
 import org.lflang.generator.LFGeneratorContext;
 import org.lflang.generator.LFGeneratorContext.BuildParm;
 import org.lflang.generator.MainContext;
+import org.lflang.generator.GeneratorCommandFactory;
```
lhstrh, line 51: Unused import?

lhstrh (approved):

PR #1534, opened 2023-01-10
Improved CLI argument handling using `picocli`

patilatharva:
#### Objective
Use [Picocli](https://picocli.info/) instead of Apache Commons CLI to parse command line options and parameters in Lfc.java, Lff.java and CliBase.java.

#### Reason
[Migrating from Commons CLI to picocli](https://picocli.info/migrating-from-commons-cli.html)

#### Tasks:

- [x] Refactor `CliBase.java` to implement the `Runnable` interface, to support Picocli.
- [x] Remove `runMain` in `CliBase.java` because Picocli handles simple validation and `--help`/`--version` options.
- [x] Declare options as annotated fields and remove the `CLIOption` enum (`Lfc.java`/`Lff.java`).
- [x] Implement the `run()` method as an entry point to post-argument parsing execution, in accordance with the `Runnable` interface (`Lfc.java`/`Lff.java`).
- [x] Rewrite function `filterProps` into a new function `filterPassOnProps` to create code gen properties from Picocli options (`CliBase.java`).
- [x] Refactoring for clarity: split `runTool` into `runTool` + `invokeGenerator` in `Lfc.java`, and `runTool` into `runTool` + `formatAllFiles` in `Lff.java`.
- [x] Tweak CLI tests to support Picocli output syntax (e.g. "Unrecognized option" → "Unknown option").
- [x] Ensure all tests on CI are passing.


org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -52,83 +56,98 @@ public Lfc() {
...
+    private String scheduler;
+
+    @Option(
+        names = {"-t", "--threading"},
+        arity = "0",
+        description = "Specify whether the runtime should use multi-threading"
+                    + " (true/false).")
+    private boolean threading;
+
+    // TODO: could be int?
```
cmnrd, line 146: Ye, I think this should be processed as an integer

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -52,83 +56,98 @@ public Lfc() {
...
+        arity = "0",
+        description = "Do not invoke target compiler.")
+    private boolean noCompile;
+
+    @Option(
+        names = {"-o", "--output-path"},
+        defaultValue = "",
+        fallbackValue = "",
+        description = "Specify the root output directory.")
+    private String outputPath;
```
cmnrd, line 114: Also here we might want to leave it to picocli to check if the directory exists (if possible)

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -52,83 +56,98 @@ public Lfc() {
...
+    @Option(
+        names = "--target-compiler",
+        description = "Target compiler to invoke.")
+    private String targetCompiler;
+
+    @Option(
+        names = "--external-runtime-path",
+        description = "Specify an external runtime library to be used by the"
+                    + " compiled binary.")
+    private String externalRuntimePath;
```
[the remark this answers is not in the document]
patilatharva: Thanks for the suggestion! Picocli does support `java.nio.file.Path` directly - I just changed the input files parameter and the `outputPath` option to be interpreted as `Path`s instead of strings in Lfc/Lff. I've left all other options as Strings (or booleans to distinguish options that take no args) though, because later on they're passed into `LFGeneratorContext` through a `java.utils.Properties` object, which expects all option values to be strings.

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: As mentioned above - I realized it gets converted back to a string anyway, so I deleted the comment and left it as is.

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: Unfortunately Picocli doesn't do any validation beyond type conversion, so we'll still have to implement it ourselves. On the other hand, your comment gave me the idea of abstracting away redundant path parsing/validation code in `Lfc.java` and `Lff.java` into separate functions in `CliBase.java`, which made the code a lot cleaner.

lhstrh:
Just resolved conflict to trigger a CI run. Please check if I did it correctly :-)

lhstrh (commented):
This looks great, @patilatharva! I understood that you had started working on unit tests. Is that coming along or do you need help with those?

lhstrh:
There are some strange test failures that are popping up, so this definitely needs some more work. 

You can ignore the build failure in Maven for now. Either I'll push a fix for that or we punt on it and wait until the Maven build support gets removed from this repository (we only need it to build Epoch, which will soon be hosted in its own repo).

patilatharva:
> Just resolved conflicts to trigger a CI run. Please check if I did it correctly :-)

Thanks, looks good to me!

> There are some strange test failures that are popping up, so this definitely needs some more work.

Definitely, looking into it now.

org.lflang/src/org/lflang/cli/Lfc.java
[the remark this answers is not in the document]
cmnrd: Is there a fundamental reason for using `java.utils.Properties`? Maybe we should just create our own record class that passes the options in the correct types. Seems like converting to String and then back to the data type is redundant. This doesn't need to happen in this PR though, we can also leave for future work.

org.lflang/src/org/lflang/cli/CliBase.java
```diff
@@ -82,136 +113,144 @@ protected CliBase(String toolName) {
...
-        main.runMain(args);
+        // Parse arguments and execute main logic.
+        CommandLine cmd = new CommandLine(main)
+            .setOut(new PrintWriter(io.getOut()))
+            .setErr(new PrintWriter(io.getErr()));
+        int exitCode = cmd.execute(args);
+        io.callSystemExit(exitCode);
     }
 
+    public abstract void run();
```
oowekyala, line 131: ??

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -152,57 +163,53 @@ public static void main(Io io, final String... args) {
...
 
-        for (Path path : files) {
-            if (!Files.exists(path)) {
-                reporter.printFatalErrorAndExit(path + ": No such file or directory");
-            }
-        }
+    /**
+     * Invoke the code generator on the given validated file paths.
+     */
+    protected void invokeGenerator(
```
oowekyala, line 187: Why protected?

org.lflang/src/org/lflang/cli/Lfc.java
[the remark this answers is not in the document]
oowekyala: Tbh even if the options need to be converted to a string later, it is cleaner to use the correct data types as early as possible in the data flow. It allows us to take advantage of built-in validation and improves the CLI help. Converting to a string is also very easy.

org.lflang/src/org/lflang/cli/Lfc.java
oowekyala: You can implement a custom validating converter very easily: https://picocli.info/#_custom_type_converters

org.lflang/src/org/lflang/cli/Lfc.java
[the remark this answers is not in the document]
patilatharva: That's true, I'll use the correct data types (with custom type convertors wherever needed) to make sure all option values are validated at the entry point of the CLI.

org.lflang/src/org/lflang/cli/CliBase.java
patilatharva: Piccoli applications are designed to be implemented with the `Runnable`/`Callable` interfaces, and I chose to make our `CliBase` class implement the `Runnable` interface, which is why it needed a `public void run()` method. `Lfc` and `Lff` required different `run()` methods, so I made the one in `CliBase` abstract and wrote custom versions for `Lfc` and `Lff`.

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: You're right, there's no reason why it shouldn't be private. I'll fix it now.

org.lflang/src/org/lflang/cli/CliBase.java
oowekyala: Hey, thanks for the patient explanation! I think a short javadoc comment here would be nice (just mentioning that PicoCli requires this so that it doesn't get refactored away). Also an Override annotation is missing.

Also I apologize that my comments are so unpolished, I intended them as notes to myself, but apparently github submitted everything when I answered the other discussion above.

revol-xut (approved):
Looks good to me

org.lflang/src/org/lflang/cli/CliBase.java
patilatharva: No worries, I agree a javadoc there would be helpful. Thanks for your comments!

org.lflang/src/org/lflang/cli/CliBase.java
```diff
@@ -41,34 +45,61 @@
...
+     */
+    @Spec CommandSpec spec;
 
+    /**
+     * Options and parameters present in both Lfc and Lff.
+     */
+    @Parameters(
+        arity = "1..",
+        paramLabel = "FILES",
+        description = "Paths of the files to run the formatter on.")
```
cmnrd, line 63: As I understand, the FILES parameter is inherited by all cli tools. Thus, the description should be more generic.

org.lflang/src/org/lflang/cli/CliBase.java
cmnrd: Could you add the javadoc? I think it would be really helpful.

org.lflang/src/org/lflang/cli/CliBase.java
```diff
@@ -82,136 +113,144 @@ protected CliBase(String toolName) {
...
-            io.callSystemExit(1);
-            return;
-        }
-
-        // If requested, print help and abort
-        if (cmd.hasOption(helpOption.getOpt())) {
-            printHelp(options, formatter, io.getOut());
-            io.callSystemExit(0);
-        }
+    protected Properties filterPassOnProps() {
```
cmnrd, line 157: This appears to be specific to lfc. Why is it located in the base class?

org.lflang/src/org/lflang/cli/Lfc.java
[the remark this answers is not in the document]
cmnrd: Looks like you missed the option here, it is still a String and not a Path.

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -152,57 +163,53 @@ public static void main(Io io, final String... args) {
...
             if (resource == null) {
-                reporter.printFatalErrorAndExit(
-                    path + " is not an LF file. Use the .lf file extension to denote LF files.");
-            }
-            else if (cmd.hasOption(CLIOption.FEDERATED.option.getOpt())) {
+                reporter.printFatalErrorAndExit(path 
+                    + " is not an LF file. Use the .lf file extension to"
+                    + " denote LF files.");
+            } else if (federated) {
                 if (!ASTUtils.makeFederated(resource)) {
```
cmnrd, line 200: It looks wrong to me to invoke a AST transformation at this point. This should really happen in the generators. This is not an issue with this PR though, as the code already existed before.

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -34,7 +31,14 @@
...
  * @author Marten Lohstroh
  * @author Christian Menard
+ * @author Atharva Patil
  */
+@Command(
+    name = "lfc",
+    // Enable usageHelp (--help) and versionHelp (--version) options.
+    mixinStandardHelpOptions = true,
+    // TODO: Import version from StringsBundle.properties. 
+    version = "lfc 0.3.1-SNAPSHOT")
```
cmnrd, line 41: This should use `LocalStrings.VERSION` to retrieve the current version.

org.lflang/src/org/lflang/cli/Lff.java
```diff
@@ -27,219 +30,173 @@
...
  * @author Christian Menard
  * @author Billy Bao
+ * @author Atharva Patil
  */
+@Command(
+    name = "lff",
+    // Enable usageHelp (--help) and versionHelp (--version) options.
+    mixinStandardHelpOptions = true,
+    // TODO: Import version from StringsBundle.properties. 
+    version = "lff 0.3.1-SNAPSHOT")
```
cmnrd, line 40: This should use `LocalStrings.VERSION` to retrieve the current version.

cmnrd (commented):
This looks like a great improvement. I added a couple of comments, mostly small things. I hope we can resolve them quickly and merge this soon!

cmnrd (changes requested):
Updating review status to "request changes".

org.lflang/src/org/lflang/cli/Lff.java
patilatharva: Done!

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: Done!

org.lflang/src/org/lflang/cli/CliBase.java
patilatharva: True, just fixed it.

org.lflang/src/org/lflang/cli/CliBase.java
patilatharva: Just fixed it.

org.lflang/src/org/lflang/cli/CliBase.java
patilatharva: Just added it!

org.lflang/src/org/lflang/cli/Lfc.java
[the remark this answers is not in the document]
patilatharva: Just fixed it! Also, I wanted to write custom type converters/validators (e.g. checking if a path actually exists), but that would require me to write a separate class implementing the `picocli.CommandLine.ITypeConverter` interface. However, our current validation logic uses protected CliBase members, like `io` and `reporter`, which I don't think I'll be able to access from the separate type converter class. Do you have any ideas for workarounds?

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -32,7 +29,14 @@
  *
  * @author {Marten Lohstroh <marten@berkeley.edu>}
  * @author {Christian Menard <christian.menard@tu-dresden.de>}
+ * @author {Atharva Patil <atharva.patil@berkeley.edu>}
  */
+@Command(
+    name = "lfc",
+    // Enable usageHelp (--help) and versionHelp (--version) options.
+    mixinStandardHelpOptions = true,
+    // TODO: Import version from StringsBundle.properties.
```
lhstrh: This `TODO` actually needs fixing before we can merge this because otherwise the version number with be incorrect (as it now already is).

lhstrh (changes requested):
Please address the version `TODO` and once fixed we can merge this. Thanks, @patilatharva!

org.lflang/src/org/lflang/cli/Lfc.java
lhstrh: Also, shouldn't it just read the version string (without the `lfc ` part?) Not sure though.

patilatharva: It should be fixed with this commit: ([2d0662d](https://github.com/lf-lang/lingua-franca/pull/1534/commits/2d0662da45aee7c19e6e32bdc12e587df9e23f35)).

patilatharva: Maybe an older version of the PR was shown in the review?

lhstrh: Yes, I was looking at an old version. Sorry for the noise! It's unclear to be what `IVersionProvider,getVersion()` is supposed to return though. Should it include the name of the command?

patilatharva: True, I just added comments to make it more clear: [4768d34](https://github.com/lf-lang/lingua-franca/pull/1534/commits/4768d3405e94ba9d4bc2bbe6ba4265ecaeaeae30).

lhstrh (approved):

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: Sorry, I completely missed your point about whether we want to include the command name in the version info. To clarify, yes, based on previous tests.

PR #1542, opened 2023-01-12
rs/AsyncCallback: Join thread before creating new one

### LINKED ISSUE #1539: AsyncCallback failing in Rust
See: https://github.com/lf-lang/lingua-franca/actions/runs/3897293996/jobs/6654823077

oowekyala:
I think this should fix #1539, at least I can't reproduce it locally anymore. 

Weirdly enough I was only able to reproduce this when running the whole `RustTest` test class. Neither using `runSingleTest` nor running just `RustTest.runConcurrentTests` showed anything odd.

Let's see what the CI says

### DIFF
--- a/test/Rust/src/concurrent/AsyncCallback.lf
+++ b/test/Rust/src/concurrent/AsyncCallback.lf
@@ -4,7 +4,7 @@ target Rust {
     cargo-features: ["cli"]
 }
 
-main reactor AsyncCallback(period: time(1 msec)) {
+main reactor AsyncCallback(period: time(10 msec)) {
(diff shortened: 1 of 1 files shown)


test/Rust/src/concurrent/AsyncCallback.lf
```diff
@@ -4,7 +4,7 @@ target Rust {
     cargo-features: ["cli"]
 }
 
-main reactor AsyncCallback(period: time(1 msec)) {
+main reactor AsyncCallback(period: time(10 msec)) {
```
oowekyala, line 7: this didn't make any difference btw

PR #1546, opened 2023-01-13
Fix #1544 - up to date check for updateRustRuntime

### LINKED ISSUE #1544: Rust submodule gets modified when running build-lf-cli
When I run
```
> build-lf-cli -c
```
and then do
```
> git status
```
I get
```
	modified:  org.lflang/src/lib/rs/runtime-version.properties
```
What is this about?  Should I commit this?
git diff gives this:
```
diff --git a/org.lflang/src/lib/rs/runtime-version.properties b/org.lflang/src/lib/rs/runtime-version.properties
index f45320cae..9d66df0bf 100644
--- a/org.lflang/src/lib/rs/runtime-version.properties
+++ b/org.lflang/src/lib/rs/runtime-version.properties
@@ -1 +1 @@
-rs = cfb690bd427f33b5567058092fd5e175e294e697
+rs = 868c834f0de46b2cd633892ccafb7768357d3bd9
```

oowekyala:
Fix #1544

### DIFF
--- a/org.lflang/build.gradle
+++ b/org.lflang/build.gradle
@@ -78,6 +78,7 @@ configurations {
 task getSubmoduleVersions(type: Exec) {
     description('Run a Git command to get the current status of submodules')
     workingDir project.rootDir
+    outputs.upToDateWhen { false }
 
     def command = "git submodule status"
     if (System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows')) {
@@ -106,6 +107,7 @@ task updateRustRuntime {
     description('Record the VCS revisions of the language runtimes into a properties file available at runtime.')
 
     dependsOn getSubmoduleVersions
+    inputs.files(getSubmoduleVersions.outputs)
     ext.outputFile = file("$project.projectDir/src/lib/rs/runtime-version.properties")
     outputs.file(outputFile)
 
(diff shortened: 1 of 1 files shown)


org.lflang/build.gradle
```diff
@@ -78,6 +78,7 @@ configurations {
 task getSubmoduleVersions(type: Exec) {
     description('Run a Git command to get the current status of submodules')
     workingDir project.rootDir
+    outputs.upToDateWhen { false }
```
oowekyala, line 81: This will make gradle execute `git submodule status` every time updateRustRuntime is called.

org.lflang/build.gradle
```diff
@@ -106,6 +107,7 @@ task updateRustRuntime {
     description('Record the VCS revisions of the language runtimes into a properties file available at runtime.')
 
     dependsOn getSubmoduleVersions
+    inputs.files(getSubmoduleVersions.outputs)
```
oowekyala, line 110: But if the output of the `git submodule status` did not change (the getSubmoduleVersions task), then this task is considered up to date.

org.lflang/build.gradle
cmnrd:
```suggestion
   // This will make gradle execute git submodule status every time updateRustRuntime is called
    outputs.upToDateWhen { false }
```

org.lflang/build.gradle
cmnrd:
```suggestion
    // If the output of the git submodule status did not change (the getSubmoduleVersions task), then this task is considered up to date.
    inputs.files(getSubmoduleVersions.outputs)
```

cmnrd (approved):
Great! I will add the comments you made on the diff  directly as comments to the gradle file.

PR #1547, opened 2023-01-13
Fix port access to contained generic reactors in C++

cmnrd:
This provides a simple bugfix and adds a new testcase


petervdonovan (approved):
LGTM

I also saw your request response example that requires it and will look at it. Thanks!

PR #1548, opened 2023-01-14
Token payload free

edwardalee:
This PR refactors the token handling code considerably and delivers clean memory reports from valgrind (or the Mac equivalent, leaks). It still includes many deprecated APIs for working with tokens, but the plan is to eliminate those in a separate PR.

The token-payload-free branch of reactor-c and reactor-c-py should be merged first, and then the pointers will need to be adjusted.


test/C/src/token/include/array.h
```diff
@@ -0,0 +1,62 @@
+/**
```
lhstrh, line 1: Does this file belong here, or would it be better to make it available through reactor-c?

lhstrh (approved):
This looks great, but I still have to take a closer look at the changes in reactor-c, which I will do next.

test/C/src/token/include/array.h
edwardalee: This is only a test example, not a thoroughly thought-through array implementation in C, so I think it's better to leave here.

PR #1552, opened 2023-01-17
Clean up attributes in fed-gen

cmnrd:
https://github.com/lf-lang/lingua-franca/pull/1221 introduces several new attributes. Some of them are overly general (`@languange`) and others are redundant. This PR cleans up the newly introduced attributes.

This PR also includes some additional small cleanups including resetting .idea/codeStyles/Project.xml to the version from master.


lhstrh:
I think the changes you made here make sense. That said, I think that it might be cleaner to just reserve a single attribute for code-gen-related stuff and just have key-value pairs in it. I think that's neater than a `_` prefix, which looks ugly (let's not use underscores in these at all). We can change this later because this only concerns cosmetics and isn't user-facing.

lhstrh (approved):

PR #1556, opened 2023-01-18
Fix docker tests in fed-gen

cmnrd:
This fixes several problems in the docker test execution. It simplifies and generalizes the script that is generated in TestBase for running federated docker test. The script has now a little "smartness" build into it. It simply starts all containers and uses grep to find if any of them failed (as opposed to start each container separately as was done before).

I also noticed that the non-federated docker tests were broken, as any errors within the container were not picked up by our test framework. Both test categories now use the same script. This also allowed a significant cleanup of the test code.

We might not be fully in the green yet. During my testing I observed multiple times that federates could not connect to the RTI. The log looked always looked smth like this:
```
DistributedCountContainerized-c    | Federate 0: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-c    | Federate 0: Failed to connect to RTI on port 15065. Trying 15066.
DistributedCountContainerized-c    | Federate 0: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-c    | Federate 0: Failed to connect to RTI on port 15066. Trying 15067.
DistributedCountContainerized-rti  | WARNING: RTI failed to accept the socket. Resource temporarily unavailable. Trying again.
DistributedCountContainerized-c    | Federate 0: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-c    | Federate 0: Failed to connect to RTI on port 15067. Trying 15068.
DistributedCountContainerized-c    | Federate 0: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-c    | Federate 0: Failed to connect to RTI on port 15068. Trying 15069.
DistributedCountContainerized-c    | Federate 0: Could not connect to RTI at rti. Will try again every 2 seconds.
DistributedCountContainerized-c    | Federate 0: Failed to connect to RTI on port 15069. Trying 15070.
``` 
I am not sure if this is a problem in my setup or a problem with our tests. I always had to restart the docker daemon to make it work again.

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/TestBase.java
+++ b/org.lflang.tests/src/org/lflang/tests/TestBase.java
@@ -558,29 +558,29 @@ public abstract class TestBase {
      * @param dockerComposeFilePath The path to the docker compose file.
      */
     private String getDockerRunScript(List<Path> dockerFiles, Path dockerComposeFilePath) {
-        var dockerComposeCommand = "docker compose";
-        StringBuilder shCode = new StringBuilder();
-        shCode.append("#!/bin/bash\n");
-        shCode.append("pids=\"\"\n");
-        shCode.append(String.format("%s run -f %s --rm -T rti &\n",
-            dockerComposeCommand, dockerComposeFilePath));
-        shCode.append("pids+=\"$!\"\nsleep 3\n");
-        for (Path dockerFile : dockerFiles) {
-            var composeServiceName = dockerFile.getFileName().toString().replace(".Dockerfile", "");
-            shCode.append(String.format("%s run -f %s --rm -T %s &\n",
-                dockerComposeCommand,
-                dockerComposeFilePath,
-                composeServiceName));
-            shCode.append("pids+=\" $!\"\n");
-        }
-        shCode.append("for p in $pids; do\n");
-        shCode.append("    if wait $p; then\n");
-        shCode.append("        :\n");
-        shCode.append("    else\n");
-        shCode.append("        exit 1\n");
-        shCode.append("    fi\n");
-        shCode.append("done\n");
-        return shCode.toString();
+        return """
+            #!/bin/bash
+            
+            # exit when any command fails
+            set -e
+            
+            docker compose -f "$1" rm -f
+            docker compose -f "$1" up --build | tee docker_log.txt
+            docker compose -f "$1" down --rmi local
+            
+            errors=`grep -E "exited with code [1-9]" docker_log.txt | cat`
+            rm docker_log.txt
+            
+            if [[ $errors ]]; then
+                echo "===================================================================="
+                echo "ERROR: One or multiple containers exited with a non-zero exit code."
+                echo "       See the log above for details. The following containers failed:"
+                echo $errors
+                exit 1
+            fi
+             
+            exit 0
+            """;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cmnrd:
Looks like the RTI issue also occurs in CI: https://github.com/lf-lang/lingua-franca/actions/runs/3949937855/jobs/6761805738

lhstrh:
The `rti` problem is puzzling and needs looking into, but this PR sure is a major improvement!

org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -555,118 +548,104 @@ static public String stackTraceToString(Throwable t) {
...
      */
-    private List<ProcessBuilder> getNonfederatedDockerExecCommand(LFTest test) {
-        if (!checkDockerExists()) {
-            System.out.println(Message.MISSING_DOCKER);
-            return List.of(new ProcessBuilder("exit", "1"));
+    private void checkDockerExists() throws TestError {
+        if (LFCommand.get("docker", List.of()) == null) {
+            throw new TestError("Executable 'docker' not found" , Result.NO_EXEC_FAIL);
+        }
+        if (LFCommand.get("docker-compose", List.of()) == null) {
```
lhstrh, line 614: Hasn't the API changed to just `docker` (no `docker-compose`)?

lhstrh (approved):
Good to merge but left a question.

lhstrh:
Much cleaner indeed! Thank you so much for these fixes :-)

org.lflang.tests/src/org/lflang/tests/TestBase.java
lhstrh: I think we can just remove this check...

cmnrd: You need to install docker-compose separately, though (even if you use `docker compose`). This also puzzled me at first, but they shipp as separate packages

cmnrd: Should we revert the removal?

edwardalee:
There is a long-standing flaw in the RTI/federate mechanism for handling ports.  The RTI tries to get a default port, and if is unavailable, it tries a port number that is one larger, and if that fails, it tries one more, etc.  The federates go through a similar sequence, trying the default port number first, and if failing, trying one more.  

However, this really doesn't work. In particular, if you start a federate before the RTI, it skips the default port, and it takes a very long time for it to circle around to try that default port again.

The problem this was trying to address is that when a program releases a port, the OS does not make the port available to other programs for some time. There is a good reason for this: the OS wants to prevent a program from grabbing a port and then receiving messages that were intended for a program that has exited.  It therefore holds the port long enough that any messages that were in flight die before it releases the port.

This feature was making CI fail because it runs many federated programs in sequence.

I think a better solution is just that the RTI should just use a fixed port, perhaps optionally specified as a command-line argument (which the federates will also need to be told). Then we just have to figure out how to make CI work (wait long enough between federated tests?).

cmnrd:
Why not pass the port to use as a command line argument to the RTI and the federates? Then we can even have multiple instances running in parallel on the same host using different ports. What we are missing is an actual deployment.

lhstrh:
I personally think we need a service/daemon that listens at default and recognizable port (like `80` for `http`) with the sole purpose of brokering connections between participants in federations. We never built this and implemented this kind of the port knocking scheme instead, but I think this issue belongs in the category of problems solved by existing frameworks...

edwardalee:
Yes, you can improve what we have or start over with another framework. That choice seems clear.

edwardalee:
> Why not pass the port to use as a command line argument to the RTI and the federates? Then we can even have multiple instances running in parallel on the same host using different ports. What we are missing is an actual deployment.

Yes, this would be easy to implement and would delete a bunch of code.

lhstrh:
Update: resolving the race condition between RTI and federates by declaring a dependency in `docker-compose.yml` seems to have done the trick. We have some remaining errors, but those are of a different nature.

cmnrd:
I converted this discussion to an issue https://github.com/lf-lang/reactor-c/issues/146

PR #1573, opened 2023-01-26
Useful error message upon encountering missing runtime sources

cmnrd:
We used to have an error message in lfc, that notified the user if the runtime files are missing (i.e. the submodules are not checked out). It seems that with the introduction of `FileUtil` this feature got lost. It currently only complains if a resource cannot be found at all, but not if the resource is an empty directory. Aside from a bit of cleanup, this fix ensures that a useful error message is also produced if a submodule is missing.

This depends on https://github.com/lf-lang/lf-lang.github.io/pull/121. Without merging it, the link in the error message will not be available.


lhstrh (approved):
This looks great! I like the solution of having a dedicated exception for this.

PR #1574, opened 2023-01-26
Don't build the RTI docker image in CI tests

cmnrd:
Our tests pull the image from docker hub and hence there is no need to build the image locally.

### DIFF
--- a/.github/workflows/c-tests.yml
+++ b/.github/workflows/c-tests.yml
@@ -60,9 +60,6 @@ jobs:
           cd arduino-1.8.19/
           sudo ./install.sh
         if: ${{ runner.os == 'Linux' }}
-      - name: Build RTI docker image
-        uses: ./.github/actions/build-rti-docker
-        if: ${{ runner.os == 'Linux' }}
       - name: Perform tests for C target with default scheduler
         run: |
           ./gradlew test --tests org.lflang.tests.runtime.CTest.*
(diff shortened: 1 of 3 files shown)


lhstrh:
I'm surprised this doesn't break the non-containerized tests. Why doesn't it?

cmnrd:
There is still the "Install RTI" task that installs the RTI locally. This PR only removes the redundant buiding of the container that is not used later.

Jakio815:
Are these also needed?

https://github.com/lf-lang/lingua-franca/blob/ci-no-rti-image/.github/actions/build-rti-docker/action.yml#:~:text=name%3A%20Build,shell%3A%20bash

https://github.com/lf-lang/lingua-franca/blob/ci-no-rti-image/.github/actions/build-rti-docker/build-rti-image.sh#:~:text=%23!/bin/bash,f%20rti.Dockerfile%20../../../

Jakio815 (approved):

lhstrh (approved):
:+1:

lhstrh:
The files you point to, @Jakio815, are where the RTI gets installed on the system (without Docker), so we need to keep those.

cmnrd:
> The files you point to, @Jakio815, are where the RTI gets installed on the system (without Docker), so we need to keep those.

Are they? It looks to me like they are only used for the docker image

lhstrh:
@Jakio815 was right. We no longer need those files, although it doesn't hurt to keep them. We have talked about putting the rti in its own repo, and then having these workflows as tests might be appropriate?

PR #1575, opened 2023-01-28
[ts] Bugfix that re-enables the use of custom configuration files

lhstrh:
A custom `package.json` can be provided alongside an LF file. The compiler finds it and copies it to the generated source directory. This capability was broken and this PR fixes it.


hokeun:
Thanks, Marten!

PR #1578, opened 2023-01-30
Fix to let relative include helper ignore non-C files

### LINKED ISSUE #1576: Arduino utility for include files is failing
The [Arduino test failure](https://github.com/lf-lang/lingua-franca/actions/runs/4036378872/jobs/6938932629#step:9:152) is due to the [`FileUtil.relativeIncludeHelper`](https://github.com/lf-lang/lingua-franca/blob/e3a735d51af0546ef0d75d4a54c31ccd0fc28655/org.lflang/src/org/lflang/util/FileUtil.java#L448), which is used in `CGenerator` for code generation for Arduino.

This function comment is quite unclear. It seems to be walking a directory tree and creating a Map from any filename it finds to the path to that file. That can't work if the directory tree has files with the same name, which it often will. Then it seems to be doing something very risky, which is to overwrite the files it finds in that directory tree, replacing `#include` statements.

A better design, if we really want to do this, would be to first scan the files with appropriate extensions (`.lf`?) for `#include` statements that do not have the right form, and then find the appropriate path to those. But why not just require the correct path in the first place?  Or fix the search path for `#include` statements?

First, I suggest changing the following [line](https://github.com/lf-lang/lingua-franca/blob/e3a735d51af0546ef0d75d4a54c31ccd0fc28655/org.lflang/src/org/lflang/util/FileUtil.java#L459):
```
                throw new IOException("Directory has different files with the same name. Cannot Relativize.");
```
to
```
                throw new IOException("Directory has different files with name " + fileName + ". Cannot Relativize.");
```
Then you will discover that the failure is because of a `README.md` file.  But this is not included by any Arduino source file that I could find.  I don't think we want to limit the number of `README.md` files. It looks like the function has a hack to avoid this error with `CMakeLists.txt` files, but what more name collisions might be lurking?

lhstrh:
Fixes #1576.


arengarajan99:
Looks like the problem is stemming from multiple definitions of the main function (one gets done automatically by Arduino to call setup and loop and the other is coming from a test file for the hashset implementation.) Ideally, the test file should not be copied over and it should resolve.

lhstrh:
Makes sense to have tests either have the word "test" in them or group them in a directory called "test." That way, we can easily ignore them.

lhstrh:
> Looks like the problem is stemming from multiple definitions of the main function (one gets done automatically by Arduino to call setup and loop and the other is coming from a test file for the hashset implementation.) Ideally, the test file should not be copied over and it should resolve.
> 
> @edwardalee Are you able to move that test file into https://github.com/lf-lang/reactor-c/tree/main/test/general/utils ? It would be the easiest fix as far as Peter and I know.

I went ahead and moved the file. It now also gets run as part of the unit tests.

PR #1580, opened 2023-02-02
Add equals initializer syntax

### LINKED ISSUE #1650: CLI gives null-pointer exception
The following valid LF program gives an NPE coming from the CLI code.
Contents of file `ExtendsBugInside.lf`:
```
target C
reactor C(
    period: time(5 ms)
) {
    timer t(0, period)
}

reactor A(
    period: time(5 ms)
) {    
    a = new C(period = period)
}
```
Contents of file `ExtendsBug.lf`:
```
target C

import A from "ExtendsBugInside.lf"

reactor B extends A {}

main reactor {
    a = new B()
}
```
Result of trying to compile:
```
ealmac:C eal$ lfc src/rosace/ExtendsBug.lf 
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.common.util.URI.isPlatform()" because "uri" is null
	at org.lflang.util.FileUtil.toIPath(FileUtil.java:88)
	at org.lflang.util.FileUtil.toPath(FileUtil.java:71)
	at org.lflang.cli.CliBase.validateResource(CliBase.java:225)
	at org.lflang.cli.Lfc.invokeGenerator(Lfc.java:193)
	at org.lflang.cli.Lfc.run(Lfc.java:165)
	at org.lflang.cli.CliBase.cliMain(CliBase.java:103)
	at org.lflang.cli.Lfc.main(Lfc.java:150)
	at org.lflang.cli.Lfc.main(Lfc.java:140)
```

oowekyala:
- Introduce the new `=` initializer syntax to replace the parens syntax in all targets except C++. This works exactly like parentheses do, except in C++ it calls the assignment operator.
- Using the parens syntax emits a warning now
- The formatter converts to the new syntax automatically for ease of migration (for C++ it doesn't touch anything).

**Updated 2023-03-02** This does in fact add a syntax for C array initializers (which can be used also for struct initializers). Summary:
- Using `a: int(0)` or `a: int(0,0)` produces a warning (except in C++). they can be replaced with `a: int = 0` resp. `a: int = {0, 0}` (the latter only in C/C++).
  - the braced list syntax is not available in Python or TS, so if you were declaring a python list/tuple with `a(1,2)`, you can convert it to `a = {= [1, 2] =}` to remove the warning, or just wait until we allow writing `a = [1, 2]` directly.
- In all targets, reactor instantiations have to use the equal syntax: write `new R(a = 1)` instead of `new R(a (1))`.
  - In C++, if you want to call the constructor with parameters, then you have to write out the type of the constructed object explicitly: `new R(a = {= std::vector<int>(1,2) =}` or so (whereas you don't need to repeat the type when you write a default parameter value or a state var initializer). This is so because we would get problems with generic type parameters otherwise (and C++ itself has the same behavior).
  - In C, if you were writing `new R(a(1,2))` or `new R(a = (1,2))` to mean an array, then use `new R(a = {1, 2})`
  - In Python and TS, the same caveat applies, that you can't write a python/TS bracket-delimited list without target escaping yet.

Remaining action items:
- [x] make build pass
- [x] update doc and website pages see 
	- [ ] https://github.com/lf-lang/lf-lang.github.io/pull/130
	- [ ] https://github.com/lf-lang/examples-lingua-franca/pull/24
- [x] @cmnrd write a test for C++
- [x] Fix issues in test/Cpp/src/target/InitializerSyntax.lf where `{}` needs to be escaped to not be omitted.

Fixes #1650  and #1651


lhstrh:
Only 654 files changed? :rofl:

lhstrh:
This looks good to me, though it would be great to get some more :eyes: on this. My main concern is that this does not only affect our tests, but also our examples. We should probably set up CI in https://github.com/lf-lang/examples-lingua-franca first, and prepare a PR that makes all the required adjustments...

lhstrh:
Just to clarify, does using braces now trigger a warning or an error?

edwardalee:
This also requires an update of all the examples and text on the website.

oowekyala:
> Just to clarify, does using braces now trigger a warning or an error?

Using parentheses triggers a warning (except in C++), braces is still an error in all targets except C++. Also in all targets including C++, writing `new R(a(1))` or `a {1}` instead of `a = 1` or eg `a = {= std::vector<int>(1) =}` is now a warning.

> This also requires an update of all the examples and text on the website.

I've been working on this and will submit a PR today.

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -757,14 +760,26 @@ public MalleableString caseSerializer(Serializer object) {
...
         .get();
   }
 
+  @Override
+  public MalleableString caseBracedListExpression(BracedListExpression object) {
+    if (object.getItems().isEmpty()) {
+      return MalleableString.anyOf("{}");
+    }
+    // Note that this strips the trailing comma. There is no way
+    // to implement trailing commas with the current set of list() methods AFAIU.
```
cmnrd, line 779: This not a problem for brace lists in C/C++, but it might become a problem for tuples in Python. That is a problem for another day though...

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -831,21 +843,36 @@ public MalleableString caseInitializer(Initializer object) {
     return initializer(object, false);
   }
 
+  private boolean shouldOutputAsAssignment(Initializer init) {
```
cmnrd, line 846: The name here seems a bit strange. What exactly does this do? Decide if () assignment should be replaced with = assignment? Maybe a short description would be helpful

org.lflang/src/org/lflang/generator/LfExpressionVisitor.java
```diff
@@ -0,0 +1,160 @@
...
+import org.lflang.lf.Literal;
+import org.lflang.lf.ParameterReference;
+import org.lflang.lf.Time;
+
+/**
+ * A visitor for expressions in LF.
+ *
+ * @author Clément Fournier &lt;clement.fournier@tu-dresden.de&gt;
+ */
+public interface LfExpressionVisitor<P, R> {
```
cmnrd, line 41: Cool! Do I understand that this is currently only used for copying expressions? I think this will become handy also for formatting expression and generating target code from expressions once the expression syntax becomes more complex.

org.lflang/src/org/lflang/generator/TargetTypes.java
```diff
@@ -68,6 +69,12 @@ default String getTargetParamRef(ParameterReference expr, InferredType typeOrNul
         return escapeIdentifier(expr.getParameter().getName());
     }
 
+    /** Translate the braced list expression into target language syntax. */
+    default String getTargetBracedListExpr(BracedListExpression expr, InferredType typeOrNull) {
```
cmnrd, line 73: I wonder if the `TargetTypes` class is the right place for this, but I am Ok with keeping it here for now. I think will need something like a `ExpressionGenerator` sooner or later (this is where the expression visitor will be very useful).

org.lflang/src/org/lflang/ast/ToLf.java
```diff
@@ -757,14 +760,26 @@ public MalleableString caseSerializer(Serializer object) {
...
   }
 
+  @Override
+  public MalleableString caseBracedListExpression(BracedListExpression object) {
+    if (object.getItems().isEmpty()) {
+      return MalleableString.anyOf("{}");
+    }
+    // Note that this strips the trailing comma. There is no way
+    // to implement trailing commas with the current set of list() methods AFAIU.
+    return list(", ", "{ ", " }", false, false, object.getItems());
```
cmnrd, line 780: Why insert the extra space after `{` and before `}`? This does not look idiomatic to me if I look at the examples.

test/C/src/ArrayAsParameter.lf
```diff
@@ -3,9 +3,9 @@
 // encode their own length.
 target C
 
-reactor Source(sequence: int[](0, 1, 2), n_sequence: int(3)) {
+reactor Source(sequence: int[] = { 0, 1, 2 }, n_sequence: int = 3) {
```
cmnrd, line 6: This looks so much cleaner! I only find the extra spaces after `{` and before `}` a bit odd.

cmnrd (approved):
Amazing work! Thanks for taking care of this. I obviously haven't checked all 650 test programs, but what I have seen is a significant improvement in all programs that I have had a look at.

cmnrd:
There are still a couple of test failures. Do you need help with those?

org.lflang/src/org/lflang/ast/ToLf.java
oowekyala: It looked nicer to me, but I don't care either way. If it's more idiomatic without the spaces then let's remove them

org.lflang/src/org/lflang/ast/ToLf.java
oowekyala: That's true. I'm not sure it makes sense to add a dedicated syntax for Python tuples anyway... It sounds like it would be rarely used.

org.lflang/src/org/lflang/generator/LfExpressionVisitor.java
oowekyala: Yes currently it's used to replace parameter references with a value in the instance tree. The visiting and rebuilding is needed because a BracedListExpression is potentially recursive and can contain other list expressions. This will extend easily to other expressions, most of which are also recursive.

org.lflang/src/org/lflang/generator/TargetTypes.java
oowekyala: Yes I think you're right that this class is getting too complex, we probably can split it later using an expression visitor

org.lflang/src/org/lflang/generator/c/CParameterGenerator.java
```diff
@@ -37,8 +32,18 @@ public static String getInitializer(ParameterInstance p) {
             return CUtil.bankIndex(p.getParent());
         }
 
+        // Handle overrides in the intantiation.
+        // In case there is more than one assignment to this parameter, we need to
+        // find the last one.
+        Assignment lastAssignment = null;
+        for (Assignment assignment: p.getParent().getDefinition().getParameters()) {
```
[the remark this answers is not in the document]
oowekyala: Also I don't think it should be allowed to have several assignments to the same parameter

petervdonovan:
I do not object to merging this right away even though I have not reviewed the code. If other people think that more review is important and can wait for a few days for me to do it, I can, but I also do not mind trusting you and Clement so that this big PR does not linger any longer.

petervdonovan:
The one failing test, `NativeListsAndTimes`, [relies on a C23 feature](https://en.cppreference.com/w/c/language/initialization#Empty_initialization), so it shouldn't be surprising that it does not pass on Windows. (The failing CLI tests are just a network issue.)

In C, I am not aware of any use cases for zero-length arrays _except_ in the case when you are code-generating C, using macros, etc., and the array of length zero is the continuation of some pattern. I am not sure I can envision such situations cropping up in LF programs. For sure it is awkward and weird that C does not permit initializers of length zero before 2023, but maybe that is C's fault, not ours. We have the ability to patch it up in LF, but I am not sure that that is our responsibility.

org.lflang.tests/src/org/lflang/tests/compiler/FormattingUnitTests.java
```diff
@@ -56,16 +65,34 @@ public void testState() {
                 reactor Destination {
                     state  one_init: tag_t( {= NEVER_TAG_INITIALIZER =})
                     state no_init:   tag_t
-                     state list_init(1,2)
+                     state list_init(1,2) // this syntax is deprecated
```
petervdonovan, line 68: Why not change this to the new syntax?
```suggestion
                     state list_init = (1,2)
```

org.lflang.tests/src/org/lflang/tests/compiler/FormattingUnitTests.java
```diff
@@ -56,16 +65,34 @@ public void testState() {
...
                 }
                 """,
             """
                 target Python
 
+                reactor Destination {
+                    state one_init: tag_t = {= NEVER_TAG_INITIALIZER =}
+                    state no_init: tag_t
+                    state list_init(1, 2)  # this syntax is deprecated
```
petervdonovan, line 77:
```suggestion
                    state list_init = (1, 2)  # this syntax is deprecated
```

org.lflang/src/org/lflang/Target.java
```diff
@@ -502,6 +500,20 @@ public boolean buildsUsingDocker() {
         };
     }
 
+    /**
+     * Whether the target requires using an equal sign to assign a defaul value to a parameter,
```
petervdonovan, line 504:
```suggestion
     * Whether the target requires using an equal sign to assign a default value to a parameter,
```

petervdonovan (approved):
LGTM, I have no important comments. Looking forward to seeing this get merged soon!

test/C/src/NativeListsAndTimes.lf
```diff
@@ -2,22 +2,25 @@ target C
...
-    p: int[](1, 2, 3, 4),      // List of integers
-    q: interval_t[](1 msec, 2 msec, 3 msec),  // list of time values
-    g: time[](1 msec, 2 msec)  // List of time values
+    x: int = 0,
+    y: time = 0,                   // Units are missing but not required
+    z = 1 msec,                    // Type is missing but not required
+    p: int[] = {1, 2, 3, 4},       // List of integers
+    q: interval_t[] = {1 msec, 2 msec, 3 msec},  // list of time values
+    g: time[] = {1 msec, 2 msec},  // List of time values
+    g2: int[] = {}
```
petervdonovan, line 11:
```suggestion
```
See above comment; this is C23 syntax that we should not rely on

org.lflang.tests/src/org/lflang/tests/compiler/FormattingUnitTests.java
oowekyala: I want to make sure that the formatter does not rewrite this syntax, even if it is deprecated. The validator emits a warning but the update here should be manual. `(1, 2)` is also not yet a valid expression in the python target

test/C/src/NativeListsAndTimes.lf
oowekyala: Thanks for your input on this, I had no idea about this limitation of C.

Do you think the LF validator should forbid this/ emit a warning, or just ignore it and leave it up to the C compiler to decide whether that's valid?

cmnrd: I think the latter. We should not try to do the job of the target compiler.

PR #1582, opened 2023-02-03
Link reactor-c lib with zephyr kernel lib

erlingrj:
Addresses compilation errors when building threaded LF programs targeting Zephyr

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
@@ -371,6 +371,9 @@ public class CCmakeGenerator {
         var code = new CodeBuilder();
         code.pr("add_subdirectory(core)");
         code.pr("target_link_libraries(core PUBLIC zephyr_interface)");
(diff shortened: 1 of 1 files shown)


org.lflang/src/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -371,6 +371,9 @@ private static String setUpMainTargetZephyr(
         var code = new CodeBuilder();
         code.pr("add_subdirectory(core)");
         code.pr("target_link_libraries(core PUBLIC zephyr_interface)");
+        // FIXME: Linking the reactor-c corelib with the zephyr kernel lib
+        //  resolves linker issues but I am not yet sure if it is safe
```
lhstrh, line 375: @erlingrj do you recommend we merge this?

erlingrj: This seems to work, however I could not find any references that suggests to do something like this. I say that we merge it and pay attention to it. I could remove the FIXME, but I think it is okay to have it and we can re-evaluate it later?

lhstrh (approved):

Issue #616, opened 2021-10-15
Physical connections should acquire a tag from the physical clock


lhstrh:
Currently, for the C target, physical connections do not work correctly. See https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/PhysicalConnection.lf

This test passes, but it is implemented incorrectly. The test should read:

```

/** Test physical connections. */
target C;
reactor Source {
    output out:int;
    reaction(startup) -> out {=
        SET(out, 42);
    =}
}
reactor Destination {
    input in:int;
    reaction(in) {=
        interval_t time = get_elapsed_logical_time();
        printf("Received %d at logical time %lld.\n", in->value, time); // This line changed!
        if (time <= 0LL) {
            fprintf(stderr, "ERROR: Logical time should have been greater than zero.\n");
            exit(1);
        }
    =}
}
main reactor PhysicalConnection {
    source = new Source();
    destination = new Destination();
    source.out ~> destination.in; 
}
```

This test does _not_ pass. This is because the physical connection doesn't assign events it relays a tag from the physical clock. This is probably because it is using logical action, not a physical action.

PR #1583, opened 2023-02-03
Use of delayed and physical connections as provided by the C++ runtime

cmnrd:
Currently, after delays are implemented as an AST transformation. With https://github.com/lf-lang/reactor-cpp/pull/38, the C++ runtime provides a native solution for implementing delayed and physical connections. This PR disables the AST transformation in the C++ target and instead uses the native runtime implementations. It also adds a couple of tests for various corner cases that have not been tested before.

In short, the runtime provides to classes `DelayedConnection` and `PhysicalConnection` that are specializations of the `Action` class (which better should be called Trigger). By implementing `Action`, the connections may add new events to the event queue. At the same time, they provide API for binding to one upstream and multiple downstream ports. Due to a callback mechanism, the connection gets notified when the upstream port is set and it will schedule an event. When the scheduler retrieves the event from the event queue, its `setup()` function will also set all downstream ports and accordingly trigger downstream reactions.

This PR probably makes the C++ target the first target to correctly implement physical connections, as physical connections are currently not properly handled by the AST transformation (see https://github.com/lf-lang/lingua-franca/issues/616)


revol-xut (approved):
Dont see any big mistakes

edwardalee (approved):
Didn't check carefully, but LGTM!

org.lflang/src/lib/cpp/lfutil.hh
```diff
@@ -101,4 +101,42 @@ void bind_multiple_ports(
...
 }
 
+template<class T>
+void bind_multiple_connections_with_ports(
+    std::vector<reactor::Connection<T>*>& connections,
+    std::vector<reactor::Port<T>*>& ports,
+    bool repeat_left) {
+
+  if (repeat_left) {
+    size_t l_size = connections.size();
```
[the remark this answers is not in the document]
cmnrd: Here it's probably best to go with auto actually. Thanks!

cmnrd:
@petervdonovan Could you take a look at the failing LSP tests? If you have any tips on where I can start looking myself in the future (I already struggle with finding the actual error in the log), I would appreciate that. Or even better: it would be great if we could have some instructions on how to handle the LSP tests in the developer docs.

petervdonovan:
The LSP tests appear to be passing

> it would be great if we could have some instructions on how to handle the LSP tests in the developer docs.

I will try to do this tomorrow

cmnrd:
> The LSP tests appear to be passing

Oh, then I guess it was just one of the regular C++ tests that failed before causing the problem.

> 
> I will try to do this tomorrow or the day after tomorrow

That would be great! Thanks!

org.lflang/src/org/lflang/generator/cpp/CppConnectionGenerator.kt
```diff
@@ -0,0 +1,59 @@
...
+            }
+        } else null
+
+    private fun generateConstructorInitializer(connection: Connection): String? =
+        if (connection.requiresConnectionClass && !connection.hasMultipleConnections) {
+            val delay = connection.delay.toCppTime()
+            val name = connection.name
+            """, $name{"$name", this, $delay}"""
+        } else null
+}
```
lhstrh, line 59:
```suggestion
}

```

org.lflang/src/org/lflang/generator/cpp/CppConnectionGenerator.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import org.lflang.*
+import org.lflang.generator.cpp.CppConnectionGenerator.Companion.cppType
+import org.lflang.generator.cpp.CppPortGenerator.Companion.dataType
+import org.lflang.lf.Connection
+import org.lflang.lf.Port
+import org.lflang.lf.Reactor
+import org.lflang.lf.VarRef
+
+class CppConnectionGenerator(private val reactor: Reactor) {
```
lhstrh, line 11: This class needs some documentation...

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -167,30 +170,49 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
...
-            else                     -> "reactor::Port<std::remove_reference<decltype($name)>::type::value_type>*"
-        }
+        get() = "reactor::Port<${dataType}>*"
 
     private fun declareMultiportConnection(c: Connection, idx: Int): String {
         // It should be safe to assume that all ports have the same type. Thus we just pick the
         // first left port to determine the type of the entire connection
         val portType = c.leftPorts[0].portType
 
+        // TODO this does not work with generics
```
lhstrh, line 180: Replace with link to issue?

lhstrh (approved):
Looks good to me!

org.lflang/src/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
cmnrd: Good catch! That comment is actually outdated and generics work fine (there is even a test for it).

lhstrh, line 180:
```suggestion
```

lhstrh:
```suggestion
```

PR #1584, opened 2023-02-07
More robust ROS2 setup in CI

cmnrd:
Setting up ROS2 in CI can be flaky and recently ROS2 tests have been failing consistently. The problem is also mentioned in the README of the ROS action: https://github.com/ros-tooling/setup-ros/

This PR attempts at fixing this by applying the solution mentioned here: https://github.com/ros2/rmw_cyclonedds/pull/134

The PR further factors out the ROS2 setup in our own customized action, so that we don't need to duplicate code.


lhstrh (approved):
Thanks for taking care of this!

PR #1587, opened 2023-02-14
Documentation about LSP tests in `README.md`

petervdonovan:
This adds a description of the output of the LSP tests.

I am optimistic that we will eventually be able to remove the target language diagnostic reporting feature together with the LSP tests by reducing reliance on inlined target code. However unless/until that happens the LSP tests require documentation.


cmnrd (approved):
Thanks!

PR #1588, opened 2023-02-15
Update rust runtime

oowekyala:
Update to use lf-lang/reactor-rs#40


lhstrh:
Do we want this to be part of v0.4.0?

lhstrh:
Do we want this to be part of v0.4.0?

oowekyala:
@lhstrh your call. There is no new feature motivating the bump, it's just to keep in sync with refactorings of the internals.

lhstrh (approved):
:rocket:

PR #1589, opened 2023-02-15
Fix failing building of tracing tools

erlingrj:
1. Add building of the tracing tools to the CI (However this wont run until it is in `master`)
2. Pass `LF_UNTHREADED=1` to the building of the tracing tools. Using `LF_THREADED=1` does not work unless we add  some threading specific files to the build..


edwardalee (approved):
LGTM!

lhstrh:
Thanks, @erlingrj! I took the liberty to decouple the trace tool build test from the c-tests because I think it is better that way...

lhstrh (approved):

PR #1590, opened 2023-02-15
Use of newer actions to reduce number of produced warnings in CI

lhstrh:
There are currently a lot of warnings printed by GitHub actions due to the use of outdated Node.js versions and the use of now deprecated GitHub Actions features. The changes in this PR are meant to reduce the number of warnings.
 - [x] Nodejs warnings
 - [x] save-state (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
 - [x] set-output (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### DIFF
--- a/.github/actions/prepare-build-env/action.yml
+++ b/.github/actions/prepare-build-env/action.yml
@@ -18,4 +18,4 @@ runs:
         echo $(mvn -v)
       shell: bash
     - name: Validate Gradle wrapper
-      uses: gradle/wrapper-validation-action@v1.0.4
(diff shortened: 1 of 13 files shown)


lhstrh:
The following third-party actions are still throwing warnings but have no newer version available:
 - arduino/setup-arduino-cli@v1
 - ATiltedTree/setup-rust@v1
 - ros-tooling/setup-ros@v0.5

PR #1593, opened 2023-02-16
Draft proposal for making the grammar unambiguous

lhstrh:
This PR makes the grammar unambiguous, but I'm not at all certain about these changes.
1. With the optional reaction name, we need to make the triggers mandatory, or else we won't be able to distinguish between the given name and the sources.
2. In the absence of a body, we need some placeholder. We could use `;`, but that would be awkward in the presence of a deadline. Using `.` in this case might make sense, but we can't because this collides with hierarchical effects. In this PR I use `{` `}` (and require `;` if there is no body and there are no deadlines).

### DIFF
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -202,7 +202,7 @@ Reaction:
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
     (sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    ((code=Code ('named' ID)?) | 'named' ID)(stp=STP)?(deadline=Deadline)?
+    ((('named' ID)? code=Code) | 'named' ID)(stp=STP)?(deadline=Deadline)?
     ;
 
 TriggerRef:
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -496,7 +496,7 @@ Token:
     'mutable' | 'input' | 'output' | 'timer' | 'action' | 'reaction' |
     'startup' | 'shutdown' | 'after' | 'deadline' | 'mutation' | 'preamble' |
     'new' | 'federated' | 'at' | 'as' | 'from' | 'widthof' | 'const' | 'method' |
-    'interleaved' | 'mode' | 'initial' | 'reset' | 'history' |
+    'interleaved' | 'mode' | 'initial' | 'reset' | 'history' | 'named' |
     // Other terminals
     NEGINT | TRUE | FALSE |
     // Action origins
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -202,10 +202,8 @@ Reaction:
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
     (sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    (((code=Code)(stp=STP)?(deadline=Deadline)?) |
-     ('{' '}' (stp=STP)?(deadline=Deadline)?) |
-     ';'
-    );
+    ((code=Code ('named' ID)?) | 'named' ID)(stp=STP)?(deadline=Deadline)?
+    ;
 
 TriggerRef:
     BuiltinTriggerRef | VarRef;
--- a/org.lflang/src/org/lflang/LinguaFranca.xtext
+++ b/org.lflang/src/org/lflang/LinguaFranca.xtext
@@ -199,12 +199,13 @@ Reaction:
     (attributes+=Attribute)*
     (('reaction') | mutation ?= 'mutation')
     (name=ID)?
-    ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')?
+    ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
     (sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    (code=Code)?
-    (stp=STP)?
-    (deadline=Deadline)?;
+    (((code=Code)(stp=STP)?(deadline=Deadline)?) |
+     ('{' '}' (stp=STP)?(deadline=Deadline)?) |
+     ';'
+    );
 
 TriggerRef:
     BuiltinTriggerRef | VarRef;


petervdonovan:
> With the optional reaction name, we need to make the triggers mandatory

It seems like no triggers is currently interpreted as "everything is a trigger." I like the idea of omitting that "feature" until it has a use case. It seems like an unnecessary special case.

> In the absence of a body, we need some placeholder. 

I do not understand why this is necessary. If the next thing following the reaction declaration is not `{=`, then there is no body, right? This seems like it should be unambiguous.

lhstrh:
> > In the absence of a body, we need some placeholder.
> 
> I do not understand why this is necessary. If the next thing following the reaction declaration is not `{=`, then there is no body, right? This seems like it should be unambiguous.

The problem is that if you have something like `reaction foo(x) y -> z ` it is ambiguous as to whether this denotes a reaction with source `y` and effect `z` or it is reaction without sources or effects (and no body) followed by a connection statement that connects `y` to `z`.

petervdonovan:
Okay, I get it now.

> We could use `;`, but that would be awkward in the presence of a deadline.

I personally like the offsides rule, but it seems like the easiest thing right now would be to separate everything with semicolons. Is it good enough to require the semicolon only after the deadline and STP handlers etc.?

edwardalee:
I would prefer that we remove the reaction name and simply assign a name using the reaction number.  This will emphasize the reaction ordering and discourage calling reactions yourself, which I think is going to happen anyway with this proposal.  

Adding yet another ad-hoc special case where a semicolon is required seems ugly to me.  I would prefer reverting to requires semicolons everywhere.

lhstrh:
The problem with removing the name is that when you reorder reactions you will inadvertently mix up your implementations, which seems terribly inconvenient...

edwardalee:
> The problem with removing the name is that when you reorder reactions you will inadvertently mix up your implementations, which seems terribly inconvenient...

This seems like the least of the problems, honestly.  If a user wants a specific name for the reaction body, they can add one line of code in the .lf file.  Note also that methods are available if you want to name functions.

I think with this change, we are going to find users calling reactions themselves and bypassing LF semantics routinely.

lhstrh:
> > The problem with removing the name is that when you reorder reactions you will inadvertently mix up your implementations, which seems terribly inconvenient...
> 
> This seems like the least of the problems, honestly. If a user wants a specific name for the reaction body, they can add one line of code in the .lf file. Note also that methods are available if you want to name functions.
> 
> I think with this change, we are going to find users calling reactions themselves and bypassing LF semantics routinely.

The point of this is not to name reactions because we want to call them, but because we want to generate conveniently-named prototypes. If we output each reactor class in its own file, my guess is that those prototypes won't even be in scope for other reactors.

petervdonovan:
> If we output each reactor class in its own file, my guess is that those prototypes won't even be in scope for other reactors.

Other reactors will be able to bring the prototypes in scope easily by `#include`ing the generated header files. I agree that this is a valid concern which I had forgotten about/not thought of. However
* In order to call these functions corresponding to some reactor A, another reactor B would need to have access to instances of structs that have all the state variables and parameters of A, and probably instances of structs that exactly match the port structs of A. To do this inadvertently, reactor A would probably have to send pointers to itself to reactor B. When B is given direct access to such pointers, it is already a huge problem, regardless of whether B accesses that data using the reaction functions of A.
* It isn't clear to me whether requiring inlined reaction bodies would alleviate this problem.

Perhaps less important, but still worth pointing out, is that based on our conversations with Magnition, I think that they understand LF pretty well. Users that are like them would probably not make such a mistake, and in the short term I think it's OK to target them as our audience before we worry about making things idiot-proof.

---

In response to other comments:

> Adding yet another ad-hoc special case where a semicolon is required seems ugly to me. I would prefer reverting to requires semicolons everywhere.

I agree, let's not have individual special cases where semicolons are required. If we cannot resolve the ambiguity in a way that completely avoids the use of special one-off constructs or semicolons, we should require semicolons everywhere.

edwardalee:
> > > The problem with removing the name is that when you reorder reactions you will inadvertently mix up your implementations, which seems terribly inconvenient...
> > 
> > 
> > This seems like the least of the problems, honestly. If a user wants a specific name for the reaction body, they can add one line of code in the .lf file. Note also that methods are available if you want to name functions.
> > I think with this change, we are going to find users calling reactions themselves and bypassing LF semantics routinely.
> 
> The point of this is not to name reactions because we want to call them, but because we want to generate conveniently-named prototypes. If we output each reactor class in its own file, my guess is that those prototypes won't even be in scope for other reactors.

The code Peter showed puts them in scope. You just #include the .h file, and voila, you can call them.  People will figure this out and write their own schedulers.  Or even better, a reaction in one reactor can call a reaction in another!

My point is that your observation that re-ordering reactions will create chaos if the reactions are not named is just the tip of the iceberg of round-tripping problems.  I think that just about any change you make to the .lf file will create problems. If users really want to use this feature (I will advise against it), then they had better get used to not changing the .lf file.

cmnrd:
I think we need to avoid the semicolon. One could argue that it is similar to C/C++, where you put a `;` after a function that you declare, but that it doesn't apply to the other targets and it is not idiomatic for LF. It seems like the discussion above is short on proposals on how to actually resolve the ambiguity with a semicolon. Maybe we can do something like this:
```
reaction (foo) bar -> baz stub name
```
where `stub` is a keyword and `name` is an arbitrary identifier. This however, only allows names for 'stub' reactions.

Generally, the ambiguity in the grammar, that we need to deal with here, is a symptom of a deeper problem. We are using the `->` operator for two very different things. Perhaps we should rethink this syntax choice. (This is probably material for a later discussion though)

I don't see the problem with naming reactions. I think giving them names actually resolves a lot of issues. For instance, it makes the generated code easier to understand, and it makes it easier to refer to reactions by an identifier (think about how you would document a reactor; referring to reactions by their order is fragile). Also if you use debuggers or profilers having named reactions is a tremendous help. Whenever I debug smth, I only see that an error occurred in `reaction_2`, but I don't know what this reaction does without going back to the LF code and counting reactions. Even with mandatory reaction bodies, I only see advantages for named reactions.

I don't think that hiding reaction bodies by giving them obscure names is a good solution anyway. If a user decides it is a good idea to call reactions directly, they will do it and it does not matter how the function that they call is named. If anything, the numeric names give us "security by obscurity". There are other ways, though, to make it hard to directly call reaction bodies. As Peter pointed out, reactions need to be given some context information anyway, and this context cannot easily be created out of the blue. We can also use the context to apply additional checks (e.g. a reaction could get a token that it needs to show when setting a port or scheduling an action).

> My point is that your observation that re-ordering reactions will create chaos if the reactions are not named is just the tip of the iceberg of round-tripping problems. I think that just about any change you make to the .lf file will create problems. If users really want to use this feature (I will advise against it), then they had better get used to not changing the .lf file.

I disagree with that comment. Most of the changes that you can make in LF files do not require updating the target code. The only changes that require an update of the reaction bodies, are changes where the reaction signature changes or where one of the state variables that a reaction uses changes. These are exactly the changes that would also require an update to the reaction bodies if they are embedded in the LF code. So the only real difference is that now two files need to be modified. I don't see where the problem with this is. For C/C++ programmers, this is already what they are used to. If we change the function declaration in the header file, we also have to update the definition in the source files. Also in any other language, if the programs are sufficiently large, a modification in one file will likely require more modifications in other files.

For me, the beauty of removing the target code lies in the fact that a reaction definition if LF becomes a pure interface specification, and there can be multiple implementation of that interface (potentially even in different languages). Of course, if the interface changes, the implementations also need to be updated. But again, this needs to be done in any programming language.

lhstrh:
Is [39a4219](https://github.com/lf-lang/lingua-franca/pull/1593/commits/39a42199a16f78243d88088f3906606107920f13), inspired by @cmnrd's suggestion, any better?

Basically, one could write `reaction(foo) bar -> baz named foobar` and also `reaction(foo) bar -> baz {= // stuff =} named foobar` or  `reaction(foo) bar -> baz {= // stuff =}` but not  `reaction(foo) bar -> baz`

cmnrd:
It is a bit verbose, but I like the proposal! There is one small change that I would like to make: I find it cleaner to write `reaction(foo) bar -> baz named foobar {= // stuff =}`. So in the grammar rule with the code body, the optional name and the code body would switch positions. This should also be unambiguous and puts the complete reaction signature on one line.

PR #1594, opened 2023-02-16
Non-destructive ReactionInstanceGraph.assignLevels

marcusrossel:
For the Lean target, I need to generate a schedule as part of code generation ([source](https://github.com/lf-lang/lingua-franca/blob/de44a7f736ceb84e52045a33c388aa1d174299b8/org.lflang/src/org/lflang/generator/lean/LeanGenerator.kt#L248)). For this I need to get a topological ordering on the reaction instance graph. The `ReactionInstanceGraph.nodesInReverseTopologicalOrder` method seems to be intended for this use case. 
The problem is that in its current form this method always returns an empty list. This seems to be a result of the fact that `ReactionInstanceGraph.assignLevels` is a destructive operation. 
This pull request implements a non-destructive version of `assignLevels`. This comes with some issues/caveats of its own though:
1. As far as I can tell, the C target relies on the fact that `assignLevels` is destructive.
2. The `PrecedenceGraph` type has a `sortNodes` method which seems to implement the same algorithm as I did (but the recursive version). I'm not sure what to do with that.

I'm confused about the graph API and how to proceed on this, so any feedback would be greatly appreciated.

### DIFF
--- a/org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
+++ b/org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
@@ -27,15 +27,19 @@ THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package org.lflang.generator;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.Stack;
 import java.util.stream.Collectors;
 
 import org.lflang.generator.ReactionInstance.Runtime;
 import org.lflang.generator.c.CUtil;
 import org.lflang.graph.PrecedenceGraph;
 import org.lflang.lf.Variable;
+
+import kotlin.Pair;
 
 /**
  * This class analyzes the dependencies between reaction runtime instances.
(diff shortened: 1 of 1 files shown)


org.lflang/src/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -27,16 +27,20 @@ STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
...
 import java.util.Set;
+import java.util.Stack;
 import java.util.stream.Collectors;
 
 import org.lflang.generator.ReactionInstance.Runtime;
 import org.lflang.generator.c.CUtil;
 import org.lflang.graph.PrecedenceGraph;
 import org.lflang.lf.Variable;
 
+import kotlin.Pair;
```
edwardalee, line 42: Do we really need a dependence on kotlin here? This precludes compiling in Eclipse, and this extra dependence looks out of proportion with needing a pair.

edwardalee (changes requested):
I'm generally supportive of having a non-destructive algorithm here, but I think there is more work to do. This algorithm is also used to highlight dependency cycles in the LF visualization, which is a pretty important feature. I believe that changing this method to throw an exception instead will break this functionality.

Also, this adds a dependence on Kotlin in a Java file (in order to import a Pair class, which seems like real overkill).

marcusrossel:
> I believe that changing this method to throw an exception instead will break this functionality.

Yes, throwing an exception is currently a placeholder as I didn't know what to do upon detecting a cycle. Do you know where this info would actually need to be written to?

PR #1596, opened 2023-02-21
Physical connections implemented as AST transformation 

erlingrj:
Addresses #616. A simple way of achieving this is to reuse the AfterDelay transformation, just with a Physcial Action instead. This PR does the following:
1. Rename AfterDelayTransformation -> DelayedConnectionTransformation. This takes two paramters `boolean transformAfterDelays` and `transformPhysicalConnections` which enables each of the two different transformations. 
2. Add some extra control logic to transform the physical connection using a generated delay reactor with a physical action instead of a logical one.

The main discussion point is whether we would prefer to make a separate Transformation class for the PhysicalConnections. I initially decided not to, since it is so similar to the AfterDelays. If we decide this is OK, I will make another pass and add documenting comments.


lhstrh (commented):
:+1: This looks pretty clean to me. I don't see the need for a separate class... Thanks for taking on this issue! It has lingered for a long time...

edwardalee (approved):
LGTM!

org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -103,7 +103,7 @@ class TSGenerator(
      */
     override fun doGenerate(resource: Resource, context: LFGeneratorContext) {
         // Register the after delay transformation to be applied by GeneratorBase.
-        registerTransformation(AfterDelayTransformation(TSDelayBodyGenerator, targetTypes, resource))
+        registerTransformation(DelayedConnectionTransformation(TSDelayBodyGenerator, targetTypes, resource, true, false))
```
cmnrd, line 106: Shouldn't we enable the physical connections also in TS?

cmnrd (commented):
This looks good indeed. I think this is exactly how physical connections should be handled. Having a separate transformation would only duplicate code. The only comment I have, is that we should probably also enable physical connection in TS and at a test for it.

hokeun:
@byeong-gil kindly agreed to work on this. Thank you!

erlingrj:
> @byeong-gil kindly agreed to work on this. Thank you!

Not sure I understand. This PR is more-or-less done?

test/TypeScript/src/PhysicalConnection.lf
```diff
@@ -0,0 +1,26 @@
...
+    output y: number
+
+    reaction(startup) -> y {= y = 42 =}
+}
+
+reactor Destination {
+    input x: number
+
+    reaction(x) {=
+        let time =util.getElapsedLogicalTime;
```
lhstrh, line 14:
```suggestion
        let time =util.getElapsedLogicalTime();
```

test/TypeScript/src/PhysicalConnection.lf
```diff
@@ -0,0 +1,26 @@
...
+        console.log("Received event at logical time: " +  time);
+        if (time.isEqualTo(0)) {
```
lhstrh, line 16:
```suggestion
        if (time.isZero()) {
```

lhstrh:
It's because the GitHub Action from ros-tools was broken. I submitted a PR with them and it is fixed now. Updating the branch and rerunning should do the trick.

lhstrh:
The LSP tests are also failing in `master`, but that's being worked on...

PR #1344, opened 2022-09-01
Address`after` delay in Federated Execution of TypeScript

byeong-gil:
Relevant reactor-ts PR: [`reactor-ts#125`](https://github.com/lf-lang/reactor-ts/pull/125)

This PR adds logic for handling `after` delay in federated execution.
`additionalDelay` is reflected when sending `taggedMessage`.

Some functions in TSExtension.kt are modified and added like getNetworkDelay() and getProcessDelay() etc.

The function names and locations might be modified for a semantic reason.

PR #1607, opened 2023-02-27
Upstream connection delays now properly handled in the TypeScript federates

byeong-gil:
The previous version of this PR is [lingua-franca/pull/1344](https://github.com/lf-lang/lingua-franca/pull/1344)

Relevant reactor-ts PR: [reactor-ts/#138](https://github.com/lf-lang/reactor-ts/pull/138)

This PR adds upstream connection delays in the generated TypeScript federate files.


byeong-gil:
It seems like there is a problem with handling `after` delays in docker. The test [TypeScript/src/docker/federated/DistributedCountContainerized.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/TypeScript/src/docker/federated/DistributedCountContainerized.lf) fails. I'll turn this PR into a draft.

petervdonovan:
Apparently the federates are not even being built because something in the generated code does not type check
```
    ERROR: never[]; }' is not assignable to type 'FederateConfig'. [c.ts:17:280]
    ERROR: never[]; }' is not assignable to type 'FederateConfig'. [c.ts:17:280]
    ERROR: TimeValue[][]; }' is not assignable to type 'FederateConfig'. [p.ts:17:294]
    ERROR: TimeValue[][]; }' is not assignable to type 'FederateConfig'. [p.ts:17:294]
```

byeong-gil:
That's weird... Because I cannot find any error on my local computer when I compile and run the test `test/TypeScript/src/docker/federated/DistributedCountContainerized`.
The CI is run with a `master` branch of the `reactor-ts`, right? If it is, there will be no type check error.

petervdonovan:
I was able to reproduce the problem. In my local clone of `reactor-ts` (in my case, located at `/home/peter/lf/reactor-ts`), I checked out `ts-common-timevalue`, and in my local clone of `lingua-franca`, I checked out `ts-upstream-delays`. Then I ran the command 
```
lfc --external-runtime-path /home/peter/lf/reactor-ts test/TypeScript/src/docker/federated/DistributedCountContainerized.lf 
```

and was able to reproduce the problem. Maybe double check to make sure that you passed the correct path to LFC using `--external-runtime-path`?

lhstrh (approved):
Looks OK to me, but it was marked as "draft" so it escaped my attention. Could you please add JavaDoc to the new function that you added in `TSExtension`? Thanks!

Issue #1610, opened 2023-02-28
Building TypeScript files is slow


lhstrh:
Most of the time is wasted during installation `(p)npm install`. It is unclear why it takes so long (`npm install` in `reactor-ts` is not slow). What dependencies are causing the long wait?

axmmisaka:
I tried some TS examples on Ubuntu in https://github.com/lf-lang/examples-lingua-franca and the `npm install` time is, imo, reasonable (given that `npm install` works like the video below I wouldn't say it's too slow); I did saw TS CICD is still running on MACos when Windows/Ubuntu finished running; are there any examples that are particularly worrying?

Uploading 348be066-f46a-11e8-8038-0a580a4b2d84.mp4…

lhstrh:
Interestingly, something funny happens to `jest` with these changes...

axmmisaka:
Speaking of that change, do you think building CICD to auto update `packages.json` would be a good idea?

lhstrh:
Also, see https://github.com/lf-lang/lingua-franca/pull/1611.

lhstrh:
> Speaking of that change, do you think building CICD to auto update `packages.json` would be a good idea?

Do you mean a bot to suggest dependency updates?

lhstrh:
It occurs to me that most of the installation overhead is due to `babel`. We started using `babel` because `tsc` had performance issues and it was considered good practice to only run the type checker when really needed. There were also a few nice-to-have experimental language extensions that were only available through babel at the time. We should probably re-evaluate this choice and revert back using `tsc` to output our `.js` files. @axmmisaka is this something you'd be willing to look into?

axmmisaka:
> > Speaking of that change, do you think building CICD to auto update `packages.json` would be a good idea?
> 
> Do you mean a bot to suggest dependency updates?

Yes, like a…… Dependabot



> It occurs to me that most of the installation overhead is due to `babel`. We started using `babel` because `tsc` had performance issues and it was considered good practice to only run the type checker when really needed. There were also a few nice-to-have experimental language extensions that were only available through babel at the time. We should probably re-evaluate this choice and revert back using `tsc` to output our `.js` files. @axmmisaka is this something you'd be willing to look into?

If `babel` installation is slow it might be because of cache, but I think `tsc` works better for a research project (iirc `babel`‘S advantages are mostly considered front-end facing); that said, considering the reason we included ts, using `ts-node` might be a better idea

[...]

PR #1611, opened 2023-02-28
Faster building of TypeScript

lhstrh:
This PR changes the process for building to stop relying on Babel and instead use plain `tsc`. Also see: https://github.com/lf-lang/reactor-ts/pull/141.

Fixes https://github.com/lf-lang/lingua-franca/issues/1610.

This PR also addresses the fact that `npm install` of `reactor-ts` no longer invokes `tsc`. This is not a problem when a published version of `reactor-ts` is used, but it is a problem when the `reactor-ts` dependency is given as a link the GitHub repository. We not detect it when an unpublished version of `reactor-ts` is used and assume a "dev mode" operation. In dev mode, when `npm` is used for building, an extra step is done to make sure the JavaScript sources are produced. When `pnpm` is used, this is not necessary because `pnpm` apparently runs the `prepublish` script on its own.


lhstrh:
Why are we downloading Gradle (and failing at it)? We are supposed to cache it now, no?

org.lflang/src/lib/ts/package.json
```diff
@@ -3,35 +3,24 @@
...
         "@typescript-eslint/eslint-plugin": "5.33.0",
         "@typescript-eslint/parser": "^5.8.1",
         "eslint": "^8.5.0",
         "typescript": "~4.8.2",
-        "ts-protoc-gen": "^0.12.0"
+        "ts-protoc-gen": "^0.12.0",
+        "rimraf": "^3.0.2"
     },
     "scripts": {
         "check-types": "tsc",
```
lhstrh, line 23: Can we remove this?
```suggestion
```

.github/actions/prepare-build-env/action.yml
```diff
@@ -1,5 +1,10 @@
 name: Prepare build environment (Linux only)
 description: Set up Java, Maven, Gradle, etc.
+inputs:
```
lhstrh, line 3: I don't remember why I added this, but it appears to be no longer used (or useful).

.github/workflows/ts-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         if: ${{ runner.os == 'macOS' }}
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#cleanup-deps"
```
lhstrh, line 31:
```suggestion
          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
```

PR #1619, opened 2023-03-03
Fix for language server "Build and Run" command

petervdonovan:
Closes https://github.com/lf-lang/vscode-lingua-franca/issues/99.


lhstrh (approved):
Great! Thanks for fixing this!

PR #1629, opened 2023-03-09
Fix validation of target properties

cmnrd:
This includes two fixes:
 - Use `equals()` to check equality of strings
 - Fix a hidden nullpointer exception in `validateRos2TargetProperties`


petervdonovan (approved):

lhstrh (approved):
:+1:

PR #1630, opened 2023-03-09
Infer keeplive natively in the C++ runtime

cmnrd:
This PR pulls in https://github.com/lf-lang/reactor-cpp/pull/42 which allows the C++ runtime to automatically infer if keepalive is needed. Consequently, the `keepalive` target property in LF is disabled for the C++ target. If it is used anyway, a warning will be printed stating that the property is ignored.


revol-xut (approved):
Looks good to me

cmnrd:
> Why not do this for all targets?

We should probably do it for all targets. But this of course requires support by the different runtimes and code generators.

Also, I think there are still some open semantic questions. For instance, the behavior of timeout in the C target is influenced by the value of `keepalive` (see discussion in https://github.com/lf-lang/lingua-franca/issues/1496). I am not sure if we would want to keep it this way if `keepalive` is inferred and not set explicitly.

petervdonovan (approved):
LGTM
The TS test failure looks like just a network issue. Let's ignore it.

lhstrh (approved):

PR #1631, opened 2023-03-09
Fix passing of command line options from lfc to the generator

cmnrd:
Currently, all command line options that need to be passed on from lfc to the code generator do not work. The reason for this, is that the long option names (e.g. `--clean`) are compared with the property names (e.g. `clean`).  This change provides a fix. Concretely, it replaces the automagic passing in a for loop from before with explicit code for each individual option. While it is a bit more verbose, this has some advantages:
- The code is readable and understandable
- IntelliJ produces a useful warning when an option is not used and therefore not passed on

This change further removes some unused code and includes other small cleanups. All in all, this removes all warnings shown bu IntelliJ.

### DIFF
--- a/org.lflang/src/org/lflang/cli/CliBase.java
+++ b/org.lflang/src/org/lflang/cli/CliBase.java
@@ -1,22 +1,15 @@
 package org.lflang.cli;
 
 import java.io.IOException;
-import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


patilatharva (approved):
Thank you for this fix!

lhstrh (approved):

lhstrh:
This bugfix warrants a path release...

cmnrd:
Did we merge the picocli PR before the release? I think it wasn't in there. It also does not show in the release notes.

lhstrh:
Oh, right, we didn't.

PR #1632, opened 2023-03-09
Tracing of federate interactions

edwardalee:
This PR (co-authored by @ChadliaJerad) adds tracepoints for interactions between federates and the RTI plus a visualization tool for the resulting traces.  A major goal is to enable optimizing coordination of federated programs by reducing the number of RTI-federate interactions. It adds a command-line utility called `fedsd` that generates sequence diagrams (augmented with physical time) for federate interactions with the RTI (see an example below).

This PR relies on the tracing-federates branch of reactor-c, which should be merged first. To use it, PR #1642 must also be merged first (or else you have to manually edit the launch script to add the `-t` option to the RTI).  

The `fedsd` utility and the other utilities in `util/tracing` should probably be split off into a separate repo dedicated to analysis and visualization of LF programs, but that should be done in a separate PR and should be accompanied with setting up regression tests for that repo.

Example federated program:

image

Running this generates three `.lft` (trace) files. Then running:

```
> fedsd *.lft
```

produces a visualization of the trace. A segment is shown below showing the one message flowing from ping to pong and back again:

image


cmnrd (changes requested):
This PR should not modify the reactor-cpp submodule

lhstrh (approved):

PR #1633, opened 2023-03-09
Set external runtime path

petervdonovan:
When we make changes to reactor-ts and accompanying changes in
lingua-franca we cannot point to the version of reactor-ts that is
published on NPM. We need to clone a specific ref and point to it.

Locally, pointing to the correct runtime version fixes the problem that
I observed in https://github.com/lf-lang/lingua-franca/pull/1607. If this change to the yaml file is correct then
this should make that PR mergeable.

This PR also fixes a bug in the way we pass CLI args because this bug was inhibiting my ability to debug #1607. I decided to fix the bug rather than working around it

### DIFF
--- a/.github/workflows/ts-tests.yml
+++ b/.github/workflows/ts-tests.yml
@@ -34,7 +34,7 @@ jobs:
           ref: master
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="./reactor-ts"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master
       - name: Report to CodeCov
         uses: codecov/codecov-action@v3.1.1
         with:
--- a/.github/workflows/ts-tests.yml
+++ b/.github/workflows/ts-tests.yml
@@ -26,9 +26,15 @@ jobs:
         run: |
           brew install coreutils
         if: ${{ runner.os == 'macOS' }}
+      - name: Clone reactor-ts
+        uses: actions/checkout@v3
+        with:
+          repository: lf-lang/reactor-ts
+          path: ./reactor-ts
+          ref: master
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.*
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="./reactor-ts"
       - name: Report to CodeCov
         uses: codecov/codecov-action@v3.1.1
         with:
--- a/.github/workflows/ts-tests.yml
+++ b/.github/workflows/ts-tests.yml
@@ -34,7 +34,7 @@ jobs:
           ref: master
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
       - name: Report to CodeCov
         uses: codecov/codecov-action@v3.1.1
         with:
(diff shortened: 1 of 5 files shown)


org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -261,27 +261,26 @@ protected Properties filterPassOnProps() {
...
             BuildParm.WORKERS)
-        .map(param -> param.getKey())
+        .map(BuildParm::getKey)
         .collect(Collectors.toUnmodifiableSet());
 
         Properties props = new Properties();
 
         for (OptionSpec option : spec.options()) {
             String optionName = option.longestName();
+            if (optionName.startsWith("--")) optionName = optionName.substring(2);
```
lhstrh, line 271:
```suggestion
```

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -261,27 +261,26 @@ protected Properties filterPassOnProps() {
...
             BuildParm.THREADING,
             BuildParm.WORKERS)
-        .map(param -> param.getKey())
+        .map(BuildParm::getKey)
         .collect(Collectors.toUnmodifiableSet());
 
         Properties props = new Properties();
 
         for (OptionSpec option : spec.options()) {
             String optionName = option.longestName();
```
lhstrh, line 270:
```suggestion
            String optionName = option.descriptionKey();
```

.github/workflows/ts-tests.yml
```diff
@@ -26,9 +26,15 @@ jobs:
...
+      - name: Clone reactor-ts
+        uses: actions/checkout@v3
+        with:
+          repository: lf-lang/reactor-ts
+          path: ./reactor-ts
+          ref: master
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.*
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="./reactor-ts"
```
lhstrh, line 37: Why are we doing this? If you want to point to `master`, you can also write:
```suggestion
          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master
```
(See https://github.com/lf-lang/lingua-franca/tree/master/test/TypeScript)

I am a little confused though that you found that `--external-runtime` didn't work, because I used it successfully earlier this week :thinking:

petervdonovan: Earlier this week I mentioned that I was able to reproduce a bug using Gradle but not using `--external-runtime-path`, so I assumed that the bug fixed in this PR was the reason. `lfc` was silently failing to change the runtime in the generated `package.json` from `3.2.0` to a file system path; I confirmed this by locally running `lfc` before and after my changes.

Are you sure that `--external-runtime-path` was actually working, and not just silently doing nothing? If so, then there must be a recent change that is in `master` but not in `ts-upstream-delays` that fixed this problem.

cmnrd: The passing of command line options was broken in master and is fixed by https://github.com/lf-lang/lingua-franca/pull/1631

petervdonovan: Thanks Christian! That makes the first change in this PR unnecessary.

org.lflang/src/org/lflang/cli/Lfc.java
cmnrd: I think I commented this earlier, but I couldn't see my comment anymore. This issue was fixed on master https://github.com/lf-lang/lingua-franca/pull/1631 in a different way.

.github/workflows/ts-tests.yml
```diff
@@ -26,9 +26,15 @@ jobs:
...
+      - name: Clone reactor-ts
+        uses: actions/checkout@v3
+        with:
+          repository: lf-lang/reactor-ts
+          path: ./reactor-ts
+          ref: master
       - name: Perform TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.*
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master
```
petervdonovan, line 37:
```suggestion
          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
```

Issue #1492, opened 2022-11-23
Epic: LF syntax enhancements


oowekyala:
This is an attempt to give an overview of planned changes to the LF syntax, and also of those changes that are still in the discussion/design state. (I'll fork off some new tickets to have more room for discussing individual items)

The goals of most of these changes are
- Make the syntax more expressive (new expression forms, constructor enhancements, required parameters)
- Make the syntax more consistent, both internally, and w.r.t. our target languages.
- Better specify some parts of code generation (eg parameter scope, preamble scope). This makes it easier to understand LF code, to implement code generators, and to build comprehensive test suites.

## In progress/planned

- [ ] #986
    - [ ] #1441: cleanup initializer grammar
    - [ ] Introduce `=` initializer syntax, deprecate other syntaxes in targets other than C++
    - [ ] Introduce new expression forms, including
        - [ ] Constructor call shorthand for Rust/TS/Python
        - [ ] List expressions
- [ ] #623 
- [ ] #657 
- [ ] @cmnrd C++ code generation change: use a struct to pass parameters with designated initializers. Initializer lists don't work properly as reactor parameters currently.

## Design work needed

Currently unplanned or not discussed at all

TODO @oowekyala: for those items that have no ticket, I'll open one soon to add details and call for discussion.

- [ ] #1238
- [ ] #1479
    - [ ] Proposals: `token<t>`, or just `@token` on the port?
- [ ] Preamble semantics
- [ ] Parameter semantics
- [ ] LF Instance declaration syntax

PR #1639, opened 2023-03-10
Types allowed in reactor type args

oowekyala:
This allows writing `new R<std::vector<int>>(...)` without needing curly braces around the type argument. This is appropriate since it's a type argument (so, a type), not a type parameter. It was surprising that it's possible to write eg `std::optional<int>` without escaping in a type annotation anywhere in LF, except in that place. See for example a use case here:

https://github.com/lf-lang/examples-lingua-franca/blob/e1dd212f28a4ccaf7ff6d717a488a113c5b466a4/Cpp/src/RequestResponse/MAC.lf#L26

Also rename type params to type args where appropriate, for consistency's sake.

Ref #1492 (grammar cleanups)


cmnrd:
Thanks for taking a shot at this! I have struggled with it a couple of times, but never made the change.

lhstrh (approved):
Fantastic! This is a great improvement. Thank you, @oowekyala!

PR #1643, opened 2023-03-12
Fix in validation so that warnings are not reported as errors

petervdonovan:
This does not change the output of `lfc` even when an error appears in an imported reactor.

The code just was not doing much except for redundant work, since it was being invoked after a round of validation checks had already passed. Actually it was even worse than that, because it was running the validator without the right `ValidationIssueAcceptor` injected, resulting in behavior that was not consistent with that implemented in `StandaloneIssueAcceptor`. This is why warnings were (wrongly) being converted to errors.

I decided to delete the method in `GeneratorUtils` rather than a) make this code work properly and b) spend some unknown number of hours fighting with Guice. I think this is an acceptable solution because it is doubtful whether the behavior that the method was supposed to implement is what we want. Specifically, if an error is reported on an imported reactor, it is not obvious to me that the error should also be reported on import statements that import the reactor.

### DIFF
--- a/org.lflang/src/org/lflang/generator/GeneratorBase.java
+++ b/org.lflang/src/org/lflang/generator/GeneratorBase.java
@@ -267,7 +267,6 @@ public abstract class GeneratorBase extends AbstractLFValidator {
         // to validate, which happens in setResources().
         setReactorsAndInstantiationGraph(context.getMode());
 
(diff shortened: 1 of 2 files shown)


lhstrh:
Looks reasonable to me...

cmnrd (approved):
Thanks! I also noticed recently that the validation runs twice, but I didn't know why until now.

PR #1644, opened 2023-03-12
Disable the "dubious ownership" thing.

petervdonovan:
See https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9 for explanation of the problem. We are not accessing repositories on a shared drive that we do not control. I do not see how this security issue could apply to us.

I am changing to a wildcard `*` in place of what Erling wrote 4 days ago because I was still having issues running CI from `reactor-c`, and because the error message I got told me to do something different:
`git config --global --add safe.directory /__w/reactor-c/reactor-c`
I am trying to avoid having to play whack-a-mole with the directories that we need to mark as safe.


erlingrj:
Looks good to me!

erlingrj (approved):

cmnrd (changes requested):
I don't fully understand what is going on here, but using a wildcard seems too permissive. If we do this, we need to fully understand the consequences.

The strange thing is that the checkout action is already supposed to mark the cloned directory as safe. Quote from the README:
```
# Add repository path as safe.directory for Git global config by running `git
# config --global --add safe.directory <path>`
# Default: true
set-safe-directory: ''
```
It should be set to true by default. But maybe setting it explicilty helps?

lhstrh (approved):

cmnrd (approved):
Thanks!

PR #1648, opened 2023-03-13
Fix zephyr tests

erlingrj:
Use a known version of the Zephyr build docker image. Was using `latest` and some incompatible changes were released yesterday


cmnrd (approved):
Thanks!

PR #1655, opened 2023-03-15
Fix NPE in lfc error reporting

cmnrd:
The lfc tool currently fails with an NPE if an issue is reported without an associated resource (see https://github.com/lf-lang/lingua-franca/issues/1650). This PR fixes this NPE, but not the source of the error in the program mentioned in https://github.com/lf-lang/lingua-franca/issues/1650.


lhstrh (approved):
This looks good to me. We should still figure out why the NPE was triggered in the first place...

PR #1660, opened 2023-03-16
Remove klighd dependency and excludes org/lflang/diagram from gradle compilation

cmnrd:
This is a temporary workaround for the gradle compilation issues that currently appear as the klighd dependency cannot be resolved. I hope we can fix it soon in other ways, but if there is no quick fix we might need to merge this workaround. This will break the language server though...

### DIFF
--- a/gradle/source-layout.gradle
+++ b/gradle/source-layout.gradle
@@ -6,8 +6,14 @@ if (name.endsWith(".tests")) {
 			resources.srcDirs = []
 		}
 		test {
-			java.srcDirs = ['src', 'src-gen']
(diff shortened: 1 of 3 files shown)


lhstrh:
We're not testing the diagrams, which also explains the dip in code coverage that we had after factoring out Epoch...

cmnrd:
Yes, but the lsp has a dependency on the diagramms, so I am surprised that it compiles without them. Vor are wie still using the old build mechanism and not gradle for building the lsp?

lhstrh:
Merging this so that things can build again...

PR #1665, opened 2023-03-18
Enclave connections and enclave coordination in the C++ target

cmnrd:
This PR extends the C++ target with support for connections between enclave. Most of the functionality is added in https://github.com/lf-lang/reactor-cpp/pull/44 which most importantly introduces three new connection types: `EnclaveConnection`, `DelayedEnclaveConnection`, and `PhysicalEnclaveConnection`. This PR mostly updates the code generator such that the corresponding enclave connection type is inserted into the generated code if an enclave is involved in the connection. This PR also pulls in https://github.com/lf-lang/reactor-cpp/pull/45 which prevents a race condition between staring enclaves.

There remain open issues which will be covered in feature PRs:
- Zero delay loops between enclaves are not supported. This itself is not much of an issue, as the sequential execution implied by a zero delay loop cancels the benefits we hope to get from using enclaves in the first place. Consequently, a program that requires a zero delay loop should not use enclaves in this loop. However, the actual problem is, that there is currently no detection mechanism for zero delay loops implemented and no warning or error is printed. An application with a zero delay loop between enclaves silently deadlocks. We should add a detection mechanism to the validator to the extend possible and also produce warnings at runtime (for corner cases that are not statically detectable).
- Cyclic dependencies with a small delay are inefficient. Currently, loops are not detected (neither statically nor dynamically). A loop between enclaves will create a chain of acquire requests. This chain might get very long if the time decrements (given by the delay on the loop) are small.


org.lflang/src/lib/cpp/lfutil.hh
```diff
@@ -27,6 +27,8 @@
 #include <reactor-cpp/reactor-cpp.hh>
 #include <reactor-cpp/logging.hh>
 
+#include <list>
```
revol-xut, line 30: I guess `std::list<T>`  is already in use but I can tell u from experience `std::list<T>`  ist a big performance killer

org.lflang/src/org/lflang/generator/cpp/CppConnectionGenerator.kt
```diff
@@ -42,18 +69,24 @@ class CppConnectionGenerator(private val reactor: Reactor) {
...
-        if (connection.requiresConnectionClass) {
-            if (connection.hasMultipleConnections) {
-                "std::vector<${connection.cppType}> ${connection.name};"
-            } else {
-                "${connection.cppType} ${connection.name};"
+        with(connection) {
+            if (requiresConnectionClass) {
+                when {
+                    hasMultipleConnections && isEnclaveConnection  -> "std::list<${connection.cppType}> ${connection.name};"
+                    hasMultipleConnections && !isEnclaveConnection -> "std::vector<${connection.cppType}> ${connection.name};"
```
revol-xut, line 76: why do you generate here different containers ?

org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
```diff
@@ -36,29 +36,49 @@ class CppInstanceGenerator(
...
+    }
+
+    private fun Instantiation.generateWrapper(): String = """
+            |struct $enclaveWrapperClassName {
+            |  ${if (!hasEachParameter) "inline static " else ""}std::unique_ptr<reactor::Environment> __lf_env{nullptr};
+            |  std::unique_ptr<$reactorType> __lf_instance{nullptr};
+            |  
+            |  $enclaveWrapperClassName(const std::string& name, reactor::Reactor* container, $reactorType::Parameters&& params) {
+            |    if (__lf_env == nullptr) {
+            |      __lf_env = std::make_unique<reactor::Environment>(container->fqn() + name, container->environment());
```
revol-xut, line 66: noice, I like the std::unique_ptr

revol-xut (approved):
Looks good so far just left a couple of questions

org.lflang/src/lib/cpp/lfutil.hh
cmnrd: It is just used during instantiation, so it shouldn't be too bad

org.lflang/src/org/lflang/generator/cpp/CppConnectionGenerator.kt
cmnrd: I don't remember exactly, but I think the problem was that vector requires that the elements are movable. I will check this again and should add a comment here

cmnrd: The problem was indeed that connections are not default constructable and not safely movable at the moment. However, I fall back on using a vector of unique pointers to connections now.

revol-xut (approved):

org.lflang/src/org/lflang/generator/cpp/CppInstanceGenerator.kt
```diff
@@ -36,29 +36,49 @@ class CppInstanceGenerator(
...
+        val Instantiation.enclaveWrapperClassName get() = "EnclaveWrapper_$name"
+    }
+
+    private fun Instantiation.generateWrapper(): String = """
+            |struct $enclaveWrapperClassName {
+            |  ${if (!hasEachParameter) "inline static " else ""}std::unique_ptr<reactor::Environment> __lf_env{nullptr};
+            |  std::unique_ptr<$reactorType> __lf_instance{nullptr};
+            |  
+            |  $enclaveWrapperClassName(const std::string& name, reactor::Reactor* container, $reactorType::Parameters&& params) {
+            |    if (__lf_env == nullptr) {
```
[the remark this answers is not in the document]
cmnrd: We use the `__lf_` prefix for some variables to avoid name clashes. It is unlikely, but if someone would name their state variable or some other reactor component `lf_env_`, we would see compile errors. `__` is kind of a protected namespace in C/C++ that is not supposed to be used by users.

Issue #1641, opened 2023-03-10
Test the arguments of all our command line tools


cmnrd:
The bug fixed in https://github.com/lf-lang/lingua-franca/pull/1631 showed that our current CLI tests are merely smoke tests and not sufficient to detect if arguments are passed properly and have the intended effect. We should add thorough tests for all command line arguments.

patilatharva:
Would it be possible to run a select few unit tests from `org.lflang.tests` using the existing `CliToolTestFixture`? Or do you think we should just write new tests specifically for the CLI?

cmnrd:
I think so, yes. When writing the issue, I forgot that @oowekyala already made a start at this.  I think it only makes sense to build on this and add tests for all the command line arguments. Most important is, that the tests indeed verify the outcome of the argument.

PR #1668, opened 2023-03-20
Test for parsing CLI arguments in `lfc`

patilatharva:
This PR adds a test that tests if the CLI arguments passed into Lfc match the values in the target properties `Properties` object which is passed to the generator. The test only checks that the desired values are passed to the generator and doesn't test the functionality of those values, because I believe that would belong in the LF unit tests and be out of scope for the CLI tests.

**Rationale:** https://github.com/lf-lang/lingua-franca/issues/1641 - The bug fixed in https://github.com/lf-lang/lingua-franca/pull/1631 showed that our current CLI tests are merely smoke tests and not sufficient to detect if arguments are passed properly.

**Implementation notes:** I preserved the structure of the existing `CliToolTestFixture.java` as much as possible, but chose to test by creating an Lfc object rather than using static methods. It seemed like the simplest way to gain visibility into the instance methods (which are necessary for Picocli), but let me know if there's a better way around this in Java.


org.lflang.tests/src/org/lflang/tests/cli/CliToolTestFixture.java
```diff
@@ -87,6 +92,36 @@ public ExecutionResult run(Path wd, String... args) {
...
 
+    /**
+     * Run and return and Lfc instance with the given arguments, in the given
+     * working directory.
+     *
+     * @param wd working directory
+     * @param args Arguments
+     * @return The execution result
+     */
+    public LfcObjExecutionResult runLfcObj(Path wd, String... args) {
```
oowekyala, line 103: I don't think we need to add that much code for this. We could just subclass `CliToolTestFixture` and save the `Lfc` instance in there. Otherwise there is a lot of code duplication.

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
```diff
@@ -101,10 +106,56 @@ public void testGenInSrcDir(@TempDir Path tempDir) throws IOException {
...
+            "--scheduler", "2",
+            "--threading", "false",
+            "--workers", "1",
+        };
+
+        lfcTester.runLfcObj(tempDir, args)
+            .verify(result -> {
+                result.checkOk();
+                // Get the properties method.
+                Method getPropsMethod =
```
oowekyala, line 137: Calling methods by reflection in test code is an anti pattern. It makes usage search and static type checking useless. Even though the method is an implementation detail, I think the lesser evil is definitely to make it public.

oowekyala (commented):
Thanks for taking this on! I will push a couple of changes in the direction of my comments

patilatharva:
@oowekyala thanks for the changes, it looks much better now!

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
```diff
@@ -98,17 +105,71 @@ public void testGenInSrcDir(@TempDir Path tempDir) throws IOException {
...
+            "--build-type", "Release",
+            "--clean",
+            "--target-compiler", "gcc",
+            "--external-runtime-path", "src",
+            "--federated",
+            "--logging", "4",
+            "--lint",
+            "--no-compile",
+            "--quiet",
+            "--rti", "-1",
```
lhstrh, line 126: The description of this switch is "Specify the location of the RTI." Then `-1` should not be a valid entry...

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
```diff
@@ -98,17 +104,71 @@ public void testGenInSrcDir(@TempDir Path tempDir) throws IOException {
...
+                // Don't validate execution because args are dummy args.
+                Properties properties = fixture.lfc.cliArgsToBuildParams();
+                assertEquals(properties.getProperty(BuildParm.BUILD_TYPE.getKey()), "Release");
+                assertEquals(properties.getProperty(BuildParm.CLEAN.getKey()), "true");
+                assertEquals(properties.getProperty(BuildParm.EXTERNAL_RUNTIME_PATH.getKey()), "src");
+                assertEquals(properties.getProperty(BuildParm.LINT.getKey()), "true");
+                assertEquals(properties.getProperty(BuildParm.LOGGING.getKey()), "4");
+                assertEquals(properties.getProperty(BuildParm.TARGET_COMPILER.getKey()), "gcc");
+                assertEquals(properties.getProperty(BuildParm.QUIET.getKey()), "true");
+                assertEquals(properties.getProperty(BuildParm.RTI.getKey()), "-1");
```
lhstrh, line 145: A negative integer is not a valid path...

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
```diff
@@ -98,17 +104,71 @@ public void testGenInSrcDir(@TempDir Path tempDir) throws IOException {
                     .check("bin", isDirectory())
                     .check("src-gen/File/File.py", isRegularFile());
             });
+    }
 
+    @Test
+    public void testBuildParams(@TempDir Path tempDir)
```
lhstrh, line 110: Another test vector that I think we should have is a set of known-bad inputs and the handling of those.

org.lflang/src/org/lflang/cli/Lfc.java
```diff
@@ -229,7 +229,7 @@ private Path getActualOutputPath(Path root, Path path) {
      *
      * @return Properties for the code generator.
      */
-    protected Properties filterPassOnProps() {
+    public Properties cliArgsToBuildParams() {
```
lhstrh, line 232: With this name I would expect the return type to be `List<BuildParm>` or something along those lines. Maybe it should just be called `getCliArgs`?

lhstrh (changes requested):
This looks pretty good! I left some minor requests for changes. Perhaps beyond the scope of this PR but good to do in the future would be to extend the tests to check the resulting `TargetConfig` as well.

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
patilatharva: True, I meant to put garbage values in all the arguments but it would be a good idea to add better CLI argument validation to this PR too.

org.lflang/src/org/lflang/cli/Lfc.java
patilatharva: It doesn't return all the CLI args either—just those required by the generator—so I called it `getGeneratorArgs` instead.

org.lflang.tests/src/org/lflang/tests/cli/LfcCliTest.java
patilatharva: Done!

PR #1672, opened 2023-03-22
Wait for event

edwardalee:
This PR adds two tests for the `lf_wait_for_event()` function introduced in PR https://github.com/lf-lang/reactor-c/pull/183.


edwardalee:
Never mind. It turns out tracing doesn't really need this, so let's not add it.

PR #1674, opened 2023-03-28
TSDockerGenerator and TSTypes ported to Java for Eclipse developers

lhstrh:
This patch works around the fact that Eclipse cannot deal with Kotlin code, which leads to build problems as the ones reported in https://github.com/lf-lang/epoch/issues/13.


lhstrh:
Please observe the following problems with this fix:
 - this very same problem will re-emerge sooner or later;
 - this change uglifies the Kotlin code; and
 - after addressing the build problems in Eclipse,I needed to edit the project in IntelliJ in order to find the error that my changes had introduced for the Kotlin code.

Is this a productive use of developer time?

edwardalee (approved):
These changes look fine to me. The java files are pretty simple.

PR #1678, opened 2023-03-28
Run regression test suite on Zephyr emulation

erlingrj:
This PR enables running parts of the regression test suite on an QEMU-emulation with Zephyr in CI.
To this we have the following changes:
- Add a bash script for compiling and running the emulator and parsing its output and killing the emulator either after a timeout or after the LF program has exited and printed "exit"
- Decouple "TestLevel" from test categories. THis was a change introduced by @arengarajan99, but I don't think it makes sense to say that the GenericTests are bundled with a TestLevel.EXECUTE and ArduinoTests are bundled with TestLevel.BUILD. This change makes the TestLevel and input to the test. This was we can have the buildZephyrGenericTests which configures and builds the GEnericTests for a Zephyr platform.
- To get this to work I need to configure the regression tests for Zephyr by changing the target property. I need to disable tracing and threading (currently) and also set platform to Zephyr. However, the TestBase.java includes a line that reloads the target properties after the Configurator is applied and thus overwrites these changes. I have commented out this line. But I am guessing that this will lead to some other test failures. So we need to find a workaround there.


org.lflang.tests/src/org/lflang/tests/TestBase.java
```diff
@@ -430,7 +433,10 @@ private void configure(LFTest test, Configurator configurator, TestLevel level)
             if (!configurator.configure(test)) {
                 throw new TestError("Test configuration unsuccessful.", Result.CONFIG_FAIL);
             }
-            context.loadTargetConfig(); // Reload in case target properties have changed.
+            // FIXME: The line below is commented out because it reloads the target properties
+            //  and thus overwrites changes done by the configurator. I am sure it was there 
+            //  for a reason. Thus the FIXME...
+            // context.loadTargetConfig(); // Reload in case target properties have changed.
```
erlingrj, line 439: OBS. What to do about this. I don't understand the comment either. Why would we want to reload and overwrite the target properties if they have  changed?

erlingrj:
Also fixes this: https://github.com/lf-lang/reactor-c/issues/180

.github/workflows/c-zephyr-tests.yml
```diff
@@ -58,4 +58,7 @@ jobs:
           ref: ${{ inputs.runtime-ref }}
         if: ${{ inputs.runtime-ref }}
       - name: Perform Zephyr tests for C target with default scheduler
-        run: ./gradlew test --tests org.lflang.tests.runtime.CZephyrTest.runZephyrTests
+        run: |
+          ./gradlew test --tests org.lflang.tests.runtime.CZephyrTest.buildZephyrTests
+          ./gradlew test --tests org.lflang.tests.runtime.CZephyrTest.buildGenericTests
```
lhstrh, line 63: Could we just a single line `./gradlew test --tests org.lflang.tests.runtime.CZephyrTest.*` instead of two lines? I recall that wildcards are accepted.

.github/workflows/ci.yml
```diff
@@ -20,15 +20,15 @@ jobs:
   cancel:
     uses: lf-lang/lingua-franca/.github/workflows/cancel.yml@master
 
-  # Test the Gradle build.
-  build:
-    uses: lf-lang/lingua-franca/.github/workflows/build.yml@master
-    needs: cancel
+  # # Test the Gradle build.
```
lhstrh, line 23: The changes to this file need to be reverted before we can merge this in.

org.lflang.tests/src/org/lflang/tests/TestBase.java
lhstrh: I agree that in its current form it makes no sense, but I think the intent here is to "reset" the configuration in between test runs. Hence, if this line is just moved to first line of the block starting with `if (configurator != null)`, I think it would work as intended?

test/C/src/zephyr/HelloZephyr.lf
```diff
@@ -1,10 +1,4 @@
-target C {
-    threading: false,
-    platform: {
-        name: Zephyr,
-        board: qemu_cortex_m3
-    }
-}
+target C
```
lhstrh, line 1: Why is the `platform` property removed?

lhstrh (changes requested):

lhstrh (approved):

lhstrh (approved):

PR #1684, opened 2023-04-02
No more use of unordered reactions in federated programs

### LINKED ISSUE #1708: Ambiguous reaction precedence rules cause deadlock
See #1707 for a draft of a test that reproduces this. Let's not lose track of @lhstrh's comment that
> I just want us to resist the temptation to treat this as just another bug that needs a quick fix.

Basically there are a number of people on the team who are in favor of rethinking the architecture a little bit rather than making minimal patches. This bug is of a sort that justifies careful thought because it is easy to create a partial solution that gets this one test to pass but that might not solve all of the ambiguities.

For the record, I should admit that the `fed-gen` refactoring, which I was in favor of and which I believed to be basically correct, probably did introduce this problem. Before, when we assigned levels globally, a problem of this kind could not occur. However, I and others are still unsure that global level assignments are the best solution.

arengarajan99:
The current runtime support for federated execution relies on a set of blocking unordered "control reactions" for each input port that to block the handling of reactions until the required port statuses are known. The drawbacks of this design are discussed in [Rengarajan 2023](https://www2.eecs.berkeley.edu/Pubs/TechRpts/2023/EECS-2023-185.html), along with a solution that no longer relies on the use of blocking unordered reactions. This PR implements that solution.

This work was initiated by @arengarajan99 and completed by @petervdonovan and @byeong-gil.

### Companion PRs
- https://github.com/lf-lang/reactor-c/pull/191
- https://github.com/lf-lang/reactor-ts/pull/186

### Tasks completed
- [x] Use a less hacky mechanism for marking ports that receive from delayed connections as not part of the MLAA calculation (@petervdonovan)
- [x] Find dead/unnecessary code in the code generator and delete it (split this into sub-tasks as needed) (the CodeCov reports may help with this)
- [x] Pass the other C tests besides the C federated tests
- [x] Ensure IntelliJ does not emit warnings related to the code that is touched by this and the other related PRs
- [x] Get this branch up to date with `master` (@petervdonovan)
- [x] Check that this also works for the Python target (@petervdonovan)
- [x] Factor C-specific code out of FedASTUtils and into CExtension or CExtensionUtils (@petervdonovan)
- [x] Get all tests in reactor-c to pass (@petervdonovan)
- [x] Get the TypeScript runtime to be compatible with this branch (@byeong-gil)
- [x] Get the Python tests to pass (before fixing the Python tests) (@arengarajan99)
- [x] ~~Fix the broken C tests and check that they still pass~~ See https://github.com/lf-lang/lingua-franca/pull/1865#issuecomment-1636513838 (EDIT: This was explained by Edward [here](https://github.com/lf-lang/lingua-franca/pull/1684#issuecomment-1637055585) and has since been properly addressed.)
- [x] Fix the broken Python tests and check that they still pass or that they are at least not regressing wrt `master` (@petervdonovan)
- [x] Check that all code added has JavaDoc (@petervdonovan)
- [x] Everywhere rename `outputControlReaction` -> `portAbsentReaction`
- [x] Do not require lots of threads to be available in order to correctly execute federated programs
- [x] Find dead/unnecessary code in the C runtime and delete it

For remaining non-required tasks, see: https://github.com/lf-lang/lingua-franca/issues/1922

Fixes #1708.


petervdonovan:
It seems like only 3 federated tests are failing. This is surprising to me because I commented out some things that seemed important in order to get anything to compile :sweat_smile:

Right now, the problem seems to be that after receiving only a PTAG, a federate finishes a tag and sends an LTC even though it might receive more data on one of its ports.

petervdonovan:
@arengarajan99 mentioned to me that tests with multiports should probably be failing, and indeed, 54/57 tests passing was too good to be true. In the commit where I got that result, if I run `BroadcastFeedback`, the output includes the following:
```
Federate 0: Starting timestamp is: 1686333071005126060.
Federate 1: Starting timestamp is: 1686333071005126060.
Federate 1: FATAL ERROR: Received a message at tag (0, 0) that has a tag (0, 0) that has violated the STP offset. Centralized coordination should not have these types of messages.
---- Elapsed physical time (in nsec): 47,917,422
Federate 1 has resigned.
Federate 0: FATAL ERROR: Received a message at tag (0, 0) that has a tag (0, 0) that has violated the STP offset. Centralized coordination should not have these types of messages.
---- Elapsed physical time (in nsec): 48,131,777
Federate 0 has resigned.
```
But the result of running `echo $?` afterward is `0`.

petervdonovan:
@lhstrh To answer your question I have double-checked that the variant of the `SpuriousDependency` test that is added in this branch does deadlock in `master`.

petervdonovan:
In the latest run the failing tests were
1. DistributedCountDecentralizedLate
2. DistributedCountDecentralizedLateDownstream
3. DistributedLoopedAction
4. DistributedStopZero

In order to find this information I had to click "view raw log" and fiddle with the `Ctrl+F` search tool a little in the browser -- there is so much output (about 11000 lines) from DistributedCountDecentralizedLateDownstream that the usual tools for searching do not work well.

Note that two of these (DistributedLoopedAction and DistributedStopZero) are also failing in https://github.com/lf-lang/lingua-franca/pull/1835, which means they are as good as failing in `master`.

petervdonovan:
`DistributedCountDecentralizedLate` gives invalid results in both this branch in `master`, although in `master`, at least, it does not fail because of #1835. I am not sure I understand the test, but it looks like it needs to advance logical time as fast as possible in order to force an STP violation. This seems hard to do reliably, period, and also hard to do without producing an unacceptable volume of debug messages when the log level is set to debug (because whenever time is advanced, debug messages are printed). The volume of debug messages may be the cause of the timeout, but I'm not sure.

`DistributedCountDecentralizedLateDownstream` appears to be similar.

It is possible that the best choice will turn out to be not to use these tests in CI. I wasn't the one who added the tests, though, so I am not sure why they were originally considered to be important.

Overall I think that at this point it is dubious whether the Python test failures we are currently seeing indicate any actual regressions wrt `master`.

edwardalee:
> `DistributedCountDecentralizedLate` gives invalid results in both this branch in `master`, although in `master`, at least, it does not fail because of #1835. I am not sure I understand the test, but it looks like it needs to advance logical time as fast as possible in order to force an STP violation. This seems hard to do reliably, period, and also hard to do without producing an unacceptable volume of debug messages when the log level is set to debug (because whenever time is advanced, debug messages are printed). The volume of debug messages may be the cause of the timeout, but I'm not sure.
> 
> `DistributedCountDecentralizedLateDownstream` appears to be similar.
> 
> It is possible that the best choice will turn out to be not to use these tests in CI. I wasn't the one who added the tests, though, so I am not sure why they were originally considered to be important.

The test does appear to be trying to cause STP violations, although the test should pass even if no violations occur.  I think the test reveals a true problem, however. In master, I get output like this:
```
Federate 1: ERROR: STP violation occurred in a trigger to reaction 2, and there is no handler.
**** Invoking reaction at the wrong tag!
At tag (0, 1), message has violated the STP offset by (0, 1).
```
The test passes despite the error message. But the error message should not be printed. The reaction that is missing the STP handler is the generated `networkMessage` reaction, which indeed has no STP handler. Indeed it should **not** have an STP handler and should instead let the downstream reactions handle the STP violation.

I think the right fix has two parts:

1. The code that issues the above error message (reactor_threaded.c, line 844) should ignore the error if the reaction is a generated `networkMessage` reaction.  The tricky part is how to recognize that it is a such reaction. Perhaps we can use the reaction naming feature for this?
2. The test can safely reduce the debug output size by increasing the period of the timer to, say, 100ms.  On my machine at least, that still reliably results in STP violations. But the test should pass even if there are no STP violations, which I think is OK. (I've gone ahead and pushed this change to master)

For `DistributedCountDecentralizedLateHierarchy.lf`, the above error message should be printed once each second because there is a reaction in Print that has no STP handler.  Currently, the message is printed twice because it is first printed for the `networkMessage` reaction and then for the `Print` reactor.  The same fix 1 above would address this so that the error message is printed only once per second.

petervdonovan:
I thought we were almost done but Python's ChainWithDelay just failed with a flaky segfault. Working on it; it will probably be fixed soon.

core/src/main/java/org/lflang/generator/ReactionInstanceGraph.java
```diff
@@ -251,11 +286,57 @@ protected void addNodesAndEdges(ReactorInstance reactor) {
...
    * levels.
    */
-  private List<Integer> numReactionsPerLevel = new ArrayList<>(List.of(0));
+  private final List<Integer> numReactionsPerLevel = new ArrayList<>(List.of(0));
 
   ///////////////////////////////////////////////////////////
   //// Private methods
 
+  /** A port and an index of a reaction relative to the port. */
+  public record MriPortPair(MixedRadixInt index, PortInstance port) {}
```
lhstrh, line 295: What does `Mri` mean?

lhstrh: Nvm, mixed radix integer...

petervdonovan:
The Python test failure in Python's `PingPongDistributed` is the same as the one observed in `ChainWithDelay` where there is a segfault on line 427 of `gcmodule.c` in CPython due to the current `PyGC_Head` being `null` -- that is, the problem is that what should be a circular doubly linked list is actually not circular, because one of the nodes has a null pointer instead of pointing to either the next node or the sentinel node.

Furthermore, I have reproduced this segfault in `master`. To reproduce, it is sufficient to increase the number of iterations by increasing the `count` variable by a couple orders of magnitude; this increases the probability of running into the error. **So this bug shouldn't stop us from merging into `master`.**

The fact that only containers (Python lists, dictionaries, etc.) are included in the linked list of possibly GC-able objects narrows down the scope of what we have to search for. Furthermore, because we have multiple tests (`ChainWithDelay` and `PingPongDistributed`) that fail in the same way consistently, this makes it seem like this is related to what is happening in this GC list, rather than just being general undefined behavior from an unrelated part of the program. The only problem is that the segfault cannot be reproduced using `python3-dbg`, which makes it hard to view the state of the GC list in a debugger.

lhstrh:
> Furthermore, I have reproduced this segfault in `master`. To reproduce, it is sufficient to increase the number of iterations by increasing the `count` variable by a couple orders of magnitude; this increases the probability of running into the error. **So this bug shouldn't stop us from merging into `master`.**

Thank you, @petervdonovan for this extremely helpful report. Let's mark this test as `failing` and create a separate issue to address this problem.

petervdonovan:
We could alternatively just disable garbage collection for the time being. [This page may be a little out of date](http://www.arctrix.com/nas/python/gc/), but I think that its comment that only pathological programs need garbage collection is still valid.

lhstrh:
> We could alternatively just disable garbage collection for the time being. [This page may be a little out of date](http://www.arctrix.com/nas/python/gc/), but I think that its comment that only pathological programs need garbage collection is still valid.

Would we disable it across the board, or just for these particular tests?

petervdonovan:
Maybe just for federated programs? The problem with disabling it only for these particular tests is that that will not stop users from running into the bug in practice. If we disable garbage collection in all federated programs (I have checked that it does not happen in the unfederated version of `PingPongDistributed`, then that gives us some assurance that the bug will not appear for users.

lhstrh:
> Maybe just for federated programs? The problem with disabling it only for these particular tests is that that will not stop users from running into the bug in practice. If we disable garbage collection in all federated programs (I have checked that it does not happen in the unfederated version of `PingPongDistributed`), then that gives us some assurance that the bug will not appear for users.

That sounds like a reasonable solution to me, if well documented and explained in the issue.

petervdonovan:
Status update on **C federated tests:** In addition to `LoopDistributedCentralizedPhysicalAction` and `LoopDistributedDouble` on macOS, there was a flaky failure of `PingPongDistributedPhysical` which seems Bad and which can be reproduced (inconsistently) in the latest version of this branch, e.g. by running the test a hundred times:
> Federate 0: FATAL ERROR: Mismatched use of _lf_increment_tag_barrier() and  _lf_decrement_tag_barrier_locked().

This means the tag barrier went negative. **I was able to reproduce this in `master`** by using the fixes in the federated launch script (so that errors are treated as errors) and running the test many times.

petervdonovan:
**Update regarding the C federated tests:** In addition to the failures in `LoopDistributedCentralizedPhysicalAction` and `LoopDistributedDouble` on macOS, there was a flaky test failure in `PingPongDistributedPhysical`:

> Federate 0: FATAL ERROR: Mismatched use of _lf_increment_tag_barrier() and  _lf_decrement_tag_barrier_locked().

This can be reproduced in `master` by enabling the changes in #1835 and running `PingPongDistributedPhysical` perhaps a hundred times until it fails.

petervdonovan:
`DecentralizedP2PUnbalancedTimeout` and `DecentralizedP2PUnbalancedTimeoutPhysical` are designed to cause one federate to fall far behind logical time while the other keeps up with logical time. The federate that keeps up with logical time terminates and closes its socket connection, and the federate that falls behind logical time fails with nonzero exit code because it tries to write to its socket after the connection is closed. This is causing errors in this branch.

Do you have an opinion about the desired behavior, @edwardalee? These tests are designed to have nondeterministic behavior with dropped messages, so it is not obvious what kind of behavior we want.
- Should we declare the behavior in such circumstances to be completely **undefined**, such that it does not need to be tested?
- Or should the behavior should be defined, such that the program is supposed to terminate with **nonzero** exit code?
- Or should the behavior is defined, such that the program is supposed to terminate with **zero** exit code?

Obviously it is simplest to declare the behavior undefined. It would be complex to coordinate shutdown in a well-defined way under these adverse circumstances after there have already been many STP violations.

petervdonovan:
Even when SSHing into the macOS GH actions runner I am having a really hard time reproducing the test failures that we are observing in C, in `LoopDistributedDouble` and `LoopDistributedCentralizedPhysicalAction` (I have successfully run both dozens of times, with debug prints enabled like in CI). But the failure in `LoopDistributedDouble` occurred three times in a row in CI, and the failure in `LoopDistributedCentralizedPhysicalAction` occurred twice in a row. I'm not sure what to do about this; has @arengarajan99 found anything out?

petervdonovan:
**Update:** in the `find-races` branch, I have found that several (all?) of the C federated tests that involve physical actions can fairly easily be made to fail in `master` by sprinkling the runtime with random delays averaging perhaps 1-10 milliseconds in length. It is possible that these "failures" do not actually indicate bugs in `master` because programs that use physical actions are inherently dependent on execution times in order to process the physical actions in any particular order relative to anything else.

However, if this is the case, then it is also possible that the physical actions that we are currently observing in C tests in this branch also do not indicate bugs. If these tests are inherently built with races, then we can just get rid of them or relax their timing so that they flake less often.

petervdonovan:
@lhstrh I anticipate that the tests will pass now (they pass locally and have been passing in recent commits). I know that this has been marked "ready for review" for some time just to trigger CI, but now I have cleaned it up and believe it is truly ready for review.

test/TypeScript/src/federated/failing/SpuriousDependency.lf
lhstrh, line 1: I think this needs to be moved out of `failing`, no?

lhstrh (approved):
I am only able to spot check this PR because it is so large, but my impression is that this yielded significant improvement in overall code quality with the many fixes and refactoring steps carried out in this effort. Thank you a ton, @petervdonovan and @byeong-gil, for bringing this to the finish line. Considering all the troubles you encountered on the way, I'm thoroughly impressed that you managed to get all the tests to pass! Congrats on reaching this milestone, everyone!

test/TypeScript/src/federated/failing/SpuriousDependency.lf
lhstrh: Sorry, didn't notice this was `TypeScript`, not `C`.

PR #1695, opened 2023-04-13
Transient federates

ChadliaJerad:
This is PR enables the support of transient federates, introduced in discussion [discussion](https://github.com/lf-lang/lingua-franca/discussions/1504).
This is still work in progress. The PR is meant for checking backward compatibility using the CI.

This branch works with `transient-federates` branch of `reactor-c`. Currently:
 - `@transient` annotation is added.
 - a federate has an attribute saying if it is transient (`is_transient`), which is sent to the RTI within the `FED_ID`.
 - the generated launcher automatically sets `-n` (number of persistent federates) and `-nt` (number of transient federates) when calling `RTI`.
 - working examples will be under `test/C/src/federated/transient`.
 - the `Timers.lf` example, when executed from different prompts, showcases the ability of a(non-connected) federate to join and leave the federation.
 - the tracing mechanism produces a mess! This is due to the different `start_time`(s).  

The semantics will be detailed in [discussion](https://github.com/lf-lang/lingua-franca/discussions/1504).


ChadliaJerad:
Closed as replaced with https://github.com/lf-lang/lingua-franca/pull/2213.

PR #1696, opened 2023-04-13
Avoid directly using global variables in tests

edwardalee:
This PR replaces the use of global variables `start_time` and `current_tag` with function calls that are meant to be part of the API.  This is part of a broader cleanup of the C target that is happening in `docs` branch of reactor-c, but these changes have no dependence on that branch and should be merged before that branch is merge.


edwardalee:
Note that I have already pushed a similar cleanup to the examples, benchmarks, and website repos.

petervdonovan (approved):

Issue #1666, opened 2023-03-19
Authenticated federation is not working


lhstrh:
On `master`, the `-a` flag is being ignored. In my feature branch (see #1663), this is fixed, but now it turns out that the implementation is broken. Here is a diff between the launch script generated by `lfc` on `fed-target-conf` vs. `master`:

```
# The RTI is started first to allow proper boot-up              # The RTI is started first to allow proper boot-up
# before federates will try to connect.                         # before federates will try to connect.
# The RTI will be brought back to foreground                    # The RTI will be brought back to foreground
# to be responsive to user inputs after all federates           # to be responsive to user inputs after all federates
# are launched.                                                 # are launched.
RTI -i ${FEDERATION_ID} \                                       RTI -i ${FEDERATION_ID} \
                        -a \                               <
                        -n 2 \                                                          -n 2 \
                        -c init \                                                       -c init \
exchanges-per-interval 10 \                                     exchanges-per-interval 10 \
&                                                               &
# Store the PID of the RTI                                      # Store the PID of the RTI
RTI=$!                                                          RTI=$!
```
I'm going to mark the tests that use authentication as "failing" so that we can move on, but this issue needs attention.

hokeun:
I think this was added by Dongha. @Jakio815, can you take a look at this when you get a chance?

Jakio815:
Oops, sorry, I saw this message last time and forgot it. I'll look into this!

lhstrh:
@Jakio815: any insights regarding this?

Jakio815:
I'll figure this out this weekend. Sorry for being late.

Jakio815:
I fixed the problem, but the mac CI tests are not passing yet. However it works on mac machines, so it looks like the runner environment has some problems yet. I'll fix it as soon as possible.

The error is as below.

```
Failed: src/federated/SimpleFederatedAuthenticated.lf
    -----------------------------------------------------------------------------
    Reason: Test timed out.
    Compilation output:
    ##### Generating code for federate fed1 in directory /Users/runner/work/lingua-franca/lingua-franca/test/C/fed-gen/SimpleFederatedAuthenticated/src
    ##### Generating code for federate fed2 in directory /Users/runner/work/lingua-franca/lingua-franca/test/C/fed-gen/SimpleFederatedAuthenticated/src
    ******** Using 2 threads to compile the program.
[...]

PR #1698, opened 2023-04-14
Auhenticated federation fixed

Jakio815:
This resolves lf-lang/lingua-franca#1666.

Please merge together with lf-lang/reactor-c#196.


Jakio815:
@petervdonovan Hi Peter. I really don't understand what's the problem and asking you for some insights.

The problem is the CI test runner can't find openssl headers, when making federate binaries.

The problem is this part.

```
/Users/runner/work/lingua-franca/lingua-franca/test/C/fed-gen/SimpleFederatedAuthenticated/src-gen/fed1/core/federated/federate.c:65:10: fatal error: 'openssl/rand.h' file not found
```

and I see it's happening when the `make` command occurs. But I don't understand why it's not working, because the CMake's OpenSSL linking has worked above. Also, the RTI build also works which also needs openssl headers.

I really don't know why it's not working. I have a bunch of tries on my commits, so you may check it out if the CI test takes too long.

I hope for your help!

petervdonovan:
It looks like you're doing basically the right things here, and I do not know off the top of my head what the issue is likely to be. However, here are some ideas:
* where you write `if (osName.contains("mac")) {` in CMakeLists.txt, make sure that it isn't returning `darwin`, or something like that, instead of `mac`. You can add debug prints to the CMakeLists.txt if you need to in order to check that the branch is being executed.
* You can also try to SSH into the GitHub Actions runner. This is surprisingly easy and is briefly mentioned [here](https://github.com/lf-lang/lingua-franca/blob/009d9f404a04f27701303aa9b5e41d7b043bbe55/.github/workflows/README.md#debugging-tests).

Jakio815:
Wow. Never knew I could ssh into the runner. I'll try that! Thanks!

Jakio815:
@lhstrh  It's finally ready. Thanks to @petervdonovan telling me the ssh trick!

lhstrh (approved):
Looks good! Thanks for fixing this :+1:

Jakio815:
@lhstrh Please merge this together with lf-lang/reactor-c#196.

PR #1700, opened 2023-04-15
Revised mechanism for handling the `files` target property

### LINKED ISSUE #1685: Library files no longer work in federated execution
The following used to work before the fed-gen branch was merged and no longer works:
```
target C {
    files: "/lib/c/reactor-c/util/sensor_simulator.h",
    timeout: 1s
}
reactor Foo {
    reaction(startup) {=
        lf_print("Hello World");
    =}
}
federated reactor {
    f = new Foo()
}
```
The `sensor_simulator.h` file is not found.

### LINKED ISSUE #1719: `lfc` incorrectly reports that a binary can be found when it aborted due to a compilation error
```
make: *** [Makefile:136: all] Error 2
Compiled binary is in /home/marten/lingua-franca-master/git/lingua-franca/test/C/bin
lfc: error:  failed with error code 2


```

lhstrh:
This PR cleans up the way we handle the `files` target property.

- [x] Add test case to cover the problem described in #1685

Fixes #1685.
Fixes #1719.


edwardalee (approved):
LGTM.  I have verified that (after merging in master) examples/C/rhythm/RhythmDistributed.lf now works.

test/C/src/federated/FederatedFileInclude.lf
```diff
@@ -0,0 +1,13 @@
+/** Success here is just successfully compiling. */
```
lhstrh, line 1: We already have this test?

lhstrh: I already put a very similar test under `target` because the `files` target property is only supported by the C target currently.

PR #1707, opened 2023-04-16
Spurious dependency example

petervdonovan:
This PR adds a failing test that exposes a bug in the current implementation of federated execution. See  #1708 for additional information.


edwardalee (commented):
This looks like it could become another good test for federated execution, but first it needs to become a test. It should have a timeout and it should check that values are properly conveyed through the Passthrough instances.

Also, I do not understand the comment.  Ports don't execute.  And what is ambiguous? I don't see any ambiguity in the program.

edwardalee (commented):
Ah, OK, this looks like a legitimate bug, so this should be an issue rather than a PR, right?

lhstrh:
It was my suggestion to create a draft PR to 1) confirm the issue and 2) don't loose track of the example. We can build our fix and then merge in this example.

lhstrh:
Shall we put the test in `federated/failed` and merge this + create an issue?

lhstrh (approved):

Issue #1636, opened 2023-03-10
Change Lfc `threading` property to `single-thread` or `non-threaded`


patilatharva:
Currently, the Lfc CLI has a `--threading` option which accepts a "true"/"false" value.
This is inconsistent with all of our other boolean options (e.g. `--quiet`, `--clean`), which don't require explicit arguments because they're `false` by default and become `true` when supplied.

The `--threading` option seems to be designed this way because unlike the other boolean options, we want it to be `true` by default:

https://github.com/lf-lang/lingua-franca/blob/0e00a815dc1d75e91471b015cc095eaaa5d61f31/org.lflang/src/org/lflang/generator/c/CGenerator.java#L406-L409

I think it might be better if we changed it to something like `--single-thread` so it can be `false` by default, and users won't have to explicitly supply an argument for an otherwise boolean option. What are your thoughts?

`--threading=false` → `--single-thread` or `--non-threaded`

edwardalee:
This is a good idea.  Maybe '--no-threads'.

lhstrh:
I also like `--no-threads`.

lhstrh:
Alternatively, we could have `--single-threaded`. It is longer, but more technically correct...

PR #1710, opened 2023-04-18
`lfc` CLI option `--threading` changed to `--single-threaded`

patilatharva:
As mentioned in #1636.


lhstrh (changes requested):
This looks good to me, but there are some things missing:
 - the corresponding target property should probably be renamed as well
 - we need to update the docs (e.g., https://www.lf-lang.org/docs/handbook/target-declaration/)

Issue #1431, opened 2022-10-28
Remove the build-lfc script and the buildLfc gradle task


cmnrd:
https://github.com/lf-lang/lingua-franca/pull/1422 renames the primary script used for building the CLI tools to `build-lf-cli` and the primary gradle task to `buildAll`. The old `build-lfc` script and `buildLfc` task are still available, but are marked as deprecated. They should be removed latest with the 0.5.0 release.

Note that also https://github.com/lf-lang/website-lingua-franca/issues/106 should be fixed before removing the script/task.

PR #1714, opened 2023-04-21
Removal of deprecated build-lfc scipt and buidLfc gradle task

cmnrd:
Fix https://github.com/lf-lang/lingua-franca/issues/1431


lhstrh (approved):
:+1:

PR #1721, opened 2023-04-28
Make USER_THREADS possible to specify in platform options

siljesu:
In reference to this PR: https://github.com/lf-lang/reactor-c/pull/194

USER_THREADS can now be specified under platform options in target properties:

platform: {
       name: Zephyr,
       user_threads: 3
}

### DIFF
--- a/org.lflang/src/org/lflang/TargetProperty.java
+++ b/org.lflang/src/org/lflang/TargetProperty.java
@@ -432,6 +432,9 @@ public enum TargetProperty {
                         case PORT:
                             pair.setValue(ASTUtils.toElement(config.platformOptions.port));
                             break;
+                        case USERTHREADS:
+                            pair.setValue(ASTUtils.toElement(config.platformOptions.userThreads));
+                            break;
                     }
                     kvp.getPairs().add(pair);
                 }
@@ -471,6 +474,9 @@ public enum TargetProperty {
                                 break;
                             case PORT:
                                 config.platformOptions.port = ASTUtils.elementToSingleString(entry.getValue());
+                                break;
+                            case USERTHREADS:
+                                config.platformOptions.userThreads = ASTUtils.toInteger(entry.getValue());
                                 break;
                             default:
                                 break;
@@ -1659,7 +1665,8 @@ public enum TargetProperty {
         BAUDRATE("baud-rate", PrimitiveType.NON_NEGATIVE_INTEGER),
         BOARD("board", PrimitiveType.STRING),
         FLASH("flash", PrimitiveType.BOOLEAN),
-        PORT("port", PrimitiveType.STRING);
+        PORT("port", PrimitiveType.STRING),
+        USERTHREADS("user_threads", PrimitiveType.NON_NEGATIVE_INTEGER);
 
         public final PrimitiveType type;
 
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -1962,6 +1962,14 @@ public class CGenerator extends GeneratorBase {
             System.out.println("To enable compilation for the Arduino platform, you must specify the fully-qualified board name (FQBN) in the target property. For example, platform: {name: arduino, board: arduino:avr:leonardo}. Entering \"no-compile\" mode and generating target code only.");
             targetConfig.noCompile = true;
         }
+
+        if (targetConfig.platformOptions.platform == Platform.ZEPHYR && targetConfig.platformOptions.userThreads >= 0) {
+            targetConfig.compileDefinitions.put(
+                "USER_THREADS",
+                String.valueOf(targetConfig.platformOptions.userThreads)
+            );
+        }
+
         if (targetConfig.threading) {  // FIXME: This logic is duplicated in CMake
             pickScheduler();
             // FIXME: this and pickScheduler should be combined.
--- a/org.lflang/src/org/lflang/TargetConfig.java
+++ b/org.lflang/src/org/lflang/TargetConfig.java
@@ -478,6 +478,11 @@ public class TargetConfig {
          * port values depending on the infrastructure you use to flash the boards.
          */
         public boolean flash = false;
+
+        /** 
+         * The int value is used to determine the number of needed threads for the user application in Zephyr. 
+         */
+        public int userThreads = 0;
     }   
 
     /**


erlingrj:
I think this looks okay. For now it is only Zephyr which needs this so its okay to only expose it (i.e. add the NUM_USER_THREADS compile def) when Zephyr is the target platform.

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -472,6 +475,9 @@ public enum TargetProperty {
                             case PORT:
                                 config.platformOptions.port = ASTUtils.elementToSingleString(entry.getValue());
                                 break;
+                            case USERTHREADS:
```
lhstrh, line 478:
```suggestion
                            case USER_THREADS:
```

org.lflang/src/org/lflang/TargetProperty.java
```diff
@@ -1659,7 +1665,8 @@ public enum PlatformOption implements DictionaryElement {
         BAUDRATE("baud-rate", PrimitiveType.NON_NEGATIVE_INTEGER),
         BOARD("board", PrimitiveType.STRING),
         FLASH("flash", PrimitiveType.BOOLEAN),
-        PORT("port", PrimitiveType.STRING);
+        PORT("port", PrimitiveType.STRING),
+        USERTHREADS("user_threads", PrimitiveType.NON_NEGATIVE_INTEGER);
```
lhstrh, line 1669:
```suggestion
        USER_THREADS("user-threads", PrimitiveType.NON_NEGATIVE_INTEGER);
```

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1962,6 +1962,14 @@ protected void setUpGeneralParameters() {
             System.out.println("To enable compilation for the Arduino platform, you must specify the fully-qualified board name (FQBN) in the target property. For example, platform: {name: arduino, board: arduino:avr:leonardo}. Entering \"no-compile\" mode and generating target code only.");
             targetConfig.noCompile = true;
         }
+
+        if (targetConfig.platformOptions.platform == Platform.ZEPHYR && targetConfig.platformOptions.userThreads >= 0) {
+            targetConfig.compileDefinitions.put(
+                "USER_THREADS",
```
lhstrh, line 1968: We should not use a plain string here as we have an enum to our disposal. In fact, this particular string doesn't match the enum, which is exactly why we shouldn't do this.

erlingrj:
Did you get a chance to look at Martens suggestion, @siljesu ?

siljesu:
Yes, take a look at the newest commits.

test/C/src/zephyr/UserThreads.lf
```diff
@@ -0,0 +1,36 @@
...
+    =}
+    reaction(startup) {=
+        int res;
+        int thread_id;
+
+        for (int i = 0; i < 3; i++) {
+            res = lf_thread_create(&thread_id, &func, NULL);
+            if (res != 0) {
+                printf("FAILURE: Expected to be able to create thread.\n");
+                exit(1);
```
erlingrj, line 24:
```suggestion
lf_print_error_and_exit("Could not create thread");
```

test/C/src/zephyr/UserThreads.lf
```diff
@@ -0,0 +1,36 @@
...
+            }
+        }
+
+        res = lf_thread_create(&thread_id, &func, NULL);
+        if (res == 0) {
+            printf("FAILURE: Expected to not be able to create thread.\n");
+            exit(2);
```
erlingrj, line 31:
```suggestion
lf_print_error_and_exit("Could create more threads than specified.");
```

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -1933,6 +1934,18 @@ protected void setUpGeneralParameters() {
...
+
+        if (targetConfig.platformOptions.platform == Platform.ZEPHYR && targetConfig.platformOptions.userThreads >= 0) {
+            targetConfig.compileDefinitions.put(
+                PlatformOption.USER_THREADS.name(),
+                String.valueOf(targetConfig.platformOptions.userThreads)
+            );
+        }
+
+        if (targetConfig.platformOptions.platform != Platform.ZEPHYR && targetConfig.platformOptions.userThreads >= 0) {
+            System.out.println("Specifying user threads is only for the Zephyr platform. This option will be ignored.");
```
erlingrj, line 1946: Isn`t there an errorReporter object somewhere which can be used to print Warnings.

erlingrj:
Try running `bin/lff test/C` from the root of the project to run the Lingua Franca Formatter on all the tests. This will format your newly added tests

lhstrh (approved):
LGTM!

erlingrj:
The remaining problem is that, in the test system, I explicitly set the threading=false on all tests executed by Zephyr. 

In the file `Configurators.java` you can look at `makeZephyrCompatible`, This is used in `CZephyrTests.java`. I propose to make two separate. One:
`makeZephyrCompatibleUnthreaded` and `makeZephyrCompatible`. Use the former for `runGenericTests` and use the altter for `runZephyrTests`. This way your test will run, but runGenericTests will still be with unthreaded.

siljesu:
OK, take a look at the newest commit. 

In addition, threaded Zephyr will generate an error "Threaded support on Zephyr is not supported" as of now (from lf_zephyr_support.c). This should probably be removed soon, or else threaded Zephyr tests will always fail regardless. I propose replacing it with a warning "Threaded support on Zephyr is still experimental". If you agree, I can make that change in https://github.com/lf-lang/reactor-c/pull/194.

erlingrj:
Yes, change it into a warning about it being experimental. These two PRs should be merged together. I will then get the threaded tests running on Zephyr, at which point we can remove that warning

erlingrj:
OK. I have merged the changes into reactor-c. Now you only have to update your reactor-c pointer to the most recent commit on main and we can actually run the test in CI

erlingrj:
Silje, there is something not working with the the joining of zephyr threads. The UserThreads program does not exit as expected, rather the main thread is stuct at `lf_thread_join` even after the worker has completed. I wonder if this could be a QEMU problem. Could you test the UserThreads.lf program on your NXP and verify that it indeed terminates. It should print `exit` at the end of the execution

siljesu:
Yes, the program exits as expected when running on an NXP.

erlingrj:
@lhstrh Lets merge this, finally...

lhstrh:
Thanks you for your persistence, @erlingrj and @siljesu, and getting this to pass all the checks.

cmnrd:
I noticed that this PR checked in the binary file `test/C/lib/libapp.a` but it does not seem to be referenced anywhere. Was this on purpose? If so, why is it needed? I think we should avoid checking in binary files.

erlingrj:
Ops. That is a mistake!

erlingrj:
I removed it again in #1803

PR #1725, opened 2023-05-02
Bugfixes in handling of `files` target property

lhstrh:
Resources identified by some string and to be looked up relative to a source file, a package, or on the class path, could be files or directories, and there is no way to tell what it is until a matching entry has been found. Therefore, it is most convenient if we provide a lookup mechanism that will work with strings, regardless of whether they point to files or directories. PR https://github.com/lf-lang/lingua-franca/pull/1700 that was merged recently, attempted to furnish this API, but introduces some bugs. This PR fixes those bugs and tries to make the public APIs a bit more friendly improves the documentation.

One issue is that file interactions are different depending on whether tests are executed through JUnit or using `lfc`. We need to add a few unit tests to get full coverage of `FileUtil`, but we should merge this in first because it address the current CI failure on `master`.

### DIFF
--- a/org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
+++ b/org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
@@ -235,7 +235,7 @@ class TSGenerator(
                     "No '" + configFile + "' exists in " + fileConfig.srcPath +
                             ". Using default configuration."
                 )
-                FileUtil.copyFileFromClassPath("$LIB_PATH/$configFile", configFileDest)
+                FileUtil.copyFromClassPath("$LIB_PATH/$configFile", configFileDest, true)
             }
         }
     }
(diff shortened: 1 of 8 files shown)


org.lflang/src/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -223,21 +214,12 @@ class TSGenerator(
      * as the source file, copy a default version from $LIB_PATH/.
      */
     private fun copyConfigFiles() {
-        for (configFile in CONFIG_FILES) {
-            val configFileDest = fileConfig.srcGenPath.resolve(configFile)
-            val configFileInSrc = fileConfig.srcPath.resolve(configFile)
-            if (configFileInSrc.toFile().exists()) {
-                println("Copying $configFileInSrc to $configFileDest")
-                Files.createDirectories(configFileDest.parent)
-                Files.copy(configFileInSrc, configFileDest, StandardCopyOption.REPLACE_EXISTING)
```
petervdonovan, line 232: Why did this feature get removed? It used to copy user-supplied build files.

lhstrh: Oh, you're right. That happened by accident. I didn't realize that it was also doing that. Thanks for the catch

Issue #1652, opened 2023-03-15
Produce an error message if 'extends' is used in a target that does not support it


cmnrd:
AFAIK, the inheritance mechanism is only supported by the C target (and probably the Python target). However, using the `extends` keyword in another target does not produce an error message. We should add a simple validator rule that produces a useful error for all targets that do not explicitly support inheritance.

lhstrh:
Hi @yugborana, thanks for offering help! The proper place to put this check would probably be [Target.java](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/Target.java) (for targets to report whether they support inheritance) and [this method](https://github.com/lf-lang/lingua-franca/blob/4c4b4119b41590876abad97f4a0bb0d9cb1751d9/org.lflang/src/org/lflang/validation/LFValidator.java#LL831C67-L831C67).

yugborana:
ok but I am not well versed with inheritance mechanism.  May be I could help if you could add specifically there? Sorry for disturbance, I am a beginner in this

lhstrh:
The point here is not to work on the inheritance mechanism itself, but to throw produce an error message in the validator when the `extends` keyword is being used while the target does not support inheritance. If you are interested in implementing support for inheritance in the targets that currently do not support it, that would be welcomed as well, but that would be another issue.

yugborana:
ok I think I will not be able to contribute in this issue

PR #1726, opened 2023-05-04
Validator rules to check if target suppports federation or inheritance

cmnrd:
Fixes https://github.com/lf-lang/lingua-franca/issues/1652.

This also updates the checkReactor rule in order to mitigate potential exception and avoid weird control flow (i.e. returning from a nested if).

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
@@ -281,7 +281,7 @@ public class LinguaFrancaValidationTest {
     public void disallowUnderscoreReactorDef() throws Exception {
         String testCase = """
                 target TypeScript;
-                main reactor __Foo {
+                reactor __Foo {
                 }
             """;
         validator.assertError(parseWithoutError(testCase), LfPackage.eINSTANCE.getReactor(), null,
--- a/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
@@ -821,7 +821,7 @@ public class LinguaFrancaValidationTest {
                     reactor B extends A{}
                 """.formatted(target));
 
-            if(target.supportsInheritance()) {
+            if (target.supportsInheritance()) {
                 validator.assertNoIssues(model);
             } else {
                 validator.assertError(model, LfPackage.eINSTANCE.getReactor(), null,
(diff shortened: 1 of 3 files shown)


org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -852,29 +852,22 @@ public void checkReaction(Reaction reaction) {
     // FIXME: improve error message.
     }
 
-    @Check(CheckType.FAST)
-    public void checkReactor(Reactor reactor) throws IOException {
```
edwardalee, line 856: Why remove this check?

edwardalee (changes requested):

org.lflang/src/org/lflang/validation/LFValidator.java
lhstrh: I don't think it is removed. See line 900.

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -812,6 +812,41 @@ public void testPreambleVisibility() throws Exception {
...
+    @Test
+    public void testInheritanceSupport() throws Exception {
+        for (Target target : Target.values()) {
+            Model model = parseWithoutError("""
+                    target %s
+                    reactor A{}
+                    reactor B extends A{}
+                """.formatted(target));
+
+            if(target.supportsInheritance()) {
```
lhstrh, line 824:
```suggestion
            if (target.supportsInheritance()) {
```

org.lflang.tests/src/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -812,6 +812,41 @@ public void testPreambleVisibility() throws Exception {
...
+
+    @Test
+    public void testFederationSupport() throws Exception {
+        for (Target target : Target.values()) {
+            Model model = parseWithoutError("""
+                    target %s
+                    federated reactor {}
+                """.formatted(target));
+
+            if(target.supportsFederated()) {
```
lhstrh, line 841:
```suggestion
            if (target.supportsFederated()) {
```

org.lflang/src/org/lflang/validation/LFValidator.java
```diff
@@ -969,6 +984,11 @@ public void checkReactor(Reactor reactor) throws IOException {
         }
 
         if (reactor.isFederated()) {
+            if(!target.supportsFederated()) {
```
lhstrh, line 987:
```suggestion
            if (!target.supportsFederated()) {
```

lhstrh (approved):
Looks great! Thanks for adding the tests also.

cmnrd:
Looks like my refactoring also revealed some other issues in the tests. This should be fixed now.

PR #1729, opened 2023-05-06
Compare Google AOSP autoformatter to the regular Google one

lhstrh:


petervdonovan:
IIRC the only difference was that AOSP uses 4 spaces instead of 2?

lhstrh:
> IIRC the only difference was that AOSP uses 4 spaces instead of 2?

Looking at the diff, that sure is the most substantial one.

Issue #1712, opened 2023-04-19
preamble is not inherited


edwardalee:
A reactor-level preamble is not inherited. Minimal example:
```
target C
reactor A {
    preamble {=
        #define FOO 2
    =}
    reaction(startup) {=
        printf("FOO: %d\n", FOO);
    =}
}
reactor B extends A {}
main reactor {
    b = new B()
}
```
This gives a compile error:
```
error: use of undeclared identifier 'FOO'
```

Also, file-level preambles are not inherited either. In file `Bug2.lf`, put this:
```
target C
preamble {=
    #define FOO 2
=}
reactor A {
    reaction(startup) {=
        printf("FOO: %d\n", FOO);
    =}
}
```
Then in another file, put this:
```
target C
import A from "Bug2.lf"
reactor B extends A {}
main reactor {
    b = new B()
}
```
The same compile error results.

edwardalee:
Closed via #1732.

PR #1731, opened 2023-05-06
Inherit preambles

edwardalee:
This PR addresses the first problem in #1712, where a preamble of base-class reactor was not inherited by a subclass of the reactor. It adds the test mentioned in the issue.


edwardalee:
Replaced by #1732.

PR #1732, opened 2023-05-06
Preamble file level

edwardalee:
This fixes the problems identified in #1712, where reactor-level preambles were not inherited and file-level preambles for base classes are not included in the generated code.


petervdonovan (approved):
Thanks for fixing this!

lhstrh (approved):
Looks great! Thanks for fixing this!

PR #1737, opened 2023-05-15
Clean `Correspondence` tags out of generated C code

### LINKED ISSUE #1701: Correspondence tag appears in error message
The following minimal program:
```
target C
main reactor {
    reaction(startup){=
        lf_set(out)
    =}
}
```
generates the following error message from `lfc`:
```
/Users/eal/lingua-franca-intellij/test/C/src/Bug.lf:4:160: error: too few arguments provided to function-like macro invocation
    /*Correspondence: Range: [(3, 8), (3, 19)) -> Range: [(0, 0), (0, 11)) (verbatim=true; src=/Users/eal/lingua-franca-intellij/test/C/src/Bug.lf)*/lf_set(out)
                                                                                                                                                               ^
```
It used to be, I think, that the leading "Correspondence" comment was used to construct a more accurate error report, like this:
```
- error: use of undeclared identifier 'lf_set'
    lf_set(out)
      ^
```
In fact, that is what my version of Epoch generates (which is not up to date with master), so this mechanism has been recently broken.

petervdonovan:
Closes #1701.

This is one of several cleanups after breaking changes introduced by #1599.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -28,7 +28,6 @@ import static org.lflang.ASTUtils.allActions;
 import static org.lflang.ASTUtils.allPorts;
 import static org.lflang.ASTUtils.allReactions;
 import static org.lflang.ASTUtils.allStateVars;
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
:+1:

PR #1741, opened 2023-05-17
Prohibit reactor classes in the same file with the same name, up to case differences

### LINKED ISSUE #1702: Name collisions between files generated for federated programs
Another recently introduced error that I think may be a consequence of the bodiless reactions merge:
 
The following minimal program:
```
target C
reactor A {
    reaction(startup){=
        printf("Hello\n");
    =}
}
federated reactor {
    a = new A()
}
```
Generates the following warning when compiled with lfc:
```
/Users/eal/lingua-franca-intellij/test/C/fed-gen/Bug/src-gen/a/a1452928096_main.c:3:10: warning: non-portable path to file '"include/a/A.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
#include "include/a/a.h"
         ^~~~~~~~~~~~~~~
         "include/a/A.h"
1 warning generated.
```
There are apparently now **two** header files generated for each reactor, one for the class `A.h` and one for the instance `a.h`. Why is a header file being generated for the instance?  On MacOS, these are the same file.

There used to be only one header file generated.  Moreover, there are now a slew of other new files that have appeared with names like `a1704353662.c`.  Where are these numbers from? What are these files?  Do we really need all these files?  How did they appear?

petervdonovan:
Fixes #1702.

The core issue in #1702 is that we consider reactor classes to have distinct names using a **case-sensitive** comparison, but we want to generate files corresponding to the reactor classes that will be compared **case-insensitively** on macOS.

We can't just mangle these file names because users might refer to them in their code.

Therefore, we should avoid creating reactor classes whose names are the same, up to case differences.

### DIFF
--- a/org.lflang.tests/src/org/lflang/tests/compiler/TargetConfigTests.java
+++ b/org.lflang.tests/src/org/lflang/tests/compiler/TargetConfigTests.java
@@ -94,7 +94,7 @@ class TargetConfigTests {
         generator.doGenerate(resource, fileAccess, context);
 
         String lfSrc = Files.readAllLines(
-            ((FedFileConfig)context.getFileConfig()).getSrcPath().resolve("a.lf")
+            ((FedFileConfig)context.getFileConfig()).getSrcPath().resolve("__a.lf")
         ).stream().reduce("\n", String::concat);
         Model federate = parser.parse(lfSrc);
         assertHasTargetProperty(federate, "tracing");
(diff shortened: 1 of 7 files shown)


petervdonovan:
The only test failure is the Zephyr test, which is broken right now.

lhstrh (approved):
This looks good to me! Would be nice to have a test, but that's not critical.

petervdonovan:
OK. I will choose to merge this now without testing the validation rule.

PR #1745, opened 2023-05-18
Delete unnecessary complexity from `build-lf-cli`

petervdonovan:
I have not been able to get this script to detect changes in my file system. At first I thought this was a Gradle issue, but it is just a problem in the bash script.

If anyone wishes to make this mechanism work properly, then I will not stop them. However, Gradle is already supposed to detect when nothing needs to be rebuilt, so if we do it this way and something needs to be rebuilt, then we are just doing the same check twice, which is a waste of time. Furthermore, if nothing needed to be rebuilt, then we would not be running this script. Let us not make an incremental build system out of bash scripts.

### DIFF
--- a/lib/scripts/build.sh
+++ b/lib/scripts/build.sh
@@ -114,18 +114,4 @@ if [ "${OSTYPE}" = "msys" ]; then
 fi
 
 jar_path="$(get_jar_path)"
-
-if [ ! -f "${jar_path}" ] || ! "${find_cmd}" "${base}" \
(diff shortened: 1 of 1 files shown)


lhstrh (commented):
I see no problem with this and your argument makes sense. I'd like @cmnrd to have a look at it though.

cmnrd (approved):
I agree that we should let Gradle decide what needs to be rebuilt. Thanks for the cleanup!

PR #1750, opened 2023-05-19
Report the RTI absence error properly in TypeScript

byeong-gil:
The gradle test cannot catch the absence of the RTI because the script exits with 0, not 1. 

Also, there is no code for installing the RTI in the [ts-test.yml](https://github.com/lf-lang/lingua-franca/blob/master/.github/workflows/ts-tests.yml). This [line](https://github.com/lf-lang/lingua-franca/blob/2ef47cd3b82d9c99d2185d30f301bdb4d738070c/.github/workflows/c-tests.yml#L49) in `c-tests.yml` has to be included.

This led to passing every federated test in TypeScript without doing actual tests.


lhstrh:
Closing this because we already have https://github.com/lf-lang/lingua-franca/pull/1752.

PR #1752, opened 2023-05-19
Federated launch script to error when RTI is missing

byeong-gil:
After fixing the launch script so that its exit code reflects that when the RTI cannot be found this amounts to an error, it became clear that quite a few of the federated tests were failing.

Of those failing tests, the following were successfully repaired:

- [x] `ChainWithDelay.lf`
- [x] `DistributedCount.lf`
- [x] `DistributedCountPhysical.lf`
- [x] `DistributedDoublePort.lf`
- [x] `DistributedLoopedAction.lf`
- [x] `DistributedLoopedPhysicalAction.lf`
- [x] `DistributedStop.lf`
- [x] `DistributedStopZero.lf`
- [x] `HelloDistributed.lf`
- [x] `StopAtShutdown.lf`
- [x] `TopLevelArtifacts.lf`

The following tests need more work and have temporarily been marked as `failing`:
- `DistributedCountPhysicalAfterDelay.lf`
- `LoopDistributedCentralized.lf`
- `LoopDistributedDouble.lf`
- `PingPongDistributed.lf`
- `PingPongDistributedPhysical.lf`
- `SimpleFederated.lf`

Relevant PR: https://github.com/lf-lang/reactor-ts/pull/152


byeong-gil:
Known error: The `keepAlive` should be set to `true` in a federation, but it isn't. I think several tests can pass after fixing this.

lhstrh:
I don't think we should move the tests to `failing`. That is meant for tests that we are unwilling to fix. Step 1 is to get the RTI installed on the runner so that the tests don't all fail for that reason.

hokeun:
> Known error: The `keepAlive` should be set to `true` in a federation, but it isn't. I think several tests can pass after fixing this.

I fixed this by preventing the federateConfig.keepAlive from being overwritten by the app parameter in https://github.com/lf-lang/lingua-franca/pull/1752/commits/3b1ad06262bfbccb88a3088d1fa630b28a6af7d4

There are still errors in `PingPongDistributed.lf` and `LoopDistributedDouble.lf`due to the PTAG handling. This might be related to the FIXME here, which temporarily disabled PTAG handling: https://github.com/lf-lang/reactor-ts/blob/master/src/core/federation.ts#L1379

lhstrh:
The check boxes in the PR description are meant to track progress of the PR, so we should only lists the tests that are failing, and check off the ones that we fixed...

byeong-gil:
I suggest merging this PR first by moving failed tests to the failing folder. Because fixing all failed tests pass requires lots of time and makes this PR too heavy. Currently, all Python federated tests are failing. Also, I realized that we have to add PTAG handling to fix the remaining failing tests in `reactor-ts`. And it requires handling of the `input_control_reaction`, the `output _control_reaction`, and `port_absent` messages. 

Because we get some improvements (finding the fault of the script and fixing some TypeScript federated Tests) with this PR, I think it's fine to merge this PR first and create other PRs that solve Python tests and TypeScript tests.

lhstrh:
@edwardalee, it looks like all the Python tests are failing. We didn't know this because the failures were being masked due to an error in the launch script... I think that fixing this might be one of the highest-priority issues, particularly given the various changes to the federated infrastructure that are coming down the pike.

lhstrh (approved):
Assuming that @byeong-gil will document the issues with the TypeScript tests now marked as failing, let's merge this so that we regain testing ability for federated execution in Python and TypeScript.

PR #1755, opened 2023-05-20
Fix ROS serialization

petervdonovan:
The problem was that the relevant #includes were behind an extern "C".


lhstrh (approved):
I think the actual test failures are masked by the issue described in #1752, but if this fixes it locally for you, then let's merge it.

petervdonovan:
Yes, it fixes it locally for me.

petervdonovan:
The tests are failing because of the "prepublish" thing, but that is a separate issue, so I'll just merge.

PR #1756, opened 2023-05-20
Struct refactoring for actions and ports

edwardalee:
This PR refactors the structs associated with ports and actions to make a clearer distinction between fields that are user-visible and those that are only used internally by the runtime. It also adds a parent pointer to the struct for actions. There is a corresponding PR in reactor-c: https://github.com/lf-lang/reactor-c/pull/216.


org.lflang/src/org/lflang/generator/c/CActionGenerator.java
```diff
@@ -104,7 +104,8 @@ public static void generateDeclarations(
             var actionName = action.getName();
             body.pr(action, CGenerator.variableStructType(action, reactor, false)+" _lf_"+actionName+";");
             // Initialize the trigger pointer in the action.
```
erlingrj, line 106:
```suggestion
            // Initialize the trigger pointer and the parent pointer in the action.
```

erlingrj (commented):
Looks good!

PR #1759, opened 2023-05-21
Automatically generated .vscode/settings.json file

edwardalee:
This PR generates a .vscode/settings.json file in the src-gen directory so that when you open VSCode in that directory, it can compile the generated program. (Recommend selecting Clang 11 kit when prompted on opening code) .  The CMake debugger also will work within code.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CGenerator.java
+++ b/org.lflang/src/org/lflang/generator/c/CGenerator.java
@@ -36,6 +36,7 @@ import static org.lflang.util.StringUtil.addDoubleQuotes;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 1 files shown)


erlingrj (commented):
Great, lets merge!

erlingrj (approved):

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -547,6 +548,29 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
...
+            if (!Files.exists(vscodePath))
+            Files.createDirectory(vscodePath);
+            FileUtil.writeToFile(
+                settings,
+                Path.of(fileConfig.getSrcGenPath()
+                    + File.separator + ".vscode"
+                    + File.separator + "settings.json")
+            );
+        } catch (IOException e) {
+            Exceptions.sneakyThrow(e);
```
lhstrh, line 571: The `sneakyThrow` pattern was inherited from generated Xtend code, and AFAIK there is no good reason to use it. If we can use the error reporter, that would be best, but just an ordinary `RuntimeException` would also do, I guess.

lhstrh (approved):
:rocket:

org.lflang/src/org/lflang/generator/c/CGenerator.java
lhstrh: Create #1761.

PR #1760, opened 2023-05-21
Improved line adjustment logic for federated programs

petervdonovan:
This fixes a bug in which, when a reaction body of a federated reactor all fits on one line, it does not get a (correct) `CodeMap.Correspondence` tag. This issue was observed [here](https://github.com/lf-lang/lingua-franca/actions/runs/5030655964/jobs/9023219148?pr=1749).

It is bad that I had to modify ToLf to do this since the logic isn't really in ToLf. The problem was that the line adjustment relies on tagging the target code (without the fat braces), and the target code is not recorded separately from the fat braces if they are entered into the MalleableString machinery as all one string.


lhstrh:
The resulting unit test failures are a bit puzzling...

lhstrh (approved):
Looks good to me!

PR #1762, opened 2023-05-21
Reduced testing

petervdonovan:
This PR changes the CI configuration to run tests selectively according to the files that have been modified.

**Rationale:**
* If we allow tests to be manually triggered, we can always run additional tests if we think they are necessary.
* Target-specific tests are only triggered by target-specific code, but I think this is OK because PRs with any substantial changes will likely touch the target-specific code.

> Could you provide a bit more explanation on how this works?

The "on" part of the workflow files specifies when the workflows are run.

Here is an example:
```yaml
on:
  workflow_dispatch:
  pull_request:
    branches-ignore:
      - 'nightly'
    paths:
    - 'org.lflang/src/org/lflang/generator/cpp/**'
    - 'org.lflang/src/lib/cpp/**'
```

This says that the C++-specific tests are only run when you manually trigger them, submit a PR to a branch other than "nightly," or modify a file in `org.lflang/src/org/lflang/generator/cpp/` or `org.lflang/src/lib/cpp/`.

`all.yml` has the trigger
```yaml
on:
  workflow_dispatch:
  push:
    branches:
      - master
```
so that all tests are run when we push to master.

> And what happens if core components change?

So, if you modify something in `ASTUtils.java` but nothing in `org.lflang/src/org/lflang/generator/cpp/` or `org.lflang/src/lib/cpp/`, and it breaks something in the C++ generator, then you will not find out until you have already merged your breaking change to `master` *unless* you manually trigger the appropriate CI workflow before merging.

So, with this change, we might have to be more proactive about running the tests that we need to run in order to avoid breaking `master`. I think this is problematic, but I also think that it is preferable to constantly running all tests regardless of whether the tests are relevant to the changes. Excessive testing
* wastes energy/computing resources
* uses up all of our macOS runners
* takes a long time to execute (esp. the TS tests, which are usually the last tests to finish)
* is not very helpful, because irrelevant tests either pass or, if they fail, end up getting ignored because they just flaked or failed due to network issues

@lhstrh has also mentioned that our `master` branch and `release` branch are separate, so ~~an occasionally broken master might not be so bad.~~ See discussion with Christian; our goal is still to run all tests before merging into master.

> The main problem that I see is that our repository isn't nicely separated in modules with clear responsibilities. There are so many classes that are only ever used in the C target, but how do we know?

Some parts of the repository do have clear responsibilities. In particular, the subpackages of `org.lflang.generator` are clearly target-specific. But I agree that those are the exception rather than the rule, so this mechanism will not work perfectly. I am working on the assumption that if a change affects a particular target, it probably will touch files that are known to be specific to that target, even if other files that the change touches do not have such clear responsibilities.


petervdonovan:
Rationale:
* If we allow tests to be manually triggered, we can always run additional tests if we think they are necessary.
* Target-specific tests are only triggered by target-specific code, but I think this is OK because PRs with any substantial changes will likely touch the target-specific code.

cmnrd (commented):
Could you provide a bit more explanation on how this works? How do you detect if only something target specific changed? And what happens if core components change?

The main problem that I see is that our repository isn't nicely separated in modules with clear responsibilities. There are so many classes that are only ever used in the C target, but how do we know?

petervdonovan:
> Could you provide a bit more explanation on how this works? How do you detect if only something target specific changed?

The "on" part of the workflow files specifies when the workflows are run.

Here is an example:
```yaml
on:
  workflow_dispatch:
  pull_request:
    branches-ignore:
      - 'nightly'
    paths:
    - 'org.lflang/src/org/lflang/generator/cpp/**'
    - 'org.lflang/src/lib/cpp/**'
```

This says that the C++-specific tests are only run when you manually trigger them, submit a PR to a branch other than "nightly," or modify a file in `org.lflang/src/org/lflang/generator/cpp/` or `org.lflang/src/lib/cpp/`.

`all.yml` has the trigger
```yaml
on:
  workflow_dispatch:
  push:
    branches:
      - master
```
so that all tests are run when we push to master.

> And what happens if core components change?

So, if you modify something in ASTUtils but nothing in `org.lflang/src/org/lflang/generator/cpp/` or `org.lflang/src/lib/cpp/`, and it breaks something in the C++ generator, then you will not find out until you have already merged your breaking change to `master` *unless* you manually trigger the appropriate CI workflow before merging.

So, with this change, we might have to be more proactive about running the tests that we need to run in order to avoid breaking `master`. I think this is problematic, but I also think that it is preferable to constantly running all tests regardless of whether the tests are relevant to the changes. Excessive testing
* wastes energy/computing resources
* uses up all of our macOS runners
* takes a long time to execute (esp. the TS tests, which are usually the last tests to finish)
* is not very helpful, because irrelevant tests either pass or, if they fail, end up getting ignored because they just flaked or failed due to network issues

Marten has also mentioned that our `master` branch and `release` branch are separate, so an occasionally broken master might not be so bad.

> The main problem that I see is that our repository isn't nicely separated in modules with clear responsibilities. There are so many classes that are only ever used in the C target, but how do we know?

Some parts of the repository do have clear responsibilities. In particular, the subpackages of `org.lflang.generator` are clearly target-specific. But I agree that those are the exception rather than the rule, so this mechanism will not work perfectly. I am working on the assumption that if a change affects a particular target, it probably will touch files that are known to be specific to that target, even if other files that the change touches do not have such clear responsibilities.

lhstrh (commented):
I do worry about replication between `all.yml` and the various target-specific workflows, especially considering reusable workflow invocations need to point to a specific ref, which already is a headache to deal with currently, without such replication.

petervdonovan:
The refactoring to move reused parts into separate workflows is done. I have also removed all unnecessary `@master` version markers. I think this is ready to merge.

lhstrh:
This will probably also break the CI that we have in our satellite repos like `reactor-c` and `reactor-ts`, so before merging this, we need to check those.

petervdonovan:
> This will probably also break the CI that we have in our satellite repos like reactor-c and reactor-ts, so before merging this, we need to check those.

I just looked at `reactor-c` and `reactor-ts` and `reactor-cpp` briefly. I don't think they'll be affected. The only existing workflow that I deleted, modified, or renamed were `ci.yml`, `cancel.yml`, and `nightly-build.yml`, and the changes to the latter two of these are trivial.

lhstrh:
What happens when multiple targets have been touched? Will there be duplicate `cancel` and `unit-tests` jobs?

petervdonovan:
I think you're right. That is not ideal.

Maybe we can fix this problem by using [this feature](https://docs.github.com/en/actions/using-jobs/using-concurrency). I just stumbled across it because it is mentioned in the [documentation of the cancel action](https://github.com/styfle/cancel-workflow-action).

lhstrh:
> I think you're right. That is not ideal.
> 
> Maybe we can fix this problem by using [this feature](https://docs.github.com/en/actions/using-jobs/using-concurrency). I just stumbled across it because it is mentioned in the [documentation of the cancel action](https://github.com/styfle/cancel-workflow-action).

Hm, that notice "you probably don't need this action" definitely wasn't there when I found it :smile:

petervdonovan:
Now I'm worried that the overlap is still a problem, because the concurrently running tasks will cancel each other's jobs and cause each other to fail. I probably need to think about this a little more.

cmnrd (changes requested):
> So, if you modify something in ASTUtils.java but nothing in org.lflang/src/org/lflang/generator/cpp/ or org.lflang/src/lib/cpp/, and it breaks something in the C++ generator, then you will not find out until you have already merged your breaking change to master unless you manually trigger the appropriate CI workflow before merging.

To me this is a no-go. The release branch should be stable. On master all tests should pass. We already see how difficult it is to handle situations where tests on master do not pass (because of a bug or because they are flaky).

It is a great initiative to reduce the number of tests run, but this should be done safely. I think it is Ok to reduce the number of tests run for intermediate commits on PRs, but before merging to master there needs to be a strict requirement that **all** tests have passed.

petervdonovan:
That which Christian described can apparently be done. Here is a discussion among other people who had this problem: https://github.com/reviewdog/action-eslint/issues/29

.github/workflows/c-arduino-tests.yml
```diff
@@ -16,12 +16,16 @@ on:
...
+        required: false
+        default: true
+        type: boolean
 
 jobs:
   run:
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest]
+        platform: ${{ (inputs.all-platforms && fromJSON('["ubuntu-latest", "macos-latest"]')) || 'ubuntu-latest' }}
```
lhstrh, line 28: I see you're using this pattern in most of the target-specific tests. What does it do, exactly?

petervdonovan: I copied it from here: https://github.com/actions/runner/issues/409

petervdonovan:
I am marking this ready for review just to trigger the tests. Please ignore.

petervdonovan:
It looks like old (cancelled) runs are not cleared from the bottom of the PR unless you push again. This could get a little confusing, but I think we can tolerate it.

I think this PR is once again ready for review/comments.
Note that once it is merged, we do plan to make `CI (non-target-specific)` and `CI (target-specific)` required in order to merge a branch. (These are two separate workflows because it is not possible to trigger more than 20 reusable workflows from one workflow.)

petervdonovan:
Let's make Clement's suggestion (to sometimes run some tests many times, in case they are flaky) a separate issue.

I also still think that because the number of jobs is greatly reduced here for draft PRs (with no macOS jobs until it is ready for review), we could make the jobs smaller to increase the amount of parallelism. But we can also make that a separate issue and address it later based on how the new CI configuration works for us.

cmnrd:
Another idea would be to add some dependencies between tests. For example: only run the target specific tests once the base tests have passed. https://github.com/lf-lang/lingua-franca/pull/1779 might also help with this, as there is now a single CI job that builds, checks formatting, and runs unit tests. All other jobs could be started only conditionally if the build task succeeds.

lhstrh:
Great idea!

petervdonovan:
My personal preference is not to run the formatting tests until the PR is ready for review. Formatting checks are a form of "trivial/vacuous feedback," and the point of automating the formatting checks, AFAICT, is to _reduce_ the time spent giving and receiving such feedback. So, my feeling is that we should run them only when doing so is necessary in order to spare a human the effort.

Otherwise I am all in favor of running the unit tests first. We can even run a subset of the integration tests (for example, maybe the most complicated test in each `CATEGORY`) before running an entire suite.

.github/workflows/only-ts.yml
```diff
@@ -0,0 +1,30 @@
+name: Rust
```
lhstrh, line 1:
```suggestion
name: TypeScript
```

.github/workflows/serialization-tests.yml
```diff
@@ -8,7 +8,7 @@ on:
         type: string
 
 jobs:
-  run:
+  serialization:
```
lhstrh, line 11:
```suggestion
  serialization-tests:
```

lhstrh (commented):
I think this is a huge improvement -- thanks for pursuing this, @petervdonovan! There is an art to naming these workflows so that they are rendered sensibly in this rather poorly design GitHub Actions UI, so I tweaked things a bit to make them more readable. I intentionally named the two "catch-all" workflows identically, so that there is a clear grouping between the workflows that should run in draft mode and the ones that should run when the PR is ready for review.

lhstrh (commented):
Was the submodule pointer for `reactor-c` meant to be changed?

lhstrh:
It also caught my eye that we're still running CodeCov in draft mode. I think we shouldn't.

lhstrh:
Here's an issue: when the PR is in draft mode and is marked ready for review while some of the tests are still running, the outcome of the PR will be that it has failed because of certain jobs having been canceled. I'm not sure if there is a way around this problem...

.github/workflows/non-target-specific.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  # Run the unit tests.
+  junit:
+    if: ${{ inputs.all || github.event.pull_request.draft }}
+    uses: ./.github/workflows/unit-tests.yml
+    with:
+      all-platforms: ${{ !github.event.pull_request.draft  }}
+
+  # Run the serialization tests
+  serialization:
```
cmnrd, line 30: I think the serialization test should be grouped together with the C tests. The serialization syntax is only supported by the C target, and the test itself work very similar to our standard integration tests.

cmnrd (commented):
This is fantastic! I only have two comments:
- The serialization test is in fact C target specific and should be grouped with the C tests.
- Marten mentioned this before: The ref of reactor-c changed. Is this intentional?

cmnrd:
> My personal preference is not to run the formatting tests until the PR is ready for review. Formatting checks are a form of "trivial/vacuous feedback," and the point of automating the formatting checks, AFAICT, is to _reduce_ the time spent giving and receiving such feedback. So, my feeling is that we should run them only when doing so is necessary in order to spare a human the effort.
> 
> Otherwise I am all in favor of running the unit tests first. We can even run a subset of the integration tests (for example, maybe the most complicated test in each `CATEGORY`) before running an entire suite.

With the new gradle setup, building includes running all checks, also the format checks. This is just the workflow gradle uses, and I think it makes sense. If we should really implement dependencies between CI jobs and block other jobs just because build fails due to a formatting issue is of course another question. Personally, I wouldn't mind it, mostly because it would force us to build in gradle and spotless in our workflows and only commit things that are properly formatte

We should probably hold off with this idea for now and discuss it first. We can always add it later.

cmnrd:
> Here's an issue: when the PR is in draft mode and is marked ready for review while some of the tests are still running, the outcome of the PR will be that it has failed because of certain jobs having been canceled. I'm not sure if there is a way around this problem...

Yeah, that could be a bit annoying, but it is probably something we have to deal with. It shouldn't be too bad.

cmnrd (approved):

cmnrd:
I just noticed another thing. In the "Details" view for the C tests, it lists "os-tests" 6 times (twice for each OS). Any Idea why this is? I also find the name a bit strange as we are not testing the OS ;) Maybe we could name it `base-tests` or `standard-tests`?

lhstrh:
> I just noticed another thing. In the "Details" view for the C tests, it lists "os-tests" 6 times (twice for each OS). Any Idea why this is? I also find the name a bit strange as we are not testing the OS ;) Maybe we could name it `base-tests` or `standard-tests`?

The issue is that we cannot name the job based on how the workflow is called. We call the c-tests workflow twice; once with normal compilation and once using the `CCpp` target. I named the job `os-tests` because we're testing on an operating system as opposed to the Zephyr and Arduino tests. I'm fine calling them `default-tests` instead if you think that's better.

.github/workflows/non-target-specific-extended.yml
```diff
@@ -0,0 +1,61 @@
...
+    - 'org.lflang/src/lib/rs/**'
+    - 'org.lflang/src/lib/ts/**'
+
+env:
+  # 2020.11
+  vcpkgGitRef: 0bf3923f9fab4001c00f0f429682a0853b5749e0
+
+concurrency:
+  group: only-non-target-specific-extended-${{ github.ref }}-${{ github.event_path }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```
lhstrh, line 29: I think this means that if the job is run from `master`, it won't be canceled, but this suggests that jobs from subsequent while the PR is not in draft mode, will not get cancelled. This is different from the current behavior. I think they _should_ be canceled in that situation...

This relates to my earlier point about dealing with canceled workflows when switching from "draft" to "ready-for-review" while tests are running. I think that we should pass into the `env` whether the workflow is `catch-all` or `only` and use that in the definition of the `group`. I think that we could then just have `cancel-in-progress` set to `true` and achieve the desired behavior; `only` jobs will not be canceled if a `catch-all` job is spawned (meaning it won't invalidate the end result, as it does now), but subsequent `catch-all` jobs will cancel each other.

lhstrh (changes requested):
See comment about cancellation behavior.

.github/workflows/non-target-specific-extended.yml
petervdonovan: OK. I think that makes sense.

An alternative solution (which might solve other problems as well) would be to just have the `catch-all` workflows be unconditionally triggered, and let them conditionally trigger the more targeted jobs. It would do this by using `git` to determine which files are different from in master rather than by using the special workflow trigger.

lhstrh: > An alternative solution (which might solve other problems as well) would be to just have the `catch-all` workflows be unconditionally triggered, and let them conditionally trigger the more targeted jobs. It would do this by using `git` to determine which files are different from in master rather than by using the special workflow trigger.

I'm fine with either approach, but, indeed, I think the approach you outline here would have advantages; we won't have any "skipped" workflows anymore.

lhstrh: We could even do fancier things when we don't rely on the built-in GH Actions feature; we could e.g., skip doing CI altogether when only a `.md` file is edited.

petervdonovan: Great! I think that this more elaborate approach will work.

petervdonovan:
> The serialization test is in fact C target specific and should be grouped with the C tests.

Thanks for the catch. It looks like C++ and Rust do not have serialization tests.

Based on which targets have a `serialization` directory in their area of the `test` directory, I am under the impression that the serialization tests are specific to those targets which support federated execution. So I think I can just have the serialization tests be run only when any of C, Python, and TypeScript code is modified.

petervdonovan:
This PR has changed somewhat since it was last reviewed:
- `check-diff.yml` uses a bash script, `check-diff.sh`, to check if the current commit differs from `origin/master` in a given group of files. It exports the results using [outputs](https://docs.github.com/en/enterprise-server@3.4/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow).
  - the current commit is the one checked out by the default behavior of `actions/checkout` (this is the merge commit between the current feature branch and the target branch)
  - there is a "group of files" for each of the 5 targets and for tracing.
- All workflows run in response to a PR come from the workflows which were named "catch-all." The "only" workflows no longer are triggered without being called by these parent workflows.
- The "catch-all" workflows run the target-specific workflows conditionally according to the result of the `check-diff` workflow. It is still true that all workflows run regardless of the output of the `check-diff` workflow if a PR has been marked ready for review.

This change has the following advantages:
- There are only 2 top-level CI workflows, so if you go to the "actions" tab it will not be necessary to sift through several workflows to find the one which has the tests that you are interested in.
- The behavior is relatively uniform regardless of whether the PR is marked ready for review in the sense that the same top-level workflows are triggered. The difference in behavior (in terms of which workflows are enabled) is easy to see by looking at the top-level workflows in GitHub's UI.
- When a PR is marked ready for review, the new runs will overwrite the older runs at the bottom of the PR because they are instances of the same top-level workflows. This addresses the cancellation issue that Marten mentioned.
- Because the conditional triggering of workflows is handled using a bash script, it might be more flexible than using only the trigger conditions that are explicitly supported by GitHub's API.

petervdonovan:
I think that with the most recent changes, it no longer makes sense to mark the largest top-level workflows as "catch-all" to emphasize their similarity because they are now the only top-level workflows. It is also useful to distinguish them, since the name of the workflow is the only identification shown in the "Actions" tab. So, I will rename them again.

lhstrh (approved):

.github/scripts/check-diff.sh
```diff
@@ -0,0 +1,10 @@
+changes() {
+  git diff --name-only --diff-filter=AMDR --cached origin/master
+}
+
+if changes | grep -q $1; then
```
lhstrh, line 5: Wonderfully simple! :astonished:

.github/workflows/non-target-specific-extended.yml
lhstrh: It works beautifully!

.github/workflows/non-target-specific.yml
lhstrh: Hm, there are serialization tests for other targets as well. I think this is the right place to have it.

lhstrh (approved):
:rocket: :rocket: :rocket:

petervdonovan:
@cmnrd @lhstrh I have added all checks in the `CI (by target)` and `CI (misc)` workflows to the branch protection rule for `master`, so it should once again be safe to use the auto-merge feature.

cmnrd:
Hmm, it looks like only the C target tests are listed in the branch protection rules. Is that on purpose?

petervdonovan:
No, definitely not. Sorry, I have had some trouble with GitHub's UI. Let me fix that.

cmnrd:
I was hoping that it would be possible to list entire groups of tests instead of individual ones... I guess we have to be careful when adding or removing CI jobs and update the list accordingly.

petervdonovan:
Yes, that is a problem. Fortunately at the bottom of a PR the required checks are labeled as "Required," so if the list is out of date we will probably notice quickly.

PR #1763, opened 2023-05-21
Submodule renamed to match repo name

lhstrh:
This PR is because the `reactor-c-py` submodule was renamed to reflect that it doesn't host a runtime implementation but only provides a support package.


petervdonovan (approved):
I'm not sure why the watchdog test is failing, but it looks unrelated.

lhstrh:
Jep, unrelated. Both C++ enclaves and C watchdog tests are flaky because of timing issues in GH Actions...

PR #1764, opened 2023-05-22
Flaky tests

edwardalee:
This PR moves SimpleFederatedAuthenticated.lf into failing because it deadlocks when running tests locally.

This PR also increases the times in Watchdog.lf by 10x to try to prevent flaky test failures.


lhstrh (approved):
LGTM!

PR #1769, opened 2023-05-23
Check spotless formatting in CI

cmnrd:
Since we now formatted all the code with spotless, we should make sure that it stays this way. This PR re-enables the CI check that ensures that everything is correctly formatted.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -1,4 +1,4 @@
-name: Test the automatic code formatting task
+name: Check if all files are formatted correctly with spotless
 
 on:
   workflow_call:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,11 +30,9 @@ jobs:
     uses: lf-lang/lingua-franca/.github/workflows/build-trace-tools.yml@master
     needs: cancel
 
-# Check that automatic code formatting works.
-# TODO: Uncomment after fed-gen is merged.
-# format:
-#   uses: lf-lang/lingua-franca/.github/workflows/format.yml@master
-#   needs: cancel
+  check-format:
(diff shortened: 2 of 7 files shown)


petervdonovan (approved):
LGTM. As usual we should change `check-format.yml@ci-check-formatting` to `check-format.yml@master` before merging.

lhstrh:
I recommend doing it _after_ merging. Just don't delete the branch or else CI will fail immediately.

lhstrh (approved):
LTGM. @petervdonovan does this mean that we can remove the formatting check from the unit tests? It would be nice if that were the case...

petervdonovan:
I deleted the part of `RoundTripTests.java` that causes many headaches.

I did not create `FormattingUnitTests` -- I think Clement created it to try to isolate individual problems, which makes them more like unit tests, whereas the formatting tests are larger, more expensive, and harder to use for debugging. Probably we should keep `FormattingUnitTests` since they solve a different problem from the formatting check.

.github/workflows/check-format.yml
```diff
@@ -1,4 +1,4 @@
-name: Test the automatic code formatting task
+name: Check if all files are formatted correctly with spotless
```
lhstrh, line 1:
```suggestion
name: Spotless check
```

lhstrh: Note that this is rendered in the left-hand menu, and if it is too long it just gets cut off, so it's better to keep this short.

lhstrh (approved):

.github/workflows/check-format.yml
lhstrh: :rofl: @petervdonovan you just merged this. Could you apply this change in `master`? Thanks!

PR #1771, opened 2023-05-23
Replaced all build option context getargs string with getkey (again)

axmmisaka:
This is supposed to be done in 81499a13 (this replaces https://github.com/lf-lang/lingua-franca/pull/1520#event-9316630760) yet it took a while and master was changed drastically. 

Fortunately, most of the conflicts arise from d872d4f where most of the magic strings were replaced, and I figured it would be a better idea to make a new branch and change accordingly.


lhstrh:
Subsumed by #2008.

PR #1773, opened 2023-05-23
Refactoring of the RTI implementation in the C runtime

ChadliaJerad:
Just to get CI running and check if the RTI refactoring breaks something.


cmnrd:
I am not sure if all our workflows use the latest RTI. I think the container tests for instance use only the published version.

lhstrh:
> I am not sure if all our workflows use the latest RTI. I think the container tests for instance use only the published version.

Yup.

ChadliaJerad:
Thanks for the hint,  @cmnrd!
All tests that require the RTI need to build and use the latest one. The published version is useful for the releases.
And maybe have a subset of tests that pull from DockerHub.
What do you think?

ChadliaJerad:
@petervdonovan, @lhstrh, several CIs were skipped in the latest push.
I know that this is a desired behavior you were investigating lately.
But since the commits in `reactor-c` update the RTI, this PR was meant to check if something breaks in the targets that implement federated execution using the RTI.
I think this is the case for `Python` and `TypeScript` targets.
If this does make sense, how to force `py` and `ts` CI to run? Couldn't do that by re-running the jobs.

petervdonovan:
The "right" way to do that (without temporarily editing the `if` condition in `.github/workflows/all-targets.yml`, which actually might be the most efficient thing to do) is to manually trigger the workflows by going, for example, [here](https://github.com/lf-lang/lingua-franca/actions/workflows/only-py.yml) and using the "run workflow" dropdown on the right:
image

But this does not currently work because of a mistake in `only-py.yml`, which I will now fix.

ChadliaJerad:
> The "right" way to do that (without temporarily editing the `if` condition in `.github/workflows/all-targets.yml`, which actually might be the most efficient thing to do) is to manually trigger the workflows by going, for example, [here](https://github.com/lf-lang/lingua-franca/actions/workflows/only-py.yml) and using the "run workflow" dropdown on the right: 

Thank you very much for the hint, @petervdonovan!

> But this does not currently work because of a mistake in `only-py.yml`, which I will now fix.

It doesn't for TypeScript either.

ChadliaJerad:
After merging @petervdonovan's fix, both triggered Python and TypeScript CI workflows succeded ([py](https://github.com/lf-lang/lingua-franca/actions/runs/5203886045), [ts](https://github.com/lf-lang/lingua-franca/actions/runs/5203894238/jobs/9387369614)). 
This `reactor-c` [PR](https://github.com/lf-lang/reactor-c/pull/212) is then ready to be merged (modulo what will be discussed tomorrow on scheduling enclaves, I guess)...

ChadliaJerad:
No longer needed, as the companion [`reactor-c` PR](https://github.com/lf-lang/reactor-c/pull/224) was merged.

PR #1774, opened 2023-05-24
Optimized Gradle configuration for faster building

axmmisaka:
@petervdonovan mentioned that build time is slow with gradle and most of the time spent is related not to building/execution but initialisation and configuration. While I didn't find a profiler that tells me what exactly causes the long configuration time, it does appear that some time-consuming configurations were carried out regardless of whether or not they are needed (i.e. same time is needed for `buildAll` and `help`). 

This commit applied some of the optimations suggested by IntelliJ that prevents unnecessary configurations from running, and drastically improved the configuration time for `buildAll` locally.

I do not yet know if they will pass CI/CD though. Once CI/CD is passed, I will mark this ready for review.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,7 +57,7 @@ subprojects {
     sourceCompatibility = JavaVersion.VERSION_17
     targetCompatibility = JavaVersion.VERSION_17
 
-    configurations.all {
+    configurations.configureEach {
         exclude group: 'asm'
     }
 
@@ -115,4 +115,4 @@ spotless {
         formatAnnotations()
     }
 }
-tasks.withType(SpotlessTask) { it.dependsOn(":org.lflang:jarCliTools") }
+tasks.withType(SpotlessTask).configureEach { it.dependsOn(":org.lflang:jarCliTools") }
--- a/org.lflang/build.gradle
+++ b/org.lflang/build.gradle
@@ -42,7 +42,7 @@ dependencies {
     mwe2 "org.eclipse.xtext:org.eclipse.xtext.xtext.generator:${xtextVersion}"
 }
 
-task generateXtextLanguage(type: JavaExec) {
+tasks.register('generateXtextLanguage', JavaExec) {
     main = 'org.eclipse.emf.mwe2.launch.runtime.Mwe2Launcher'
     classpath = configurations.mwe2
     inputs.file "src/org/lflang/GenerateLinguaFranca.mwe2"
@@ -90,10 +90,10 @@ configurations {
     }
 }
 
-task getSubmoduleVersions(type: Exec) {
+tasks.register('getSubmoduleVersions', Exec) {
     description('Run a Git command to get the current status of submodules')
     workingDir project.rootDir
-   // This will make gradle execute git submodule status every time updateRustRuntime is called
+    // This will make gradle execute git submodule status every time updateRustRuntime is called
     outputs.upToDateWhen { false }
 
     def command = "git submodule status"
@@ -119,7 +119,7 @@ task getSubmoduleVersions(type: Exec) {
     }
 }
 
-task updateRustRuntime {
+tasks.register('updateRustRuntime') {
     description('Record the VCS revisions of the language runtimes into a properties file available at runtime.')
 
     dependsOn getSubmoduleVersions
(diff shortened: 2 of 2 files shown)


lhstrh:
Interesting! One other thing that about our Gradle configuration is that the default `build` does not work. Is there a way to get it to work and be an alias for `buildAll`?

axmmisaka:
> Interesting! One other thing that about our Gradle configuration is that the default `build` does not work. Is there a way to get it to work and be an alias for `buildAll`?

For now I have not discovered a way to override the default `build` and make it solely run `buildAll`; nevertheless......

I think the default `build` might "work" (I didn't check) as `gradle`'s `build` is defined as such:
```
assemble - Assembles the outputs of this project.
build - Assembles and tests this project.
```
Which matches the behavior as spotless is being configured and takes forever.......
The one that matches `buildAll` the most should instead be `./gradlew assemble` I suppose, and it does produce lf jar rather quickly.

Windows C++ tests seem to be failing, let me see what is going on......

lhstrh:
The C++ tests are flaky due to a known bug. @cmnrd is on top of this, so it's safe to ignore these test failures -- they have nothing to do with your changes.

lhstrh (approved):
LGTM! :rocket:

lhstrh:
> > Interesting! One other thing that about our Gradle configuration is that the default `build` does not work. Is there a way to get it to work and be an alias for `buildAll`?
> 
> For now I have not discovered a way to override the default `build` and make it solely run `buildAll`; nevertheless......
> 
> I think the default `build` might "work" (I didn't check) as `gradle`'s `build` is defined as such:
> 
> ```
> assemble - Assembles the outputs of this project.
> build - Assembles and tests this project.
> ```
> 
> Which matches the behavior as spotless is being configured and takes forever....... The one that matches `buildAll` the most should instead be `./gradlew assemble` I suppose, and it does produce lf jar rather quickly.

I guess my concern is that we should have a configuration that when used by someone who is familiar with Gradle and did not read our docs will be able to build the project without problems. If that means we should rename `buildAll` to `assemble`, then I would be fine with that, but when I current call `./gradlew assemble` it already does a bunch of stuff, so I'm guessing it runs the `buildAll` task?

axmmisaka:
I don't think so - I'm not gradle expert but from what this [article](https://gradlehero.com/gradle-assemble-task-essentials/#tasks-that-assemble-depends-on) told me it appears the following is run:

```
> Task :org.lflang:getSubmoduleVersions
> Task :org.lflang:updateRustRuntime
> Task :org.lflang:processResources
> Task :org.lflang.tests:processResources NO-SOURCE
> Task :org.lflang:compileKotlin
> Task :org.lflang:compileJava
> Task :org.lflang:classes
> Task :org.lflang:inspectClassesForKotlinIC
> Task :org.lflang:jar
> Task :org.lflang:startScripts
> Task :org.lflang:distTar
> Task :org.lflang:distZip
> Task :org.lflang:shadowJar
> Task :org.lflang:startShadowScripts
> Task :org.lflang:shadowDistTar
> Task :org.lflang:shadowDistZip
> Task :org.lflang:assemble
> Task :org.lflang.tests:compileKotlin NO-SOURCE
> Task :org.lflang.tests:compileJava NO-SOURCE
> Task :org.lflang.tests:classes UP-TO-DATE
> Task :org.lflang.tests:inspectClassesForKotlinIC
> Task :org.lflang.tests:jar
> Task :org.lflang.tests:assemble
```

In contrast, `buildAll` runs the following:
```
> Task :org.lflang:getSubmoduleVersions
> Task :org.lflang:updateRustRuntime UP-TO-DATE
> Task :org.lflang:processResources
> Task :org.lflang:compileKotlin
> Task :org.lflang:compileJava
> Task :org.lflang:classes
> Task :org.lflang:jarCliTools
```

I think `assemble` is somewhat auto-generated so we can't replace it; we could make it dependent on `buildAll` but my experiment shows that it doesn't work. (there'd be dependency issues)

Honestly, I believe I would need some time to digest how our gradle settings were designed and see if they fit what gradle wants us to do......

cmnrd (approved):
Great! Thanks for these fixes!

> My concern is that we should have a configuration that when used by someone who is familiar with Gradle and did not read our docs will be able to build the project without problems. If that means we should rename buildAll to assemble, then I would be fine with that, but when I current call ./gradlew assemble it already does a bunch of stuff, so I'm guessing it runs the buildAll task?

For this we need a more significant refactoring of the gradle setup. I started a little side project where I experiment with the setup. I will open a draft PR as soon as I have smth substantial.

PR #1775, opened 2023-05-24
Fixed validator check for C generics

### LINKED ISSUE #1770: [C-Generics] [Validator] Typename missmatch
With the recent changes for C-Generics branch which is now in master. There's a big issue in Validator

The Validator is simply text matching the Generic TypeName and isn't allowing us to compile the code at all

So if i have something like this

```
reactor A<T1> {
  input in:T1;
}

reactor B<T2> {
  output out:T2;

  timer t(0, 1sec);
  reaction(t) -> out {=
    lf_set(out, rand());
  =}
}


main reactor {
  a = new A<int>();
  b = new B<int>();

  b.out -> a.in;
}
```
Screenshot 2023-05-23 at 8 34 08 PM

petervdonovan:
Fixes  #1770.

### DIFF
--- a/org.lflang/src/org/lflang/validation/LFValidator.java
+++ b/org.lflang/src/org/lflang/validation/LFValidator.java
@@ -216,6 +216,7 @@ public class LFValidator extends BaseLFValidator {
     }
   }
 
+  /** Resolve the port type if it is known. */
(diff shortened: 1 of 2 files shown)


lhstrh (approved):
This looks good. Only nitpick is that `portTypeIfResolvable` needs a comment.

PR #1776, opened 2023-05-24
`SimpleFederatedAuthenticated.lf` test passing

Jakio815:
@edwardalee reported this test is failing. #1777 I opened this issue because it keeps breaking due to the same problem.

It is a simple OpenSSL problem happening on Mac. I tested Ubuntu environments and it works well.

I'll report when it is fixed.


lhstrh:
@Jakio815: this is still marked as draft. Is it ready for review?

Jakio815:
Yes. It works fine currently. However, as mentioned in #1777 the protocol needs to be fixed, as both RTI and federate waits for other's message when RTI is not compiled with `-a` option, and federate is. So, I'll fix the protocol, and make a new PR for this.

lhstrh (approved):

Issue #1514, opened 2022-12-21
Cleanup the gradle files and support the `assemble` task


cmnrd:
The current repository structure is weird for a gradle project. Once https://github.com/lf-lang/lingua-franca/pull/1482 is merged, we should be able to restructure the repository in a more natural style for gradle projects and also let gradle manage *all* the building. Then we can also properly support gradle's `assemble` task (see comment [here](https://github.com/lf-lang/lingua-franca/issues/1447#issuecomment-1335555008) and hopefully fix other gradle related issues like https://github.com/lf-lang/lingua-franca/issues/1284.

PR #1779, opened 2023-05-24
Redesign of the repository layout and gradle configuration

cmnrd:
## Overview and Motivation

This PR presents a major refactoring of the source layout in our repository as well as the gradle configuration that we use for building. Our current layout was created by Xtext and does not follow the Gradle/Maven layout conventions that are typical for Java projects (https://github.com/lf-lang/lingua-franca/issues/1801). Moreover, our build configuration (gradle) grew over time, without ever considering the idiomatic gradle way of solving certain problems. This lead to a situation where we don't support gradle's standard tasks like `build` and `assemble` (https://github.com/lf-lang/lingua-franca/issues/1514). Moreover, there were several issues with the setup like Kotlin classes missing on a rebuild (https://github.com/lf-lang/lingua-franca/issues/1284) and whenever we tried to modify something in the configuration something else broke. For this reason, this PR rewrites the gradle scripts from scratch using an idiomatic approach.

## Goals

- Establish a source layout that follows Gradle/Maven conventions (https://github.com/lf-lang/lingua-franca/issues/1801)
- Create an idiomatic gradle setup and split code into subprojects where appropriate (https://github.com/lf-lang/lingua-franca/issues/1514)
- Embrace gradle as a first-order citizen of our tooling. Use gradle instead of trying to implement redundant functionality in bash scripts.

### Project structure

The new project structure looks like this:
```
org.lflang
├── cli
│   ├── base
│   │   ├── build.gradle
│   │   └── src/
│   ├── lfc
│   │   ├── build.gradle
│   │   └── src/
│   ├── lff
│   │   ├── build.gradle
│   │   └── src/
├── core
│   ├── build.gradle
│   └── src/
└── lsp
    ├── build.gradle
    └── src/
 ```
In the root of the repository (the root project) we have one sub-project `org.lflang` that is divided into multiple nested sub-projects. This is in contrast to the single project that we had before. Gradle doesn't like if certain plugins are combined (e.g. library and application) and instead one should be explicit about which sub-projects are libraries (with public API or internal) and which are executable applications.

In the new setup, we have the `core` projects which is a library and holds the largest part of our code base. The `lsp` project contains is application and holds sources that are specific to the language server. The `cli/lfc` and `cli/lfc` are both applications and of course correspond to our cli tools. `cli/base` is a library containing logic that is shared between the two.

### Source Layout

Each of the `src/` directories follows the Gradle/Maven source tree layout and looks something like this:
```
src
├── integrationTest
│   └── java
├── main
│   ├── java
│   ├── kotlin
│   └── resources
├── test
│   └── java
└── testFixtures
    └── java
```
Note that we divide the source tree by function (`main` i.e main sources, `test`, `testFixtures` and `integrationTest`) and then by language (`java`, `kotlin`, `resources`). `resources` contains all additional resources that should be on the classpath like our runtime submodules.

### Testing

Instead of having a separate test project (the `org.lflang.tests` from before) we now make the tests part of each project (i.e. we place the tests close to the code they are testing). In the source layout from above, the `test` subdirectory contains all unit tests. We make use or gradle's test suite feature and also define an additional test suite `integrationTest`. These are our runtime tests that compile and execute each test LF program. Separating unit and integration tests gives us more control about which types of tests we run. The `test` task will only run all unit tests and not the slow integration tests.

The integration tests can be run using `./gradlew integrationTest`. We can also specify the precise test as before. For instance, the following command will run all concurrency tests for the C target:
```
./gradlew org.lflang:core:integrationTest --tests org.lflang.tests.runtime.CTest.runConcurrentTests
```
Note that we qualify the `integrationTest` task with the project, so that we only search for matching tests in this project.

The root project defines a convenient helper tasks that allows to run all tests for a specific target. For instance, you can use the following command to run all Rust tests:
```
./gradlew targetTest -Ptarget=Rust
```
You can specify any valid target. If you run the task without specifying the target property `./gradlew tagetTest` it will produce an error message and list all available targets.

Another root task allows running only a single specific test:
```
./gradlew singleTest -DsingleTest=test/C/src/Minimal.lf
```

Note that this replaces the `runSingleTest` task.

To invoke only the C tests in the `concurrent` category, for example, do this:
```
./gradlew org.lflang:core:integrationTest --tests org.lflang.tests.runtime.CTest.runConcurrentTests
```

Note that the description above can also be found in the updated `test/README.md`.

### Building and running with Gradle

To build everything, simply run `./gradlew build`. Note that the build task in Gradle also includes running all checks. So `build` will also execute the `test` and `checkFormat` tasks. If you don't want the checks to run, `./gradelw assemble` can be used instead. However, I think it is great to always make sure that the checks pass when building and Gradle is quite smart in figuring out which checks to run. It will only rerun those affected by a change. Overall, build should be quite fast.

Both the `build` and the `assemble` task will create executable `lfc` and `lff` scripts in `build/install/lf-cli/bin`.

As before, we also define custom tasks `runLfc` and run `runLff` that can be used to directly execute the cli applications:
```
./gradlew runLfc --args=test/C/src/Minimal.lf
```

### Packaging

Packaging is trivial now. Gradle's distribution plugin takes care of it. It creates zip and tar archives in `build/distributions` that are ready to be uploaded as artifacts. In fact, the installed scripts from above are the unpacked version of these archives.

For nightly builds, the `nightly` project property can be specified. For instance:
```
./gradlew runLfc -Pnightly
```
This will create an archive called something like `lf-cli-0.4.1-SNAPSHOT-nightly-20230531150822.zip`. Note that the name now also includes the snapshot version, as before it was named `lf-cli-nightly-20230531150822.zip`. This shouldn't be an issue, though.

### Scripts in bin

We have collected several bash scripts in `bin/`, many of which have rarely been used or are redundant.
- `measure-lf-time` and `run-multiple-times` are simply not needed as this functionality is covered by the benchmark runner. 
- `build-lf-cli` and `run-lf-tests` are also not needed as we can instead invoke the corresponding gradle task directly.
- `bump_versions` we do need, but I moved it to `util/scripts/bump-versions`. It should only be used rarely and should probably not be located in the prominent bin directory.
- `lfc` and `lff` see below

#### `bin/lfc` and `bin/lff` scripts

I decided to make the `bin/lfc` and `bin/lff` scripts as simple as possible. They simply call the corresponding gradle run tasks. Running `./bin/lfc <arg1> <arg2> <...>` is equivalent to running `./gradlew runLfc --args="<arg1> <arg2> <...>"`.

Advantages:
- The scripts are as simple as it gets. We don't have to check if the code was actually build and don't have to produce an error message if the user needs to build first.
- Gradle will make sure that everything is build up to date. So lfc and lff will always run using the latest code, which is what we (at least I ;) ) want in a developer setup.
- Incremental builds are typically very fast for the run tasks (gradle will only do what is necessary for running).
- Tab-Completion on the command line works when using the scripts but not when using the gradle run tasks (at least for me)

Disadvantages:
- Gradle will produce output before and after the actual output from lfc/lff.

Note that users and developers can still use the lfc/lff scripts in `build/install/lf-cli/bin` if they prefer. Those invoke the commands directly (not through gradle).

### CI Flows

Since the `build` task includes running unit tests and formatter checks, we don't need separate CI jobs for the latter two anymore.

### Idea Plugin

I also added the idea plugin to our root project. This allows to run `./gradlew openIdea` to open the project in IntelliJ IDEA.

## Open questions

- ~Should we eliminate the `org.lflang`  directory and have `cli`, `core`, `lsp` directly as sub-projects of the root project?~
- ~Is changing the behavior of `bin/lfc` and `bin/lff` controversial or is it Ok for everyone?~

## TODO

- [x] Figure out why ominous directory `tmp/fed-gen` is created in project root after running unit tests.
- [x] Move `util/scripts/bump-versions` to epock repo ([still works and fix if needed. It should also consider the versions in ](https://github.com/lf-lang/epoch/pull/16)
- ~The release tooling needs to be updated~  ( after merging this)
- [x] There is a file in `org.lflang/todo/log4j.xml` where it is unclear what it was needed for (the current location in master is `org.lflang/todo/log4j.xml`). Was this needed for the LSP or for the CLI tools? Either way, it should be added as a resource to the corresponding application projects.
- [x] Add dummy tasks for our old tasks (`buildAll`, `runSingleTest`) with an informative error message indicating what to do instead.
- [x] Update benchmark CI tests.
- [x] The paths used to check which CI tests need to run need to be updated. (@petervdonovan)
- [x] Update documentation on the website (https://github.com/lf-lang/lf-lang.github.io/pull/140)
- [x] Test the LSP distribution.
- ~Integrate changes in epoch repo.~ -> do later

## Fixed issues

- Fixes https://github.com/lf-lang/lingua-franca/issues/1801
- Fixes https://github.com/lf-lang/lingua-franca/issues/1514
- Fixes https://github.com/lf-lang/lingua-franca/issues/1284
- Fixes https://github.com/lf-lang/lingua-franca/issues/1342
- Fixes https://github.com/lf-lang/lingua-franca/issues/567

### CODE REVIEW

**Comment by cmnrd** [comment]
There is still stuff that needs fixing, but the overall structure is set and this should be ready for first reviews. I will also update the PR description when I find the time.

**Comment by lhstrh** [comment]
One problem I see with putting `lfc` in `./lfc/build/install/lfc/bin/` is that if we do the same for `lff` and `lfc`, we'll end up with a lot of `bin` directories to put on people's `PATH`. Is there a way to avoid this?

**Comment by petervdonovan** [comment]
> Is there a way to avoid this?

I do not know about an idiomatic Gradle solution, but we could work around this by using symlinks or by providing a script that updates the PATH environment variable so that people do not have to do it by hand.

**Comment by lhstrh** [comment]
> > Is there a way to avoid this?
> 
> I do not know about an idiomatic Gradle solution, but we could work around this by using symlinks or by providing a script that updates the PATH environment variable so that people do not have to do it by hand.

I don't think we should mess with people's `PATH` unrequested. If we use symlinks, should we check in symlinks or have Gradle create them?

**Comment by petervdonovan** [comment]
> I don't think we should mess with people's PATH unrequested.

Well, people can choose whether or not to use the script.

**Comment by axmmisaka** [comment]
> One problem I see with putting `lfc` in `./lfc/build/install/lfc/bin/` is that if we do the same for `lff` and `lfc`, we'll end up with a lot of `bin` directories to put on people's `PATH`. Is there a way to avoid this?

I think we could just make current `./bin/lfc` a symlink that points to `./org.lflang/cli/lfc/build/install/lfc/bin/lfc`...... And I think making PATH very complicated is...... kinda normal if the user wants to

**Comment by cmnrd** [comment]
> One problem I see with putting lfc in ./lfc/build/install/lfc/bin/ is that if we do the same for lff and lfc, we'll end up with a lot of bin directories to put on people's PATH. Is there a way to avoid this?

That is already solved. The description above is a bit outdated... First of all, we have to distinguish the distribution version of our project (e.g. the downloadable archive that we make available for releases or nightly builds) and the developer version.
- **distribution**: The root project creates a combined archive bundling both lfc and lff. Check out the contents of `build/distribution` after running `./gradlew build`. It contains an archive ready for distribution. And `build/install/lf-cli/bin` contains both runnable scripts.
- **development**: For the scripts in the `bin` directory which we use in a developer setup, I decided to simply make them call gradle. So invoking `./bin/lfc <args>` is in fact equivalent to running `./gradlew org.lflang:cli:lfc --args <args>` (which in turn does the same as the `runtlfc` gradle task now). This has the main advantage, that the lfc run is always up to date. Also incremental building is super fast when using gradles run tasks. Also we don't need a separate build script and can just let gradle handle things for us.

So in fact, the user/developer may choose which version to use.

**Comment by cmnrd** [comment]
>     1. `./gradlew runLfc --help` outputs gradle help, not lfc help (I guess we might need better ways to pass in args)

This is just the way gradle works (we have the same behaviour in master).  `./gradlew runLfc --args --help` should produce the expected output.

>     2. Plainly calling `./bin/lfc` results in error: `> No argument was provided for command-line option '--args' with description: 'Command line arguments passed to the main class.'` (idem)

That is a good catch. Thanks! It should be an easy fix though.

>     3. For `runLfc` and `org.lflang:cli:lfc:run` gradle will run some long configuration tasks...... kinda like "pre-attack animation" in moba games......

This is of course a side effect of running lfc/lff through gradle. This, however, only effects the developer setup and for a developer I think it is desired anyway to build and execute the latest code. If the user/developer wants to run the plain version they can simply fall back to the distribution in build/install. I unterstand that running lfc/lff through gradle os porentially a controversial change and we should also debate it in the wider team.

**Comment by cmnrd** [comment]
Actually `./gradlew runLfc --args --help` does not work either, but `./gradlew runLfc --args="--help"` does work. I also updated the scripts in bin accordingly.

File: util/scripts/launch.sh [hunk 1]
```diff
@@ -69,5 +67,11 @@ else
   exit 2
 fi
 
-# Launch the compiler.
-run_cli_tool_with_args "$@"
+gradlew="${base}/gradlew"
+
+# Launch the tool.
+if [ $# -eq 0 ]; then
+    "${gradlew}" -p "${base}" "org.lflang:cli:${tool}:run"
```
lhstrh, line 74: :thinking: Does this mean that Gradle is needed as a dependency to be able to run `lfc`?

lhstrh (approved):
This reorganization looks great to me. I'm a little confused by the use of Gradle in the launch script, but I'm guessing this script is only to be used in a development context, not for a release artifact...please confirm :-)

.github/workflows/c-zephyr-tests.yml
```diff
@@ -41,12 +41,14 @@ jobs:
...
+      - name: Collect code coverage
+        run: ./gradlew jacocoTestReport
       - name: Report to CodeCov
         uses: codecov/codecov-action@v3.1.1
         with:
-          file: org.lflang.tests/build/reports/xml/jacoco
+          files: org.lflang/core/build/reports/xml/jacoco
           fail_ci_if_error: false
           verbose: true
-        if: ${{ !inputs.runtime-ref && runner.os == 'Linux' }}  # i.e., if this is part of the main repo's CI
```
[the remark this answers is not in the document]
lhstrh: Nvm this, I see that the check is redundant.

lhstrh:
It would be great if you could document a summary of the changes in the PR description (which will get linked from the release notes), including a list of frequently used commands or workflows that are now different.

util/scripts/launch.sh
lhstrh: Now, when I run `lfc` without arguments, instead of getting output directly from `lfc`, I get a message from `gradle`. I don't think this is what we want.

lhstrh: I now see this is also what you were discussing above with @axmmisaka. What motivated this change to begin with?

cmnrd: > Does this mean that Gradle is needed as a dependency to be able to run lfc?

Gradle is only needed in the developer setup, and since we use the wrapper it will install itself.

> Now, when I run lfc without arguments, instead of getting output directly from lfc, I get a message from gradle. I don't think this is what we want.

The output of lfc is printed directly. Gradle only adds a message to the bottom.

> I now see this is also what you were discussing above with @axmmisaka. What motivated this change to begin with?

To get rid of the launch shell scripts that search for a compiled jar and then invoke it and the build script. I think we should just embrace that we use gradle and not add additional logic around it. There are a few other reasons to, which I will add to my summary.

.github/workflows/all-misc.yml
```diff
@@ -21,11 +21,6 @@ jobs:
   check-diff:
     uses: ./.github/workflows/check-diff.yml
 
-  junit:
-    uses: ./.github/workflows/unit-tests.yml
```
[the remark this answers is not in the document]
cmnrd: The build task of gradle implies running all checks (including both tests and formatting). So it is sufficient to have a single build CI job.

cmnrd:
Yes, that is desperately needed and on its way.

lhstrh:
@petervdonovan, it just occurred that the paths used to check which tests need to run will need to be updated in this branch.

petervdonovan:
> There is a file in `org.lflang/todo/log4j.xml` where it is unclear what it was needed for (the current location in master is `org.lflang/todo/log4j.xml`). Was this needed for the LSP or for the CLI tools? Either way, it should be added as a resource to the corresponding application projects.

Its current location in `master` is `org.lflang/src/log4j.xml`. It was needed for the language server. Originally a large volume of what appeared to be debug logs from the XText framework were appearing in the output of the LF language server, which is unacceptable because that output is what we use to display the results from the build and to help users debug their build problems.

> Test the LSP distribution. Do we need to create a fat jar?

How else would we do this? The VS Code extension is essentially distributed as a `.zip` file, and all the Java resources required to operate the language server have to be packed in that file.

I suppose it is possible to ship classes in multiple separate jars; that is what we used to do with the SWT jars. But I do not know much about this.

lhstrh:
@a-sr mentioned during the call that @soerendomroes thought that the way we deal with the jar in VS Code was somehow non-standard and unnecessarily complicated, but I didn't understand what the exact problem was. Perhaps we should arrange a chat with @soerendomroes to discuss this?

petervdonovan:
> Perhaps we should arrange a chat with @soerendomroes to discuss this?

I would be fine with that. The goal should be to minimize the amount of time spent fiddling with the language server build, so if there is a way to make it even simpler, then that would be great.

cmnrd:
> Its current location in master is org.lflang/src/log4j.xml. It was needed for the language server.

Ok, great, then I will move it to the resources of the lsp subproject.

> How else would we do this? The VS Code extension is essentially distributed as a .zip file, and all the Java resources required to operate the language server have to be packed in that file.
I suppose it is possible to ship classes in multiple separate jars; that is what we used to do with the SWT jars. But I do not know much about this.

Currently, we use gradle's distribution plugin for packaging the LSP (similar to lfc and lff). This creates a zip (or tar archive) containing all the individual jars and a launch script. I don't really know how the language server is started from VS code. If we can simply run the `bin/lsp` script, we should be good with what we have right now. If it is more convenient to have a single jar and directly call java, I can add the shadow jar plugin back and create a fat jar.

cmnrd:
@petervdonovan I added a the shadow jar plugin back to the lsp supbroject. Now `./gradlew lsp:assemble` or `./gradlew lsp:build` will also create a fat jar. This can be found in `lsp/build/libs/lsp-0.4.1-SNAPSHOT-all.jar`. I assume that the VS code extension build needs to be changed, as the paths changed. Could you take care of updating the extension and testing if the lsp jar that we produce works as before?

cmnrd:
@erlingrj Do you have any idea why the zephyr tests are failing?

erlingrj:
I think the problem is that my `RunZephyrTests.sh` script requires that all the directories below `test/C/src-gen` are built for Zephyr. It recursively executes each test on the QEMU emulator. There seems to be some leftover src-gen directories, from previous workflows here. Can I pass an argument to gradlew inorder to pass `-c` to lfc and delete all the previous src-gen directories?

cmnrd:
Ok, that gave me the right pointer. I fixed it in https://github.com/lf-lang/lingua-franca/pull/1779/commits/75a05787135de68580d5186e4c63bd3748defac8 (hopefully). The problem was that the new CI setups runs all test of the `CZephyrTests` class. Since that was based on `RuntimeTest` it also inherited all the default C tests.

petervdonovan:
Because Christian added back a way to construct a fat jar for the language server, it is trivial to update `vscode-lingua-franca` to be compatible with these changes (https://github.com/lf-lang/vscode-lingua-franca/pull/120). In any case, we don't have to update `vscode-lingua-franca` right away since it still tracks the same commit as a submodule, so that should not block our merging this.

cli/lfc/src/test/kotlin/org/lflang/cli/LfcIssueReportingTest.kt
```diff
@@ -155,6 +151,6 @@ class LfcIssueReportingTest {
         trim().replace(Regex("\\s+\\R"), "\n")
             // Replace file path with placeholder. File path is not
             // rendered the same on linux and windows (forward/backward slash)
-            .replace(lfFile.toString(), "%%%PATH.lf%%%")
+            .replace("%%%PATH.lf%%%", lfFile.toString())
```
petervdonovan, line 154: What happened here? It seems like the original intent was to replace the value of `lfFile.toString()` with `"%%%PATH.lf%%%"` and not the other way around.

petervdonovan (approved):
I am far from expert in Gradle, but I think this is probably ready to merge.

Thanks Christian!

petervdonovan:
In order to get this to work in IntelliJ, I needed to go to `Help > Edit Custom VM Options` and add the following to the bottom of the `.vmoptions` file:
```
-DsingleTest=test/C/src/Minimal.lf
-Ptarget=C
```

We should make sure people know about this to save them some trouble.

lhstrh:
> In order to get this to work in IntelliJ, I needed to go to `Help > Edit Custom VM Options` (...)

@cmnrd, do you think there is a way to avoid this extra step?

cli/lfc/src/test/kotlin/org/lflang/cli/LfcIssueReportingTest.kt
cmnrd: I am not entirely sure either. The test used to fail and worked fine after switching this.

petervdonovan:
> Or was your problem that IntelliJ would throw errors when the tasks are loaded initially? If that was the case for you, it should be fixed by https://github.com/lf-lang/lingua-franca/commit/b2c70d2ed199c50ea2be458080b4d7f60f1674ad.

You are right, the problem seems to be solved now!

Issue #1761, opened 2023-05-21
Remove the `sneakyThrow` pattern that was inherited from generated Xtend code


lhstrh:
> The `sneakyThrow` pattern was inherited from generated Xtend code, and AFAIK there is no good reason to use it. If we can use the error reporter, that would be best, but just an ordinary `RuntimeException` would also do, I guess.

_Originally posted by @lhstrh in https://github.com/lf-lang/lingua-franca/pull/1759#discussion_r1199682095_

PR #1787, opened 2023-05-26
Fix for #1785

### LINKED ISSUE #1785: Compiling a federation succeeds even if one federate fails
If you have a federation where one of the federates fails to compile, the overall process succeeds. Here is a simple example (assumed to be in `test/C/federated/Foo.lf`), which fails to compile because of issue #1733:
```
target C {
    timeout: 5 sec,
    coordination: decentralized
}
import Count from "../lib/Count.lf"
import TestCount from "../lib/TestCount.lf"

reactor PrintTimer(STP_offset:time = 10 msec) extends TestCount {
    timer t(0, 1 sec);
    reaction(t) {=
        lf_print("Timer ticked at (%lld, %d).", 
            lf_time_logical_elapsed(), lf_tag().microstep
        );
    =}
}
federated reactor {
    c = new Count();
    p = new PrintTimer();
    c.out -> p.in;
}
```
The compiler reports an NPE when compiling one federate:
```
Copied 'include/federate__p_extension.cmake' from the file system.
Exception in thread "pool-1-thread-2" java.lang.NullPointerException: Cannot invoke "java.util.LinkedHashSet.stream()" because the return value of "org.lflang.ast.ASTUtils.allFileLevelPreambles(org.lflang.lf.Reactor)" is null
	at org.lflang.generator.c.CGenerator.lambda$generateTopLevelPreambles$11(CGenerator.java:2052)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:273)
```
but then reports success:
```
[100%] Built target federate__c
Install the project...
-- Install configuration: "Release"
SUCCESS: Compiling generated code for federate__c finished with no errors.
Compiled binary is in /Users/eal/lingua-franca-intellij/test/C/fed-gen/STPParameter/bin
##### Generating launcher for federation  in directory /Users/eal/lingua-franca-intellij/test/C/bin
Code generation finished.
```

lhstrh:
Note that this is a symptom of the problem described in #1761. Fixes #1785.

### DIFF
--- a/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
+++ b/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
@@ -161,6 +161,12 @@ public class FedGenerator {
             subContexts -> {
               createDockerFiles(context, subContexts);
               generateLaunchScript();
+              // If an error has occurred during codegen of any federate, report it.
+              subContexts.forEach(c -> {
+                if (c.getErrorReporter().getErrorsOccurred()) {
+                  context.getErrorReporter().reportError("Failure during code generation of " + c.getFileConfig().srcFile);
+                }
+              });
             });
 
     context.finish(Status.COMPILED, codeMapMap);
@@ -321,7 +327,7 @@ public class FedGenerator {
     try {
       compileThreadPool.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
     } catch (Exception e) {
-      Exceptions.sneakyThrow(e);
+      context.getErrorReporter().reportError("Failure during code generation.");
     } finally {
       finalizer.accept(subContexts);
     }
--- a/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
+++ b/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
@@ -26,7 +26,6 @@ import org.eclipse.xtext.generator.JavaIoFileSystemAccess;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.resource.XtextResourceSet;
 import org.eclipse.xtext.util.RuntimeIOException;
-import org.eclipse.xtext.xbase.lib.Exceptions;
 import org.lflang.ErrorReporter;
 import org.lflang.FileConfig;
 import org.lflang.LFStandaloneSetup;
@@ -162,11 +161,15 @@ public class FedGenerator {
               createDockerFiles(context, subContexts);
               generateLaunchScript();
               // If an error has occurred during codegen of any federate, report it.
-              subContexts.forEach(c -> {
-                if (c.getErrorReporter().getErrorsOccurred()) {
-                  context.getErrorReporter().reportError("Failure during code generation of " + c.getFileConfig().srcFile);
-                }
-              });
+              subContexts.forEach(
+                  c -> {
+                    if (c.getErrorReporter().getErrorsOccurred()) {
+                      context
+                          .getErrorReporter()
+                          .reportError(
+                              "Failure during code generation of " + c.getFileConfig().srcFile);
+                    }
+                  });
             });
 
     context.finish(Status.COMPILED, codeMapMap);
--- a/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
+++ b/org.lflang/src/org/lflang/federated/generator/FedGenerator.java
@@ -330,7 +330,8 @@ public class FedGenerator {
     try {
       compileThreadPool.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
     } catch (Exception e) {
-      context.getErrorReporter().reportError("Failure during code generation.");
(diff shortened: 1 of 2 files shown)


org.lflang/src/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -321,7 +330,7 @@ public TargetConfig getTargetConfig() {
     try {
       compileThreadPool.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
     } catch (Exception e) {
-      Exceptions.sneakyThrow(e);
+      context.getErrorReporter().reportError("Failure during code generation.");
```
petervdonovan, line 333: To my knowledge, it "should be impossible" for errors to bubble up to this point, in the sense that if there is an exception here, it indicates that the compiler has a bug, and that the problem is not merely a matter of invalid user input.

If that is the case, then I think that we should print the full stack trace of the error message and/or re-throw it as a `RuntimeException`. The resulting stack trace will make the output of the compiler look bad, but if the compiler has a bug then it should admit that by allowing its output to look bad. This will be more helpful for debugging than an error message like "Failure during code generation," which is not very specific and which does not provide any hints about where in the code to look for the error.

petervdonovan (changes requested):
Certainly I am happy that we are phasing out this Xtext magic. I left a quibble regarding how the error message is presented.

petervdonovan (approved):

org.lflang/src/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -330,7 +330,8 @@ public TargetConfig getTargetConfig() {
     try {
       compileThreadPool.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
     } catch (Exception e) {
-      context.getErrorReporter().reportError("Failure during code generation.");
+      context.getErrorReporter().reportError("Failure during code generation: " + e.getMessage());
```
lhstrh, line 333: This is actually not the exception triggered by the example in #1785 (it is the other one). I'm thinking we should do the same there?

lhstrh: Oh, never mind; line 170 is not in response to an exception but an error reported in one of the code generators.

org.lflang/src/org/lflang/generator/c/CGenerator.java
```diff
@@ -416,9 +416,8 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
       copyTargetFiles();
       generateHeaders();
       code.writeToFile(targetFile);
-    } catch (IOException e) {
-      //noinspection ThrowableNotThrown,ResultOfMethodCallIgnored
-      Exceptions.sneakyThrow(e);
+    } catch (Exception e) {
+      errorReporter.reportError(e.getMessage());
```
lhstrh, line 420: This is the line that is hit by the example though. Should we add a `reportError` method that allows passing in an exception so that we can preserve the stack trace and print it later?

petervdonovan: OK, maybe we should handle the stack trace here too.

Why not just print the stack trace immediately? We should never have to print a stack trace anyway except when something else is already horribly wrong.

I also just noticed that this catches a general `Exception` instead of an `IOException`, which seems unnecessary.

petervdonovan: Oh, now I remember why I was OK with this. This is actually not a situation when something is horribly wrong. This line can be reached when the user provides an invalid path to the TargetConfig. In that case, probably just the error message will usually be sufficient to figure out what the problem is, _and_ it isn't really the compiler's fault, so maybe the compiler shouldn't have to print out an ugly stack trace.

lhstrh: If we only catch an `IOException` as you suggest, the example that Edward provides in #1785 is no longer handled correctly.

lhstrh: I suppose we can catch the runtime exception later, but we weren't doing that either...

petervdonovan: Ah, okay. Then since we are catching more general errors such as NPEs here, I guess I agree with you that we should do something with the stack trace.

lhstrh: Reverting this to a catch for `IOException` and handling runtime exceptions further up the call stack might be better though...

petervdonovan: That also makes sense. I am happy as long as we never silence exceptions by catching them without handling them properly. If we forget to handle some exceptions that "should be impossible", then we will still get a stack trace, which I would be satisfied with.

lhstrh: OK, narrowed it down and added a separate catch for `RuntimeException`.

edwardalee (approved):
LGTM! Thanks for fixing this.

PR #1788, opened 2023-05-26
Amortize slow LFF startup time

### LINKED ISSUE #1342: `LFFormatter` is slow

petervdonovan:
Fixes #1342.

Benefits of correctly incorporating LFF into Spotless include:
- uniform framework for formatting tests in this repository, including spotlessCheck, for which we have not implemented an analogue
- use of Spotless's [PaddedCell](https://github.com/diffplug/spotless/blob/main/PADDEDCELL.md) features
- use of Spotless's ability to run only on changed files (I have verified that using the Spotless task is empirically much faster when only one file has been modified since the last formatter run)

This solution works by giving the formatter the ability to listen in its input stream for the paths that it needs to format. This lets us use one instance of the formatter for the entire lifetime of the calling Gradle daemon.

To my knowledge, Spotless does not provide an API for releasing resources used by the formatting step across multiple invocations of the formatter. However the JVM does provide a shutdown hook, which I think is an acceptable though imperfect way to release the formatter. I have verified on Ubuntu that the formatter subprocess does indeed terminate when the enclosing Gradle daemon is killed with a SIGKILL.

### DIFF
--- a/buildSrc/src/main/java/lfformat/LfFormatStep.java
+++ b/buildSrc/src/main/java/lfformat/LfFormatStep.java
@@ -1,23 +1,15 @@
 package lfformat;
 
-import com.diffplug.spotless.FormatterFunc;
 import com.diffplug.spotless.FormatterStep;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.io.OutputStream;
 import java.io.OutputStreamWriter;
-import java.io.PrintWriter;
-import java.io.Reader;
-import java.io.Serializable;
 import java.io.Writer;
-import java.lang.ProcessBuilder.Redirect;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
@@ -41,6 +33,7 @@ public final class LfFormatStep {
     // The use of the static keyword here is a workaround for serialization difficulties.
     /** The path to the lingua-franca repository. */
     private static Path projectRoot;
+
     private static Step instance;
 
     private static Process formatter;
@@ -56,16 +49,19 @@ public final class LfFormatStep {
 
     private Step() throws IOException {
       initializeFormatter();
-      Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-        try {
-          writer.close();
-          formatter.waitFor();
-          reader.close();
-          error.close();
-        } catch (IOException | InterruptedException e) {
-          throw new RuntimeException(e);
-        }
-      }));
+      Runtime.getRuntime()
+          .addShutdownHook(
+              new Thread(
+                  () -> {
+                    try {
+                      writer.close();
+                      formatter.waitFor();
+                      reader.close();
+                      error.close();
+                    } catch (IOException | InterruptedException e) {
+                      throw new RuntimeException(e);
+                    }
+                  }));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/lfformat/LfFormatStep.java
```diff
@@ -21,38 +23,75 @@ public final class LfFormatStep {
...
     /** The path to the lingua-franca repository. */
     private static Path projectRoot;
 
+    private static Step instance;
+
+    private static Process formatter;
+    private static Writer writer;
+
+    private static BufferedReader reader;
+    private static BufferedReader error;
```
petervdonovan, line 43: It is a little bit of a red flag that these variables are static. It is necessary because `Step` needs to be `Serializable`. I do not fully understand why this is required by Spotless, but I am under the impression that it relates to an optimization that we are not trying to use at the moment.

cmnrd (commented):
Great! I changed the base to the `gradle` branch. It's probably best to merge it there first. 

There is one major issue with this though. As lff runs as a daemon in the background, any changes we make to the formatter do not have an effect until we kill the gradle daemon and then do a clean build. Can we just spawn a new lff process when `create` is called and kill the old one?

cmnrd:
Since you are the first to work with the new repo structure in https://github.com/lf-lang/lingua-franca/pull/1779. Do you think it makes sense?

petervdonovan:
> Do you think it makes sense?

Yes, I think so. One minor issue that I see is that `RunSingleTestMain` and `RuntimeTest` are separated from everything else that is in the `org.lflang.tests` package (why?), and for some reason IntelliJ does not seem to know how to build those two files (code help does not work for them).

petervdonovan:
On line 26, where it says
```groovy
spotlessLinguaFranca.dependsOn(":org.lflang:cli:lff:installDist")
```

There seems to be a circular dependency, for running `./gradlew org.lflang:cli:lff:installDist` triggers an error
```
* Where:
Build file '/home/peter/vscode-lingua-franca/lingua-franca/build.gradle' line: 19

* What went wrong:
A problem occurred configuring root project 'org.lflang'.
> Failed to notify project evaluation listener.
   > Cannot run program "org.lflang/cli/lff/build/install/lff/bin/lff": error=2, No such file or directory
```

which must mean that we ran `spotlessLinguaFranca` in the process of running `installDist`, before `installDist` had produced its output, `lff`.

petervdonovan:
> Can we just spawn a new lff process when create is called and kill the old one?

Yes, good idea. Now we always run `terminateFormatter` when `create` is called and we initialize the formatter as needed.

cmnrd:
> Yes, I think so. One minor issue that I see is that RunSingleTestMain and RuntimeTest are separated from everything else that is in the org.lflang.tests package (why?), and for some reason IntelliJ does not seem to know how to build those two files (code help does not work for them).

This is just an artifact of the porting process. The plan is to eliminate `org.lflang.tests` and move all its contents. The only thing remaining to be ported is `RunSingleTestMain`.

cmnrd (approved):
This looks great! I think the static variables and the fact that  we need to run a separate lff daemon, are just artifacts of the hacky thing we as doing, where a build flow (formatting) also depends on our build output (lff). It seems to work fine though!

I made a small change and added the outputs of the task that insalls lff to the inputs of our formatting tasks. This ensures that gradle automatically reformats all lf files when lff changed.

PR #1793, opened 2023-05-28
Fix warnings reported on CliBase.

petervdonovan:
The comparison of strings to string literals looks like a critical error which I think would have been caught by sufficient tests.

One of these warnings is incorrect, but we might as well silence it with an assertion.

The warning about toArray is merely a matter of best practice. A human couldn't be expected to know about this, but the warning emitted by IntelliJ is correct.

I'm not sure I have time to add tests for this at the moment or to do a close second read of this code but we also don't want to leave wrong code sitting around.


lhstrh (approved):
Thanks for the fixes. These are pretty bad :facepalm:

cmnrd:
Now that we have formatting addressed, we should also introduce linting and enforce that the code is warning free. I opened https://github.com/lf-lang/lingua-franca/issues/1806

PR #1794, opened 2023-05-28
Fixes for federated tests

petervdonovan:
This is an effort to partially correct some of the Python federated test failures that we are observing.


petervdonovan:
I think that the zombie RTI problem is probably fixed now. I didn't even have to do anything terribly hacky.

Looking at the logs that we are now getting in CI, it seems possible that many (most/all???) of the errors that we are getting could be trivial to fix.

petervdonovan:
The Python federated tests all pass for me locally. CI is failing due to the expired Klighd certificate.

org.lflang/src/org/lflang/federated/generator/FedTargetConfig.java
```diff
@@ -57,11 +58,20 @@ private void mergeImportedConfig(
...
-            this, convertToEmptyListIfNull(targetProperties.getPairs()), errorReporter);
+            this,
+            convertToEmptyListIfNull(targetProperties.getPairs()),
+            getRelativePath(mainResource, federateResource),
+            errorReporter);
       }
     }
   }
 
+  private Path getRelativePath(Resource source, Resource target) {
```
lhstrh, line 69: Isn't there a method in `FedFileConfig` that does something like this?

lhstrh (approved):
This generally looks good to me! If the tests pass locally, then let's go ahead and merge this into @byeong-gil's branch. Thanks a bunch for helping out with this, @petervdonovan!

org.lflang/src/org/lflang/federated/generator/FedTargetConfig.java
petervdonovan: Yes, it is similar, and it is better in that it handles important cases that I did not address here. However, it produces paths that are relative to the generated LF file, whereas here we need paths relative to the original main LF file.

The "right way" to fix this is to require the projects consisting of multiple files to specify target properties in a separate Lingo config file. I would be OK with merging this as a stopgap measure until we can do it the right way.

lhstrh: Sounds fair to me...

byeong-gil (approved):
LGTM! Thank you for helping to address these issues, @petervdonovan! I'll merge this to my PR ([lingua-franca#1752](https://github.com/lf-lang/lingua-franca/pull/1752)).

PR #1795, opened 2023-05-29
Update extract-ref.yml to use the new label to output result 

erlingrj:
I believe this should work, but cannot test until it is in master


lhstrh (approved):
This looks right to me. Apologies for the inconvenience this rename caused...

petervdonovan (approved):
Thanks!

PR #1798, opened 2023-05-30
Pass CLI args to federates

petervdonovan:
When debug mode is set using a CLI argument, it does not seem to be passed on to federates generated in `fed-gen`. This amends that.


lhstrh (approved):
Nice catch! Thanks, @petervdonovan.

edwardalee (approved):
Ah, this could explain it.  I guess this has been happing since fedgen.

PR #1804, opened 2023-05-31
Generic params as arguments

petervdonovan:
Allow parameters of generic reactors to syntactically appear as arguments of contained reactors.

### DIFF
--- a/org.lflang/src/org/lflang/generator/c/CUtil.java
+++ b/org.lflang/src/org/lflang/generator/c/CUtil.java
@@ -498,9 +498,9 @@ public class CUtil {
    */
   public static String selfType(TypeParameterizedReactor reactor) {
     if (reactor.reactor().isMain()) {
-      return "_" + CUtil.getName(reactor) + "_main_self_t";
+      return CUtil.getName(reactor) + "_main_self_t";
     }
-    return "_" + CUtil.getName(reactor) + "_self_t";
+    return CUtil.getName(reactor) + "_self_t";
   }
 
   /** Construct a unique type for the "self" struct of the class of the given reactor. */
--- a/org.lflang/src/org/lflang/generator/c/CUtil.java
+++ b/org.lflang/src/org/lflang/generator/c/CUtil.java
@@ -28,6 +28,7 @@ package org.lflang.generator.c;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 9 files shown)


org.lflang/src/org/lflang/generator/c/TypeParameterizedReactor.java
```diff
@@ -76,9 +90,23 @@ public InferredType resolveType(InferredType t) {
     return InferredType.fromAST(resolveType(t.astType));
   }
 
+  public String uniqueName() {
+    String name = reactor.getName().toLowerCase();
+    if (uniqueNames.containsKey(this)) return uniqueNames.get(this);
+    if (nameCounts.containsKey(name)) {
+      int currentCount = nameCounts.get(name);
+      nameCounts.put(name, currentCount + 1);
```
lhstrh, line 98: I noticed that there is one shared counter for all names. Not that it matters, but would it be worth it to have a counter per name for the same of readability?

org.lflang/src/org/lflang/generator/c/TypeParameterizedReactor.java
```diff
@@ -76,9 +90,23 @@ public InferredType resolveType(InferredType t) {
     return InferredType.fromAST(resolveType(t.astType));
   }
 
+  public String uniqueName() {
```
lhstrh, line 93: Comment missing.

org.lflang/src/org/lflang/generator/c/CUtil.java
```diff
@@ -135,7 +137,7 @@ public static String channelIndexName(PortInstance port) {
    * (to allow for instantiations that have the same name as the main reactor or the .lf file).
    */
   public static String getName(TypeParameterizedReactor reactor) {
-    String name = reactor.reactor().getName().toLowerCase() + reactor.hashCode();
+    String name = reactor.uniqueName();
```
lhstrh, line 140: This is a nice improvement!

lhstrh (commented):
Looks good! Only nitpicks...

org.lflang/src/org/lflang/generator/c/TypeParameterizedReactor.java
petervdonovan: But I think I do use different counters for different names. `currentCount` is retrieved from a hash table the keys of which are the names.

org.lflang/src/org/lflang/generator/c/TypeParameterizedReactor.java
petervdonovan: Done!

org.lflang/src/org/lflang/generator/c/TypeParameterizedReactor.java
lhstrh: Oh, you're right. I missed that. Great!

lhstrh (approved):

PR #1807, opened 2023-06-01
Adjustable port sides for reactors

a-sr:
Introduces a new attribute `@side` that influences the side of a reactor port in the diagram.

```
reactor IO {
    input I1:int
    @side("east")
    input I2:int
    
    output O1:int
    @side("west")
    output O2:int
}

main reactor {
    a = new IO()
    b = new IO()
    
    a.O1 -> b.I1
    b.O1 -> a.I1
    a.O2 -> b.I2
    b.O2 -> a.I2
}
```
port-side

Additionally, there is a new synthesis option (in the Layout category) named "Fixed Reactor Port Side". If inactive, the layout algorithm will decide the best location for ports. However, this will ignore the user attribute because the current layout algorithm does not support mixed constraints.
The result for the example above will look like this:

port-side-free


edwardalee (approved):
Nice!!  Minor suggestion made for docs. Also, I guess the formatter needs to be applied to get the test to pass.

lhstrh:
I'm guessing that the strings "north", "east", etc. are the ones required by ELK? I would have opted for "top", "left", etc. but I guess it's easy enough to get used to...

lhstrh (approved):
This is pretty cool! Haven't yet had a chance to try it out, but this looks good :-)

Issue #1800, opened 2023-05-31
Get rid of injection of Kotlin-based generators in `LFGenerator`


cmnrd:
Currently, we use a weird mechanism to manually load the classes for Kotlin based generators. This was needed to make the Eclipse developer setup happy.  We should get rid of the manual loading and just import the classes directly.

PR #1813, opened 2023-06-02
Remove odd mechanism for loading target generators dynamically

cmnrd:
This removes the manual loading of Kotlin classes in `LFGenerator`. We did this as a workaround, as the developer Eclipse setup would not allow importing Kotlin classes in Java code. Therefore, this change will definitely break the Eclipse developer setup. If someone wants to maintain it, then we could probably write an alternative implementation of `LFGenerator` that is only used in the Eclipse setup, or create dummy implementations of the Kotlin classes. This, however, would be only a temporary solution as I think we should really consider porting more code to Kotlin.

The Maven build for epoch should not be affected by this change.

Fix https://github.com/lf-lang/lingua-franca/issues/1800


edwardalee (approved):
I have no opinion about this. I've abandoned the Eclipse dev environment.

petervdonovan (approved):
LGTM. I do not know why the tests except C and Rust tests ran, but it could have something to do with changes happening in `master` between forking off this branch and triggering CI.

lhstrh (approved):
:+1: Thanks for doing this cleanup, @cmnrd!

a-sr (approved):

PR #1814, opened 2023-06-02
CI updates

petervdonovan:
~~If CI fails due to a formatting error, the formatter will have to be
    applied certainly, so to spare humans the annoyance we can just automate
    this.~~

This includes a couple of small formatting and CI-related chores.
- This updates `CONTRIBUTING.md` with suggestions about how to automate code formatting
- This corrects the paths appearing in `check-diff.yml`, which were wrong (oops)
- This addresses @cmnrd's [comment](https://github.com/lf-lang/lingua-franca/pull/1813#issuecomment-1573942291) that it would make sense to run the `build` task unconditionally.


cmnrd:
I am not a big fan of applying formatting automatically. I think it would be much better if people just get used to applying formatting directly in their IDEs. Unfortunately, we need a bit of annoyance to change habits ;)

I am a bit afraid that with this change evey other commit would be an automatic formatting commit, but maybe I am too cautious here. I would certainly not oppose this change if others would like to include it.

petervdonovan:
Thanks for voicing this concern early @cmnrd -- what you are saying makes sense. Let's see what @lhstrh thinks.

cmnrd:
And another thought: Many contributors are not used to debase their changes. So of we apply commits automatically in CI while the contributor continues working in their branch, we will also likely get an additional merge commit. And in addition to that, the formatting is likely to happen on code that is still being worked on, and hence is likely to create a conflict

lhstrh:
Every other commit being a formatting commit tends to happen already, except it requires manual effort. The advantage I see is that they would be authored by a bot and easily distinguishable from material changes. If the formatting checks and changes only happen once the PR is marked for review, then I think the impact would be negligible...

cmnrd:
My experience is that if you annoy people by having CI fail if they don't format, they will learn to format before they push and eventually also before they commit. It is also easy to set up formatting in IDEs. I haven't looked into this for spotless and IntelliJ yet, but it certainly is possible.

petervdonovan:
Just talked to Marten and I think we are all in agreement that it would be better to distribute instructions on IDE configuration for formatting. I will retract the changes relating to automatic commits.

lhstrh:
@petervdonovan should we rename this PR and still merge what's left?

axmmisaka:
I think this would be best achieved not with a bot but instead a [pre-commit hook](https://pre-commit.com/hooks.html), like this one: https://github.com/jguttman94/pre-commit-gradle

CONTRIBUTING.md
```diff
@@ -76,15 +76,50 @@ To address feedback from code review, implement changes and push to your feature
...
+You can enable formatting on save by going to `Settings` and searching for "format" and checking the box
+`Editor: Format on Save`. This can be either a user-level or workspace-level setting.
+
+_Checking using Git hooks_
+
+If you prefer not to use formatting on save but would still like to avoid committing unformatted code, you can add a
+file `.git/hooks/pre-commit` with the contents:
+
+```
+./gradlew spotlessCheck
```
lhstrh, line 120: Maybe better to suggest `spotlessApply` here?

CONTRIBUTING.md
```diff
@@ -76,15 +76,50 @@ To address feedback from code review, implement changes and push to your feature
...
 - To check that modified files are formatted correctly, run:
+
 ```
 ./gradlew spotlessCheck
 ```
+
 - To apply the changes recommended by the formatter, run:
+
 ```
 ./gradlew spotlessApply
```
lhstrh, line 91: I've always wondered why `./gradlew spotlessCheck` suggests to run `./gradlew :spotlessApply`. What is the `:` for?

CONTRIBUTING.md
petervdonovan: I have tried that, but the problem is that the changes that come from `spotlessApply` will not be staged. You could run `git add -a` in the hook, of course, but that would be dangerous because people would end up accidentally committing changes that they did not mean to commit.

lhstrh: Ah, I see. Does [this](https://github.com/jguttman94/pre-commit-gradle) solve that?

lhstrh (approved):
These changes look good to me!

CONTRIBUTING.md
petervdonovan: I do not know the details, but I believe the colon `:` is used in Gradle to delimit segments of hierarchical namespaces. So, maybe starting with a colon means that something is in the root namespace.

I think you get the same results either way.

CONTRIBUTING.md
petervdonovan: I do not know. If anyone wants to investigate that, we can update `CONTRIBUTING.md`.

petervdonovan:
Unfortunately, I also have found that the `google-java-format` plugin actually has [at least one serious bug](https://github.com/google/google-java-format/issues/911). It is possible that we should not recommend it. The Spotless plugin works fine, and the only issue with it is that it cannot be configured to execute on save.

CONTRIBUTING.md
lhstrh: Maybe @axmmisaka can have a look. Marking this as resolved for now.

CONTRIBUTING.md
cmnrd: Yes, you can also use `./gradlw cli:lfc:spotlessApply` to only format the lfc subproject. This works for all gradle tasks including `build`, `test`, `clean` etc.. If you run a task without a qualifier, then gradle will search for tasks with this name in all subprojects and execute them all.

CONTRIBUTING.md
```diff
@@ -76,15 +76,45 @@ To address feedback from code review, implement changes and push to your feature
...
 ```
 
+_Java code formatting in IntelliJ_
+
+Follow the directions here to install the Spotless Gradle plugin. We are not currently aware of a way to run Spotless
```
cmnrd, line 96: Looks like this is incomplete. I didn't see a description of the plugin in the diff.

CONTRIBUTING.md
axmmisaka: We could do something like this (inspired [here](https://medium.com/@mmessell/apply-spotless-formatting-with-git-pre-commit-hook-1c484ea68c34)):

```bash
f = $(git diff --name-only --cached --diff-filter=AM)
for file in $f; do
  if test -f "$file"; then # this if might be redundant, as we applied filter in `git diff`
    git add $file
  fi
done
git commit
```

cmnrd (commented):
It looks like the spotless plugin instructions are still missing.

The [Idea Gradle Plugin](https://docs.gradle.org/current/userguide/idea_plugin.html) allows us to modify the project configurations that developers use. (I see this as similar to the Oomph installer for Eclipse, but with much less overhead and probably also less power). I haven't really investigated yet how this works and what is possible. I doubt that we could install a plugin this way (but I am not sure either). It might be possible to setup a predefined keybinding though.

CONTRIBUTING.md
petervdonovan: Thanks for the catch! I just added a link.

Issue #1765, opened 2023-05-22
Fully support generics for the C target


petervdonovan:
- [ ] handle type arguments that are more syntactically complex than just a single identifier
- [ ] propagate values for type arguments across multiple levels of instantiation hierarchy
- [ ] use `#define`s instead of string substitution so that behavior is consistent with that of C preprocessor and so that types in LF files can be defined in terms of generics? (This was discussed multiple times and I vacillated but I think this will be the better choice)
- [ ] add more integration tests for generics

PR #1816, opened 2023-06-04
Fix bugs in name mangling

petervdonovan:
~~The assignment of mangled names is too complex when compiling several distinct programs in the same JAR execution because it causes the different programs to interact with each other by changing the number of reactors that have the same name. We can patch this up (by parameterizing the `uniqueName` function by the current `FileConfig`, for example), but it is safer and easier to just eliminate the global variables. It will make the generated code uglier, but maybe life is too short to worry about that.~~

This PR fixes the assignment of unique names mechanism so that it does not use global variables, which could result in interactions between multiple distinct compilation processes that affect the names used in generated code. This ensures that if one compilation only uses one reactor class with a certain name, then no numbers will be appended to the reactor class's name. It also appends the concrete types used to the reactor class's name, which may be useful for debugging.

The order in which reactor classes with the same name are numbered is based on the alphabetical ordering of the string representations of their URIs. Since different reactor classes with the same name cannot appear in the same file, I believe that this should work properly.

Here are some examples of unique names produced by the method implemented in this PR:
```
_superlong_double_char_int_float
_modal
_modal1
```

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/TypeParameterizedReactor.java
+++ b/core/src/main/java/org/lflang/generator/c/TypeParameterizedReactor.java
@@ -3,14 +3,15 @@ package org.lflang.generator.c;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
+
+import com.google.common.collect.ImmutableMap;
+import org.eclipse.emf.common.util.URI;
 import org.lflang.InferredType;
 import org.lflang.ast.ASTUtils;
 import org.lflang.generator.CodeBuilder;
-import org.lflang.lf.Instantiation;
-import org.lflang.lf.Reactor;
-import org.lflang.lf.Type;
-import org.lflang.lf.TypeParm;
+import org.lflang.lf.*;
 
 /**
  * A reactor class combined with concrete type arguments bound to its type parameters.
@@ -21,6 +22,8 @@ public class TypeParameterizedReactor {
   /** The type arguments associated with this particular variant of the reactor class. */
   private final Map<String, Type> typeArgs;
   private final List<String> typeParams;
+  private final ImmutableMap<String, Map<URI, Integer>> nameMap;
+
   /**
    * Construct the TPR corresponding to the given instantiation which syntactically appears within
    * the definition corresponding to {@code parent}.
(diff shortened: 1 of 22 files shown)


petervdonovan:
This PR includes some changes in CI that were necessary in order to make CI output more useful. I can move them into a separate PR if needed, but this is not a big PR anyway so maybe it is OK. Note that there is a commit cherry-picked between this and https://github.com/lf-lang/lingua-franca/pull/1814.

lhstrh:
> This PR includes some changes in CI that were necessary in order to make CI output more useful. I can move them into a separate PR if needed, but this is not a big PR anyway so maybe it is OK. Note that there is a commit cherry-picked between this and #1814.

I'm not sure I agree those CI changes are a good idea. What motives them?

petervdonovan:
- Currently the tests stop running when one test category fails. This prevents CI from uncovering all of the errors that it could uncover.
- Often I find that I am the only person submitting tasks to the GH actions queue. Therefore, it is in fact useful to increase the amount of parallelism in the workflow. With this change, many of the test categories which might fail complete their tests within 3 minutes.
- It is easy to find which test categories are failing by just looking at the list of jobs (on the left). Otherwise, you have to search the log, the interface to which is quite laggy.

petervdonovan:
I realize that if we want these changes, we would need to update the "required jobs" in the repo settings, which can be done.

petervdonovan:
The CI changes are too controversial and do not belong in this PR, so I reverted them. @lhstrh, I think we should merge this soon so that master does not continue to be broken.

core/src/main/java/org/lflang/generator/c/TypeParameterizedReactor.java
lhstrh, line 1: This file has some undocumented methods.

core/src/testFixtures/java/org/lflang/tests/TestBase.java
```diff
@@ -140,6 +140,7 @@ public static class Message {
     /* Descriptions of collections of tests. */
     public static final String DESC_SERIALIZATION = "Run serialization tests.";
     public static final String DESC_GENERIC = "Run generic tests.";
+    public static final String DESC_GENERICS = "Run generics tests.";
```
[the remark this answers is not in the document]
petervdonovan: Well, we already use "default" for "not Arduino nor Zephyr," so maybe we can use something else, like "miscellaneous."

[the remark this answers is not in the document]
lhstrh: How about "basic"?

core/src/main/java/org/lflang/generator/c/TypeParameterizedReactor.java
```diff
@@ -94,18 +145,13 @@ public InferredType resolveType(InferredType t) {
...
-    }
-    nameCounts.put(name, 1);
-    uniqueNames.put(this, "_" + name);
-    return uniqueName();
+  public String uniqueName() {
+    var resolved = ASTUtils.toDefinition(reactor);
+    return "_"
+        + uniqueName(resolved)
+        + typeParams.stream()
+            .map(it -> typeArgs.get(it).getId()) // FIXME: may be more than just an ID
```
lhstrh, line 153: Mysterious comment. What does it mean?

lhstrh (approved):
LGTM!

petervdonovan:
There is something wacky happening in the Rust tests. The test framework is throwing an NPE and I am not sure why.

petervdonovan:
OK, fixed. The problem was that there already was a test for generics, but it was called `runTypeParameterTests`.

core/src/main/java/org/lflang/generator/c/TypeParameterizedReactor.java
petervdonovan: It means that a type can consist of multiple words or can be escaped in `{= =}`. There is already an open issue for this: it is the first item listed in https://github.com/lf-lang/lingua-franca/issues/1765.

PR #1817, opened 2023-06-04
The 'threading' CLI option/build param/target property changed to 'single threaded'

### LINKED ISSUE #2080: Merging of target properties in federations is brittle and incomplete
The code in `FederateTargetConfig.update()` only performs the merging and path adjustment of the `files` target property; all other properties are ignored. This code needs thorough review and better testing.

patilatharva:
Todo:
- [x] Change docs accordingly.
- [x] Rename all occurrences of "unthreaded" to "single threaded".
- [x] Add validation check to ensure mutual exclusivity between `workers` and `single-threaded`
- [x] Return error upon specifying `single-threaded` on a federated program.

Also fixes #2080 in the sense that we now adjust paths for any target property that extends `FileListProperty`.


cmnrd:
I am not sure if "single threaded" is a good name. I expect that it is difficult for users to see the difference between `--single-threaded` and `--threads=1`. Perhaps `--unthreaded` would be clearer?

lhstrh:
> I am not sure if "single threaded" is a good name. I expect that it is difficult for users to see the difference between `--single-threaded` and `--threads=1`. Perhaps `--unthreaded` would be clearer?

What would actually happen if both `--single-threaded` and `--threads=1` are used?

core/src/main/java/org/lflang/generator/LFGeneratorContext.java
```diff
@@ -35,7 +35,7 @@ enum BuildParm {
     RUNTIME_VERSION("Specify the version of the runtime library used for compiling LF programs."),
     SCHEDULER("Specify the runtime scheduler (if supported)."),
     TARGET_COMPILER("Target compiler to invoke."),
-    THREADING("Specify whether the runtime should use multi-threading (true/false)."),
+    SINGLE_THREADED("Specify whether the runtime should use multi-threading (true/false)."),
```
petervdonovan, line 38: This makes it seem like you have to pass "true" or "false" as in `--single-threaded true`. But isn't the arity now zero?

cli/lfc/src/test/java/org/lflang/cli/LfcCliTest.java
```diff
@@ -293,7 +293,7 @@ public void testGeneratorArgs(@TempDir Path tempDir) throws IOException {
       "rs",
       "--scheduler",
       "GEDF_NP",
-      "--threading",
+      "--single-threaded",
       "false",
```
petervdonovan, line 297: I'm confused. Are we supposed to write "false" here?

patilatharva: Thanks for catching that - the "false" should have been deleted because single-threaded is now a switch. Will push a fix right now.

core/src/main/java/org/lflang/generator/LFGeneratorContext.java
patilatharva: That's true, I'd left it because the `SINGLE_THREADED` build param is still boolean (as opposed to the CLI option that now has zero-arity), but I'll remove the "true/false" part so it's consistent with the other boolean build params.

lhstrh (changes requested):
Looks good!

The absence of changes to the validator suggests that the combination of the `single-threaded` and `threads` target property isn't being checked for, which we should do, as @cmnrd suggested. Would you be able to fix this?

patilatharva:
> The absence of changes to the validator suggests that the combination of the `single-threaded` and `threads` target property isn't being checked for, which we should do, as @cmnrd suggested. Would you be able to fix this?

I just added checks in the validator and CLI to make the options `single-threaded` and `workers` mutually exclusive. To clarify, this makes the combination of `--single-threaded` and `--workers=1` illegal as well.

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
```diff
@@ -107,16 +108,22 @@ public class Lfc extends CliBase {
...
-      names = {"-t", "--threading"},
-      paramLabel = "<true/false>",
-      description = "Specify whether the runtime should use multi-threading" + " (true/false).")
-  private String threading;
+  /** Mutually exclusive options related to threading. */
+  static class ThreadingMutuallyExclusive {
+    @Option(
+        names = "--single-threaded",
+        arity = "0",
+        description = "Specify whether the runtime should be single-threaded.")
```
edwardalee, line 116: If the arity is zero, then this description is not correct. Should it be:
```suggestion
        description = "Specify that the runtime should be single-threaded.")
```

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
```diff
@@ -107,16 +108,22 @@ public class Lfc extends CliBase {
...
+    private boolean singleThreaded;
+
+    @Option(
+        names = {"-w", "--workers"},
+        description = "Specify the default number of worker threads.")
```
edwardalee, line 121: How does this specify the "default" number of worker threads? What will be the actual number of worker threads? Shouldn't the "default" number of worker threads be the number you get if you do _not_ specify this option?

test/C/src/arduino/BlinkAttemptThreading.lf
```diff
@@ -8,7 +8,7 @@ target C {
     name: "arduino",
     board: "arduino:avr:mega"
   },
-  threading: true
+  single-threaded: false
```
edwardalee, line 11: Should this instead be a `workers` property?

test/C/src/zephyr/threaded/UserThreads.lf
```diff
@@ -5,7 +5,7 @@ target C {
     name: Zephyr,
     user-threads: 3
   },
-  threading: true,
+  single-threaded: false,
```
edwardalee, line 8: Should this instead be a `workers` property?

edwardalee (changes requested):
This is nice! However, I'm not sure what `workers` means. It says it specifies the "default" number of workers, but I would think that the default number of workers would be what you get when you do _not_ specify the workers property. Also, we will now have an inconsistency with the website.  I suggest starting a `release-4` branch of the lf-website repo and update the documentation in this following locations:

https://www.lf-lang.org/docs/handbook/target-declaration#threading

https://www.lf-lang.org/docs/handbook/target-declaration#workers

https://www.lf-lang.org/docs/handbook/target-declaration#command-line-arguments

https://www.lf-lang.org/docs/handbook/target-language-details?target=c#single-threaded-implementation

https://www.lf-lang.org/docs/handbook/target-language-details?target=c#multithreaded-implementation

cmnrd:
> I just added checks in the validator and CLI to make it illegal to set workers when single-threaded is enabled. To clarify, this makes the CLI combination of --single-threaded and --workers=1 illegal, but in LF syntax single-threaded: false, workers: 2 is still allowed.

What is the reason for allowing single-threaded: false, workers: 2 in the target properties? Shouldn't this be illegal too?

core/src/main/java/org/lflang/generator/c/CPreambleGenerator.java
```diff
@@ -78,15 +78,10 @@ public static String generateDefineDirectives(
     //         targetConfig.clockSyncOptions
     //     ));
     // }
-    if (targetConfig.threading) {
-      targetConfig.compileDefinitions.put("LF_THREADED", "1");
-    } else {
+    if (targetConfig.singleThreaded) {
       targetConfig.compileDefinitions.put("LF_UNTHREADED", "1");
```
erlingrj, line 82: We should change this macro from `LF_UNTHREADED` to `LF_SINGLE_THREADED` in `reactor-c`.

test/C/src/arduino/BlinkAttemptThreading.lf
erlingrj: I think we can just remove the property also since threading is the default option?

test/C/src/zephyr/threaded/UserThreads.lf
erlingrj: I would just remove the property and use the default number of workers.

core/src/main/java/org/lflang/federated/extensions/CExtension.java
```diff
@@ -94,8 +94,8 @@ public void initializeTargetConfig(
     // Also, create the RTI C file and the launcher script.
     // Handle target parameters.
     // If the program is federated, then ensure that threading is enabled.
-    federate.targetConfig.threading = true;
-    federate.targetConfig.setByUser.add(TargetProperty.THREADING);
+    federate.targetConfig.singleThreaded = false;
+    federate.targetConfig.setByUser.add(TargetProperty.SINGLE_THREADED);
```
erlingrj, line 98: I know that this isn't introduced by this PR. But, specifying `single-threaded` on a federated program should, IMO, be an error condition, not be silently overridden.

erlingrj (approved):
Looks good, couple of comments. We should do some renaming in `reactor-c` also since "unthreaded" is used quite alot

patilatharva:
> What is the reason for allowing single-threaded: false, workers: 2 in the target properties? Shouldn't this be illegal too?

@cmnrd I was thinking it's technically correct to explicitly disable the `single-threaded` property when multiple workers are specified (even if the `single-threaded: false` is redundant), but do you think it's better design to just make the two properties mutually exclusive?

test/C/src/arduino/BlinkAttemptThreading.lf
patilatharva: True, I meant to preserve the original logic but in this case the `single-threaded: false` is just redundant. I'll remove the property.

test/C/src/zephyr/threaded/UserThreads.lf
patilatharva: Same as the other one - I'll remove the property.

cli/lfc/src/main/java/org/lflang/cli/Lfc.java
patilatharva: True, I'll remove the word "default" the CLI description because it's misleading.

After the reading the documentation, it seems like we meant to say the `workers` property in the LF target statement defines the "default" number of threads, as opposed to the `--workers` CLI option which overrides the former if supplied. So either way, the CLI option shown here shouldn't be called "default".
Source: https://www.lf-lang.org/docs/handbook/target-language-details?target=c#multithreaded-implementation

core/src/main/java/org/lflang/generator/c/CPreambleGenerator.java
patilatharva: True, I'll change all occurrences of "unthreaded" to "single threaded".

edwardalee (approved):
This looks good to me.  I have made some updates in the [release-5](https://github.com/lf-lang/lf-lang.github.io/pull/145) branch of the lf-lang website and requested your review to make sure I got it right. Thanks!

lhstrh:
@patilatharva, I did not yet merge this PR due to bigger changes that were happening in #2008. I will fix this PR after #2008 has been merged. (Expect force pushes on `single-threaded`.)

lhstrh (approved):

lhstrh:
I think this PR is in a good shape now, but there are still some federated test failures that I'm not sure what to do with.

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -1958,16 +1960,16 @@ protected void setUpGeneralParameters() {
...
-      if (targetConfig.get(ThreadingProperty.INSTANCE)
+      if (!targetConfig.get(SingleThreadedProperty.INSTANCE)
           && platformOptions.platform() == Platform.ARDUINO
           && (platformOptions.board() == null || !platformOptions.board().contains("mbed"))) {
         // non-MBED boards should not use threading
         messageReporter
             .nowhere()
             .info(
-                "Threading is incompatible on your current Arduino flavor. Setting threading to"
+                "Threading is incompatible plain (non-MBED) Arduino. Setting threading to"
```
erlingrj, line 1970:
```suggestion
                "Threading is incompatible on plain (non-MBED) Arduino. Setting threading to"
```

core/src/main/java/org/lflang/generator/c/CPreambleGenerator.java
```diff
@@ -87,13 +87,12 @@ public static String generateDefineDirectives(TargetConfig targetConfig, Path sr
     //         targetConfig.clockSyncOptions
     //     ));
     // }
-
-    if (targetConfig.get(ThreadingProperty.INSTANCE)) {
+    if (!targetConfig.get(SingleThreadedProperty.INSTANCE)) {
       definitions.put("LF_THREADED", "1");
     } else {
       definitions.put("LF_UNTHREADED", "1");
```
erlingrj, line 93: I guess these will have to be updated to work with the reactor-c changes?

erlingrj (commented):
It seems many of these changes here are not related to moving from `threading` to `single-threaded`. They seem to part of the general target-property updates? I did not look closely on them because I think I am lacking that context. Otherwise looks good. But we are still code-generating `LF_THREADED` and `LF_UNTHREADED` so I guess we should move to `LF_SINGLE_THREADED` to be compatible with the reactor-c PR?

core/src/main/java/org/lflang/generator/c/CPreambleGenerator.java
lhstrh: Ah, thanks for spotting this. Those changes got lost somehow. I'll reintroduce them.

lhstrh:
> It seems many of these changes here are not related to moving from `threading` to `single-threaded`. They seem to part of the general target-property updates? I did not look closely on them because I think I am lacking that context. Otherwise looks good. But we are still code-generating `LF_THREADED` and `LF_UNTHREADED` so I guess we should move to `LF_SINGLE_THREADED` to be compatible with the reactor-c PR?

Right, there is some overflow from #2008 that crept into this PR, but that's OK; they are closely related and I had to quite drastically alter the code in this PR to accommodate the changes in #2008.

erlingrj:
@lhstrh, there is a segfault in one of the federated Python programs, not reproducible on my end. Also very hard to see how the seg-fault in Python is due to doing this renaming

lhstrh:
> @lhstrh, there is a segfault in one of the federated Python programs, not reproducible on my end. Also very hard to see how the seg-fault in Python is due to doing this renaming

I also think it's unlikely that this is due to these changes in this PR. Maybe it's related to https://github.com/lf-lang/reactor-c/pull/293.

lhstrh:
The problem is that our `reactor-c` CI doesn't run the Python tests...so we don't find out that something is broken until we update the `reactor-c` submodule pointer in `lingua-franca`...

PR #1821, opened 2023-06-07
Changes to workflow dispatch

petervdonovan:
The current behavior in master is that workflows are skipped when manually triggered.

This PR also adds `merge-group` triggers to the main CI workflows in order to enable the merge queue that has now been activated on `master`.

### DIFF
--- a/.github/workflows/all-misc.yml
+++ b/.github/workflows/all-misc.yml
@@ -8,6 +8,7 @@ on:
       - master
   pull_request:
     types: [synchronize, opened, reopened, ready_for_review, converted_to_draft]
+  merge_group:
 
 env:
   # 2020.11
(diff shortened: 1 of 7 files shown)


petervdonovan:
@ChadliaJerad ran into an issue where `workflow_dispatch` (manually triggering workflows) does not work. I think that this change should fix the problem, although I don't think we can check until this is in `master`.

lhstrh (approved):
This looks good to me, although I never really understood what the use of the `all` input was. Let's try it and find out!

petervdonovan:
The `all` input is a vestige from the earlier version of the CI configuration in which some workflows were sometimes triggered by other workflows, and sometimes triggered by pull requests. Its purpose was to mark whether the workflow was triggered by another workflow. However, in our current configuration, it has had no purpose.

PR #1822, opened 2023-06-07
New `--check` flag for `lff`

cmnrd:
This implements a `--check` option for lff. In this check mode, lff reports all files that it would reformat as errors and only returns with a zero exit code if it would not change any files.


lhstrh (approved):
Looks fantastic! Thanks for adding the tests also :rocket:

PR #1829, opened 2023-06-08
Check to avoid repeated CI runs

lhstrh:
This PR intends to change the CI workflows in the following ways:
 - when a previous run can be found that already pass all the checks, checks are skipped
 - when `.txt` or `.md` files are checked in, we don't run tests, we just build (which includes doing a Spotless check)
 - a summary is added to let users better understand what is happening (and why)

We will probably have to get this merged in order to see whether it really works as expected...

### DIFF
--- a/.github/scripts/check-diff.sh
+++ b/.github/scripts/check-diff.sh
@@ -3,7 +3,7 @@ changes() {
 }
 
 if changes | grep -q $1; then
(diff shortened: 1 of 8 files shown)


cmnrd (commented):
Looks good, but I am not sure what exactly this does now. Could you provide a summary? I think it would be great to also document our entire CI setup somewhere.

PR #1830, opened 2023-06-08
Static scheduler

lsk567:
This is a PR for keeping track of the progress on a static scheduler, more concretely, a static schedule generator.


cmnrd:
@ChadliaJerad it looks like several commits from master got rebased and pushed to this branch and are now duplicated. Please use git merge to update a future branch or rebase the future branch on top of master (not the other way around). This way, we can keep a clean history.

ChadliaJerad:
> @ChadliaJerad it looks like several commits from master got rebased and pushed to this branch and are now duplicated. Please use git merge to update a future branch or rebase the future branch on top of master (not the other way around). This way, we can keep a clean history.

Yes! I was indeed wondering why are all these commits there.

lsk567:
Thanks @cmnrd for catching this issue. I just cherry-picked the relevant commits and force-pushed to this branch. The history should be clean now.

core/src/main/java/org/lflang/analyses/evm/Instruction.java
```diff
@@ -0,0 +1,40 @@
+package org.lflang.analyses.evm;
+
+public abstract class Instruction {
```
oowekyala, line 3: A suggestion to get rid of the boilerplate: use Kotlin :)

```kotlin
enum class OpCode {
    ADV,
    ADV2,
    BIT
}

// Sealed classes are abstract by default, and the compiler
// knows all their subtypes to allow for exhaustivity checking
// when pattern-matching.
sealed class EvmInstruction(val opcode: OpCode)

// Data classes get a nice toString generated
data class InstructionADV(
    val reactor: ReactorInstance,
    val nextTimeValue: TimeValue
) : EvmInstruction(OpCode.ADV2)

// Since this instruction has no fields it can be a singleton
object InstructionBIT : EvmInstruction(OpCode.BIT)

// ...
```

This entire package could fit in 100 lines in a single file :)

You can refer to Kotlin classes from Java as if they were regular Java classes. But if you want to stick to Java I would recommend making the supertype sealed (`public abstract sealed class Instruction`). ([doc](https://docs.oracle.com/en/java/javase/17/language/sealed-classes-and-interfaces.html#GUID-0C709461-CC33-419A-82BF-61461336E65F))

Also I would suggest giving this `Instruction` class a more specific name like `EvmInstruction` for clarity

core/src/main/java/org/lflang/analyses/statespace/EventQueue.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * An event queue implementation that sorts events by time tag order
+ *
+ * @author{Shaokai Lin <shaokai@berkeley.edu>}
+ */
+package org.lflang.analyses.statespace;
+
+import java.util.PriorityQueue;
+
+public class EventQueue extends PriorityQueue<Event> {
```
oowekyala, line 10: I would suggest not extending this class but hiding it inside. As it is you inherit also the `offer` method which inserts an element, but does not constrain uniqueness as your implementation claims.

core/src/main/java/org/lflang/analyses/statespace/StateSpaceNode.java
```diff
@@ -0,0 +1,103 @@
...
+          != n2.eventQ.get(i).tag.timestamp - n2.tag.timestamp) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /** Two methods for pretty printing */
+  public void display() {
+    System.out.println("(" + this.time + ", " + reactionsInvoked + ", " + eventQ + ")");
```
oowekyala, line 48:
```suggestion
    System.out.println(this);
```
This will call toString

core/src/main/java/org/lflang/analyses/statespace/StateSpaceNode.java
```diff
@@ -0,0 +1,103 @@
...
+
+    // Generate hash for the time differences.
+    List<Long> timeDiff =
+        this.eventQ.stream()
+            .map(
+                e -> {
+                  return e.tag.timestamp - this.tag.timestamp;
+                })
+            .collect(Collectors.toList());
+    result = 31 * result + timeDiff.hashCode();
```
oowekyala, line 99:
```suggestion
    long timeDiffHash =
        this.eventQ.stream()
            .mapToLong(e -> e.tag.timestamp - this.tag.timestamp)
            .reduce(1, (a, b) -> 31 * a + b);
    result = 31 * result + (int) timeDiffHash;
```

core/src/main/java/org/lflang/analyses/statespace/StateSpaceNode.java
```diff
@@ -0,0 +1,103 @@
...
+    // Generate hash for the reactions invoked.
+    result = 31 * result + reactionsInvoked.hashCode();
+
+    // Generate hash for the triggers in the queued events.
+    List<String> eventNames =
+        this.eventQ.stream()
+            .map(Event::getTrigger)
+            .map(TriggerInstance::getFullName)
+            .collect(Collectors.toList());
+    result = 31 * result + eventNames.hashCode();
```
oowekyala, line 89:
```suggestion
    int eventsHash =
        this.eventQ.stream()
            .map(Event::getTrigger)
            .map(TriggerInstance::getFullName)
            .mapToInt(Object::hashCode)
            .reduce(1, (a, b) -> 31 * a + b);
    result = 31 * result + eventsHash;
```

core/src/main/java/org/lflang/analyses/statespace/Tag.java
```diff
@@ -0,0 +1,65 @@
+/**
+ * Class representing a logical time tag, which is a pair that consists of a timestamp (type long)
+ * and a microstep (type long).
+ *
+ * @author{Shaokai Lin <shaokai@berkeley.edu>}
+ */
+package org.lflang.analyses.statespace;
```
oowekyala, line 7: Javadoc comments should be before the class declaration, not before the package declaration. As it is these comments are dangling and will be ignored by the javadoc engine, including IDEs.

core/src/main/java/org/lflang/analyses/statespace/Tag.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.lflang.TimeValue;
+
+public class Tag implements Comparable<Tag> {
```
oowekyala, line 11: This class looks like it could be a record
```java
public record Tag(long timestamp, long microstep, boolean forever)
   implements Comparable<Tag> { // ...
```

This would spare you the fields and ctor declaration, and the equals declaration.

core/src/main/java/org/lflang/analyses/dag/Dag.java
```diff
@@ -0,0 +1,319 @@
...
+      bufferedReader = new BufferedReader(fileReader);
+    } catch (IOException e) {
+      System.out.println("Problem accessing file " + dotFileName + "! " + e);
+      return false;
+    }
+
+    String line;
+
+    // Pattern with which an edge starts:
+    Pattern pattern = Pattern.compile("^((\s*)(\\d+)(\s*)->(\s*)(\\d+))");
```
oowekyala, line 277: Constant patterns should be put into constant fields to avoid recompiling them

core/src/main/java/org/lflang/analyses/dag/Dag.java
```diff
@@ -0,0 +1,319 @@
...
+   */
+  public boolean updateDag(String dotFileName) throws IOException {
+    FileReader fileReader;
+    BufferedReader bufferedReader;
+    // Read the file
+    try {
+      fileReader = new FileReader(dotFileName);
+      // Buffer the input stream from the file
+      bufferedReader = new BufferedReader(fileReader);
+    } catch (IOException e) {
```
oowekyala, line 269: These should be put in a try-with-resources (TWR). The close method should be in a finally block, which isn't necessary with a TWR.

```java
try (BufferedReader bufferedReader = Files.newBufferedReader(Paths.get(dotFileName))) {
 // your code here
} catch (IOException e) {
  // your handler here
}
// no need to close explicitly
```

The exception handler doesn't look useful either. I think you should let the IOException propagate to the caller and let the caller handle it in a more sensible way. You can just omit the catch block of your TWR to do that. The boolean return value of this method would not be needed.

core/src/main/java/org/lflang/analyses/dag/Dag.java
```diff
@@ -0,0 +1,319 @@
...
+    Matcher matcher;
+
+    // Before iterating to search for the edges, we clear the DAG edges array list
+    this.dagEdges.clear();
+
+    // Search
+    int i = 0;
+    while (bufferedReader.ready()) {
+
+      line = bufferedReader.readLine();
```
oowekyala, line 287: Looping over a buffered reader should not be done by polling `ready`.
```suggestion
    while ((line = bufferedReader.readLine()) != null) {
```

core/src/main/java/org/lflang/analyses/dag/Dag.java
```diff
@@ -0,0 +1,319 @@
...
+
+        // Get the source and sink nodes ids and add the edge
+        try {
+          srcNodeId = Integer.parseInt(st.nextToken());
+          sinkNodeId = Integer.parseInt(st.nextToken());
+          this.addEdge(srcNodeId, sinkNodeId);
+        } catch (NumberFormatException e) {
+          System.out.println("Parse error in line " + line + " : Expected a number!");
+          Exceptions.sneakyThrow(e);
+        }
```
oowekyala, line 313: Actually it occurs to me that this whole int extraction logic could be replaced by using the capture groups of the matcher. Consider rewriting your pattern
```java
"^\\s*(\\d+)\\s*->\\s*(\\d+)"
```
to remove useless capture groups. Notice also that we use `\\s` and not `\s` which is a string escape and not what we want. Then after `matcher.find()` all you need is to write
```java
int srcNodeId = Integer.parseInt(matcher.group(1));
int sinkNodeId = Integer.parseInt(matcher.group(2));
this.addEdge(srcNodeId, sinkNodeId);
```
This also makes it clear that what you're giving to your `Integer.parseInt` is a sequence of digits so you don't need to catch a NumberFormatException, that will never occur.

oowekyala (commented):
A couple of comments to improve code quality

core/src/main/java/org/lflang/analyses/evm/Instruction.java
lsk567: Thanks for the mini Kotlin tutorial here, @oowekyala ! Yes, Kotlin looks quite interesting. I will try to reimplement the instructions using your solution.

core/src/main/java/org/lflang/analyses/statespace/Tag.java
lsk567: Ah, I see. Thanks for the explanation.

lsk567:
Thanks for your detailed review, @oowekyala! I will work on the issues you pointed out and try to make the implementation more robust.

lhstrh:
> @ChadliaJerad it looks like several commits from master got rebased and pushed to this branch and are now duplicated. Please use git merge to update a feature branch or rebase the feature branch on top of master (not the other way around). This way, we can keep a clean history.

Looks like there is a similar problem with duplicate commits in https://github.com/lf-lang/lingua-franca/pull/1695.

lsk567:
It seems like the Draft status prevents the CI tests from being run. To invoke CI, I will mark the PR as ready for review.

lsk567:
Just performed a huge rebase to this branch. Now this branch adopts the latest target property registration workflow. There is an unresolved problem where nested dictionaries in target properties do not seem to get loaded. Therefore, at commit [1e236c9](https://github.com/lf-lang/lingua-franca/pull/1830/commits/1e236c951b0b89b832c326cabb89522d4ec677b4), the `MOCASIN` scheduler cannot be used and the `EGS` scheduler can only be invoked at the command line using `--static-scheduler EGS`.

PR #1831, opened 2023-06-09
Raspberry Pi Pico target support

gundralaa:
This PR adds (single-threaded only) support for the RP2040 platform.

- [x] modify main method
- [x] cmake generator
- [x] fix incorrect platform npe
- [x] vscode support

Running `lfc` runs the default `cmake` directive and populates the bin directory with an elf binary for the downloading to the RPi 2040..

Multi-threaded support is planned to be added later.


core/src/main/java/org/lflang/generator/CodeBuilder.java
```diff
@@ -68,7 +68,7 @@ public int length() {
 
   /** Add a new line. */
   public void newLine() {
-    this.pr("");
+    this.pr("\n");
```
lhstrh, line 71: This looks a little sus. Wouldn't this result in two newlines?

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -930,6 +930,16 @@ protected void copyTargetFiles() throws IOException {
       FileUtil.copyFileFromClassPath(
           "/lib/platform/zephyr/Kconfig", fileConfig.getSrcGenPath(), true);
     }
+
+    // For the pico src-gen, copy over vscode configurations for debugging
+    if (targetConfig.platformOptions.platform == Platform.RP2040) {
+      Path vscodePath = Path.of(fileConfig.getSrcGenPath() + File.separator + ".vscode");
```
lhstrh, line 936:
```suggestion
      Path vscodePath = fileConfig.getSrcGenPath().resolve(".vscode");
```

core/src/main/java/org/lflang/TargetConfig.java
```diff
@@ -383,7 +383,9 @@ public static class PlatformOptions {
     /**
      * The string value used to determine what type of embedded board we work with and can be used
      * to simplify the build process. For example, when we want to flash to an Arduino Nano 33 BLE
-     * board, we can use the string arduino:mbed_nano:nano33ble
+     * board, we can use the string arduino:mbed_nano:nano33ble TODO: move to lingo Can also be used
```
lhstrh, line 386: We generally want to move stuff like this to `lingo`, so let's not leave arbitrary `TODOs` related to this in the code.

lhstrh (commented):
This generally looks good. CI won't run because of a conflict on the submodule. Please fix that by merging in `master` and pinning the submodule to the appropriate ref.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -153,6 +168,8 @@ CodeBuilder generateCMakeCode(
             + " gcc\")");
     cMakeCode.pr("  endif()");
     cMakeCode.pr("endif()");
+    // remove warnings for making building easier
+    cMakeCode.pr("set(CMAKE_C_FLAGS \"${CMAKE_C_FLAGS} -w\")");
```
petervdonovan, line 172: I have two concerns with this:
- I do not think it is portable: One big reason to use CMake is to allow to build with different compilers. `CMAKE_C_FLAGS` bypasses CMake to pass arguments directly to the underlying C compiler. Will this work with MSVC?
- Wrong code should look wrong (e.g. by producing warnings). I know that our code base produces warnings, but lots of effort has gone into reducing the number of warnings and probably more effort should continue to go into eliminating the warnings if we want the code to work reliably.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -230,6 +257,27 @@ CodeBuilder generateCMakeCode(
     cMakeCode.pr("target_include_directories(${LF_MAIN_TARGET} PUBLIC include/core/modal_models)");
     cMakeCode.pr("target_include_directories(${LF_MAIN_TARGET} PUBLIC include/core/utils)");
 
+    // post target definition board configurations
+    if (targetConfig.platformOptions.board != null) {
+      switch (targetConfig.platformOptions.platform) {
+        case RP2040:
+          String[] bProps = targetConfig.platformOptions.board.split(":");
```
petervdonovan, line 264: This looks like ad hoc parsing. Will it silently produce wrong behavior if there is a typo, or if there is whitespace where whitespace is not expected?

core/src/main/java/org/lflang/generator/CodeBuilder.java
gundralaa: Ya that's fair. The method was essentially not doing anything due to the empty string. I changed the pr method to reflect the proper usage.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
gundralaa: Did some whitespace removal and placed reasonable defaults in case of a formatting mismatch.

core/src/main/java/org/lflang/generator/CodeBuilder.java
petervdonovan: Are you sure that it was not do anything? My understanding is that the `pr` method adds a newline after the argument that you pass to it. See line 88 of this file.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
gundralaa: This is fair but debugging application code with the warnings on is currently very difficult. I think it would be beneficial for students to not have to parse through them. I've moved the directive to only insert for the rp2040 platform which shouldn't mess with the portability. This is only a temporary fix for usability.

core/src/main/java/org/lflang/TargetConfig.java
```diff
@@ -389,7 +389,9 @@ public static class PlatformOptions {
     /**
      * The string value used to determine what type of embedded board we work with and can be used
      * to simplify the build process. For example, when we want to flash to an Arduino Nano 33 BLE
-     * board, we can use the string arduino:mbed_nano:nano33ble
+     * board, we can use the string arduino:mbed_nano:nano33ble Can also be used to specify uart
+     * output setting on rp2040 boards where arduino_nano_rp2040_connect:uart or
+     * arduino_nano_rp2040_connect:usb (usb) fully specifies the board and uart output
```
edwardalee, line 394: This is unclear. First, the string arduino:mbed_nano:nano33ble is probably not right in a target property because it will not be valid syntax (too many colons). Also, isn't this specifying where stdout is directed rather than a "uart output setting"?

core/src/main/java/org/lflang/TargetProperty.java
```diff
@@ -465,6 +465,13 @@ public enum TargetProperty {
           config.platformOptions = new PlatformOptions();
           config.platformOptions.platform =
               (Platform) UnionType.PLATFORM_UNION.forName(ASTUtils.elementToSingleString(value));
+          if (config.platformOptions.platform == null) {
+            String s =
+                "Unidentified Platform Type, LF supports the following platform types: "
+                    + Arrays.asList(Platform.values()).toString();
+            err.at(value).error(s);
+            throw new AssertionError(s);
```
edwardalee, line 473: This should be checked in the validator.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -116,25 +116,51 @@ CodeBuilder generateCMakeCode(
...
+        cMakeCode.pr("include(pico_sdk_import.cmake)");
+        cMakeCode.pr("project(" + executableName + " LANGUAGES C CXX ASM)");
+        cMakeCode.newLine();
+        // board type for rp2040 based boards
+        if (targetConfig.platformOptions.board != null) {
+          // board syntax <board_name> : <stdio_opt>
+          // ignore whitespace
+          String[] bProps = targetConfig.platformOptions.board.trim().split(":");
+          for (int i = 0; i < bProps.length; i++) {
+            bProps[i] = bProps[i].trim();
```
edwardalee, line 148: All but the first of these bProps is ignored here. Is that right?

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -230,6 +266,30 @@ CodeBuilder generateCMakeCode(
...
+            cMakeCode.pr("pico_enable_stdio_uart(${LF_MAIN_TARGET} 0)");
+          }
+        } else {
+          // default
+          cMakeCode.pr("# Enable both usb and uart stdio");
+          cMakeCode.pr("pico_enable_stdio_usb(${LF_MAIN_TARGET} 1)");
+          cMakeCode.pr("pico_enable_stdio_uart(${LF_MAIN_TARGET} 1)");
+        }
+        break;
+    }
```
edwardalee, line 291: Why isn't this done above instead of ignoring the additional options?

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -397,4 +458,35 @@ private static String setUpMainTargetZephyr(
     code.newLine();
     return code.toString();
   }
+
+  private static String setUpMainTargetRp2040(
```
edwardalee, line 462: Could this also set PICO_SDK_PATH?

core/src/main/java/org/lflang/generator/c/CMainFunctionGenerator.java
```diff
@@ -62,6 +62,9 @@ private String generateMainFunction() {
           "   int res = lf_reactor_c_main(0, NULL);",
           "   exit(res);",
           "}");
+    } else if (targetConfig.platformOptions.platform == Platform.RP2040) {
+      // Pico platform cannont use command line args.
```
edwardalee, line 66:
```suggestion
      // Pico platform cannot use command line args.
```

edwardalee (changes requested):
This is definitely getting closer, but we need to resolve PICO_SDK_PATH setting, I think.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -136,6 +137,22 @@ CodeBuilder generateCMakeCode(
         cMakeCode.newLine();
         break;
       case RP2040:
+        // Attempt to set PICO_SDK_PATH if it is not already set.
```
lhstrh, line 140: I'm not sure that this is the right approach. I don't think should be up to `lfc` to guess where the Pico SDK may be installed. The fact that an environment variable is used at all suggests that this it should be up to the user to specify it. I see that this code provides the convenience of doing that if the user forgets this, but it addresses only a single corner case, and it reinforces the idea that the output directory should contain copies of library code, which we're starting to realize has serious drawbacks. Is this all because we don't want to instruct users of the `rpi3+` template to invoke `nix develop`? Note that this also provides access to tools like `minicom`, which if you don't run `nix develop` will not be on your `$PATH`.
Perhaps we can do something more generic (and adopt the same approach for `lingo`), which is that if the package has a nix configuration, we automatically run `nix develop`, capture the environment variables, and use that environment for subsequent invocations of commands?

edwardalee: `nix develop` will have to be run every time you start a new session. It launches a new shell. It does not configure the user environment.  Thus, with that approach, every lab exercise instructions will have to be modified to remind the user to run `nix develop` before they do anything else.

If you have a better approach, feel free to modify this and also modify the instructions in `lf-embedded-lab` (Installation.md, Tools.md, and, if necessary, each lab exercise).  I have spent way too much time already on configuration issues with this, and if I spend more time on them, the labs themselves have no chance of being ready in time for Fall semester.

Also, note that we are not using `minicom`, which doesn't work for me (it requires a mysterious "meta" key, which doesn't exist). We are using `screen` instead.

lhstrh: > `nix develop` will have to be run every time you start a new session. It launches a new shell. It does not configure the user environment. I have spent way too much time already on configuration issues with this, and if I spend more time on them, the labs themselves have no chance of being ready in time for Fall semester.

Doesn't it configure the user environment _by launching a new shell?_
 
> If you have a better approach, feel free to modify this and also modify the instructions in `lf-embedded-lab` (Installation.md, Tools.md, and, if necessary, each lab exercise).

I plan to bring this up during the packaging meeting and will circle back. I'm happy to help with this!

> Also, note that we are not using `minicom`, which doesn't work for me (it requires a mysterious "meta" key, which doesn't exist). We are using `screen` instead.

Totally fine to use `screen` instead, we should just update the `nix` configuration to pull in that instead of `minicom` then. @gundralaa could you take care of this?

edwardalee: > > `nix develop` will have to be run every time you start a new session. It launches a new shell. It does not configure the user environment. I have spent way too much time already on configuration issues with this, and if I spend more time on them, the labs themselves have no chance of being ready in time for Fall semester.
> 
> Doesn't it configure the user environment _by launching a new shell?_

In the new shell, you have environment variables set. If you log out and log back in, those environment variables are gone. You have to run `nix develop` again, and then do everything in the shell it launches.

> > Also, note that we are not using `minicom`, which doesn't work for me (it requires a mysterious "meta" key, which doesn't exist). We are using `screen` instead.
> 
> Totally fine to use `screen` instead, we should just update the `nix` configuration to pull in that instead of `minicom` then. @gundralaa could you take care of this?

I thought I did that already... ???

lhstrh: My proposal is to add a target property called `pre-build`, which is like `build` but the provided command will be executed before building, and the environment of the pre-build command will be used when invoking the target tool chain (or build command). The `.lf` files in the rp3+ project would just specify `pre-build: nix develop` and they would just compile using `lfc`. I can go ahead and implement this if it sounds OK.

gundralaa: I've thought about this, and I think that it is fair to assert using nix does require extra steps for the user but, in most cases, once you run nix develop, any terminal multiplexers or ides spawned from the nix shell inherit the environment variables. For example, spawning vscode using ``code .`` from a nix shell on ubuntu, which the students will most likely be using, keeps the applications bindings such that the vscode plugins use the nix versions of the packages like cmake.

The usage of nix for this project was to provide a way for users to easily get started without installing dependencies.  That being said, there aren't many dependencies and if someone doesn't want to use nix I think its very reasonable to direct them to setting the ``PICO_SDK_PATH`` manually in their own shell env and installing packages manually. To support this case, the setup documentation for the pico-platform needs to robust.

I personally don't think there is any further work that needs to be done here and adding new properties would only bloat the existing compiler. Use nix or dont it doesn't really matter, but lfc in my opinion should focus on code generation tasks instead of searching for dependencies.

lhstrh: I generally agree with you, @gundralaa, which also motivated my initial comment. The idea behind `pre-build` is that it would also be useful for other external configurations, beyond `nix`, but I agree that it isn't a necessity but a "nice-to-have."

edwardalee: Pre-build feels like an epicycle to me.  I would rather just use build, like we did with the nrf52.  And running nix every time also feels wrong to me.  It will slow down the compile, and you will likely get an ever growing number of nested shells running.

I really don't see what wrong with having lfc deal with what will be a common usage pattern that eliminates the need for an environment variable.  And it's easy to override by just setting the environment variable.

lhstrh: > Pre-build feels like an epicycle to me. I would rather just use build, like we did with the nrf52. And running nix every time also feels wrong to me. It will slow down the compile, and you will likely get an ever growing number of nested shells running.

Running `nix develop` after the initial install runs really fast (.5 second on my machine). We can also just have `pre-build: configure-nix.sh`, which will be even faster. Either way, this won't lead to an ever growing number of nested shells (running a command in Java does not affect the shell from which `lfc` is invoked).

> I really don't see what wrong with having lfc deal with what will be a common usage pattern that eliminates the need for an environment variable. And it's easy to override by just setting the environment variable.

Dealing with these kinds of corner cases bloats the compiler. Such code will eventually not be used or break as common usage patterns change.

lhstrh: More specifically, my concern is that the current fix deals with a single _corner case_ only, not with a _usage pattern_. The `pre-build` feature would be useful for a lot more than just setting this specific environment variable.

A build tool/package manager like `npm` also has user-specifiable commands like `install` or `pre-publish` to hook in stuff that needs to happen as part of building/packaging tasks that the tool handles, so I would argue that something like this is pretty common.

lhstrh: Update on this: getting `pre-build` to work was easy, but not in combination with `nix develop`; indeed, because it spawns a new shell...

lhstrh: It would allow one to execute something ahead of a normal build sequence that is not completely taken over by the user. If you use `build`, then the user is fully in charge of building (this is like running with `--no-compile` and handling the build steps externally).

As @gundralaa pointed out, there is no _need_ for any of this, but the point is provide a convenience that not just there to handle a single corner case but has broader utility.

edwardalee: I would like to see the difference between a pre-build script and a build script. Is the diff more than one line?

lhstrh: That depends on what's in the build script? My point is that there doesn't have to be a build script at all. I would say the `pre-build` is only useful when no `build` is specified. They can be mutually exclusive.

lhstrh: I abandoned my changes because it turned out to be difficult to generalize this.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -247,12 +327,13 @@ CodeBuilder generateCMakeCode(
     }
 
     if (targetConfig.threading || targetConfig.tracing != null) {
-      // If threaded computation is requested, add the threads option.
-      cMakeCode.pr("# Find threads and link to it");
-      cMakeCode.pr("find_package(Threads REQUIRED)");
-      cMakeCode.pr("target_link_libraries(${LF_MAIN_TARGET} PRIVATE Threads::Threads)");
-      cMakeCode.newLine();
-
+      if (targetConfig.platformOptions.platform != Platform.RP2040) {
```
lhstrh, line 330: I think a check like this should be done earlier, during validation. Specifically, if the platform is `RP2040`, then `threading` must not be `true` (or an error should result).

core/src/main/java/org/lflang/TargetProperty.java
lhstrh: This pattern was copied from other target properties. I see why, because it's currently difficult to do it the "correct" way, and it introduces a bunch of code duplication. I think it might be worth to try and improve this situation... I'm currently trying to chip away at that a little bit...

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
lhstrh: I've pushed a fix, which works, but I'm not very happy about how it works. I'll try to push improvements at a later time...

core/src/main/java/org/lflang/TargetProperty.java
lhstrh: I will push my refactoring of this to another branch. Let's merge this PR first.

lhstrh (approved):
I think this is good to go!

Issue #1834, opened 2023-06-09
fedsd does not work with Pandas 2.0


edwardalee:
The `fedsd` utility for generating sequence diagrams fails with Pandas 2.0 with the following error:
```
Building the communication diagram for the following trace files: federate__c.lft federate__i.lft federate__t.lft rti.lft in trace_svg.html
Traceback (most recent call last):
  File "/Users/eal/lingua-franca-intellij/util/tracing/visualization/fedsd.py", line 137, in <module>
    trace_df = trace_df.append(fed_df, sort=False, ignore_index=True)
               ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pandas/core/generic.py", line 5989, in __getattr__
    return object.__getattribute__(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DataFrame' object has no attribute 'append'. Did you mean: '_append'?
```
According to [this page](https://stackoverflow.com/questions/75956209/dataframe-object-has-no-attribute-append), append has been removed and must be replaced with concat.

Issue #1818, opened 2023-06-04
Timeout in EnclaveBank


petervdonovan:
The following was observed in a branch that (to my knowledge) only made changes relevant to the C target. See [this run](https://github.com/lf-lang/lingua-franca/actions/runs/5171546665/jobs/9315144565?pr=1814) for details.

```
Failed: src\enclave\EnclaveBank.lf
    -----------------------------------------------------------------------------
    Reason: Test timed out.
    Compilation output:
    Generating code for: file:/D:/a/lingua-franca/lingua-franca/test/Cpp/src/enclave/EnclaveBank.lf
    ******** mode: STANDALONE
    ******** generated sources: D:\a\lingua-franca\lingua-franca\test\Cpp\src-gen\enclave\EnclaveBank
    Path: D:\a\lingua-franca\lingua-franca\test\Cpp\src-gen\enclave\EnclaveBank D:\a\lingua-franca\lingua-franca\test\Cpp\src-gen\enclave\EnclaveBank
    --- Current working directory: D:\a\lingua-franca\lingua-franca\test\Cpp\build
    --- Executing command: cmake --version
    cmake version 3.26.3

    CMake suite maintained and supported by Kitware (kitware.com/cmake).
    --- Current working directory: D:\a\lingua-franca\lingua-franca\test\Cpp\build
    --- Executing command: cmake -DCMAKE_BUILD_TYPE=Test -DCMAKE_INSTALL_PREFIX=D:/a/lingua-franca/lingua-franca/test/Cpp -DCMAKE_INSTALL_BINDIR=bin/enclave -DREACTOR_CPP_VALIDATE=ON -DREACTOR_CPP_PRINT_STATISTICS=OFF -DREACTOR_CPP_TRACE=OFF -DREACTOR_CPP_LOG_LEVEL=4 -DLF_SRC_PKG_PATH=D:\a\lingua-franca\lingua-franca\test\Cpp D:/a/lingua-franca/lingua-franca/test/Cpp/src-gen
    -- Building for: Visual Studio 17 2022
    -- The CXX compiler identification is MSVC 19.35.32217.1
    -- Detecting CXX compiler ABI info
    -- Detecting CXX compiler ABI info - done
    -- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.35.32215/bin/HostX64/x64/cl.exe - skipped
    -- Detecting CXX compile features
[...]

PR #1836, opened 2023-06-09
Fedsd compatibility with pandas 2.0

ChadliaJerad:
This PR fixes [issue #1834](https://github.com/lf-lang/lingua-franca/issues/1834).
It also makes `fedsd` correctly match non-tagged messages, which enables the visualization of exchanged messages within physical connections.

### DIFF
--- a/util/tracing/visualization/fedsd.py
+++ b/util/tracing/visualization/fedsd.py
@@ -51,15 +51,6 @@ parser.add_argument('-f','--federates', nargs='+', action='append',
                     help='List of the federates csv trace files.')
 
 
-''' Clock synchronization error '''
-''' FIXME: There should be a value for each communicating pair '''
-clock_sync_error = 0
-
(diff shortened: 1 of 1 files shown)


edwardalee (approved):
LGTM! Thanks for fixing this.

edwardalee (approved):
LGTM! Thanks for fixing this.

As usual, there are test failures that have nothing to do with this PR. Not sure what to do about that.

petervdonovan:
[The failing C++ test](https://github.com/lf-lang/lingua-franca/actions/runs/5226254458/jobs/9436713778?pr=1836) seems the same as #1818.

It is possible that the failing LSP test might have something to do with compiling in parallel such that the output gets interleaved and garbled. But I'm not sure, and this is awkward for me to reproduce since it seems like I do not have the same compiler on my system as is in CI.

I'm also not sure what our policy should be when this happens.

Issue #1464, opened 2022-11-07
Microstep alignment in logical actions


petervdonovan:
This is not so much a bug report as a question about whether the semantics as currently designed is what we really want.

We currently allow scheduled actions to pile up in the microstep dimension instead of having a "last write wins" policy like what we have for ports. The question is about whether this asymmetry between ports and actions is really necessary.

The purpose of microsteps might be
1. to prevent scheduled events from being dropped just because they conflict with other events;
1. to allow many iterations of a program to happen as quickly as possible without waiting for physical time to elapse;
1. to increase parallelism by pipelining several consecutive operations without waiting for physical time to elapse;
1. to make a conceptually fork-join program as tree-shaped as possible so that chain IDs are seldom necessary. (I believe that one of our key claims is not merely that we have determinism (other frameworks provide that) but that we have determinism in the presence of arbitrary topologies; however, I do not totally understand the benefit when you can always just schedule an event one microstep later.)

So, how useful is use case #1? Can we dispense with it?

Here is why I ask. The piling up of events in the microstep dimension can make it difficult to predict how different scheduled actions will align with each other. The logical time of the scheduled event now depends not only on the current time, the min delay, and the additional delay, but the state of the event queue as well. Furthermore, an approach to determining maximum rates at which logical actions are present that is based on the set of times when they _can_ be present becomes more complicated: If you do not know which/how many microsteps there are when an action can be present, the number of times the action can be present in a time interval is unbounded. I will try to follow up later with more details on the types of conclusions that you cannot draw because of this, but it is a work in progress...

Example program:
```
target C

main reactor {
    logical action a: int
    logical action b: int
    reaction(startup) -> a {=
        lf_schedule_int(a, 0, 0);
    =}
    reaction(startup) -> a, b {=
        // Based on this reaction body, it looks like a and b will be logically simultaneous,
        // but in fact a will be processed a microstep later. The user must know the state
        // of the event queue in order to realize this misalignment.
        lf_schedule_int(a, 0, 1);
        lf_schedule_int(b, 0, 1);
    =}
    reaction(a, b) {=
        if (a->is_present) {
            lf_print("a is present with value %d", a->value);
        }
        if (b->is_present) {
            lf_print("b is present with value %d", b->value);
        }
        lf_print("----------------------------------");
    =}
}
```

Actual output:
```
---- Start execution at time Mon Nov  7 13:43:44 2022
---- plus 87511666 nanoseconds.
---- Using 1 workers.
[...]

PR #1837, opened 2023-06-10
Top level multiports for `remove-unordered-reactions`

petervdonovan:
This is still a work in progress. I expect to be working in this branch for some more days. It is not ready for review.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
@@ -316,15 +316,19 @@ public class FedASTUtils {
     // Establish references to the involved ports.
     sourceRef.setContainer(connection.getSourcePortInstance().getParent().getDefinition());
     sourceRef.setVariable(connection.getSourcePortInstance().getDefinition());
-    destRef.setContainer(connection.getDestinationPortInstance().getParent().getDefinition());
-    destRef.setVariable(connection.getDestinationPortInstance().getDefinition());
+    destRef.setContainer(
+        connection
+            .getDstFederate()
+            .networkPortToIndexer
+            .get(connection.getDestinationPortInstance()));
+    var v = LfFactory.eINSTANCE.createVariable();
+    v.setName("port" + connection.getDstChannel());
+    destRef.setVariable(v);
     instRef.setContainer(networkInstance);
     instRef.setVariable(out);
 
     out.setName("msg");
     out.setType(type);
-    out.setWidthSpec(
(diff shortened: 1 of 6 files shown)


petervdonovan:
17/57 tests passing now. A few days ago only 5 or 10 tests were passing, so maybe this is progress.

Right now ChainWithDelay is failing because the InternalDelay reactor, `i`, runs out of work on its earliest level before it has had a chance to enqueue the network receivers, which are inserted into the current level in response to a tagged message received over the network. The workers think that they are already done with the level, so the network receiver for time 0 just doesn't get executed. I am still working on a fix for this.

petervdonovan:
Now [`DistributedNetworkOrder`](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/DistributedNetworkOrder.lf) is failing, but I'm not sure that this is a test that we should keep. It tests an internal API, but the real reason why I am not sure we should keep it is that it tests a behavior that we plan to remove: https://github.com/lf-lang/lingua-franca/issues/1464

petervdonovan:
### Status update:

Programs with physical actions deadlock. For example:
- DistributedCountPhysical deadlocks
- DistributedCountPhysicalAfterDelay also deadlocks
- DistributedLoopedPhysicalAction deadlocks on termination
- DistributedPhysicalActionUpstream deadlocks, probably on termination

Programs that use decentralized coordination seem to deadlock. For example:
- DistributedToken (which uses decentralized coordination) seems to deadlock on termination
- PhysicalSTP (decentralized) fails. Message violates STP offset. Deadlock
- STPParameter (decentralized) deadlocks on shutdown

Programs with federate-level cycles all seem to deadlock

StopAtShutdown deadlocks on termination and might be the only centralized program without federate-level cycles that deadlocks

lhstrh:
> Now [`DistributedNetworkOrder`](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/DistributedNetworkOrder.lf) is failing, but I'm not sure that this is a test that we should keep. It tests an internal API, but the real reason why I am not sure we should keep it is that it tests a behavior that we plan to remove: #1464

I agree that it seems reasonable to just remove this test...

petervdonovan:
### Status update:

- 43/57 C federated tests are passing
- I think that all of the tests that are currently failing involve federate-level cycles

lhstrh:
> ### Status update:
> 
>     * 43/57 C federated tests are passing
> 
>     * I think that all of the tests that are currently failing involve federate-level cycles

Wow, that's really encouraging!

Issue #1832, opened 2023-06-09
Bugs in formatter


cmnrd:
I applied the formatter to the examples repo today (https://github.com/lf-lang/examples-lingua-franca/pull/48). Interestingly, the CI formatting failed after I have correctly applied lff to all files. I investigated this found multiple issues with our formatter:

- In [HelloWorldPhysicalAfter.lf](https://github.com/lf-lang/examples-lingua-franca/blob/48f2e88887bc187600c06fd3a7d6cff9bd7f1d20/C/src/DistributedHelloWorld/HelloWorldPhysicalAfter.lf) the `;` after `at localhost` was removed by lff. This, however, appears to be illegal syntax.

- In [Rosace.lf](https://github.com/lf-lang/examples-lingua-franca/blob/48f2e88887bc187600c06fd3a7d6cff9bd7f1d20/C/src/rosace/Rosace.lf) lff made the following change on the first run: 
  ```
  - // Documentation
  - @icon("Variables.png")
  - reactor Variables {}
  + @icon("Variables.png")  // Documentation
  + reactor Variables {
  }
  ```
  I am not sure if we really should move the comment here. But the actual problem is that, on the next run, lff does this:
  ```
  - @Icon("Variables.png")  // Documentation
  + @icon("Variables.png")  // Documentation Documentation
  ```

- In [RosaceController.lf](https://github.com/lf-lang/examples-lingua-franca/blob/48f2e88887bc187600c06fd3a7d6cff9bd7f1d20/C/src/rosace/RosaceController.lf), does not like the following:
  ```
    az_f = new Filter(
        period = filter_period,
        init_x1 = 0.0,
        init_x2 = 0.0,
        a = (0.169118914523145, -0.518588903229759),
        b = (0.229019233988375, 0.421510777305010)
    )
  ```
  it converts this to:
  ```
  az_f = new Filter(
    period = filter_period,
    init_x1 = 0.0,
    init_x2 = 0.0,
    a = {},
    b = {}
  )
  ```
  Notice the empty assignment of `a` and `b`.
- Interestingly, when running lff again, it decides to flatten the above output to
  ```
   az_f = new Filter(period = filter_period, init_x1 = 0.0, init_x2 = 0.0, a = {}, b = {})
  ```
  Although the initial conversion from the previous point is wrong, a formatter should not change the output again when run repeatedly.
[...]

PR #1840, opened 2023-06-11
Formatter fixes

petervdonovan:
This fixes some of the problems in #1832.

- [x] (1) address interaction of semicolons with `at` clauses
- [x] (2) fix comment duplication after annotations that precede reactor definitions
- [x] (3) fix removal of the contents of parenthesized list initializer expressions in C target
- [x] (4) fix lack of idempotency
- [x] (5) do not put comments that were not in `{=` `=}` into the `{=` `=}`
- [x] (6) do not eliminate comments that appear at the start of a code block that is an initializer

Point 4 might be merely a result from point 3, since the previous point results from destructive modification of the ECore model that might somehow violate the assumptions of the line breaker. Therefore, it might be solved; I'm not sure.

I am not sure about point (5). Fixing it would likely result in more special cases in the formatter. I would also like to point out that the current behavior is safe and semantics-preserving because comments that appear right after an opening `{=` do not seem to be included in the resulting `Code` ECore object. Instead, I think they are treated as LF-level comments. As a result, a program with the following reaction compiles with the C target:
```c
    reaction(t) -> y {= # this is a test
       lf_set(y, self->count);
       self->count++;
    =}
```
whereas the following does not:
```c
    reaction(t) -> y {= # this is a test
       # this is another test
       lf_set(y, self->count);
       self->count++;
    =}
```
Note the usage of a `#` sign, which is not a valid C comment.
- [x] (7) do not copy comments to the outside of initializers that use code:
```
    init_x1 = {= Va_eq * (1.0 - 1.911199519984605 /*a[1]*/ - 0.001916104193780 /*b[1]*/) =},
```
formats to
```
    /** a[1] */
    /** b[1] */
    init_x1 = {= Va_eq * (1.0 - 1.911199519984605 /*a[1]*/ - 0.001916104193780 /*b[1]*/) =},
```


petervdonovan:
I think the only remaining problem is (5).

I do not disagree with Christian on this, but it is not exactly a trivial change (it might take an afternoon to get it right). Therefore, this PR is awaiting a consensus on problem (5).

core/src/main/java/org/lflang/ast/ParsingUtils.java
cmnrd, line 1: Nitpick: We commonly name classes utility classes `Util` without an 's'.

cmnrd (commented):
Thanks a lot for those fixes. They look good to me. I also tested this version on the examples repo and it seems to fix the problems that I discovered previously.

I think we should also add tests based on the corner cases that I listed in https://github.com/lf-lang/lingua-franca/issues/1832. We could either add new integration tests, but probably it would be better to add specific formatter unit tests like the simple ones we already have in [LffCliTest.java](https://github.com/lf-lang/lingua-franca/blob/master/cli/lff/src/test/java/org/lflang/cli/LffCliTest.java)

core/src/main/java/org/lflang/ast/ParsingUtils.java
petervdonovan: Done

petervdonovan:
I'd like to draw attention to two significant changes other than the bugfixes:
- The formatter now does an `isEqual` check to make sure that its output is semantically equivalent to its input
- Comments are no longer inlined if their target syntactic element is more than one line long.

The latter of these two changes is responsible for the big diff in our tests. The other bugfixes did not affect our tests because they only addressed corner cases that were found in the examples repo.

test/C/temp.txt
lhstrh, line 1: Another accidental check in!

lhstrh (changes requested):
Looks good to me, modulo the .txt file that I commented on.

core/src/main/java/org/lflang/ast/ParsingUtils.java
lhstrh: What about `ASTUtils`? :rofl: I wasn't aware of the consensus...

test/C/temp.txt
petervdonovan: Fixed! Thanks for the catch

lhstrh (approved):

cmnrd:
Great! Thanks a lot @petervdonovan!

PR #1841, opened 2023-06-11
Bugfix in integration tests and fixed ROS2 tests 

cmnrd:
This PR fixes a bug in our integration tests that prevented the C++ ROS2 tests from actually being compiled with the ROS2 flag. Effectively, we did not run any of the ROS2 tests for a while. Since a few small bugs crept into the ROS2 support in the meantime, this PR also fixes the problems in the ROS2 code generation and pulls in https://github.com/lf-lang/reactor-cpp/pull/50.

### DIFF
--- a/.github/workflows/check-diff.yml
+++ b/.github/workflows/check-diff.yml
@@ -69,7 +69,8 @@ jobs:
       - id: should-skip
         name: "Determine whether to skip checks"
         run: |
-          echo "skip=${{ steps.duplicate.outputs.should_skip == 'true' && github.event.action != 'ready_for_review' }}" >> $GITHUB_OUTPUT
+          echo "skip=${{ steps.duplicate.outputs.should_skip == 'true' && github.event.action != 'ready_for_review' && github.event.sender.login != 'github-merge-queue' }}" >> $GITHUB_OUTPUT
+          echo "(Run initiated by ${{ github.event.sender.login }} )"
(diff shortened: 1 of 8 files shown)


cmnrd:
@erlingrj It looks like the fix in the test suite also exposed a bug in the zephyr support (see the test log). Could you take a look at this?

cmnrd:
Ok, I think I wound the problem.

erlingrj:
Will take a look tomorrow!

lhstrh (approved):
LGTM! Thanks for the fix!

erlingrj:
Do we need a new type of Configurator which can modify the target properties and runs last?

cmnrd:
Oh, that is a good point. Thanks! I guess it was not well-defined how the configurator is supposed to be used. The configurator for the ROS2 tests was modifying the resource, so that the target config needs to be reloaded after the transformation. I will change the ROS2 generator to also directly modify the target properties.

lhstrh:
> Oh, that is a good point. Thanks! I guess it was not well-defined how the configurator is supposed to be used. The configurator for the ROS2 tests was modifying the resource, so that the target config needs to be reloaded after the transformation. I will change the ROS2 generator to also directly modify the target properties.

I had assumed that the added reload happened in between runs, but if it happens after the target property changes and before the execution, then that's no bueno.

PR #1844, opened 2023-06-15
Adjustments for running the LF compiler in Epoch

a-sr:
- Fixed access to URIs which caused NPEs in Epoch. Eclipse does not use file URIs but platform URIs hence `toFileString` will return null.
- Improved Exception handling for URI conversion. Methods no longer throw IOExceptions beause there are no IO operations involved, instead IllegalArgumentException if the URI is malformed.


lhstrh (approved):
LGTM!

PR #1845, opened 2023-06-15
Use dynamic tests for formatter round trip tests

oowekyala:
The formatter round trip tests currently take about a minute on my machine. Using concurrent execution and a small refactoring this is now down to under 20secs.

I converted the test to use Junit dynamic tests, which means each individual test file gets its own test instance. This allows JUnit to parallelize them, and the IDE to display each test and its status separately. Additionally, the test is linked to the source file, so that if you click on it from the IDE you'll jump to the test file.


core/build.gradle
```diff
@@ -14,7 +14,10 @@ sourceSets {
     }
     test {
         java {
-            srcDirs = ['src/test/java', 'test-gen']
+            srcDirs = ['src/test/java']
```
oowekyala, line 17: I changed this because both `testFixtures` and `test` used `test-gen` as a same source directory. Intellij does not support this. Since `test` depends on `testFixtures` there is no problem.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https://services.gradle.org/distributions/gradle-8.0.2-bin.zip
-distributionSha256Sum=ff7bf6a86f09b9b2c40bb8f48b25fc19cf2b2664fd1d220cd7ab833ec758d0d7
+distributionUrl=https://services.gradle.org/distributions/gradle-8.1.1-bin.zip
```
oowekyala, line 3: This is because gradle 8.0.2 had a problem with dynamic tests: https://github.com/gradle/gradle/pull/23285

core/build.gradle
cmnrd: Oh, that explains why Intellij showed all these errors. Thanks!

cmnrd (approved):
Cool! This makes a lot of sense. We'll have to convince the CI as well though ;)

gradle/wrapper/gradle-wrapper.properties
lhstrh: The reason this won't work is that the hash that we're checking against also needs to be updated.

lhstrh:
This is what CI complains about:

```
RoundTripTests > roundTripTestFactory() > initializationError FAILED
    java.lang.ExceptionInInitializerError
        at org.lflang.tests.compiler.RoundTripTests.roundTripTestFactory(RoundTripTests.java:44)

        Caused by:
        java.lang.RuntimeException: Cannot create a resource for 'file:/home/runner/work/lingua-franca/lingua-franca/test/C/src/PhysicalConnection.lf'; a registered resource factory is needed
            at org.eclipse.xtext.resource.XtextResourceSet.getResource(XtextResourceSet.java:263)
            at org.eclipse.xtext.resource.SynchronizedXtextResourceSet.getResource(SynchronizedXtextResourceSet.java:33)
            at org.lflang.tests.TestRegistry$TestDirVisitor.visitFile(TestRegistry.java:360)
            at org.lflang.tests.TestRegistry$TestDirVisitor.visitFile(TestRegistry.java:297)
            at java.base/java.nio.file.Files.walkFileTree(Files.java:2811)
            at java.base/java.nio.file.Files.walkFileTree(Files.java:2882)
            at org.lflang.tests.TestRegistry$TestDirVisitor.walk(TestRegistry.java:381)
            at org.lflang.tests.TestRegistry.<clinit>(TestRegistry.java:205)
            ... 1 more
```

core/src/test/java/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -1399,29 +1411,34 @@ private Model createModel(TargetProperty key, String value) throws Exception {
...
-  public void checkTargetProperties() throws Exception {
+  @TestFactory
+  public Collection<DynamicTest> checkTargetProperties() throws Exception {
+    List<DynamicTest> result = new ArrayList<>();
+
     for (TargetProperty prop : TargetProperty.getOptions()) {
       if (prop == TargetProperty.CARGO_DEPENDENCIES) {
         // we test that separately as it has better error messages
-        return;
+        continue;
```
oowekyala, line 1421: Notice this change: I believe some properties were not being tested because we used to exit early. Fixing this revealed that there were no checks being executed for properties which have type "file", and further, that the validator does not do any check for those properties, so there is nothing to test for those properties. I filtered out tests for properties of this type as a workaround, although maybe the better fix is to add a validator check. The thing is I don't know how that would be written because it would have to deal with relative paths.

oowekyala:
Maybe this Junit feature can be used in the future to run the target tests concurrently, which would greatly speed them up as they spend large amounts of time waiting on IO.

core/src/testFixtures/java/org/lflang/tests/LFInjectorProvider.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.Singleton;
+import org.eclipse.xtext.testing.GlobalRegistries;
+import org.eclipse.xtext.testing.IInjectorProvider;
+import org.lflang.LFRuntimeModule;
+import org.lflang.LFStandaloneSetup;
+
+/**
+ * Note: this is a copy of a file generated by XText. The main modifications is to make this a
```
lhstrh, line 16:
```suggestion
 * Note: this is a copy of a file generated by XText. The main modification is to make this a
```

lhstrh (approved):
This looks great!

core/src/test/java/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
lhstrh: Thanks for the heads up. I created an issue for this...

buildSrc/src/main/groovy/org.lflang.test-conventions.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
     testFixturesImplementation "org.junit.platform:junit-platform-commons:$jUnitPlatformVersion"
     testFixturesImplementation "org.junit.platform:junit-platform-engine:$jUnitPlatformVersion"
     testFixturesImplementation "org.opentest4j:opentest4j:$openTest4jVersion"
-    testFixturesImplementation "org.eclipse.xtext:org.eclipse.xtext.testing:$xtextVersion"
+    testFixturesImplementation platform("org.eclipse.xtext:org.eclipse.xtext.testing:$xtextVersion")
```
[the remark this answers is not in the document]
oowekyala: I don't understand it well, but it changes the version conflict resolution for dependencies. Previously, running tests from the integrationTest module in intellij failed with 
```
java.lang.SecurityException: class "org.eclipse.core.runtime.OperationCanceledException"'s signer information does not match signer information of other classes in the same package
```
which indicates there are several jars on the classpath containing classes in this package --dependency hell.
The platform incantation forces the version conflict resolver to pick the dependency versions required by the 'platform', ie the versions it has been built and tested with. Without it, gradle is free to pick a newer version of each dependency independently, which resulted in an inconsistent dependency set apparently.

You can inspect how this affects our dependencies by running `./gradle :core:dependencies --configuration integrationTestRuntimeClasspath`

oowekyala:
Thanks for merging!

buildSrc/src/main/groovy/org.lflang.test-conventions.gradle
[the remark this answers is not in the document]
lhstrh: Interesting.. Thanks for the fix and explaining this!!

core/build.gradle
cmnrd: Oh, it works because the PR adds a custom implementation of `LfParsingTestHelper`, which is the only class generated in `test-gen`.

cmnrd: In fact, we can disable it. See https://github.com/lf-lang/lingua-franca/pull/1858

PR #1848, opened 2023-06-15
Declarative Port Graph in C++ Runtime

revol-xut:
# Declarative Port Graph
This pull requests adjusts the code generator to use the new port graph interface supplied by reactor-cpp. 

Previously Ports and Connections were wired together directly by the code generator in the assembly method. This Pull-Request changes this behavior that connections between ports are not declared in the environment. This allows the environment to optimize a lot of connections away and then wire up the components. 

**Previously**
```cpp
reactor1.out.bind_to(reactor2.in);
```

**Now**
```
env.draw_connection(reactor1.out, reactor2.in, ConnectionProperties{});
```

Also different Connection Types can be annotated over this interface like `ConnectionType::EnclavedDelayed`, `ConnectionType::PhysicalConnection` and many more. 

This PR works together with the changes in lf-lang/reactor-cpp/pull/51


core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -100,6 +100,12 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
         }
     }
 
+    private val Connection.cppDelay: String 
+        get() = if (delay != null) "${delay.toCppTime()}" else "0s"
```
cmnrd, line 104: You can use the elvis operator for this. Also the magic string is redundant.
```suggestion
        get() = delay?.toCppTime() ?: "0s"
```

.gitmodules
```diff
@@ -3,7 +3,8 @@
 	url = https://github.com/lf-lang/reactor-c.git
 [submodule "org.lflang/src/lib/cpp/reactor-cpp"]
 	path = core/src/main/resources/lib/cpp/reactor-cpp
-	url = https://github.com/lf-lang/reactor-cpp
+    url = https://github.com/lf-lang/reactor-cpp.git
+    branch = "connection-optimization"
```
cmnrd, line 7: Why change the module configuration? Usually we just change the ref of the submodule

core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -153,17 +159,9 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
...
-                    ${c.name}.bind_upstream_port(&${leftPort.name});
-                    ${c.name}.bind_downstream_port(&${rightPort.name});
-                """.trimIndent()
-            else
-                """
-                    // connection $idx
-                    ${leftPort.name}.bind_to(&${rightPort.name});
-                """.trimIndent()
+            """
+                left.environment()->draw_connection(${leftPort.name}, ${rightPort.name}, ${c.properties})
```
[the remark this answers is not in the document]
revol-xut: done

cmnrd (approved):
Great! Thanks for all the hard work!

PR #1850, opened 2023-06-15
More formatter fixes

petervdonovan:
This runs another pass of optimization. This is a heuristic, but it should not affect idempotency nor other correctness criteria. This can affect performance; however, I modified the formatter to keep track of whether anything has changed so that it knows whether further optimization passes have the potential to give any improved results, so in this common case the performance is not worse. When I ran it on my machine it did not seem much slower than before (around 30 seconds).

Also, when running the formatter from VS Code, we should go ahead even if there are warning or info messages, or syntax errors in the target code. If we can understand the AST, we can pretty-print it.

### DIFF
--- a/core/src/main/java/org/lflang/ast/MalleableString.java
+++ b/core/src/main/java/org/lflang/ast/MalleableString.java
@@ -357,6 +357,14 @@ public abstract class MalleableString {
         String singleLineCommentPrefix) {
       this.width = width;
       keepCommentsOnSameLine = true;
+      // Multiple calls to optimizeChildren may be required because as parts of the textual
+      // representation are updated, the optimal representation of other parts may change.
+      // For example, if the text is wider than 100 characters, the line may only need to be
+      // broken in one place, but it will be broken in multiple places a second optimization pass
+      // is not made. This is a heuristic in the sense that two passes are not guaranteed to result
+      // in a fixed point, but since a subsequent call to the formatter will get the same AST and
+      // therefore have the same starting point, the formatter as a whole should still be
+      // idempotent.
       var everChanged = false;
       var changed =
           optimizeChildren(providedRender, badness, width, indentation, singleLineCommentPrefix);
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/lflang/ast/MalleableString.java
```diff
@@ -348,40 +349,51 @@ private List<String> getLinesOfInterest(
...
+      everChanged |= changed;
       long badnessFalse = badness.applyAsLong(providedRender.get());
       keepCommentsOnSameLine = badnessTrue < badnessFalse;
-      optimizeChildren(providedRender, badness, width, indentation, singleLineCommentPrefix);
-      optimizeChildren(providedRender, badness, width, indentation, singleLineCommentPrefix);
+      if (changed)
+        changed =
+            optimizeChildren(providedRender, badness, width, indentation, singleLineCommentPrefix);
+      if (changed)
+        optimizeChildren(providedRender, badness, width, indentation, singleLineCommentPrefix);
```
lhstrh, line 385: This call seems redundant, and you're not doing anything with the return value...

lhstrh (commented):
Spotted only one issue.

core/src/main/java/org/lflang/ast/MalleableString.java
petervdonovan: These calls have side effects. Multiple calls to the same function with the same arguments may be required because as parts of the textual representation get updated, the optimal representation of other parts may change. in particular, if the text is wider than 100 characters, the line may only need to be broken in one place, but it will be broken in multiple places if this second optimization pass is not made.

This is a heuristic because in order to get a truly optimal textual representation it might be necessary to make combinatorially many calls with different combinations of textual representations.

lhstrh: I see. There might not even be a fixed point? Perhaps a small comment with a hint would be useful because it does look a bit sus.

lhstrh (approved):

core/src/main/java/org/lflang/ast/MalleableString.java
petervdonovan: Yes, no fixed point is guaranteed, but since a subsequent call to the formatter will get the same AST and therefore have the same starting point, the formatter as a whole should still be idempotent.

cmnrd (approved):
Great! Thanks!

PR #1851, opened 2023-06-16
More formatter fixes

petervdonovan:
This addresses some issues raised in https://github.com/lf-lang/examples-lingua-franca/pull/48.

- [x] Usually put the ending delimiter of a list on the same line as the last element of the list
- [x] Do not always surround `=` signs with whitespace when they are in parameter lists
- [ ] Resolve the issues with bad formatting of comments


cmnrd (approved):
Looks good! So we now do change the indentation in the target code blocks as well? I guess it is safe most of the time, but I wonder if there might be some corner cases where we could break the target code. I guess we'll find out eventually :smile:

PR #1853, opened 2023-06-17
More natural syntax for reaction declarations

lhstrh:
This PR changes the syntax for reaction declarations (which no inline implementation) from `reaction(t) s -> e named foo` to `reaction foo(t) s -> e`. Moreover, these changes allow reactions definitions (which do have an inline implementation) to also specify a name. This can be useful for documentation purposes and for making the generated code more readable. Also see #1856.

~~Note the semicolon at the end, which is required to avoid ambiguity with connection statements. We can make the semicolon optional if we pay attention to newlines, which we currently simply filter out. There is no mandatory semicolon for concrete reactions (ones that feature a body).~~

A trailing semicolon is only required in situations where not having one leads to ambiguity. A validator check makes sure that we handle this situation.

- [x] Implement validator check
- [x] Make error message more specific
- [x] Ensure that formatter does not introduce newline between triggers and sources (@petervdonovan)
- [x] Add a validator check to avoid NPE in targets other than C
- [x] Add unit tests

Note: this PR also fixes a problem where duplicate sources were detected inaccurately (62113a4b0)

### DIFF
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -200,10 +200,13 @@ Action:
 Reaction:
     (attributes+=Attribute)*
     (('reaction') | mutation ?= 'mutation')
+    (name=ID)?
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
     (sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    ((('named' name=ID)? code=Code) | 'named' name=ID)(stp=STP)?(deadline=Deadline)?
+    (((code=Code)(stp=STP)?(deadline=Deadline)? ';'?) |
+     ((stp=STP)?(deadline=Deadline)? ';')
+    )
     ;
 
 TriggerRef:
@@ -511,7 +514,7 @@ Token:
     'mutable' | 'input' | 'output' | 'timer' | 'action' | 'reaction' |
     'startup' | 'shutdown' | 'after' | 'deadline' | 'mutation' | 'preamble' |
     'new' | 'federated' | 'at' | 'as' | 'from' | 'widthof' | 'const' | 'method' |
-    'interleaved' | 'mode' | 'initial' | 'reset' | 'history' | 'watchdog' | 'named' |
+    'interleaved' | 'mode' | 'initial' | 'reset' | 'history' | 'watchdog' |
 
     // Other terminals
     NEGINT | TRUE | FALSE |
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -204,7 +204,7 @@ Reaction:
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
     ( => sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    (code=Code)? (stp=STP)? (deadline=Deadline)? ';'?
+    (code=Code)? (stp=STP)? (deadline=Deadline)? (delimited?=';')?
     ;
 
 TriggerRef:
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -202,11 +202,9 @@ Reaction:
     (('reaction') | mutation ?= 'mutation')
     (name=ID)?
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
-    (sources+=VarRef (',' sources+=VarRef)*)?
+    ( => sources+=VarRef (',' sources+=VarRef)*)?
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
-    (((code=Code)(stp=STP)?(deadline=Deadline)? ';'?) |
-     ((stp=STP)?(deadline=Deadline)? ';')
-    )
+    (code=Code)? (stp=STP)? (deadline=Deadline)? ';'?
     ;
 
 TriggerRef:
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -643,7 +643,7 @@ public class ToLf extends LfSwitch<MalleableString> {
     if (object.getCode() != null) msb.append(" ").append(doSwitch(object.getCode()));
     if (object.getStp() != null) msb.append(" ").append(doSwitch(object.getStp()));
     if (object.getDeadline() != null) msb.append(" ").append(doSwitch(object.getDeadline()));
-    if (object.getName() != null) msb.append(";");
+    if (object.getCode() == null) msb.append(";");
     return msb.get();
   }
 
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -618,7 +618,7 @@ public class ToLf extends LfSwitch<MalleableString> {
     } else {
       msb.append("reaction");
     }
-    if (object.getName() != null) msb.append(object.getName());
+    if (object.getName() != null) msb.append(" ").append(object.getName());
     msb.append(list(true, object.getTriggers()));
     msb.append(list(", ", " ", "", true, false, object.getSources()));
     if (!object.getEffects().isEmpty()) {
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -643,7 +643,6 @@ public class ToLf extends LfSwitch<MalleableString> {
     if (object.getCode() != null) msb.append(" ").append(doSwitch(object.getCode()));
     if (object.getStp() != null) msb.append(" ").append(doSwitch(object.getStp()));
     if (object.getDeadline() != null) msb.append(" ").append(doSwitch(object.getDeadline()));
-    if (object.isDelimited()) msb.append(";");
     return msb.get();
   }
 
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -618,6 +618,7 @@ public class ToLf extends LfSwitch<MalleableString> {
     } else {
       msb.append("reaction");
     }
+    if (object.getName() != null) msb.append(object.getName());
     msb.append(list(true, object.getTriggers()));
     msb.append(list(", ", " ", "", true, false, object.getSources()));
     if (!object.getEffects().isEmpty()) {
@@ -639,10 +640,10 @@ public class ToLf extends LfSwitch<MalleableString> {
                               : doSwitch(varRef))
                   .collect(new Joiner(", ")));
     }
-    if (object.getName() != null) msb.append(" named ").append(object.getName());
     if (object.getCode() != null) msb.append(" ").append(doSwitch(object.getCode()));
     if (object.getStp() != null) msb.append(" ").append(doSwitch(object.getStp()));
     if (object.getDeadline() != null) msb.append(" ").append(doSwitch(object.getDeadline()));
+    if (object.getName() != null) msb.append(";");
     return msb.get();
   }
 
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -643,7 +643,7 @@ public class ToLf extends LfSwitch<MalleableString> {
     if (object.getCode() != null) msb.append(" ").append(doSwitch(object.getCode()));
     if (object.getStp() != null) msb.append(" ").append(doSwitch(object.getStp()));
     if (object.getDeadline() != null) msb.append(" ").append(doSwitch(object.getDeadline()));
-    if (object.getCode() == null) msb.append(";");
+    if (object.isDelimited()) msb.append(";");
     return msb.get();
   }
 
--- a/core/src/main/java/org/lflang/validation/LFValidator.java
+++ b/core/src/main/java/org/lflang/validation/LFValidator.java
@@ -712,7 +712,7 @@ public class LFValidator extends BaseLFValidator {
         && reaction.getDeadline() == null
         && reaction.getStp() == null
         && reaction.getSources() != null
-        && reaction.getSources().size() != 0) {
+        && !reaction.getSources().isEmpty()) {
       error("Missing semicolon at the end of reaction declaration.", Literals.REACTION__SOURCES);
     }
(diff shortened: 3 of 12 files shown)


cmnrd (commented):
This also allows reactions with bodies to be named, right? I think generally allowing named reactions would be a good thing, we should only adjust generators and the Diagramm synthesis to handle the names similiarly to the current labels.

lhstrh:
> This also allows reactions with bodies to be named, right? I think generally allowing named reactions would be a gold thing, we should only adjust generators and the Diagramm synthesis to handle the names similiarly to the current labels.

That is true. Prohibiting it would require an additional validator check, but I don't really see a reason for that.

lhstrh:
> > This also allows reactions with bodies to be named, right? I think generally allowing named reactions would be a gold thing, we should only adjust generators and the Diagramm synthesis to handle the names similiarly to the current labels.
> 
> That is true. Prohibiting it would require an additional validator check, but I don't really see a reason for that.

@edwardalee and @a-sr, what do you think of this?

edwardalee:
I find the occasional need for semicolons to be an ugly defect in the grammar.  I would prefer to require semicolons everywhere over requiring them sometimes.

lhstrh:
> I find the occasional need for semicolons to be an ugly defect in the grammar. I would prefer to require semicolons everywhere over requiring them sometimes.

I agree, but I think that most languages that have optional semicolons have some corner cases where a semicolon is required. JavaScript is one of them. The reason why I find the proposed solution better than the one that's current in `master` is twofold:
 - consistency with the way we declare other named syntactic elements in the language; and
 - the prospect of removing the remaining oddity of the required semicolon by using a line break instead of a semicolon to disambiguate.

Note that the semicolon is _only_ required when you do not inline the reaction body, so this will not require any changes to any LF code you've written, as opposite to requiring semicolons, which will affect every piece of LF code out there. As you can see, this PR is very small.

What I was hoping you'd comment on, was not so much the semicolons, but the ability to name _any_ reaction (also those that have inlined implementations) and use the name in the generated code and diagrams. I think this would be helpful and worth implementing, but if for some reason we want to stick with numbers and forbid the naming of reactions unless they are abstract, then a simple validator check can enforce this. WDYT? I can of course implement the check and leave the dealings with names for later... The idea would be that this would subsumes @a-sr's `@name` attribute.

edwardalee:
I agree that the ability to name reactions can help with documentation and code clarity. I suspect we will get people who will want to manually invoke a reaction, but I guess we can point them to methods.

test/C/src/no_inlining/CountHierarchy.lf
```diff
@@ -15,9 +15,9 @@ main reactor {
 
   state count: int
 
-  reaction(t.out) named increment
+  reaction increment(t.out);
 
-  reaction(t.out) named check_done
+  reaction check_done(t.out);
```
oowekyala, line 20: I wonder if the semicolon should be made optional? Here there is no parsing ambiguity. In languages with optional semis, semis are only required when they resolve an ambiguity, so it feels odd that these lines have to have a semi.

I would find it more consistent to be able to write
```
state count: int
reaction increment(t.out)
reaction check_done(t.out) -> x
```
The ambiguity is only there if the following token is an identifier:
```
reaction foo(a)
a -> b
```
We don't keep track of newlines in the parser but we have access to line numbers in the validator. We could issue a warning if we find a reaction with no body, and both a source and effect declaration which are not on the same line as the reaction. Since AFAIK sources are relatively rare, I don't think this should have a lot of false positives.

lhstrh: ANTLR doesn't like ambiguous grammars. Is there a way to implement what you suggest without this resulting in a barrage of errors/warnings from Xtext?

erlingrj:
I think this looks alot better than `reaction(a) named foo`. But also agree that it would be optimal if we could get rid of the semi-colon and use the newline as the delimiter. Parsing and grammars is however not my area of expertise so cant contribute with any ideas on how to do it.

test/C/src/no_inlining/CountHierarchy.lf
lhstrh: OK, I was able to force the parse to go down the "sources" path using the `=>` predicate (see caa69f615). This removes the ambiguity, but it also means that
```
reaction foo(x)
p -> q

```
will be parsed as:
```
reaction foo(x) p -> q
```
For that particular situation, we need to add a validator check to tell the user that it must place a `;` after the reaction declaration.

test/C/src/no_inlining/BankMultiportToReactionNoInlining.lf
```diff
@@ -20,7 +20,7 @@ main reactor {
 
   s = new[2] DoubleCount()
 
-  reaction(s.out) named check
+  reaction check(s.out);
```
lhstrh, line 23:
```suggestion
  reaction check(s.out)
```

test/C/src/no_inlining/BankToReactionNoInlining.lf
```diff
@@ -12,5 +12,5 @@ main reactor {
 
   s = new[2] Count()
 
-  reaction(s.out) named check
+  reaction check(s.out);
```
lhstrh, line 15:
```suggestion
  reaction check(s.out)
```

test/C/src/no_inlining/Count.lf
```diff
@@ -8,9 +8,9 @@ main reactor Count {
 
   state count: int
 
-  reaction(t) named increment
+  reaction increment(t);
```
lhstrh, line 11:
```suggestion
  reaction increment(t)
```

test/C/src/no_inlining/Count.lf
```diff
@@ -8,9 +8,9 @@ main reactor Count {
...
 
-  reaction(t) named check_done
+  reaction check_done(t);
```
lhstrh, line 13:
```suggestion
  reaction check_done(t)
```

test/C/src/no_inlining/CountHierarchy.lf
```diff
@@ -15,9 +15,9 @@ main reactor {
 
   state count: int
 
-  reaction(t.out) named increment
+  reaction increment(t.out);
```
lhstrh, line 18:
```suggestion
  reaction increment(t.out)
```

test/C/src/no_inlining/CountHierarchy.lf
lhstrh, line 20:
```suggestion
  reaction check_done(t.out)
```

test/C/src/no_inlining/IntPrint.lf
```diff
@@ -6,14 +6,14 @@ target C {
 reactor Print {
   output out: int
 
-  reaction(startup) -> out named sender
+  reaction sender(startup) -> out;
```
lhstrh, line 9:
```suggestion
  reaction sender(startup) -> out
```

test/C/src/no_inlining/IntPrint.lf
```diff
@@ -6,14 +6,14 @@ target C {
...
 }
 
 // expected parameter is for testing.
 reactor Check(expected: int = 42) {
   input in: int
 
-  reaction(in) named receiver
+  reaction receiver(in);
```
lhstrh, line 16:
```suggestion
  reaction receiver(in)
```

test/C/src/no_inlining/MultiportToReactionNoInlining.lf
```diff
@@ -23,7 +23,7 @@ main reactor {
   state s: int = 6
   b = new Source(width = 4)
 
-  reaction(b.out) named check
+  reaction check(b.out);
```
lhstrh, line 26:
```suggestion
  reaction check(b.out)
```

lhstrh (commented):
We can now remove the semicolons in all the tests (because none of them pose ambiguity).

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -200,10 +200,11 @@ Action:
 Reaction:
     (attributes+=Attribute)*
     (('reaction') | mutation ?= 'mutation')
+    (name=ID)?
     ('(' (triggers+=TriggerRef (',' triggers+=TriggerRef)*)? ')')
-    (sources+=VarRef (',' sources+=VarRef)*)?
+    ( => sources+=VarRef (',' sources+=VarRef)*)?
```
lhstrh, line 205: This is to tell ANTLR to always prefer going down this path rather than use the `Connection` production.

core/src/main/java/org/lflang/ast/ToLf.java
```diff
@@ -639,10 +640,10 @@ public MalleableString caseReaction(Reaction object) {
                               : doSwitch(varRef))
                   .collect(new Joiner(", ")));
     }
-    if (object.getName() != null) msb.append(" named ").append(object.getName());
     if (object.getCode() != null) msb.append(" ").append(doSwitch(object.getCode()));
     if (object.getStp() != null) msb.append(" ").append(doSwitch(object.getStp()));
     if (object.getDeadline() != null) msb.append(" ").append(doSwitch(object.getDeadline()));
+    if (object.isDelimited()) msb.append(";");
```
petervdonovan, line 646: Based on the implementation of `caseReactor`, the connections are always placed before the reactions in a reactor, so we will never have the ambiguity. Therefore, it is safe to omit the semicolon from the formatter output and to remove the `isDelimited` field from the ECore model.

lhstrh: I will change the formatter to not add semicolons, but we cannot remove `isDelimited` because it is used in the validator.

lhstrh, line 646:
```suggestion
```

cmnrd:
Great! I have a concern with terminology herer though. "Abstract" is commonly used for fully virtual (i.e. methods that are virtual and don't have a default implementation). Our reactions, however, are not virtual. So the term "abstract" is quite misleading in my opinion.

We could use the C/C++ terminology and call a bodiless reaction "reaction declaration", or we simply use bodiless reaction.

lhstrh:
> Great! I have a concern with terminology herer though. "Abstract" is commonly used for fully virtual (i.e. methods that are virtual and don't have a default implementation). Our reactions, however, are not virtual. So the term "abstract" is quite misleading in my opinion.
> 
> We could use the C/C++ terminology and call a bodiless reaction "reaction declaration", or we simply use bodiless reaction.

Why don't we make it virtual? I believe the only reason we didn't have overrides is because we didn't have names, but we do now...

cmnrd:
What would be the benefit? In C, we would have to manually implement  the virtual method table. In C++ this is automated, but virtual methods still add overhead. Without a clear concept of reactor interfaces and use-cases where concrete reactor instances are somehow injected, there is no point in making reactions virtual. We always know the precise  reactor type, so we can also always call the precise reaction.

lhstrh:
> Great! I have a concern with terminology though. "Abstract" is commonly used for fully virtual methods (i.e. methods that are virtual and don't have a default implementation). Our reactions, however, are not virtual. So the term "abstract" is quite misleading in my opinion.
> 
> We could use the C/C++ terminology and call a bodiless reaction "reaction declaration", or we simply use bodiless reaction.

OK, I renamed "abstract reactions" to "reaction declarations".

cmnrd (commented):
Adding a validator check is a great idea. This means we will only need to use a semicolon for reaction declarations (without a body) that have sources, right?

There should also be a validator check that ensures that the target code generator has support for bodyless reactions if they are used. Currently, the C++ generator fails with a nullpointer exception and I expect that also other target generators will fail ungracefully if the reaction body is missing.

cmnrd:
One more comment: Currently the error message produced by the validator is "Missing semicolon at the end of reaction declaration." Perhaps we can add a little more context as to why the semicolon is needed and how to fix it. I am not sure if we condense this into a short message, but if not, we could add a page on the website and point to it.

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -706,12 +707,20 @@ public void checkReaction(Reaction reaction) {
     if (reaction.getTriggers() == null || reaction.getTriggers().size() == 0) {
       warning("Reaction has no trigger.", Literals.REACTION__TRIGGERS);
     }
-    HashSet<Variable> triggers = new HashSet<>();
+    if (!reaction.isDelimited()
+        && reaction.getCode() == null
+        && reaction.getDeadline() == null
+        && reaction.getStp() == null
+        && reaction.getSources() != null
+        && reaction.getSources().size() != 0) {
```
oowekyala, line 715:
```suggestion
        && !reaction.getSources().isEmpty()) {
```

core/src/test/java/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -172,6 +172,52 @@ public void disallowReactorCalledPreamble() throws Exception {
...
+              reaction(inp) -> out {==}
+            }
+
+            main reactor {
+              f1 = new Foo()
+              f2 = new Foo()
+              f3 = new Foo()
+
+              reaction increment(f1.out)
+              f2.out -> f3.inp
```
oowekyala, line 193: Currently IIUC you emit an error and expect the reaction to be written
```

              reaction increment(f1.out)
              f2.out -> f3.inp;
```

which I believe doesn't help, as this still looks ambiguous.

I think the validator should not emit an error but a warning here, saying something like
```
              reaction increment(f1.out)
              f2.out -> f3.inp
              ^^^^^^^^^^^^ This line is parsed as a source and effects declaration for the reaction
                           on line 2, but it looks like a connection statement. If it is really a connection,
                           add a semicolon at the end of line 2. If it is really a dependency declaration, 
                           place it at the end of line 2 to silence this warning.
```

lhstrh: My reasoning was that "at the end of the reaction declaration" would capture the intent because it would then be up to the programmer to specify where that end is (either at the error marker, or further down). I'm on board with making the message more explicit, like you suggest. Thanks for the thoughtful feedback, @oowekyala!

petervdonovan: You can also avoid emitting the message in certain cases by digging into the node model using NodeModelUtils.

The node model has all of the whitespace information that would be needed to make a very good guess about what the user really meant. For example, you can find out if there is a `\n` between `reaction increment(f1.out)` and `f2.out -> f3.inp`. You can also find out whether the line that says `f2.out -> f3.inp` is indented.

If we go this route, then the error message can recommend changing the whitespace instead of adding a semicolon.

lhstrh: Handled this in 94e8da06f.

lhstrh:
> Adding a validator check is a great idea. This means we will only need to use a semicolon for reaction declarations (without a body) that have sources, right?
> 
> There should also be a validator check that ensures that the target code generator has support for bodyless reactions if they are used. Currently, the C++ generator fails with a nullpointer exception and I expect that also other target generators will fail ungracefully if the reaction body is missing.

Fixed in 8b920a8a1.

petervdonovan (approved):

PR #1857, opened 2023-06-19
Unknown port types handled with `unknown` instead of `Present`

lhstrh:
The type constraint `extends Present` (with `Present` being the type that included all types but `undefined`) was inconvenient and non-standard from the perspective of TypeScript. Removing this awkwardness will probably also make it more straightforward to offer support for generic reactors in LF. See https://github.com/lf-lang/reactor-ts/pull/168 for related changes made in the runtime.


.github/workflows/ts-tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform TypeScript tests
         run: |
-          ./gradlew targetTest -Ptarget=TypeScript -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
+          ./gradlew targetTest -Ptarget=TypeScript -Druntime="git://github.com/lf-lang/reactor-ts.git#axmmisaka/remove-present-replace-unknown"
```
lhstrh, line 40:
```suggestion
          ./gradlew targetTest -Ptarget=TypeScript -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
```

core/src/main/resources/lib/ts/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "LinguaFrancaDefault",
     "type": "commonjs",
     "dependencies": {
-        "@lf-lang/reactor-ts": "git://github.com/lf-lang/reactor-ts.git#master",
+        "@lf-lang/reactor-ts": "git://github.com/lf-lang/reactor-ts.git#axmmisaka/remove-present-replace-unknown",
```
lhstrh, line 5:
```suggestion
        "@lf-lang/reactor-ts": "git://github.com/lf-lang/reactor-ts.git#master",
```

lhstrh (commented):
Looks good, modulo suggested changes (to be applied after merging https://github.com/lf-lang/reactor-ts/pull/168)

axmmisaka (approved):
lgtm, especially considering it works :^)

PR #1858, opened 2023-06-19
Cleanup gradle config

cmnrd:
This is a small cleanup of the gradle configuration. 
- Remove the test source set configuration that became superfluous after https://github.com/lf-lang/lingua-franca/pull/1845.
- Backport clean-ups in the gradle configuration made in https://github.com/lf-lang/lingua-franca/pull/1271 so that we can merge them quicker.
- Disable test code generation from xtext as https://github.com/lf-lang/lingua-franca/pull/1845 made this superfluous.


oowekyala (approved):
LGTM!

PR #1864, opened 2023-06-22
Generics multiple instantiations

### LINKED ISSUE #1860: Multiple definitions of the ports when using generics in the C target
This example program falis in the target C compiler:
```
target C;

reactor R<T> {
    input in: T
}

main reactor {
    r1 = new R<int>()
    r2 = new R<int>()
}
```

The problem is that struct representing the input port of the generic `R` is generated twice. I am guessing that the two instances of R, even though using the same template argument, get different TypeParameterizedReactor objects

petervdonovan:
Fixes #1860.

This also fixes another bug aside from #1860 where if two instantiations with _different_ type parameters exist, then there is a naming collision.

I have not written detailed tests for generics in the C target and expect us to find more bugs.

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CGenerator.java
@@ -809,7 +809,8 @@ public class CGenerator extends GeneratorBase {
     @Override
     public boolean equals(Object obj) {
       // This is equivalence modulo decl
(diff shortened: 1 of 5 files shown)


lhstrh:
The example LF code that this fixes should be added as a test.

petervdonovan:
Thanks for the catch!

lhstrh (approved):
LGTM!

erlingrj:
Sorry for the delay. This fixes the issue on my end. Thanks Peter

PR #1865, opened 2023-06-24
Update level assignment algorithm

petervdonovan:
This includes changes required in order to implement TPO.

### Status update 6/24

The basic idea is almost implemented, but still only 43/57 tests pass (previously it was 41/57). I think that most of the tests with cycles are still deadlocking because the output reactions that send "port absent" messages are not being enqueued. This should be an easy fix. Also the method of assigning the TPO is not entirely correct yet, for uninteresting software engineering reasons.

### Status update 6/26

Most of the tests are passing now. 53 ish out of 57 depending on how you count, BUT some of the failures are nondeterministic, so probably there are others that will also sometimes fail.

- `FeedbackDelay` and `FeedbackDelaySimple` are failing due to falsely detecting cycles when there is actually a 1 microstep `after` delay
- `DecentralizedP2PComm` is failing with wrong functional behavior (STP violation in the generalized sense, I believe). This failed in the second-to-last run but not the last one. It may be nondeterministic, so this isn't surprising.
- `DistributedLoopedPhysicalActionDecentralized` failed with an error that I could not reproduce locally, so the situation there is probably similar.

### TODOs

Whoever chooses to pick up any of these tasks should write their name next to the task to make sure there is no duplicate work.

- [x] Use a less hacky mechanism for marking ports that receive from delayed connections as not part of the MLAA calculation (@petervdonovan)
- [x] Find dead/unnecessary code in the code generator and delete it (split this into sub-tasks as needed) (the CodeCov reports may help with this)
- [x] Pass the other C tests besides the C federated tests
- [x] Ensure IntelliJ does not emit warnings related to the code that is touched by this and the other related PRs
- [x] Get this branch up to date with `master` (@petervdonovan)
- [x] Check that this also works for the Python target (@petervdonovan)
- [ ] Factor C-specific code out of FedASTUtils and into CExtension or CExtensionUtils (@petervdonovan)
- [ ] Get all tests in reactor-c to pass (@petervdonovan)
- [ ] Get the TypeScript runtime to be compatible with this branch (@byeong-gil)
- [ ] Fix the broken C tests and check that they still pass
- [ ] Fix the broken Python tests and check that they still pass
- [ ] Check that all code added has JavaDoc

Non-required TODOs
- [ ] Use library reactors as the network senders and receivers instead of generated reactors (@petervdonovan)
- [ ] Reduce the complex hidden dependencies between different functions through their interaction with the state of mutable objects
- [ ] Do not add the special reactors for indexing into multiports
- [ ] Find dead/unnecessary code in the C runtime and delete it


petervdonovan:
I think the problem might just be the nondeterministic STP violations at this point. Right now I think this might be because the MLAA is advanced to FOREVER from the start when it shouldn't be. As a result the first tag is executed quickly without waiting for a message, and then the message is received after the logical time has already advanced.

petervdonovan:
Apparently the code does not currently expect to receive "port absent" messages after it has already assumed a port absent and labeled its status as known to be absent at the current tag (why not?). Also, FeedbackDelay now deadlocks for unknown reason.

petervdonovan:
The other day @lhstrh asked about the status of this PR. The federated C tests finally passed in this branch (:tada:), so at least we know that it will be possible to get this to work within a reasonable amount of time. However there is a lot of cleanup left to do (maybe a few weeks) before this is good enough to merge into `master`.

lhstrh:
That is fantastic news! Perhaps you could compile a list of todos for the cleanup so that others could help out with some of the items?

petervdonovan:
Today I mentioned that in this branch there is an STP violation in `LoopDistributedDecentralized` that should not occur. Just now I checked and found that this error also occurs in `master`, even when the STP offset and deadline are both very large (100 msec). Probably fixing this in this branch should not be required in order to merge this into `master`.

petervdonovan:
Work on this is being moved back to the `remove-unordered-reactions` branch.

PR #1869, opened 2023-06-27
Removed org.lflang.ide

cmnrd:
We don't need `org.lflang.ide` as part of the core repo. I am not sure, if the deleted files are needed by Epoch. If they are needed, we should add them to the epoch repo. @a-sr Do you know if Epoch needs the deleted classes?


cmnrd:
Oh, nevermind. Apparently the org.lflang.ide is actually needed for the LSP.

a-sr:
> Oh, nevermind. Apparently the org.lflang.ide is actually needed for the LSP.

Correct!

PR #1871, opened 2023-06-28
Added tests for lf_request_stop with enclaves and federates

edwardalee:
This is a companion PR to [244 in reactor-c](https://github.com/lf-lang/reactor-c/pull/244), which fixes lf_request_stop() for enclaves and federates with enclaves. This PR aligns reactor-c and adds two tests.


erlingrj (approved):
Lets get this merged! When I merge in the first version of the coordinated enclaves the test with enclaves and federates will be moved to failing

PR #1876, opened 2023-06-30
Add a `NetworkReactor` for TypeScript

byeong-gil:
This PR reflects changes of this PR ([lingua-franca/pull/1684](https://github.com/lf-lang/lingua-franca/pull/1684)) to TS targeted LF.


byeong-gil:
> I know this is still a draft, but the tests are passing and it will be desirable to merge this as soon as we can do so without significant regressions. I see that the TPO-related code is commented out in `reactor-ts`, but I think it is OK to merge with that code disabled and only enable it separately when `ts-cyclic-dependencies` is ready.

I totally agree with you, @petervdonovan. I will merge this branch into `remove-unordered-reactions` right after this [reactor-ts PR](https://github.com/lf-lang/reactor-ts/pull/186) passes every test.

core/src/main/kotlin/org/lflang/generator/ts/TSActionGenerator.kt
```diff
@@ -10,7 +10,7 @@ import java.util.*
  */
 class TSActionGenerator(
     private val actions: List<Action>,
-    private val networkMessageActions: List<String>
+    // private val networkMessageActions: List<String>
```
petervdonovan, line 13: Let's remember to delete all the commented-out code

core/src/main/kotlin/org/lflang/generator/ts/TSActionGenerator.kt
```diff
@@ -45,12 +45,21 @@ class TSActionGenerator(
...
                 }
-                if (action.name in networkMessageActions) {
+                // if (action.name in networkMessageActions) {
+                //     actionInstantiations.add(
+                //         "this.${action.name} = new __FederatePortAction<${action.tsActionType}>($actionArgs);")
+                // } else {
+                //     actionInstantiations.add(
+                //         "this.${action.name} = new __Action<${action.tsActionType}>($actionArgs);")
+                // }
+                if (action.name.take(7) == "network") {
```
petervdonovan, line 55: Let's not forget to clean this up so that if the user creates an action that starts with "network", it isn't interpreted as a a generated action that is part of the federated infrastructure. A magic attribute such as "__federatePortAction` might be sufficient to mark these.

It is a long-standing goal to do less ad-hoc string matching deep in the code generators. There are several other instances of string matching here in places where I do not think we should be doing string matching

petervdonovan (approved):
Other than the two quibbles which are easily fixed, this looks good to me. The issues that I mentioned can be fixed before or after merging into `remove-unordered-reactions` -- either is fine.

core/src/main/kotlin/org/lflang/generator/ts/TSActionGenerator.kt
byeong-gil: > if the user creates an action that starts with "network", it isn't interpreted as a generated action that is part of the federated infrastructure.

Oh, I didn't notice this. You're right. However, I don't know how to classify normal `Action` and `FederatePortAction`. Because the `Action` interface only has its name, origin, and policy. Do we have to add a new attribute in lf files?

core/src/main/kotlin/org/lflang/generator/ts/TSActionGenerator.kt
byeong-gil: This is fixed in the commit [9acb408](https://github.com/lf-lang/lingua-franca/pull/1876/commits/9acb408821453057fa1dedd652cce0c8ddf5fa9e)

core/src/main/kotlin/org/lflang/generator/ts/TSActionGenerator.kt
petervdonovan: Well, I'm not sure I remember all the details, and I'm not 100% sure that my suggestion will work, but an example of adding an attribute is [in this commit](https://github.com/lf-lang/lingua-franca/pull/1684/commits/b4a76297b444c4a317bc0aec03a2025ee6fe4538#diff-0a51816c8e589e8f0b45b48eb7857be7c6c45006c00dbbccb813d78939399ae0), where I added the `_tpoLevel` attribute. If you do not get around to fixing this today, then I can work on it overnight since I do not think it will be too important or complicated.

byeong-gil: I just added a new attribute named `_NetworkReactor` on network reactors instead of `_FederatePortAction` on federate port actions. Because the code generator needs to classify network sender and receiver reactors as well as normal `Action` and `FederatePortAction`. We can know that an action is `FederatePortAction` when it is contained by a network receiver.

petervdonovan: Great, thanks!

PR #1878, opened 2023-07-02
TS code generator adjusted to appease `eslint`

lhstrh:
See https://github.com/lf-lang/reactor-ts/pull/175.


axmmisaka (approved):
👍🏼 , and this should be necessary

PR #1891, opened 2023-07-07
Fix for inheritance problem exposed in examples

### LINKED ISSUE #1733: Inheritance not working with fedgen
The `fed-gen` branch introduced another error that makes inheritance not work. The following minimal program fails to compile:
```
target C
reactor A {
    reaction(startup) {=
        printf("Hello\n");
    =}
}
reactor B {
    a = new A()
}
reactor C extends B {}
federated reactor {
    c = new C()
}
```
In the generated `c.lf` file, we see:
```
reactor C extends B {
}
```
But reactor B is never defined.

lhstrh:
- [x] Fix problem in https://github.com/lf-lang/examples-lingua-franca/issues/53
- [x] Fix #1733 
- [x] Add comments to tests
- [x] Clean up `FederateInstance`

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
+++ b/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
@@ -271,7 +271,7 @@ public class FederateInstance {
           if (r.equals(parent)) {
             return true;
           }
-          // Check if there are instantiations of the reactor a super class
+          // Check if there are instantiations of the reactor in a super class
           if (parent instanceof Reactor p) {
             for (var inst : p.getInstantiations()) {
               if (contains(inst, reactor)) {
--- a/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
+++ b/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
@@ -259,23 +259,34 @@ public class FederateInstance {
     }
 
     if (instantiation.getReactorClass() instanceof Reactor reactorDef) {
-      // Check whether the reactor is instantiated
+      // Check if the reactor is instantiated
       for (Instantiation child : reactorDef.getInstantiations()) {
         if (contains(child, reactor)) {
           return true;
         }
       }
-      // Check whether the reactor is a parent
+      // Check if the reactor is a super class
       for (var parent : reactorDef.getSuperClasses()) {
         if (reactor instanceof Reactor r) {
-          return r.equals(parent);
+          if (r.equals(parent)) {
+            return true;
+          }
+          // Check if there are instantiations of the reactor a super class
+          if (parent instanceof Reactor p) {
+            for (var inst : p.getInstantiations()) {
+              if (contains(inst, reactor)) {
+                return true;
+              }
+            }
+          }
         }
         if (reactor instanceof ImportedReactor i) {
-          return i.getReactorClass().equals(parent);
+          if (i.getReactorClass().equals(parent)) {
+            return true;
+          }
         }
       }
     }
-
     return false;
   }
 
--- a/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
+++ b/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
@@ -211,11 +211,12 @@ public class FederateInstance {
   private final MessageReporter messageReporter;
 
   /**
-   * Return {@code true} if federate instance inherits the given reactor declaration.
+   * Return {@code true} if the class declaration of the given {@code instantiation} references the
+   * {@code reactor}, either directly or indirectly (e.g., via a superclass or a type parameter).
    *
-   * @param declaration The reactor declaration to inherit.
+   * @param declaration The reactor declaration to check if it is referenced.
    */
-  public boolean inherits(ReactorDecl declaration) {
+  public boolean references(ReactorDecl declaration) {
     return references(this.instantiation, declaration);
   }
 
@@ -224,7 +225,7 @@ public class FederateInstance {
    * {@code reactor}, either directly or indirectly (e.g., via a superclass or a type parameter).
    *
    * @param instantiation The instantiation the class of which may refer to the reactor declaration.
-   * @param declaration The potentially referenced reactor declaration.
+   * @param declaration The reactor declaration to check if it is referenced.
    */
   private boolean references(Instantiation instantiation, ReactorDecl declaration) {
     if (instantiation.getReactorClass().equals(ASTUtils.toDefinition(declaration))) {
@@ -264,13 +265,13 @@ public class FederateInstance {
   }
 
   /**
-   * Return {@code true} if this federate inherits the given import.
+   * Return {@code true} if this federate references the given import.
    *
-   * @param imp The import to inherit.
+   * @param imp The import to check if it is referenced.
    */
-  public boolean inherits(Import imp) {
+  public boolean references(Import imp) {
     for (ImportedReactor reactor : imp.getReactorClasses()) {
-      if (this.inherits(reactor)) {
+      if (this.references(reactor)) {
         return true;
       }
     }
@@ -278,11 +279,11 @@ public class FederateInstance {
   }
 
   /**
-   * Return {@code true} if this federate inherits the given parameter.
+   * Return {@code true} if this federate references the given parameter.
    *
-   * @param param The parameter to inherit.
+   * @param param The parameter to check if it is referenced.
    */
-  public boolean inherits(Parameter param) {
+  public boolean references(Parameter param) {
     // Check if param is referenced in this federate's instantiation
     var returnValue =
         instantiation.getParameters().stream()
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -97,28 +92,20 @@ public FederateInstance(
...
-
   /**
    * The position within a bank of reactors for this federate. This is 0 if the instantiation is not
    * a bank of reactors.
    */
-  public int bankIndex = 0;
-
-  /** A list of outputs that can be triggered directly or indirectly by physical actions. */
-  public Set<Expression> outputsConnectedToPhysicalActions = new LinkedHashSet<>();
+  public int bankIndex;
```
edwardalee, line 108: Why is the initialization of bankIndex to 0 removed?

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -220,79 +204,87 @@ public Instantiation getInstantiation() {
...
-    } else if (object instanceof Import) {
-      return contains((Import) object);
-    } else if (object instanceof Parameter) {
-      return contains((Parameter) object);
-    } else if (object instanceof StateVar) {
-      return true; // FIXME: Should we disallow state vars at the top level?
-    }
-    throw new UnsupportedOperationException(
-        "EObject class " + object.eClass().getName() + " not supported.");
+  public boolean inherits(ReactorDecl declaration) {
```
edwardalee, line 218: I'm confused. What does the word "inherits" mean here?  Does it have anything to do with subclassing? Please clarify the documentation and consider renaming the method.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -220,79 +204,87 @@ public Instantiation getInstantiation() {
...
+  private final MessageReporter messageReporter;
+
   /**
-   * Return true if the specified EObject should be included in the code generated for this
-   * federate.
+   * Return {@code true} if federate instance inherits the given reactor declaration.
    *
-   * @param object An {@code EObject}
-   * @return True if this federate contains the EObject.
+   * @param declaration The reactor declaration to inherit.
```
edwardalee, line 216: This comment suggests that the function causes inheritance to occur, but the function is a predicate. I don't understand what you mean by inheritance here. A reactor inherits something if that something is defined in a reactor that it extends. But this seems to have nothing to do with that.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -220,79 +204,87 @@ public Instantiation getInstantiation() {
...
     return false;
   }
 
   /**
-   * Return true if the specified parameter should be included in the code generated for this
-   * federate.
+   * Return {@code true} if this federate inherits the given parameter.
    *
-   * @param param The parameter
+   * @param param The parameter to inherit.
```
edwardalee, line 283: Again, this function doesn't cause inheritance to occur.

edwardalee (changes requested):
I'm confused about the use of the word "inherits" here.  You changed all the `contains` methods to `inherits` with comment changes like this: from
```
   * Return true if the specified timer should be included in the code generated for the federate.
    * This means that the timer is used as a trigger in a top-level reaction that belongs to this
    * federate. This also returns true if the program is not federated.
```
to
```
    * Return {@code true} if this federate inherits the given timer.
```
But these don't mean the same thing. Inheriting a timer means that the timer is defined in a class that this reactor extends. But that does not seem to be what is being checked.  Perhaps a different word is needed here?

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
lhstrh: Because it's unnecessary. The default initialization is zero.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
lhstrh: I've been struggling with finding the most appropriate name, but I landed on `inherits` because I think that's closest to what happens here.

In my view, there is an original file that defines the federation, and the federate inherits part of the definitions in it. Take for instance the method for determining whether a top-level reaction (i.e., completely one outside of the definition of the federate instance) should be associated with the federate or not; how is it appropriate to call this method `contains` and why doesn't `inherits` cover the meaning of the predicate? Another alternative is `adopts` which avoids the confusion with class inheritance. I just felt that `contains` wasn't good...

The meat of the refactoring, however, was the removal of the public "catch all" overload that used reflection to dispatch to private implementations. This is an anti-pattern because it defers type checking to runtime. Java has overloading, so this was completely unnecessary.

edwardalee: Yes, using overloading is a good improvement. But "inherents" I find very confusing. How about "includes"?

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -392,19 +386,19 @@ private boolean contains(Reaction reaction) {
   }
 
   /**
-   * Return true if the specified timer should be included in the code generated for the federate.
-   * This means that the timer is used as a trigger in a top-level reaction that belongs to this
-   * federate. This also returns true if the program is not federated.
+   * Return {@code true} if this federate includes the given timer from the top-level of the
+   * federation, which is necessary when the federate includes a reaction that uses the given timer.
```
edwardalee, line 390:
```suggestion
   * Return {@code true} if this federate include a top-level reaction that uses the given
   * timer as a trigger.
```

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -392,19 +386,19 @@ private boolean contains(Reaction reaction) {
...
    *
-   * @return True if this federate contains the action in the specified reactor
+   * <p>Specifically, this means that either the timer is used as a trigger in a top-level reaction
+   * that is included by this federate.
```
edwardalee, line 393:
```suggestion
```

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -392,19 +386,19 @@ private boolean contains(Reaction reaction) {
...
+   *
+   * @param timer The action to check if it is to be included.
```
edwardalee, line 395:
```suggestion
   * @param timer The timer to check whether it is to be included.
```

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -415,26 +409,18 @@ private boolean contains(Timer timer) {
...
   /**
-   * Return true if the specified reactor instance or any parent reactor instance is contained by
-   * this federate. If the specified instance is the top-level reactor, return true (the top-level
-   * reactor belongs to all federates). If this federate instance is a singleton, then return true
-   * if the instance is non null.
+   * Return {@code true} if this federate instance includes the given instance.
    *
-   * <p>NOTE: If the instance is bank within the top level, then this returns true even though only
-   * one of the bank members is in the federate.
+   * <p>NOTE: If the instance is bank within the top level, then this returns {@code true} even
```
edwardalee, line 418:
```suggestion
   * <p>NOTE: If the instance is a bank within the top level, then this returns {@code true} even
```

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -415,26 +409,18 @@ private boolean contains(Timer timer) {
...
+   * though only one of the bank members is included in the federate.
    *
-   * @param instance The reactor instance.
-   * @return True if this federate contains the reactor instance
+   * @param instance The reactor instance to check if it is to be included.
```
edwardalee, line 421:
```suggestion
   * @param instance The reactor instance to check whether it is to be included.
```

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -453,7 +439,7 @@ public boolean contains(ReactorInstance instance) {
    * @param federatedReactor The top-level federated reactor
    */
   private void indexExcludedTopLevelReactions(Reactor federatedReactor) {
-    boolean inFederate = false;
+    boolean inFederate;
```
edwardalee, line 442:
```suggestion
    boolean inFederate = false;
```

edwardalee (approved):
LGTM, modulo some suggestions on comments and reintroducing initialization of local variables.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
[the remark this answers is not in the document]
lhstrh: What is the source of this information? To the best of my knowledge, it's wrong.

Variables of primitive types do not need explicit initialization as they have well-defined default values (e.g., `0` for `int`, `false` for `boolean`. Non-primitive (reference) types get initialized to `null`. A program with uninitialized non-primitive variables will compile fine, but your IDE will complain about it.

In this class, the variables are initialized in the constructor. This is why IntelliJ flags their initialization upon declaration in the class body as unnecessary.

lhstrh:
The PR was already on auto-merge, and your approval pushed it onto the merge queue. Next time, when making suggestions, please hit the "request changes" button, not the "approve" one :-) ...I can create a separate PR to make your suggested changes to the comments.

lhstrh:
> The PR was already on auto-merge, and your approval pushed it onto the merge queue. Next time, when making suggestions, please hit the "request changes" button, not the "approve" one :-) ...I can create a separate PR to make your suggested changes to the comments.

Alternatively, we could enable a feature that requires all conversations to be resolved prior to merging. I think it might be a good idea because that gives reviewers the ability to approve and still give suggestions for (optional) changes.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
lhstrh: Fixed.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -306,27 +299,28 @@ private boolean contains(Parameter param) {
...
   /**
-   * Return true if the specified action should be included in the code generated for this federate.
-   * This means that either the action is used as a trigger, a source, or an effect in a top-level
-   * reaction that belongs to this federate. This returns true if the program is not federated.
+   * Return {@code true} if this federate includes the given action from the top-level of the
+   * federation, which is necessary when the federate includes a reaction that references the given
+   * action.
+   *
+   * <p>Specifically, this means that either the action is used as a trigger, a source, or an effect
+   * in a top-level reaction that is included by this federate.
```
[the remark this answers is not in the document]
lhstrh: Fixed.

PR #1894, opened 2023-07-08
C math lib always linked for C target

cmnrd:
This PR replaces https://github.com/lf-lang/lingua-franca/pull/1121. Instead of adding a new target property that is likely to cause various problems (as it is platform dependent) and that is difficult to maintain, this PR simply always links the math lib. To add other libraries, we can use the `cmake-include` property.

This PR also removes the magic matching of compiler flags in the C generator.

Replaces https://github.com/lf-lang/lingua-franca/pull/1121
Replaces https://github.com/lf-lang/benchmarks-lingua-franca/pull/11
Fixes https://github.com/lf-lang/lingua-franca/issues/1120

### DIFF
--- a/core/src/main/java/org/lflang/TargetConfig.java
+++ b/core/src/main/java/org/lflang/TargetConfig.java
@@ -184,9 +184,6 @@ public class TargetConfig {
    */
   public Map<String, String> compileDefinitions = new HashMap<>();
 
-  /** Additional libraries to add to the compile command using the "-l" command-line option. */
-  public List<String> compileLibraries = new ArrayList<>();
-
   /** Flags to pass to the compiler, unless a build command has been specified. */
   public List<String> compilerFlags = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


edwardalee (approved):
LGTM.  This might rule out some really esoteric embedded platforms that have no implementation of the math library, but I've not encountered one of those in decades, so we are probably OK.

cmnrd:
Right, it is also possible to make this conditional on specific platforms if needed.

lhstrh (approved):
I like this solution. There is an unrelated test failure in CI tied in with code coverage reporting that we should probably have a closer look at (maybe ignore such failures?).

lhstrh:
> I like this solution. There is an unrelated test failure in CI tied in with code coverage reporting that we should probably have a closer look at (maybe ignore such failures?).

I've put together a fix in #1895.

PR #1895, opened 2023-07-09
Small fixes to make CI more robust

lhstrh:
- [x] Determine when to report coverage based on `github.repository`
- [x] Replace the wretry action (which appears to be unreliable), with a custom retry mechanism based on multiple steps.

### DIFF
--- a/.github/workflows/cpp-tests.yml
+++ b/.github/workflows/cpp-tests.yml
@@ -70,4 +70,4 @@ jobs:
         uses: ./.github/actions/report-code-coverage
         with:
           files: reactor-cpp.info
-        if: ${{ !inputs.compiler-ref && matrix.platform == 'ubuntu-latest' }}  # i.e., if this is part of the main repo's CI
+        if: ${{ github.repository == 'lf-lang/lingua-franca' && matrix.platform == 'ubuntu-latest' }}
--- a/.github/workflows/cpp-tests.yml
+++ b/.github/workflows/cpp-tests.yml
@@ -53,7 +53,7 @@ jobs:
         uses: ./.github/actions/report-code-coverage
         with:
           files: core/build/reports/jacoco/integrationTestCodeCoverageReport/integrationTestCodeCoverageReport.xml
-        if: ${{ runner.os == 'Linux' }}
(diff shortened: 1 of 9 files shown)


cmnrd (changes requested):
Did you investigate the cause of the error in https://github.com/lf-lang/lingua-franca/pull/1894?

I don't think we should ignore codecov errors. If we do this, we don't notice if there are any problems in the reporting and whether we can trust the coverage report. Also see the discussion in https://github.com/lf-lang/lingua-franca/pull/1868.

I also think we should report on all platforms, not just Linux. We have some platform dependent code in our code base and will only get full coverage if we report on all platforms. https://github.com/lf-lang/lingua-franca/pull/1868 deliberately removed the platform check.

lhstrh:
Shall we then replace `if: ${{ github.repository == 'lf-lang/lingua-franca' && matrix.platform == 'ubuntu-latest' }}` with `if: ${{ github.repository == 'lf-lang/lingua-franca' }}` and merge this?

lhstrh:
Note this failure: https://github.com/lf-lang/lingua-franca/actions/runs/5512578023/jobs/10049603441?pr=1895

I think this will become a major annoyance if we don't ignore failures, tbh.

cmnrd:
This failed, because this particular report (on reactor-cpp) only works on Linux. Here we need the check on the platform (Ubuntu), and this check was there before the chanhges in this PR.

.github/workflows/cpp-tests.yml
```diff
@@ -70,4 +70,4 @@ jobs:
         uses: ./.github/actions/report-code-coverage
         with:
           files: reactor-cpp.info
-        if: ${{ !inputs.compiler-ref && matrix.platform == 'ubuntu-latest' }}  # i.e., if this is part of the main repo's CI
+        if: ${{ github.repository == 'lf-lang/lingua-franca' }}
```
lhstrh, line 73:
```suggestion
        if: ${{ github.repository == 'lf-lang/lingua-franca' && matrix.platform == 'ubuntu-latest' }}
```

cmnrd (approved):

PR #1897, opened 2023-07-09
Removal of `REQUIREMENTS.md`

lhstrh:
We have a `REQUIREMENTS.md` document prominently featured in the root of our main repository, but it hasn't been maintained at all, meaning that a lot of information is outdated, incomplete, or simply irrelevant.

For Lingua Franca, the baseline is that a Java installation >= 17 (_strictly_ 17 for those who want to build the compiler from source) and the availability of normal development dependencies for the chosen target should allow one to use our compiler. Going into a lot of detail about what normal development dependencies are will likely only scare off potential users. I therefore propose to remove this file and instead focus on presenting useful docs in the [handbook](https://www.lf-lang.org/docs/handbook).


cmnrd (approved):

edwardalee (approved):
OK with me.

PR #1903, opened 2023-07-14
Minor fixes for the README in the test directory

lsk567:
As suggested in the title.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -15,6 +15,7 @@ The `targetTest` task is essentially a convenient shortcut for the following:
 ./gradlew core:integrationTest --tests org.lflang.tests.runtime.<target>Test.*
 ```
 If you prefer have more control over which tests are executed, you can also use this more verbose version.
+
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -8,14 +8,21 @@ The simplest way to run integration tests for a specific target is the `targetTe
 ```
 ./gradlew targetTest -Ptarget=Rust
 ```
-You can specify any valid target. If you run the task without specifying the target property `./gradlew tagetTest` it will produce an error message and list all available targets.
+You can specify any valid target. If you run the task without specifying the target property `./gradlew targetTest` it will produce an error message and list all available targets.
```
lhstrh, line 11:
```suggestion
You can specify any valid target. If you run the task without specifying the target property, `./gradlew targetTest` will produce an error message and list all available targets.
```

lhstrh (approved):
Looks good modulo grammar edit. Thanks, @lsk567!

PR #1907, opened 2023-07-21
Add smoke test for clock synchronization

petervdonovan:
This makes `DistributedCountDecentralized` also serve as a smoke test for clock synchronization. The test still runs quickly and seemingly reliably on my machine, so this seems like a harmless change. This is a response to our recently discovering that programs with clock synchronization haven't been building on Ubuntu.


lhstrh (approved):
Good idea to amend the test...

PR #1910, opened 2023-07-22
[IGNORE ME] remove-unordered-reactions debugging in CI

petervdonovan:


petervdonovan:
The bug that was being investigated here seems to have been resolved. The problem was probably just that the volume of debug output combined with the use of a physical action was causing nondeterministic behavior.

PR #1914, opened 2023-07-25
Python launch scripts

cmnrd:
This PR implements the solution proposed in https://github.com/lf-lang/lingua-franca/issues/1458#issuecomment-1640218120. Instead of printing a Python command to run, we generate a shell or batch script in `bin/` which can be executed to run the Python program. This ensures that the executing Python runtime is the same as the one we linked against during compilation.


core/src/main/java/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -578,7 +573,7 @@ private static String setUpMainTarget(
             add_subdirectory(core)
             set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR})
             set(LF_MAIN_TARGET <pyModuleName>)
-            find_package(Python 3.7.0...<3.11.0 COMPONENTS Interpreter Development)
+            find_package(Python 3.10.0...<3.11.0 REQUIRED COMPONENTS Interpreter Development)
```
edwardalee, line 576: Should this be <3.12.0?

edwardalee (approved):
Great!!  Thanks for fixing this!

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
cmnrd: No, this does specifically not add Python 3.11 support. For this we should finalize https://github.com/lf-lang/lingua-franca/pull/1902.

cmnrd: I tried what happens if we just enable Python 3.11 on top of this branch in https://github.com/lf-lang/lingua-franca/pull/1915, but we are getting segfaults. So we still need the fixes in https://github.com/lf-lang/lingua-franca/pull/1902.

cmnrd:
For some reason the Python CI now runs much longer. Has anyone an idea why that could be?

core/src/main/java/org/lflang/federated/launcher/PyBuildConfig.java
```diff
@@ -12,26 +12,12 @@ public PyBuildConfig(
...
-  public String localExecuteCommand() {
-    return "python3 "
-        + fileConfig.getSrcGenPath()
-        + "/"
-        + federate.name
-        + "/"
-        + federate.name
-        + ".py -i $FEDERATION_ID";
+  public String remoteExecuteCommand() {
+    return "bin/" + fileConfig.name + "_" + federate.name + " -i '$FEDERATION_ID'";
```
lhstrh, line 16: The slash here doesn't look platform independent, but the code that this replaces also featured slashes...

lhstrh (approved):

core/src/main/java/org/lflang/federated/launcher/PyBuildConfig.java
lhstrh: Created #1917.

PR #1916, opened 2023-07-25
Fix verifier error when there is no main reactor

lsk567:
As the title suggests, the verifier no longer complains when the program has no main reactor.


lhstrh (approved):
Thanks for the quick fix!

PR #1918, opened 2023-07-26
Zephyr: Move from newlibc to picolibc

erlingrj:
We originally used newlib, but they provided no implementation of the regexp.h (which they should?). However, picolibc does and is also smaller. This PR has Zephyr use picolibc instead. It required changing a bit of the CMake infrastructure. In particular we had to remove the calls to install() in the generated CMakeLists.txt


lhstrh (approved):
LGTM!

erlingrj:
Update: When moving to v3.4.0 and SDK 0.16.1 picolibc gets linker errors and the missing regexp.h in newlib is fixed. So closing this

PR #1923, opened 2023-07-27
Fix AST Transformation to Support Platform Target Property in Federates

arengarajan99:
Allows for Platform target property to not require every key-value pair to be defined as well as overloading the platform with a literal string.


cmnrd (changes requested):
Thanks for tacking a stab at this! I don't think that throwing an `AssertionError` is much of an improvement over the nullpointer exception, though. Ideally, we would report an error in the [validator](https://github.com/lf-lang/lingua-franca/blob/master/core/src/main/java/org/lflang/validation/LFValidator.java) when checking the target proberties.

erlingrj:
Thanks Ani! This fixes my issue.

PR #1924, opened 2023-07-28
Fix for edge case in which comments are dropped

petervdonovan:
Fixes [vscode-lingua-franca/133](https://github.com/lf-lang/vscode-lingua-franca/issues/133#issuecomment-1654465957).

In the issue, the comment `// test` would be completely erased from the following program by the formatter:
```
target C

main reactor Minimal {
  // test
}
```

This fixes that issue such that the program is instead formatted like this:
```
target C

// test
main reactor Minimal {
}
```

I realize that some people might expect the comment to remain inside the reactor, but I think that that would be too much work to implement without much obvious benefit. With the current design, the comment must be associated with some AST node in order to not be dropped; in this case, it is associated with the whole reactor (no other alternatives are available), and as a result, it goes above the reactor.

### DIFF
--- a/core/src/main/java/org/lflang/ast/ToLf.java
+++ b/core/src/main/java/org/lflang/ast/ToLf.java
@@ -125,9 +125,13 @@ public class ToLf extends LfSwitch<MalleableString> {
           .forEach(allComments::add);
     } else {
       Stream<String> precedingComments =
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
LGTM!

PR #1927, opened 2023-07-29
TypeScript runtime bumped to `v0.5.0`

byeong-gil:
This PR bumps the `reactor-ts` package to `0.5.0`.


lhstrh:
This is the right idea, but I think we should quickly do a release of reactor-ts and use the npm package instead.

lhstrh (approved):

Issue #1909, opened 2023-07-22
`--json` argument is not properly handled by `lfc`


lhstrh:
Example:
```
lfc --json="{\"src\":\"/home/marten/lf-lang/lingo/sandbox/src/Main.lf\"}"
```
=>
```
lfc: fatal error: Invalid JSON string:
 {src:/home/marten/lf-lang/lingo/sandbox/src/Main.lf}
```

lhstrh:
The problem here is that in a development setup, `lfc` goes through Gradle, which strips all the quotes from the JSON string.

When using `./build/install/lf-cli/bin/lfc` directly, this does not occur.

lhstrh:
The problem is something like [this](https://discuss.gradle.org/t/javaexec-strips-double-quotes-from-property-values-on-windows/15053) but I haven't been able to find a solution to it.

lhstrh:
Actually, we might be better off using stdin instead of CLI arguments to accept JSON data...

revol-xut:
I am confused about this because normally the json also needs to include a `out` and `properties` field.

Should lfc if `out` is not specified drop the files in the current working directory ?

lhstrh:
> I am confused about this because normally the json also needs to include a `out` and `properties` field.
> 
> 
> 
> Should lfc if `out` is not specified drop the files in the current working directory ? 

This is not about the contents of the json, this is about the json itself not being formatted properly. It turns out that Gradle messes with the quotes in the dev version of lfc...

axmmisaka:
for reference: https://discuss.gradle.org/t/javaexec-strips-double-quotes-from-property-values-on-windows/15053

PR #1929, opened 2023-08-01
Add delimeter to each argument, fixes 

axmmisaka:
**Might** be a fix for https://github.com/lf-lang/lingua-franca/issues/1909

This issue specifically is caused by gradle wrapper. Specifically, it uses `org.gradle.api.tasks.JavaExec.setArgsString`, but as the doc suggested, 

> Note: the parser does not support using backslash to escape quotes. If this is needed, use the other quote delimiter around it. For example, to pass the argument 'singly quoted', use "'singly quoted'". 

This is not a good solution imo, but I guess it could work as of now? This is not touching gradlew, so I guess it's not that bad.

There's a separate issue that our scripts are named `.sh` but it is not POSIX-shell compatible (it uses `[[`, `function`, and I added `extglob` and IFS in this commit). While we reasonably expect everyone to have bash, I'm unsure if that will be the case (e.g. openwrt only have `sh` and `busybox`, maybe some day openwrt will use LF??????)

Also refer https://github.com/gradle/gradle/issues/6056 .

### DIFF
--- a/util/scripts/launch.sh
+++ b/util/scripts/launch.sh
@@ -75,5 +75,9 @@ gradlew="${base}/gradlew"
 if [ $# -eq 0 ]; then
     "${gradlew}" -p "${base}" "cli:${tool}:run"
 else
-    "${gradlew}" -p "${base}" "cli:${tool}:run" --args="$*"
+    shopt -s extglob
+    # Split through IFS
+    arr=("$*")
+    # For magic in args, see https://stackoverflow.com/a/50932265/22274983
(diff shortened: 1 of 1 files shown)


lhstrh:
> There's a separate issue that our scripts are named `.sh` but it is not POSIX-shell compatible

Doesn't the `#!/bin/bash` address this? Or is your point that the requirement for Bash should be obvious from the file extension alone?

axmmisaka:
I discussed with @lhstrh a few weeks ago regarding not using wrappers to boot up lf in development. I think it would be best to close this as of now, as that is a much more elegant solution compared to dealing with delimetre passing with gradle and banging our head on the wall.

PR #1932, opened 2023-08-04
Fix IllegalArgumentException in diagram synthesis

### LINKED ISSUE #1930: Diagram Synthesis fails wih IllegalArgumentException when enabling "Reactor State Variables"
Steps to reproduce:
1. Create a program that contains state variables
```
target Cpp

main reactor Example {
    state foo: int = 42
}
```
2. Enable the "Reactor State Variables" option in VS code

petervdonovan:
This is a runtime type error -- this is my bad, as such things should not happen in well written Java code. This commit modifies `FormattingUtil.render` to make it slightly more type-safe. It also changes the diagram synthesis to bypass the regular formatting logic -- the formatting logic seems more heavyweight than necessary since in diagrams we should not usually have to worry about how to represent comments and where to break lines (comment representation is the reason why the target was needed by the FormattingUtil.render function).

Fixes #1930.


lhstrh (approved):
:+1:

PR #1933, opened 2023-08-08
Support for named and bodyless reactions in C++

cmnrd:


petervdonovan (approved):
It is remarkable how simple this is in comparison to the C target. LGTM!

lhstrh (approved):
Nice! I wish all feature-adding PRs were this simple!

test/Cpp/src/hello_bodyless_world.cc
```diff
@@ -0,0 +1,5 @@
+#include "HelloBodylessWorld/HelloBodylessWorld.hh"
+
+void HelloBodylessWorld::Inner::hello_body([[maybe_unused]] const reactor::StartupTrigger& startup) {
```
revol-xut, line 3: so its always reaction_name + _body ? is this _body necessary ?

revol-xut (approved):

lhstrh:
Can we include this in v0.5.0?

test/Cpp/src/hello_bodyless_world.cc
cmnrd: I don't think it is required, actually.  However, there will be C++ compile errors if a reactor uses e.g. a state variable and a reaction with the same name. We  don't have proper name conflict handling.

cmnrd: Opened https://github.com/lf-lang/lingua-franca/issues/1973

PR #1935, opened 2023-08-08
Fix for STP violation

petervdonovan:
This fixes yet another bug reported by @jackykwok2024. See the corresponding reactor-c PR for details.


lhstrh (approved):
Good idea to record the time spent for each test...

Issue #1946, opened 2023-08-15
Move the C-related tracing utilities to reactor-c


erlingrj:
After the discussion at todays Package meeting we concluded that the tracing utilities should be moved from the `util` directory of `lingua-franca` and into `reactor-c`. We also concluded that copying the executables into `lingua-franca/bin` is not the right place for them. We could use `/usr/local/bin` which is where the RTI is installed.

erlingrj:
The documentation at the website must also be updated

cmnrd:
Completed by https://github.com/lf-lang/lingua-franca/pull/1948, https://github.com/lf-lang/lf-lang.github.io/pull/151 and https://github.com/lf-lang/reactor-c/pull/259.

PR #1948, opened 2023-08-15
Move tracing utils and zephyr run scripts from util folder

erlingrj:
Addresses #1946 and #1947. Mainly removing all the tracing related utilities from `lingua-franca/util` and instead putting them in `reactor-c`. We also install them to `/usr/local/bin` by default (together with the RTI) but allow overriding of this.

Related PRs:
Docs: https://github.com/lf-lang/lf-lang.github.io/pull/151
reactor-c: https://github.com/lf-lang/reactor-c/pull/259


erlingrj:
This is ready for merge. The reason why some of the CI is still pending is that these tasks are moved to reactor-c. But the CI uses the ci.yml from master always so it will not take effect until it is merged.

@ChadliaJerad FYI we are moving the trace utils to reactor-c, Marten made me aware of that you are working on some of the tracing tools in another branch.

cmnrd (approved):
Thanks for addressing this @erlingrj!

I removed the tracing CI checks from the list of required checks.

edwardalee:
Merging this into master without merging [reactor-c 259](https://github.com/lf-lang/reactor-c/pull/259) means that the tracing utilities are no longer available at all.  Can we reverse this merge until we are ready to merge both?

ChadliaJerad:
> This is ready for merge. The reason why some of the CI is still pending is that these tasks are moved to reactor-c. But the CI uses the ci.yml from master always so it will not take effect until it is merged.
> 
> @ChadliaJerad FYI we are moving the trace utils to reactor-c, Marten made me aware of that you are working on some of the tracing tools in another branch.

Sure. I will handle this.

PR #1950, opened 2023-08-16
Let the reaction number begin at 0 instead of 1.

### LINKED ISSUE #1782: Reaction number in diagram doesn't match messages
The reaction number in the diagrams starts with 1, but at least in the C target, it starts with 0. This means there is an off-by-one mismatch between error messages and the diagram.  Can we change the diagram to start at 0?  Do other targets also start at 0?

soerendomroes:
This corresponds to their internal names.

Fixes #1782


cmnrd (changes requested):
This is a trivial change, but I think it has quite significant consequences. People are just used to seeing diagrams starting with reactions at 1 (personally I also find this more pleasant). 

As I understand this change is mostly motivated by a misalignment between the name of C reactions and the diagrams. This change, however, will introduce a misalignment with other targets that use the current numbering from the diagrams (like C++).

PR #1951, opened 2023-08-16
First prototype @layout annotation.

soerendomroes:


cmnrd (approved):
Thanks! I think this will be helpful. I'll play around with this and see if it resolves some of the "issues" I had lately.

Could you run `./gradlew spottlessJavaApply` to apply the code formatting and commit the changes? I would do it myself, but I don't think I can push to your fork.

PR #1953, opened 2023-08-17
Fix for unconnected multiport and bank reactor bug

OmerMajNition:
- Unconnected multiport has compilation error, we needed an iterator which was missing
- Bank reactor iterator was missing, causing bank reactor declaration failures


edwardalee (commented):
Can you provide tests for these?

OmerMajNition:
@edwardalee, I've added test file for your review please.

edwardalee:
> @edwardalee, I've added test file for your review please.

The test seems overly complex. Is there a minimal case that illustrates the problem this was addressing? Such a case was missing in the original issue.  Also, I don't think a test should be writing to an external log file.

OmerMajNition:
@edwardalee, I've added few more test files. This compilation error occurs on unconnected output port.
For multi-port there's an error and in case of multi-bank there another error.

This PR fixes both errors.

petervdonovan:
I spoke to Marten about this but I will re-post here so we are all on the same page: Since this PR was delayed, I independently encountered and fixed the same problem while this PR was pending. This PR might still be useful because it adds more test cases and may be a somewhat cleaner solution that what I did, so it might still be useful if anyone is interested in addressing the concerns raised and the merge conflicts.

test/C/src/MultiBankToMultiPort_UnconnectedOutput_Test.lf
lhstrh, line 1: This file already compiles/runs successfully using what's in `master`.

test/C/src/MultiBank_UnconnectedOutput_Test.lf
lhstrh, line 1: This file does _not_ compile using what's in `master`.

test/C/src/MultiPort_MultiBankTest_UnconnectedOutput_Test.lf
lhstrh, line 1: Also does not compile in `master`.

test/C/src/MultiPort_UnconnectedOutput_Test.lf
lhstrh, line 1: Builds and runs in `master`.

lhstrh (changes requested):
Could you please remove the two files that are already passing in `master` and reduce the other two down to minimal examples (or one, if they both end up testing the same thing)? The change in the code generator looks OK to me...

OmerMajNition:
@lhstrh, removed the test case files that main is now already taking care of.
Please review again and let me know if anything else required from my end

lhstrh:
I'm seeing some mysterious errors on Windows happening in CI? Not sure what's going on there...

core/src/main/resources/lib/c/reactor-c
lhstrh, line 1: I don't recall that this PR originally included changes to the runtime, but now it appears that it does. Could this be an erroneous check in of the `reactor-c` submodule?

OmerMajNition: yes, rebased with the latest submodule changes of master. Please review

lhstrh (changes requested):
Looks like the use of `rand_r()` is causing trouble on Windows. The tests should only include code that is strictly necessary to expose the bug that this PR is fixing...

test/C/src/MultiBank_UnconnectedOutput_Test.lf
lhstrh, line 1: @OmerMajNition, I simplified your [MultiPort_MultiBankTest_UnconnectedOutput_Test.lf](https://github.com/lf-lang/lingua-franca/pull/1953/files#diff-60ee6f86670963e765eeb0a9c57891f347fa5d335a448b38c7c8f11945f3ae23) considerably. Now looking at this test, is there anything `MultiBank_UnconnectedOutput_Test` captures that the other test doesn't? If so, please explain. If not, let's remove this additional test.

lhstrh: Nevermind this. It was easy enough to fix the other example. If this passes CI, let's merge.

lhstrh (approved):

test/C/src/MultiBank_UnconnectedOutput_Test.lf
OmerMajNition: Yes we can go ahead with one test. [MultiPort_MultiBankTest_UnconnectedOutput_Test.lf](https://github.com/lf-lang/lingua-franca/pull/1953/files#diff-60ee6f86670963e765eeb0a9c57891f347fa5d335a448b38c7c8f11945f3ae23) does test unconnected multiport issue which already stands fixed in master branch. If we already have a test case for it then we can ignore/remove it.

Issue #1940, opened 2023-08-11
IllegalArgumentException in ASTUtils


edwardalee:
The following program throws an IllegalArgumentException in ASTUtils.
Worse, the compilation claims to have succeeded, so running the code just runs the previous version.
The stack trace is:

```
java.lang.IllegalArgumentException: Parameter null is not a parameter of reactor instance _i1in0.
        at org.lflang.ast.ASTUtils.initialValue(ASTUtils.java:1327)
        at org.lflang.ast.ASTUtils.initialValueInt(ASTUtils.java:1412)
        at org.lflang.generator.ReactorInstance.initialIntParameterValue(ReactorInstance.java:467)
        at org.lflang.generator.PortInstance.setInitialWidth(PortInstance.java:424)
        at org.lflang.generator.PortInstance.<init>(PortInstance.java:83)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:869)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:882)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:94)
        at org.lflang.ast.ASTUtils.createMainReactorInstance(ASTUtils.java:604)
        at org.lflang.generator.c.CGenerator.setUpGeneralParameters(CGenerator.java:1966)
        at org.lflang.generator.c.CGenerator.doGenerate(CGenerator.java:397)
        at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:130)
        at org.lflang.federated.generator.FedGenerator.lambda$compileFederates$2(FedGenerator.java:330)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

```
/**
 * The Lingua Franca (LF) program above represents a distributed leader election system with three
 * nodes. FIXME
 *
 * @author Edward A. Lee
 * @author Marjan Sirjani
 */
target C {
    coordination: decentralized
}

reactor Node(
    id: int = 0,
    is_initial_primary: bool = false,
    num_nodes_minus_one: int = 2,
    heartbeat_period: time = 1s
) {
  input[num_nodes_minus_one] in: int
  output out: int
  
  initial mode Idle {
[...]

PR #1956, opened 2023-08-17
Fix bug when top-level multiport width in federation depends on parameter

petervdonovan:
This simply computes the concrete numeric value of the parameter at compile time.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
@@ -657,8 +657,15 @@ public class FedASTUtils {
     Input in = factory.createInput();
     in.setName("msg");
     in.setType(type);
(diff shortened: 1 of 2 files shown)


petervdonovan:
This PR uncovered #1957 by making some ports into multiports that were not previously multiports (because we went from no width spec to a width spec of `[1]`). This is why the PR is currently delayed.

petervdonovan:
I'm not sure what is going on with the TS federated tests, so I re-ran them. I will need to take another look before this can get merged.

edwardalee (approved):
LGTM!  Thanks!

edwardalee:
It looks like this also fixes #1940

However, it exposes a new bug: #1961

petervdonovan:
I think I fixed the problem with the failing TypeScript test as well as #1961. Let's see if the tests pass.

PR #1959, opened 2023-08-17
Fix for after delays that use user-provided declarations

### LINKED ISSUE #1958: Cannot use after delay with user-defined types
If you use a typedef to define type and import a file with the type defined in a file-scoped preamble, the type is not made available to the generated delay reactors.  This makes it very difficult (impossible?) to use after delays with user-defined types unless everything is one file.

To reproduce, put the following into a file `GenDelay.lf`:

```
target C

preamble {=
  #ifndef GEN_DELAY
  #define GEN_DELAY
  typedef int message_t;
  #endif
=}

reactor Source {
  output out: message_t
  reaction(startup) -> out {=
    lf_set(out, 42);
  =}
}

reactor Sink {
  input in: message_t
  reaction(in) {=
    lf_print("Received %d at time %lld", in->value, lf_time_logical_elapsed());
  =}
}
```

Then put this into another file, e.g. `GenDelayTest.lf`:

```
target C
import Source, Sink from "GenDelay.lf"

main reactor {
  source = new Source()
  sink = new Sink()
  source.out -> sink.in after 10 ms
}
```

The file fails to compile with this error:

```
llij/test/C/src-gen/GenDelayTest/./include/GenDelayTest/_lf_GenDelay_49429363.h:39:5: error: unknown type name 'message_t'
    message_t value;
    ^
/Users/eal/lingua-franca-intellij/test/C/src-gen/GenDelayTest/./include/GenDelayTest/_lf_GenDelay_49429363.h:55:5: error: unknown type name 'message_t'
    message_t value;
    ^
In file included from /Users/eal/lingua-franca-intellij/test/C/src-gen/GenDelayTest/__lf_gendelay_49429363.c:3:
/Users/eal/lingua-franca-intellij/test/C/src-gen/GenDelayTest/./__lf_gendelay_49429363.h:13:5: error: unknown type name 'message_t'
    message_t value;
```

petervdonovan:
Fixes #1958.

### DIFF
--- a/core/src/main/java/org/lflang/ast/DelayedConnectionTransformation.java
+++ b/core/src/main/java/org/lflang/ast/DelayedConnectionTransformation.java
@@ -104,6 +104,7 @@ public class DelayedConnectionTransformation implements AstTransformation {
           Type type = ((Port) connection.getRightPorts().get(0).getVariable()).getType();
           var resource =
               connection.getLeftPorts().size() > 0
+                      && connection.getLeftPorts().get(0).getContainer() != null
                   ? connection.getLeftPorts().get(0).getContainer().getReactorClass().eResource()
                   : mainResource;
           Reactor delayClass = getDelayClass(resource, type, connection.isPhysical());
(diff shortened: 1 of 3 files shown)


edwardalee (approved):
Nice! Thanks!

PR #1960, opened 2023-08-18
Handling of caching across runs using `gradle-build-action`

### LINKED ISSUE #1898: [CI] Fix caching on Windows hosts
We use caching on GitHub Actions to avoid re-downloading dependencies for every single run. Not only does this speed things up, it also avoids running into network problems like [this one](https://github.com/lf-lang/lingua-franca/actions/runs/5501332923/jobs/10024848076). The problem is that caching doesn't appear to work correctly on Windows. This results is us having to re-run jobs on occasion due to said network issues.

lhstrh:
Fixes #1898.

The default behavior of this action is to only write the cache on `main`/`master`, but this leads to cache misses and the associated network errors that occur when downloading stuff, so we allow writing to the cache in all branches currently. We should keep and eye on this and make adjustments down the road..

### DIFF
--- a/.github/actions/prepare-build-env/action.yml
+++ b/.github/actions/prepare-build-env/action.yml
@@ -9,3 +9,5 @@ runs:
         java-version: 17
     - name: Gradle Build Action
       uses: gradle/gradle-build-action@v2.8.0
+      with:
+        - cache-read-only: false
--- a/.github/actions/prepare-build-env/action.yml
+++ b/.github/actions/prepare-build-env/action.yml
@@ -10,4 +10,4 @@ runs:
     - name: Gradle Build Action
       uses: gradle/gradle-build-action@v2.8.0
       with:
-        - cache-read-only: false
+        cache-read-only: false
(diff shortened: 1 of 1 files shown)


axmmisaka (approved):
I love this, it uses steps that are officially supported, so that this CI step is concise, updated, and eliminated the need to reinvent wheels. If CI passes then I'm all for it!

.github/actions/prepare-build-env/action.yml
```diff
@@ -1,42 +1,11 @@
...
-    - name: Bring down Gradle daemon (Windows)
-      uses: webiny/action-post-run@3.0.0
-      id: post-run-command
-      with:
-        run: ./gradlew --stop
-      if: ${{ runner.os == 'Windows' }}
+        distribution: temurin
+        java-version: 17
+    - name: Gradle Build Action
+      uses: gradle/gradle-build-action@v2.8.0
```
lhstrh, line 11:
```suggestion
      uses: gradle/gradle-build-action@v2.8.0
      with:
        - cache-read-only: false
```

PR #1974, opened 2023-08-28
New job for building `epoch` in CI

### LINKED ISSUE #1614: CI check for compatibility with Epoch
While Epoch lives in its own repository and is built using Maven (not Gradle), the introduction of new dependencies in the Gradle build configuration need to be mirrored in the Maven configuration, or else Epoch will fail to build. To catch such problems early, we need to build Epoch in the CI triggered by pull requests for Lingua Franca.

The VS Code extension is more independent and should not fail to build regardless of what dependencies are required to build the language and diagram server that it uses.

lhstrh:
Fixes #1614.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           files: core/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/base/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfc/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfd/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lff/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml
 
-  build-epoch:
+  epoch:
     uses: lf-lang/epoch/.github/workflows/build.yml@main
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/all-misc.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     with:
       all-platforms: ${{ !github.event.pull_request.draft }}
 
-  # Run language server tests.
+  # Run language server tests. foo
```
lhstrh, line 41:
```suggestion
  # Run language server tests.
```

petervdonovan (approved):

.github/workflows/build.yml
```diff
@@ -30,3 +30,9 @@ jobs:
         uses: ./.github/actions/report-code-coverage
         with:
           files: core/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/base/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfc/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfd/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lff/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml
+
+  build-epoch:
+    uses: lf-lang/epoch/.github/workflows/build.yml@nightly-latest
```
lhstrh, line 35:
```suggestion
    uses: lf-lang/epoch/.github/workflows/build.yml@main
```

PR #1979, opened 2023-08-31
Minor fixes for C++ tests

revol-xut:
## Things Done

- replaced `int` in for loop and replaced it with `auto`


cmnrd:
What is the motivation for this change? Just to get rid of warnings, or is there something additional?

revol-xut:
Just the warnings

cmnrd (approved):
Great!

PR #1984, opened 2023-09-01
Avoid squeezing reaction, method, or preamble bodies onto a single line

### LINKED ISSUE #1975: Single line reactions should not be inlined
I'm finding the formatter's insistence on inlining single-line reactions problematic.  It results in two languages being on the same line and it's rather hard to see what is going on.  Consider the following very simple TypeScript example:

```
  reaction(t) -> i.x {= i.x = 42 =}
```

I find this hard to read.  Given that the formatter already inserts quite a few blank lines, this policy does not seem to be justified by reducing the number of lines (or else we should change that policy and not insert blank lines).  Can we change this to allow:

```
  reaction(t) -> i.x {=
    i.x = 42
  =}
```

petervdonovan:
Fixes  #1975.

This is in response to readability issues. Additionally, I personally find that when statements are put on one line, that makes refactoring more cumbersome because one must add newlines before adding additional statements to a reaction.

An example of a project that made a similar decision to this is `rustfmt`, which never seems to put if statements all on one line.

~~This change has a code smell -- it does matching on raw strings without parsing them. Also it only has effects in the case where the target language has semicolons, which may or may not be what we want. I am thinking about changing the implementation of this to be based on whether the code is a reaction body instead of whether it has a semicolon.~~ This is now resolved.

Since this change is late, we do not need to squeeze it into the 0.5.0 release. However, it might behoove us to release it soon, e.g. in 0.5.1, because the changes to the appearance of LF code could be highly disruptive; ideally users could transition straight from the 0.4.0 format to the 0.5.1 format without bothering with the 0.5.0 format in between.

### DIFF
--- a/cli/lff/src/test/java/org/lflang/cli/LffCliTest.java
+++ b/cli/lff/src/test/java/org/lflang/cli/LffCliTest.java
@@ -83,7 +83,9 @@ public class LffCliTest {
 
                     /** moo */
                     // this is a humbug reaction
-                    reaction(a) -> humbug {= /* it reacts like this*/ react react =}
+                    reaction(a) -> humbug {=
+                      /* it reacts like this*/ react react
+                    =}
                   }
                   """),
           List.of(
--- a/core/src/main/java/org/lflang/ast/FormattingUtil.java
+++ b/core/src/main/java/org/lflang/ast/FormattingUtil.java
@@ -56,7 +56,7 @@ public class FormattingUtil {
    * with the assumption that the target language is {@code target}.
    */
   public static String render(EObject object, int lineLength, Target target, boolean codeMapTags) {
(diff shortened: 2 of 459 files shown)


edwardalee (commented):
I think this change is a good one, but it doesn't solve the problem for TypeScript, since semicolons are optional.  Would it be possible for the formatter to simply accept whatever the user chooses in this regard?  I still think we should reformat all our examples and tests to not inline single-line reactions, however.  I've been doing that manually for all the examples in the Handbook (please do not run the formatter on them!!).

petervdonovan:
I undid the changes that I had originally implemented and changed to a completely different implementation that I think is less kludgy. I think this PR is ready now.

edwardalee (approved):
LGTM.  This should definitely be merged soon given the number of files changed.

Issue #1981, opened 2023-08-31
Add "square bracket list literal" to grammar


edwardalee:
I'm not sure to whom to assign this. Please help.

I'm trying to update [Expressions.md](https://www.lf-lang.org/docs/handbook/expressions?target=py#collections) for Python, and I have found several problems.  Consider the following program:
```
target Python

reactor Foo(param(1, 2, 3)) {
  state x(1, 2, 3)

  reaction(startup) {=
    print("parameter param = ", self.param);
    print("type is ", type(self.param));
    print("state x = ", self.x);
    print("type is ", type(self.x));
  =}
}

main reactor {
  f = new Foo(param(3, 4, 5))
}
```
According to the documentation, `param` is supposed to be a tuple and `x` is supposed to be a list. But they are both lists. The output is:

```
parameter param =  [3, 4, 5]
type is  <class 'list'>
state x =  [1, 2, 3]
type is  <class 'list'>
```

The formatter also messes up with the above code.  It gives these warnings:

```
3 | reactor Foo(param(1, 2, 3)) {
  |                  ^ This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
  |
4 |   state x(1, 2, 3)

lfc: warning: This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
--> code/c/src/Junk.lf:4:10
  |
3 | reactor Foo(param(1, 2, 3)) {]
4 |   state x(1, 2, 3)
  |          ^ This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
  |
5 | 

lfc: warning: This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
 --> code/c/src/Junk.lf:15:20
   |
14 | main reactor {
15 |   f = new Foo(param(3, 4, 5))
   |                    ^ This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
   |
16 | }

```

I try to fix the syntax, and the formatter changes it right back! (in VS code, it does this when I save the file).

If I fix the syntax in another editor to get this:

```
target Python

reactor Foo(param = (1, 2, 3)) {
  state x = (1, 2, 3)

  reaction(startup) {=
    print("parameter param = ", self.param);
    print("type is ", type(self.param));
    print("state x = ", self.x);
    print("type is ", type(self.x));
  =}
}

main reactor {
  f = new Foo(param = (3, 4, 5))
}
```

[...]

PR #1985, opened 2023-09-01
Suppress deprecation warnings so arrays work in Python

edwardalee:
To work around bug #1981, this PR allows the old syntax for array parameters and states in the Python target, which is the only syntax that currently works.

### DIFF
--- a/core/src/main/java/org/lflang/Target.java
+++ b/core/src/main/java/org/lflang/Target.java
@@ -516,7 +516,7 @@ public enum Target {
    * reactor constructor call, regardless of this method.
    */
   public boolean mandatesEqualsInitializers() {
-    return this != CPP;
+    return this != CPP && this != Python;
   }
 
   /** Allow expressions of the form {@code {a, b, c}}. */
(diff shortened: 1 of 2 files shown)


core/src/main/resources/lib/c/reactor-c
lhstrh, line 1: This probably shouldn't be here...

lhstrh (changes requested):
This workaround appears to also affect the validation behavior for the `Cpp` target, not just the `Python` target...

petervdonovan:
> the old syntax for array parameters and states in the Python target, which is the only syntax that currently works.

No one replied to @oowekyala's comment under the issue:

> I think we should recommend using funny braces to fix this situation.

Why is this not an acceptable near-term solution?

edwardalee:
Yes, this will work. I will verify and close this.

edwardalee:
I will update the docs to use syntax `param = {= [1, 2, 3] =}`, so we don't need these changes.

PR #1987, opened 2023-09-01
Test lf_set_array and persistent inputs

edwardalee:
I noticed that the lf_set_array function in the C target is not properly tested in the test suite.  This adds a test to test it.

This also adds a test of persistent inputs that matches the documentation.


lhstrh (changes requested):
Looks good, modulo formatting. Maybe also add a high level comment that explains what this is meant to do?

lhstrh (approved):

PR #1989, opened 2023-09-02
Set federates' bank index

### LINKED ISSUE #1962: Bank of federates with multiports have the wrong bank_index
The following program creates four federates, all with `bank_index` equal to 0 (should be 0, 1, 2, 3):

```
target C
reactor Node(bank_index: int = 0, num_nodes: int = 3) {
  input[num_nodes] in: int
  output[num_nodes] out: int
    
  reaction(startup) {=
    lf_print("Bank index is %d out of %d.", self->bank_index, self->num_nodes);
  =}
}
federated reactor {
  b = new[4] Node(num_nodes = 4)
  b.out -> b.in
}
```

Removing the ports results (incorrectly) in 16 instances of Node, triggering bug #1961.

The output from the above program is:

```
Federate 3: Bank index is 0 out of 4.
Federate 1: Bank index is 0 out of 4.
Federate 2: Bank index is 0 out of 4.
Federate 0: Bank index is 0 out of 4.
```

petervdonovan:
Fixes #1962.

This is based on #1984, which should be merged first.

### DIFF
--- a/core/src/main/java/org/lflang/ast/FormattingUtil.java
+++ b/core/src/main/java/org/lflang/ast/FormattingUtil.java
@@ -56,7 +56,9 @@ public class FormattingUtil {
    * with the assumption that the target language is {@code target}.
    */
   public static String render(EObject object, int lineLength, Target target, boolean codeMapTags) {
-    MalleableString ms = new ToLf().doSwitch(object);
+    var toLf = new ToLf();
+    toLf.setTarget(target);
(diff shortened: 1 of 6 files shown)


lhstrh (approved):
This looks good to me.

I supposed that `bank_id` is not necessarily the same as `federate_id` as there could be multiple banks?

petervdonovan:
Yes, that sounds right.

PR #1991, opened 2023-09-07
Cleanup of latest-release workflow

lhstrh:
This PR fixes a bug in the `latest-release.yml` workflow where the output of the workflow wasn't being set correctly. It also factors out the bulk of the functionality into a separate action, which can also be used directly.


axmmisaka (approved):
If it works it works...

PR #1993, opened 2023-09-07
Default hostname for RTI in dockerized federation changed from "localhost" to "rti"

byeong-gil:
Dockerized federation programs should use "rti" as the hostname of the RTI as default to get a proper IP address of the RTI.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
@@ -416,6 +416,14 @@ public class FedGenerator {
         && !federation.getHost().getAddr().equals("localhost")) {
       rtiConfig.setHost(federation.getHost().getAddr());
     }
+    // If the federation is dockerized, use "rti" as the hostname.
+    // FIXME: Do we have to check the condition rtiConfig.getHost().equals("localhost")? 
+    // In other words, do we have to preserve the user-specified IP address of the RTI 
+    // when the program is dockerized? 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -416,6 +416,13 @@ private void analyzeFederates(Reactor federation, LFGeneratorContext context) {
         && !federation.getHost().getAddr().equals("localhost")) {
       rtiConfig.setHost(federation.getHost().getAddr());
     }
+    // If the federation is dockerized, use "rti" as the hostname.
+    // FIXME: Do we have to check the condition rtiConfig.getHost().equals("localhost")?
```
lhstrh, line 420: This is a good question. For a federation that runs on a single machine, using `localhost` works just fine, but this doesn't work for Docker because it creates processes in their own container, each of which has their own distinct "localhost." I think your fix is correct.

lhstrh (changes requested):
Thank for pushing this fix, @byeong-gil! Please remove the `FIXME` and mark as "ready for review." I think we can go ahead and merge this...

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
byeong-gil: Thanks for answering, @lhstrh!

lhstrh (approved):

PR #1988, opened 2023-09-01
Gradlew not longer used to run dev version of lf cli tools

axmmisaka:
This PR fixes #1909.
 
Right now, we have two ways of executing `lfc` (*mutatis mutandis* for `lfd` and `lff`):
1. Use the script *built* by gradle located in `build/install/lf-cli/bin/`
2. Use `gradlew` to invoke `cli:lfc:run`, which is handled by `./bin/lfc`

The issue is these two methods handle strings differently - specifically, `gradlew` will parse strings again. If we do not distinguish them and use them interchangeably, we will get screwed (#1909 )

This PR renames the executables in `./bin` by adding a `-dev` suffix to them, to distinguish them from any installed artifacts that may already be on the `PATH`. In addition, the scripts have been simplified by not letting Gradle handle the invocation of `lfc`, `lfd`, etc. because this leads to problems like the one reported in https://github.com/lf-lang/playground-lingua-franca/pull/81 because Gradle swallows exit codes.

PR #1995, opened 2023-09-08
More robust dev scripts and removed util directory

cmnrd:
This PR removes the util directory. By removing the utli directory, I hope to reduce the incentive on dumping unfurnished code and experimental tools in the main LF repo.

The only content of util at the moment are the launch helper scripts. This PR removes the launch scripts and instead implements the logic directly in the `lf*-dev` scripts. This adds a bit of code duplication (the readlink logic), but also simplifies other aspects of the script. Overall, the new solution should be more robust.

This PR also adds back the capability to run the `lf*-dev` scripts from anywhere in the filesystem. This was removed by  #1988.

Finally, this PR also fixes bugs in the way we handled scripts before.


cmnrd:
It also turns out that there was a major bug in the readlink logic. I honestly don't know why this passed tests before.

.github/scripts/test-lfc.sh
```diff
@@ -18,3 +18,8 @@ bin/lfc-dev test/C/src/Minimal.lf
 
 # Ensure that lfc can be invoked via symbolic links.
 test_with_links "lfc-dev"
+
+# Ensure that lfc can be invoked from outside the root directory.
+cd bin
+./lfc-dev --help
```
lhstrh, line 24: Why not skip the `cd`s and use a relative path, like this?
```suggestion
./bin/lfc-dev --help
```

.github/scripts/test-lfc.sh
```diff
@@ -18,3 +18,8 @@ bin/lfc-dev test/C/src/Minimal.lf
 
 # Ensure that lfc can be invoked via symbolic links.
 test_with_links "lfc-dev"
+
+# Ensure that lfc can be invoked from outside the root directory.
```
lhstrh, line 22: Should we also add a test to execute outside of the dev tree? (`./bin` is still inside)

lhstrh (approved):
Modulo a minor comment about the tests, this looks like the right solution to me.

lhstrh:
> It also turns out that there was a major bug in the readlink logic. I honestly don't know why this passed tests before.
> 
> Edit: Actually, I do know: The tests passed before because both the launch scripts and the link structure crated in cli-test is 2 levels deep. It resolved the wrong path, but the end result was correct coincidentally...

:open_mouth:

PR #2002, opened 2023-09-13
Test fix for STP offset violation with centralized coordination

edwardalee:
This adds a test for the problem fixed by [PR 269 in the reactor-c repo](https://github.com/lf-lang/reactor-c/pull/269).


petervdonovan (approved):
Looks good. I waffled over the length of the MRE but since it is hard to further minimize it this seems like the right thing to do.

lhstrh:
Looks like we have a new test failure:
```
   Failed: src/federated/FeedbackDelaySimple.lf in 300.00 seconds

    -----------------------------------------------------------------------------
    Reason: Test timed out.
```

edwardalee:
This rather complicated test has been subsumed by the much simpler `src/federated/AfterDelays.lf` test.

Issue #1675, opened 2023-03-28
Remove deprecated initializer syntax


cmnrd:
[`state foo: int(42)`](https://github.com/lf-lang/lingua-franca/pull/1580) introduced a new initializer syntax. Currently, the validator produces a warning if the old initializer syntax is used (e.g., `state foo: int(42)` instead of `state foo: int = 42`). While this is a good solution for the transition phase, we should convert the warning into an error in a later release.

The only exception is currently the C++ target, where the old syntax is still legal. We should also carefully consider if this syntax is required and completely remove it, if not.

lhstrh:
@cmnrd, could you outline the steps required to get this done?

PR #2003, opened 2023-09-13
Bracket list expression for initialization without escaping

### LINKED ISSUE #1981: Add "square bracket list literal" to grammar

oowekyala:
Add an expression kind for Python and TS lists, and Rust arrays. 

Fix #1981 

This requires an update to the website.


lhstrh:
Fantastic! This is _really_ nice, @oowekyala. It occurs to me that even if we were to aim for a clearer separation between target code and LF code, initializers are probably the places where it would remain to pay off to elegantly mix languages (i.e., without escaping)...

lhstrh:
Looks like our `epoch` workflow doesn't deal well with forks. It's failing because it is attempting to resolve the ref `clem.bracket-list-expr` that corresponds to this PR in the `lf-lang` repo instead of your fork... I guess the workflow should take as input a string that identifies the fork and then just replace the `lingua-franca` submodule before checking out the given ref...

lhstrh:
> Looks like our `build-epoch` workflow doesn't deal well with forks. It's failing because it is attempting to resolve the ref `clem.bracket-list-expr` that corresponds to this PR in the `lf-lang` repo instead of your fork... I guess the workflow should take as input a string that identifies the fork and then just replace the `lingua-franca` submodule before checking out the given ref...

OK, fixed this!

.github/workflows/build.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           files: core/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/base/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfc/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lfd/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml,cli/lff/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml
 
   epoch:
-    uses: lf-lang/epoch/.github/workflows/build.yml@main
+    uses: lf-lang/epoch/.github/workflows/build.yml@build-against-fork
```
lhstrh, line 35: Change back to `main` after https://github.com/lf-lang/epoch/pull/53 has been merged.

lhstrh, line 35:
```suggestion
    uses: lf-lang/epoch/.github/workflows/build.yml@main
```

oowekyala:
Thanks Marten for fixing the CI!

cmnrd (dismissed):
This looks great! The only request I have, is that more Python tests should be updated. For instance `ArrayAsParamenter.lf` still uses the old syntax. There might be more.

This is beyond this PR, but we should soon remove the old `name(<expr>*)` syntax (https://github.com/lf-lang/lingua-franca/issues/1675).

Issue #1806, opened 2023-06-01
Use linter and enforce warning free code in CI


cmnrd:
Now that we have formatting addressed, we should also introduce linting and enforce that the code is warning free. Gradle has a convenient plugin called [checkstyle](https://docs.gradle.org/current/userguide/checkstyle_plugin.html) for linting java code. It supports various style guide rule sets, including Google's style guide. This should be a tremendous help in avoiding the most common errors such as the problems fixed in https://github.com/lf-lang/lingua-franca/pull/1793. The biggest challenge, however, will be to fix all linter warnings in our existing code.

lhstrh:
Would be cool to set this up for `reactor-c` as well...

PR #2004, opened 2023-09-13
Apply checkstyle plugin

cmnrd:
This enables the checkstyle plugin for gradle using the google style template.

As it turns out, the checkstyle tool isn't quite the tool that I was after when I wrote https://github.com/lf-lang/lingua-franca/issues/1806. Checkstyle is purely a style checker and it is quite pedantic :sweat_smile:. For instance, these are the warnings produced for `TimeValue.java`:
```
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:3: Empty line should be followed by <p> tag on the next line. [JavadocParagraph]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:6: Empty line should be followed by <p> tag on the next line. [JavadocParagraph]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:9: Empty line should be followed by <p> tag on the next line. [JavadocParagraph]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:13: Empty line should be followed by <p> tag on the next line. [JavadocParagraph]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:77:3: Missing a Javadoc comment. [MissingJavadocMethod]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:91:5: switch without "default" clause. [MissingSwitchDefault]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:142:5: 'if' construct must use '{}'s. [NeedBraces]
[ant:checkstyle] [WARN] /home/cmenard/projects/lf/lingua-franca/core/src/main/java/org/lflang/TimeValue.java:206:13: Abbreviation in name 'isThisUnitSmallerThanBUnit' must contain no more than '1' consecutive capital letters. [AbbreviationAsWordInName]
```

I think this is really nice to enforce a more consistent style, but I do think that we have more pressing problems at the moment than fixing all those style warnings. Therefore, I think we shouldn't merge this PR for now. I put the PR here in hope that it will be useful at a later point.

 I will open another PR that uses the spotbugs tool, which is a static code analysis tool and produces much more useful warnings.


lhstrh:
> I think this is really nice to enforce a more consistent style, but I do think that we have more pressing problems at the moment than fixing all those style warnings. Therefore, I think we shouldn't merge this PR for now. I put the PR here in hope that it will be useful at a later point.

Agreed!

PR #2005, opened 2023-09-13
Apply spotbugs plugin

cmnrd:
This PR applies the [spotbugs](https://spotbugs.github.io/) gradle plugin. spotbugs is a static code analysis tool that can detect a lot of bug patterns in Java (and Kotlin).

This is a selection of the type of issues reported by spotbugs:
```
H V MS: org.lflang.generator.GeneratorResult.NOTHING isn't final but should be  At GeneratorResult.java:[line 14]
H C HE: org.lflang.TimeValue doesn't define a hashCode() method but it is used in a hashed context in org.lflang.federated.generator.FederateInstance.stpToNetworkActionMap  In FederateInstance.java
M P UuF: Unused field: org.lflang.analyses.c.CToUclidVisitor.unchangedTriggers  In CToUclidVisitor.java
M V EI2: new org.lflang.federated.generator.FedEmitter(FedFileConfig, Reactor, MessageReporter, RtiConfig) may expose internal representation by storing an externally mutable object into FedEmitter.messageReporter  At FedEmitter.java:[line 29]
M D NP: Possible null pointer dereference in org.lflang.generator.cpp.CppStandaloneGenerator.generatePlatformFiles() due to return value of called method  Dereferenced at CppStandaloneGenerator.kt:[line 38]
M D BC: Questionable cast from Collection to abstract class java.util.List in org.lflang.generator.cpp.CppStandaloneGenerator.compareVersion(String, String)  At CppStandaloneGenerator.kt:[line 187]
M B FS: Format string should use %n rather than \n in org.lflang.ast.FormattingUtil.lineWrapComment(String, int, String)  At FormattingUtil.java:[line 172]
M P SBSC: org.lflang.analyses.uclid.UclidGenerator.generateActionAxioms() concatenates strings using + in a loop  At UclidGenerator.java:[line 987]
```

Spotbugs is run automatically as part of the `check` and `build` gradle tasks. The plugin is currently configured to treat all issues that it finds as warnings, not errors. Thus, the build succeeds although spotbugs reports problems.

Fixing all issues reported by spotbugs will keep us busy for a while. To keep this manageable, I think we should fix warnings in several smaller PRs and treat found issues as errors as soon as we have cleaned up the codebase.

This is a first step towards resolving https://github.com/lf-lang/lingua-franca/issues/1806.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -13,6 +13,8 @@ dependencies {
     // https://mvnrepository.com/artifact/com.diffplug.spotless/spotless-lib-extra
     implementation group: 'com.diffplug.spotless', name: 'spotless-lib-extra', version: spotlessLibVersion
 
+    implementation group: 'com.github.spotbugs.snom', name: 'spotbugs-gradle-plugin', version: spotbugsPluginVersion
+
     implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
     implementation "com.diffplug.spotless:spotless-plugin-gradle:$spotlessVersion"
 
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -3,3 +3,4 @@ spotlessVersion=6.11.0
 spotlessLibVersion=2.30.0
 kotlinVersion=1.6.21
 shadowJarVersion=7.1.2
+spotbugsPluginVersion=5.1.3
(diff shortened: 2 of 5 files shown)


lhstrh:
This is a great idea. Let's go ahead and fix those bugs/warnings!

cmnrd:
I spend some time today fixing issues found by spotbugs. Some are quite easy to fix, but for others the actual issue is not so obvious and the problem is actually a bad design. Thus, spotbugs actually helps to improve the design by a lot, but it also takes a while to fix the reported issues.

I think that fixing all issues as part of this PR is not feasible. It would create a humongous PR that would be difficult to keep up to date and merge. Thus, I advocate for merging this in now, treating the reported issues as warnings. Then we can fix issues in independent PRs and eventually switch back to treating issues as errors once we cleaned up the code base.

lhstrh:
Sounds like the right approach to me as well.

lhstrh (approved):

PR #2006, opened 2023-09-14
Clean up of test code and fix all test related spotbugs warnings

cmnrd:
This is a step towards https://github.com/lf-lang/lingua-franca/issues/1806.

This builds on https://github.com/lf-lang/lingua-franca/pull/2005 and should only be merged after.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -13,6 +13,8 @@ dependencies {
     // https://mvnrepository.com/artifact/com.diffplug.spotless/spotless-lib-extra
     implementation group: 'com.diffplug.spotless', name: 'spotless-lib-extra', version: spotlessLibVersion
 
+    implementation group: 'com.github.spotbugs.snom', name: 'spotbugs-gradle-plugin', version: spotbugsPluginVersion
+
     implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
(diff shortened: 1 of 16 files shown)


cmnrd:
@a-sr, @soerendomroes I tried to add the dependencies on `com.github.spotbugs:spotbugs-annotations` and `net.jcip:jcip-annotations` to the Epoch maven configuration, but I was not able to get it to work. I suspect this also needs to be added to the target platform. But how? Could you help with this?

a-sr:
https://github.com/lf-lang/epoch/pull/54 adds the jcip and spotbugs dependency to the target platform and the LF plugin.
The PR will not update the LF subrepo. I don't know what is the best strategy in this case.

cmnrd:
Thanks to Alex the Epoch build is now fixed. As far as I am concerned, this is ready to be merged.

lhstrh (approved):
This looks good to me! The usage of `StandardCharsets.UTF_8` seems a bit verbose to me; is that really the way proper Java ought to be written these days?

cmnrd:
Yeah, I am not sure about those, actually. The problem is, that there is no well-defined default. So if you write a file and then read it back in on another machine, you may end up with garbage. This is why it is better to define the encoding we actually want. However, in our case, we don't really know the encoding used in the pipes to external processes. So maybe it would be safer to use the default... but I don't know, and I figured we could try if using UTF-8 throughout causes any troubles.

PR #2007, opened 2023-09-14
Usage of high-resolution counter device on Zephyr boards. Update to v3.4.0

erlingrj:
This PR does three things:

1. Dont generate CMake code whiich looks for Threads library and links with it. Fails for certain Zephyr boards.
2. Add a bunch of Device Tree Config overlays for different Zephyr boards to enable that we get high-resolution timers for these boards. The DTC overlays enable a timer device.
3. Add a few tests for specific boards. Particularily, the boards on which we support the high-resolution timers. We just compile these simple programs to see that we indeed was able to enable a Timer device and get all the drivers etc. working.
4. Move to Zephyr v3.4.0


lhstrh:
Could you please provide a description? I have no idea what this does :smile:

erlingrj:
@lhstrh yes! I also added some more fixes to this PR. Most importantly I added some test programs that we compile for specific boards, particularily those boards where we have support for "high-resolution timer", this support has caused some issues since it is not just using the Zephyr kernel API. 

Next step would be buying a bunch of boards and also flashing them and seeing that they can accuratly toggle GPIOs.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "org.lflang/src/lib/c/reactor-c"]
 	path = core/src/main/resources/lib/c/reactor-c
-	url = https://github.com/lf-lang/reactor-c.git
+	url = git@github.com:lf-lang/reactor-c.git
```
lhstrh: This is not a good idea because anyone who checks out the repo and does not have public-key auth set up will run into trouble. This is why we use `https` for all our submodules.

core/src/main/resources/lib/platform/zephyr/boards/nrf52dk_nrf52832.overlay
```diff
@@ -0,0 +1,3 @@
+&timer1 {
+	status = "okay";
+};
```
lhstrh, line 3:
```suggestion
};

```

lhstrh (changes requested):
This looks great! Please revert the submodule protocol change and other than that this looks ready to merge...

lhstrh (approved):
Thanks for putting emphasis on testing! This looks great.

PR #2008, opened 2023-09-15
Refactoring of target properties and their validation

### LINKED ISSUE #1558: Cleanup `TargetConfig`
Currently, there are some target and federation-specific items in `TargetConfig` that do not belong there. They need to be moved elsewhere.

### LINKED ISSUE #1859: Print warnings in `lfc` even if there are errors
The message printed when there exists ambiguity in the program (see #1853) could arguably be an warning instead of an error, but it turns out warnings are not shown by `lfc`. We should fix this.

### LINKED ISSUE #1919: Unrecognized platform causes NPE
I tried to compile a pico example without having the branch implementing the platform support installed, and got the following error:
```
   java.lang.NullPointerException: Cannot invoke "org.lflang.TargetProperty$Platform.toString()" because "this.targetConfig.platformOptions.platform" is null
           at org.lflang.generator.c.CGenerator.pickCompilePlatform(CGenerator.java:902)
           at org.lflang.generator.c.CGenerator.setUpGeneralParameters(CGenerator.java:2007)
           at org.lflang.generator.c.CGenerator.doGenerate(CGenerator.java:397)
           at org.lflang.generator.LFGenerator.doGenerate(LFGenerator.java:115)
           at org.lflang.cli.Lfc.invokeGenerator(Lfc.java:187)
           at org.lflang.cli.Lfc.doRun(Lfc.java:150)
           at org.lflang.cli.CliBase.run(CliBase.java:145)
           at org.lflang.cli.CliBase.doExecute(CliBase.java:117)
           at org.lflang.cli.CliBase.cliMain(CliBase.java:109)
           at org.lflang.cli.Lfc.main(Lfc.java:137)
           at org.lflang.cli.Lfc.main(Lfc.java:127)
```

There should be a validator check that produces a graceful error if the specified platform is not supported.

### LINKED ISSUE #2001: StateSpaceExplorer.java uses unchecked or unsafe operations.
Warning thrown during compilation.

lhstrh:
This PR reworks the way we deal with target properties. It makes them easier to specify, manipulate, and verify. It cuts down on the use of lambda functions, which makes the code a bit more verbose, but also more reusable and more readable. The `TargetProperty` class that had become a depository of random helper classes has been cleaned and brought down from 1800 LOC to less than 300.

While the goal of this refactoring is to retain the same functionality, some changes were made in the name of enforcing consistency:
 - [x] `--target-compiler` CLI argument changed to `--compiler` so that it matches the `compiler` target property
 - [x] error reporting changed to report more precisely (on the `Element` instead of `KeyValuePair`) and with simpler messages
 - [x] #1919
 - [x] #2039
 - [x] let TypeScript backend honor `--no-compile` flag/target property
 - [x] let Rust and C++ use the same build type (`Debug`)
 - [x] register target properties in `Target` class and revise "supported by" mechanism
 - [x] add method for parsing from JSON (for interop with `lingo`)
 - [x] base CLI args directly on target property names
 - [x] clear separation between configurators and transformers in the test framework (c96a794)
 - [x] add comments

Fixes #1558.
Fixes #2001.
Fixes #1919.
Fixes #1859.


core/src/main/java/org/lflang/TargetConfig.java
```diff
@@ -25,24 +25,46 @@
 package org.lflang;
```
cmnrd, line 26: Should we move this class and also `TargetProperty` and `TargetPropertyConfig` to the `target` package?

core/src/main/java/org/lflang/target/property/DockerProperty.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class DockerProperty extends TargetPropertyConfig<DockerOptions> {
+
+  public DockerProperty() {
+    super(UnionType.DOCKER_UNION);
+  }
+
+  @Override
+  public DockerOptions initialValue() {
+    return new DockerOptions(false);
```
cmnrd, line 29: I don't know why, but it seems like docker is enabled by default. When I run lfc on a C file, it prints:
```
lfc: info: Dockerfile written to /home/cmenard/projects/lf/lingua-franca/test/C/src-gen/Minimal/Dockerfile
lfc: info: #####################################
To build and run:
    pushd /home/cmenard/projects/lf/lingua-franca/test/C/src-gen/Minimal && docker compose up --build
To return to the current working directory afterwards:
    popd
#####################################
```

core/src/main/java/org/lflang/TargetProperty.java
```diff
@@ -792,165 +248,58 @@ else if (paths.size() == 1) {
...
-      PropertyParser updater) {
-    this.description = description;
-    this.type = type;
-    this.supportedBy = supportedBy;
-    this.getter = getter;
-    this.setter = setter;
-    this.updater = updater;
+  public static TargetPropertyConfig<?> getPropertyInstance(TargetProperty p) {
+    try {
+      return p.propertyClass.newInstance();
```
cmnrd, line 286: `newInstance` is deprecated.

core/src/main/java/org/lflang/TargetProperty.java
```diff
@@ -1012,6 +372,51 @@ public static TargetDecl extractTargetDecl(Target target, TargetConfig config) {
...
+      KeyValuePairs targetProperties, TargetProperty property) {
+    List<KeyValuePair> properties =
+        targetProperties.getPairs().stream()
+            .filter(pair -> pair.getName().equals(property.toString()))
+            .toList();
+    assert properties.size() <= 1;
+    return properties.size() > 0 ? properties.get(0) : null;
+  }
+
+  public static KeyValuePair getKeyValuePair(Model ast, TargetProperty property) {
```
cmnrd, line 385: The `getKeyValuePair` methods shouldn't be used anymore, right?

cmnrd (commented):
This is a great improvement! I haven't looked carefully at all the changes yet, but I left some comments in places where I noticed issues or had questions.

core/src/main/java/org/lflang/TargetConfig.java
lhstrh: Yes, good idea.

core/src/main/java/org/lflang/target/property/DockerProperty.java
lhstrh: Yep, still a bunch of bugs to squash...

core/src/main/java/org/lflang/TargetProperty.java
lhstrh: It is still used in some `validate` methods to look up key-value pairs of _other_ target properties (e.g., because they might be incompatible).

core/src/main/java/org/lflang/target/property/CompilerProperty.java
```diff
@@ -0,0 +1,12 @@
+package org.lflang.target.property;
+
+import java.util.List;
+import org.lflang.Target;
+
+public class CompilerProperty extends AbstractStringConfig {
+
+  @Override
+  public List<Target> supportedTargets() {
+    return Target.ALL;
```
cmnrd, line 10: Is this property also supported by the TypeScript target?

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  @Override
+  protected PlatformOptions fromString(String string, MessageReporter reporter) {
+    throw new UnsupportedOperationException("Not supported yet.");
+  }
+
+  @Override
+  public List<Target> supportedTargets() {
+    return Target.ALL;
```
cmnrd, line 86: afaik, this is only supported by the C target.

core/src/main/java/org/lflang/target/property/TracingProperty.java
```diff
@@ -0,0 +1,169 @@
...
+    super(UnionType.TRACING_UNION);
+  }
+
+  @Override
+  public TracingOptions initialValue() {
+    return new TracingOptions(false);
+  }
+
+  @Override
+  public TracingOptions fromAst(Element node, MessageReporter reporter) {
```
cmnrd, line 35: There is a challenge in properties like tracing. In C++, the tracing option is simply considered to be a boolean. In C, however, there is additional configuration possible (i.e. specifying a file). As far as I understand the code now, there would be no warning or error when specifying a trace file in C++ (which was probably already the case before). But how can we handle these target specific sub properties?

We have the same problem with the platform property. I think it would make sense to support this one also in C++ (and remove the dedicated `ros2` property), but all the platform options like baudrate are quite useless in this context.

It seems like the "Options" of a property, should also be properties. We could parse and validate them nested.
It seems like the "Options" should also be properties

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -94,8 +94,8 @@ class TSGenerator(
 
     init {
         // Set defaults for federate compilation.
-        targetConfig.compiler = "gcc"
-        targetConfig.compilerFlags.add("-O2")
+        targetConfig.compiler.override("gcc")
```
cmnrd, line 97: Why is the C compiler set in TypeScirpt?

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -123,15 +123,15 @@ class TSGenerator(
...
+        if (targetConfig.dockerOptions.get().enabled) {
             val dockerData = TSDockerGenerator(context).generateDockerData();
             dockerData.writeDockerFile()
             DockerComposeGenerator(context).writeDockerComposeFile(listOf(dockerData))
         }
         // For small programs, everything up until this point is virtually instantaneous. This is the point where cancellation,
         // progress reporting, and IDE responsiveness become real considerations.
 
-        if (context.mode != LFGeneratorContext.Mode.LSP_MEDIUM && targetConfig.noCompile) {
+        if (context.mode != LFGeneratorContext.Mode.LSP_MEDIUM && targetConfig.noCompile.get()) {
```
cmnrd, line 134: `NoCompileProperty` does not return TS in `supportedTargets` method.

cmnrd (commented):
I made a complete pass over all the changes and pushed several fixes and adjustments. I also left a couple of comments, where I wasn't sure what needs to be done.

One problem with reviewing this, is that there might be more that should be changed (e.g. comparing a property to null), that of course don't show up in the diff. But I guess our only option is to trust in the tests.

Overall, I think this is a really good improvement!

I tried to figure out why the Python test fails, but couldn't. It looks like there is an actual bug in the runtime, that was masked before. Maybe, the automatic setting of `keepalive` in presence of a physical action wasn't working correctly before and hence the test passed. Without `keepalive` the test will just terminate immediately and not test anything.

core/src/main/java/org/lflang/generator/rust/RustTargetConfig.java
```diff
@@ -24,83 +24,39 @@
...
-   */
-  public List<Path> getRustTopLevelModules() {
-    return rustTopLevelModules;
-  }
+  private BuildType profile = BuildTypeProperty.BuildType.DEBUG;
 
   /** The build type to use. Corresponds to a Cargo profile. */
-  public BuildType getBuildType() {
+  public BuildType getBuildType(BuildTypeProperty cmakeBuildType) {
+    // FIXME: this is because Rust uses a different default.
```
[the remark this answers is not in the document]
lhstrh: Just tell me which one to use and I'll adjust the code :smile:

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
lhstrh: Nice catch!

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+    e.setKeyvalue(kvp);
+    if (kvp.getPairs().isEmpty()) {
+      return null;
+    }
+    return e;
+  }
+
+  /** Settings related to Platform Options. */
+  public static class PlatformOptions { // FIXME: use a record for this
```
cmnrd, line 145: FIXME

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -0,0 +1,262 @@
...
+    RP2040("Rp2040", false),
+    LINUX("Linux", true),
+    MAC("Darwin", true),
+    ZEPHYR("Zephyr", true),
+    WINDOWS("Windows", true);
+
+    final String cMakeName;
+
+    private boolean multiThreaded =
+        true; // FIXME: this is never read. If we set it, we can simplify the validator method in
```
cmnrd, line 201: FIXME

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
cmnrd: That brings me to another idea: Maybe we should at an assertion to the `get()` and `toString` (maybe also others) methods of the Property class and check that the property is enabled for the current target. Then we would be able to catch those bugs more easily during testing.

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -366,8 +365,8 @@ public boolean isOSCompatible() {
   public void doGenerate(Resource resource, LFGeneratorContext context) {
     // Set the threading to false by default, unless the user has
     // specifically asked for it.
-    if (!targetConfig.setByUser.contains(TargetProperty.THREADING)) {
-      targetConfig.threading = false;
+    if (!targetConfig.threading.isSet()) {
```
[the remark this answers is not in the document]
lhstrh: I agree that it's weird, but I've seen many such enigmatic workarounds sprinkled throughout the code base, without any proper explanation, too... I tried as much as possible for my changes to remain meaning-preserving, which is why I didn't address this. I don't think that `initialValue` should take a `Target` as a parameter because that information generally isn't available upon initialization...

core/src/main/java/org/lflang/target/property/CompilerProperty.java
lhstrh: Nope. Will change this.

core/src/main/java/org/lflang/target/property/PlatformProperty.java
lhstrh: 5d499c8bf

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
lhstrh: 876491c71

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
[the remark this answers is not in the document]
lhstrh: Interestingly enough, the failing test passes when I remove this code :shrug:

core/src/main/java/org/lflang/federated/generator/FedTargetConfig.java
```diff
@@ -36,7 +36,8 @@ public FedTargetConfig(LFGeneratorContext context, Resource federateResource) {
     mergeImportedConfig(
         federateResource, context.getFileConfig().resource, context.getErrorReporter());
 
-    clearPropertiesToIgnore();
+    clearPropertiesToIgnore(); // FIXME: add boolean inherit() function to TargetPropertyConfig
```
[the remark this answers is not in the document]
lhstrh: I thought it might be helpful to declare for each target property whether it should be inherited by federates or not. On second though, I think that might be excess generality, so I'll remove the FIXME...

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
lhstrh: That would require making the property aware of the current target (e.g., by passing a reference to the associated `TargetConfig` into the constructor). I'm hesitant to do this because those methods are not final but meant to be overridden, meaning the assertion could easily be bypassed...

core/src/main/java/org/lflang/target/property/TracingProperty.java
lhstrh: That's an interesting thought... We do recursively check types, but not whether the current target has support... I think that if we do this, we should rely on reflection to figure out what "options" exist. If we go down this path, this would also remove the need for the `TargetProperty` enum...

lhstrh: I made progress toward this in #2035, as it does proper recursive checking, but a full generalization like the one you describe is a bridge too far, I think. Because all of the structural checking is done automatically through the type classes, all that needs to happen in the `validate` method of the target property is to do additional checks (such as cross checks between target properties). Hence, if we want an error message when the dictionary type is used in the C++ target, we can easily handle this corner case in the `TracingProperty.validate`.

oowekyala (commented):
Just a couple of comments on the design. I assumed the extensibility goal would be for plugins to add targets, though maybe it is to add target properties as stated on #2035... Not sure what would be the thing

core/src/main/java/org/lflang/AbstractTargetProperty.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Return {@code true} if this target property is supported by the given target, {@code false}
+   * otherwise.
+   *
+   * @param target The target to check against.
+   */
+  public final boolean isSupported(Target target) {
+    return supportedTargets().contains(target);
```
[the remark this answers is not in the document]
lhstrh: That's a good point...

core/src/main/java/org/lflang/AbstractTargetProperty.java
```diff
@@ -0,0 +1,190 @@
...
+  /** The type of values that can be assigned to this property. */
+  public final TargetPropertyType type;
+
+  /** Whether (after initialization) this property has been set. */
+  protected boolean isSet;
+
+  /**
+   * The value assigned to the target property, initialized using the {@code initialValue()} method.
+   */
+  private T value = initialValue();
```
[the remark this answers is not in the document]
lhstrh: The state has to be kept somewhere. What this does is tie the state and the name together. If you separate them, you have another mapping problem on your hands...

core/src/main/java/org/lflang/AbstractTargetProperty.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * A base class for target properties.
+ *
+ * <p>After implementing this class to create a new target property, add a corresponding entry to
+ * {@code TargetConfig} and hook it into the {@code TargetProperty} enum.
+ *
+ * @param <T> The data type of the value assigned to the target property.
+ */
+public abstract class AbstractTargetProperty<T> {
```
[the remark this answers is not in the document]
lhstrh: Right, it's in my todo to rename `AbstractTargetProperty` to `TargetProperty` (see the description of #2035). The point of having a shared base class is not to add a lot of functionality in its extensions but to reuse the shared functionality in the base class.

I like your idea of turning `TargetConfig` into a strongly-typed map. Let me think about this a bit.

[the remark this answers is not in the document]
lhstrh: Have a look at https://github.com/lf-lang/lingua-franca/pull/2039 and lmk what you think...

core/src/main/java/org/lflang/target/property/TracingProperty.java
cmnrd: But then we need to make `TracingPolicy.validate` aware of the target and have target specific validation code. This is reversed in the same way as the `isSupported` property as [pointed out by Clément](https://github.com/lf-lang/lingua-franca/pull/2008#discussion_r1345934810).

lhstrh: This is a good point, but I don't know how to fix this without spending another two weeks on this :sweat_smile: I'm happy to explore fixing this at a later time. All of the changes in these PRs are steps in the right direction already...

core/src/main/java/org/lflang/AbstractTargetProperty.java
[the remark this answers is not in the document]
lhstrh: Fixed.

core/src/main/java/org/lflang/target/Target.java
```diff
@@ -546,4 +580,84 @@ public static Target fromDecl(TargetDecl targetDecl) {
     return Target.forName(name)
         .orElseThrow(() -> new RuntimeException("Invalid target name '" + name + "'"));
   }
+
+  public void initialize(TargetConfig config) {
```
cmnrd, line 584: This still seems a bit backwards. Instead of having a switch case here, why not add a method to `GeneratorBase` that is extended by the individual target generators and to the registering their. Then the generators would be responsible for declaring the properties that they support.

core/src/main/java/org/lflang/target/Target.java
```diff
@@ -546,4 +580,84 @@ public static Target fromDecl(TargetDecl targetDecl) {
...
+          CompilerProperty.INSTANCE,
+          ExportDependencyGraphProperty.INSTANCE,
+          ExportToYamlProperty.INSTANCE,
+          ExternalRuntimePathProperty.INSTANCE,
+          NoRuntimeValidationProperty.INSTANCE,
+          PrintStatisticsProperty.INSTANCE,
+          Ros2DependenciesProperty.INSTANCE,
+          Ros2Property.INSTANCE,
+          RuntimeVersionProperty.INSTANCE,
+          TracingProperty.INSTANCE,
```
cmnrd, line 621: We should distinguish the C tracing property (which can be configured) from the C++ property (which is a plain boolean).

core/src/main/java/org/lflang/target/TargetConfig.java
```diff
@@ -0,0 +1,321 @@
...
+    // Register target properties for internal use.
+    this.register(FedSetupProperty.INSTANCE);
+
+    // Register general-purpose properties
+    this.register(
+        FastProperty.INSTANCE,
+        LoggingProperty.INSTANCE,
+        NoCompileProperty.INSTANCE,
+        TimeOutProperty.INSTANCE,
+        TracingProperty.INSTANCE);
```
cmnrd, line 88: I don't think tracing qualifies as general purpose. Also it seems to be registered again separately for the C and the C++ target.

core/src/main/java/org/lflang/TargetProperty.java
cmnrd, line 1: Should this move to `org.lflang.target.property`?

lhstrh: Good idea. Will do.

core/src/main/java/org/lflang/target/TargetConfig.java
lhstrh: Good catch. I'll remove it.

core/src/main/java/org/lflang/target/Target.java
lhstrh: OK, that works for me. Will make the change.

core/src/main/java/org/lflang/target/Target.java
lhstrh: I don't suggest we do this because it will become too complicated. What I will do is add a validation check to catch inappropriate use.

core/src/main/java/org/lflang/target/Target.java
lhstrh: Actually, I don't think this will work. The problem is that the target configuration is part of the context, which is passed into the generator and is also used during validation. We need to register the properties ahead of time, or else we cannot catch during validation whether a property is supported or not.

lhstrh: So, whether this is backwards or not, let's keep it like this for now and prioritize getting this merged. So far, we only have a single code generator for each target, so it shouldn't really matter where we do the initialization.

core/src/main/java/org/lflang/target/Target.java
lhstrh: I do consider this refactoring a success based on how easy it was to implement this check. See [8018f7f](https://github.com/lf-lang/lingua-franca/pull/2008/commits/8018f7f2744f4522878056fc6317548a6ef08b27).

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -1125,176 +1073,9 @@ public void checkTargetDecl(TargetDecl target) throws IOException {
...
-    KeyValuePair ros2 = getKeyValuePair(targetProperties, TargetProperty.ROS2);
-    KeyValuePair ros2Dependencies =
-        getKeyValuePair(targetProperties, TargetProperty.ROS2_DEPENDENCIES);
-    if (ros2Dependencies != null && (ros2 == null || !ASTUtils.toBoolean(ros2.getValue()))) {
-      warning(
-          "Ignoring ros2-dependencies as ros2 compilation is disabled",
-          ros2Dependencies,
-          Literals.KEY_VALUE_PAIR__NAME);
+    if (targetProperties.eContainer() instanceof TargetDecl) {
+      // Only validate the target properties, not dictionaries that may be part of their values.
```
lhstrh, line 1077:
```suggestion
      // Skip dictionaries that may be part of a target property value because type checking is done recursively.
```

core/src/main/kotlin/org/lflang/generator/rust/RustModel.kt
```diff
@@ -715,7 +726,7 @@ private val TypeParm.identifier: String
 val BuildType.cargoProfileName: String
     get() = when (this) {
         BuildType.DEBUG             -> "debug"
-        BuildType.TEST              -> "test"
+        BuildType.TEST              -> "lf-test"
```
lhstrh, line 729: @oowekyala, I'm unsure where this changes comes from. Does it look right to you?

core/src/integrationTest/java/org/lflang/tests/RuntimeTest.java
```diff
@@ -209,8 +221,33 @@ public void runEnclaveTests() {
...
+            || category == TestCategory.DOCKER_FEDERATED
+            || category == TestCategory.DOCKER
+            || category == TestCategory.ENCLAVE
+            || category == TestCategory.ARDUINO
+            || category == TestCategory.VERIFIER
+            || category == TestCategory.ZEPHYR_UNTHREADED
+            || category == TestCategory.ZEPHYR_BOARDS
+            || category == TestCategory.ZEPHYR_THREADED;
+
+    // SERIALIZATION and TARGET tests are excluded on Windows.
```
lsk567, line 249: Looks like `SERIALIZATION` is excluded without checking the platform. So there is a small mismatch between the comment and the implementation.

lsk567 (approved):
This is a big effort that makes target properties more robust by relying less on String operations. Looks good to me overall. Thanks for working on this @lhstrh. :-)

core/src/integrationTest/java/org/lflang/tests/RuntimeTest.java
lhstrh: Yep. I'll fix that later. No need to run CI again for a comment change. Thanks!

Issue #1943, opened 2023-08-13
Modal models with banks cause segmentation fault


edwardalee:
The following program segfaults when executed.  Below is a second program that does the same thing without banks. The second program works fine.

```
target C {
    timeout: 100 ms
}
reactor B(bank_index: int = 0) {
  input in:int
  output out:int
  reaction(in) -> out {=
    lf_set(out, self->bank_index);
  =}
}
reactor Modal {
  input in:int
  output out1:int
  output out2:int
  output out3:int
  initial mode A {
    reaction(startup) -> B {=
      lf_set_mode(B);
    =}
  }
  mode B {
    b = new[3] B()
    (in)+ -> b.in
    b.out -> out1, out2, out3
  }
}
main reactor {
  m = new Modal()
  timer t(0, 10 ms)
  reaction(t) -> m.in {=
    lf_set(m.in, 42);
  =}
  reaction(m.out3) {=
    lf_print("out3 = %d", m.out3->value);
  =}
}
```

The following variant does not segfault:

```
target C {
    timeout: 100 ms
}
reactor B(b: int = 0) {
  input in:int
  output out:int
  reaction(in) -> out {=
    lf_set(out, self->b);
  =}
}
reactor Modal {
  input in:int
  output out1:int
  output out2:int
  output out3:int
  initial mode A {
    reaction(startup) -> B {=
      lf_set_mode(B);
    =}
  }
  mode B {
    b1 = new B(b = 0)
    b2 = new B(b = 1)
    b3 = new B(b = 2)
    (in)+ -> b1.in, b2.in, b3.in
    b1.out, b2.out, b3.out -> out1, out2, out3
  }
}
main reactor {
  m = new Modal()
  timer t(0, 10 ms)
  reaction(t) -> m.in {=
    lf_set(m.in, 42);
  =}
  reaction(m.out3) {=
    lf_print("out3 = %d", m.out3->value);
  =}
}
```

a-sr:
I just realized that only port `out3` is connected and unconnected ports sometimes cause problems (#1489 & #1321) and this may be related.

lhstrh:
Could be related to #1953.

PR #2011, opened 2023-09-18
Handling of unconnected outputs to avoid segmentation fault

edwardalee:
This addresses #1943.


lhstrh (approved):
LGTM! Thanks this fix and the accompanying test, @edwardalee!

PR #2014, opened 2023-09-18
[C] Self struct made available to initializers of state variables, parameters

OmerMajNition:
When passing parameter to a reactor through an array we have the flexibility to use bank index of current declaration, but parent bank index and parent's parameters are not exposed to let user define a linear combination in nested banked reactors.
```
target C;
preamble {=
    int table[] = {4, 3, 2, 1};
=}
reactor A(bank_index:int(0), value:int(0)) {
    reaction (startup) {=
        printf("bank_index: %d, value: %d\n", self->bank_index, self->value);
    =}
}
main reactor {
    a = new[4] A(value = {= table[bank_index] =});
}
```
However for nested bank initializations we feel less equipped to form a linear combination to access elements of table based on nesting reactor indexes.
```
target C;
preamble {=
    int table[] = {6, 5, 4, 3, 2, 1};
=}

reactor B (bank_index:int(0), value:int(0)) {
    reaction (startup) {=
        printf ("B bank_index:%d, value:%d\n", self->bank_index, self->value);
    =}
}

reactor A(bank_index:int(0), n_B:int(1)) {
    b = new [n_B] B(value = {= table[ ${ bank_index } * ${ n_B } + bank_index] =});
    reaction (startup) {=
        printf("A bank_index:%d\n", self->bank_index);
    =}
}
main reactor {
    a = new[2] A(n_B = 3);
}
```
In above example we want to set up B to be able to get parameter from array based on it's position in reactor hierarchy. We have 2 problems

Parent bank index is not available
Parent reactor's parameter are not accessible
This GH issue is created to have a discussion around this feature being available in LF

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
@@ -1,7 +1,5 @@
 package org.lflang.generator.c;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import org.lflang.ast.ASTUtils;
 import org.lflang.generator.CodeBuilder;
 import org.lflang.generator.ParameterInstance;
@@ -27,36 +25,10 @@ public class CParameterGenerator {
     if (p.getName().equals("bank_index") && p.getOverride() == null) {
       return CUtil.bankIndex(p.getParent());
     }
-    String initializer = null;
 
     CTypes ctypes = CTypes.generateParametersIn(p.getParent().getParent());
     Initializer values = p.getActualValue();
-    initializer = ctypes.getTargetInitializer(values, p.getDefinition().getType());
-
-    // If Initializer is a code expression
-    // parse for parent parameter values being used inside expressions
-    // Look for <parent-param> encapsulated inside ${ <parent-param> }
-    if (ctypes.checkForCodeExp(values)) {
-      ReactorInstance parent = p.getParent().getParent();
-      if (parent != null) {
-        for (ParameterInstance parameter : parent.parameters) {
-          String search_param = "(\\$\\{\s*" + parameter.getName() + "\s*\\})";
-          Pattern pt = Pattern.compile(search_param);
-          Matcher m = pt.matcher(initializer);
-
-          if (m.find()) {
-            StringBuilder tmp_init = new StringBuilder();
-            do {
-              String replacement = CUtil.reactorRef(parent) + "->" + parameter.getName();
-              m.appendReplacement(tmp_init, replacement);
-            } while (m.find());
-            m.appendTail(tmp_init);
-            initializer = tmp_init.toString();
-          }
-        }
-      }
-    }
-    return initializer;
+    return ctypes.getTargetInitializer(values, p.getDefinition().getType());
   }
 
   /**
--- a/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
@@ -2,7 +2,6 @@ package org.lflang.generator.c;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import org.lflang.ast.ASTUtils;
 import org.lflang.generator.CodeBuilder;
 import org.lflang.generator.ParameterInstance;
--- a/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
@@ -1,8 +1,12 @@
 package org.lflang.generator.c;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import org.lflang.ast.ASTUtils;
 import org.lflang.generator.CodeBuilder;
 import org.lflang.generator.ParameterInstance;
+import org.lflang.generator.ReactorInstance;
 import org.lflang.lf.Initializer;
 import org.lflang.lf.Parameter;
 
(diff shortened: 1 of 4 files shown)


edwardalee:
The [NestedBanks](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/multiport/NestedBanks.lf) test shows how to accomplish this with no changes to LF.  Moreover, I believe that the method used in that test does not break encapsulation.

cmnrd:
I don't think that exposing an additional parameter `parent_bank_index` actually solves the problem. What if you want to access the parent's parent's parameter? Currently, our approach is to explicitly hand over parameters through the hierarchy as needed (as in the example posted by Edward). This might be a bit tedious if many parameters are just passed through, but it should solve the problem. In principle, we could think about syntactic sugar to make parameter passing simpler, if that is desired.

I would also like to point out another issue in your example. All values assigned to any parameters in the C target need to be statically computable by the LF compiler. This is a fundamental design decision in the C target. Hence, it is not possible to use a target code expression to assign a parameter value (like you do in `B(value = {= table[parent_bank_index * ${n_B} + bank_index] =});`). This is only possible in targets with a more dynamic runtime (i.e., C++, Rust, TypeScript).

OmerMajNition:
[NestedBanks](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/multiport/NestedBanks.lf) is too much of a juggling to pass on the linear combination of nested banks forward. It would add too many ports in the system, diagram would look horrible as the number of reactors grow in number.

OmerMajNition:
I've removed parent_bank_index change and replaced parent_bank_index with ${ bank_index } implying user is referring to parent's bank index parameter rather than current bank's index

edwardalee:
How about the following solution:

```
target C;

reactor B (bank_index:int(0), parent_bank_index:int = 0, n_B:int = 1) {
    preamble {=
        int table[] = {6, 5, 4, 3, 2, 1};
    =}
    state value:int = 0;
    reaction (startup) {=
        self->value = table[self->parent_bank_index * self->n_B + self->bank_index];
        printf ("B bank_index:%d, value:%d\n", self->bank_index, self->value);
    =}
}

reactor A(bank_index:int(0), n_B:int(1)) {
    b = new [n_B] B(parent_bank_index = bank_index, n_B = n_B);
    reaction (startup) {=
        printf("A bank_index:%d\n", self->bank_index);
    =}
}
main reactor {
    a = new[2] A(n_B = 3);
}
```

Running this yields the following output:

```
---- Start execution at time Tue Sep 19 14:58:28 2023
---- plus 822009000 nanoseconds
Environment 0: ---- Spawning 8 workers.
B bank_index:2, value:1
B bank_index:1, value:2
B bank_index:0, value:3
A bank_index:1
B bank_index:2, value:4
B bank_index:1, value:5
B bank_index:0, value:6
A bank_index:0
---- Terminating environment 0
---- Elapsed physical time (in nsec): 998,000
```

OmerMajNition:
Major problem for Magnition with this approach is, reactor B can no longer be generic. Every time reactor B fits into a different topology code changes are required in reactor B implementation. We want to have components written once and reuse them as much and as efficient as possible

cmnrd:
If you want to provide the complete configuration from outside and use C code to populate this configuration, then, as far as I can see, the only way this would work in the C target is sending the configuration at startup via a port. You could introduce a config struct that combines all configuration variables, so that you only need a single connection.

 I know that you are concerned about the visual clutter that this might create. An idea to resolve this, would be to add an attribute for hiding certain reactor elements. Or more general, for defining in which views an element appears.

OmerMajNition:
Currently hierarchy is 
A
  B

If we take this index linearization to startup reaction of B then when B sits in some other hierarchy as follows
X
  A
    B

Then we would have to go change reactor B's startup method. B would never be topology agnostic if we take this array linearization in reactions of B

lhstrh:
Based on what @OmerMajNition showed me today, I think the point of this PR isn't directly to expose parameters of parents to their children, but to allow accessing parameters in initializers. The `self` struct is only accessible within reactions currently. Is there any reason this is the case other than we just didn't think of implementing it? How is it in the C++ target? Is it possible to reference a parameter in an initializer?

cmnrd:
I fully understand the problem. And, yes, in the C++ target parameters can be assigned using target code expressions and the parameters of the outer reactor are in scope. We use this extensively in the benchmarks. However, the underlying problem is more fundamental than scoping.

In the C++ target parameters are considered runtime constants. They are not known at compile time, but written once during program initialization. Therefore, it is possible to assign target code expressions to a parameter. This, however, is only possible as the C++ runtime assembles the program dynamically. This has been a very conscious design decision, precisely to enable use-cases like the one described by Omer.

In the C target, parameters are considered compile time constants. They need to be fully resolvable by the LF code generator. As far as I understand, changing this either requires a significant redesign of the C target, or an LF expression syntax. Both sound like rabbit holes to me.

cmnrd:
After cross-checking, I realized like I might be partly wrong. It seems like the C code generator can handle target code expressions in parameter assignments. I am not sure if this is a bug or a consciously added feature. Either way, target code assignments can only work for parameters that do not influence the width of a multiport or bank (for this we would need the dynamic runtime). In fact, the compiler crashes with a nullpointer exception when trying to assign a target code expression to a width parameter. I don't know how we handle other cases, e.g., when a parameter is used to give the offset or period of a timer.

If we can safely consider target code expressions for parameter assignment a feature in the C target, then yes, we could probably bring the self struct in scope for these assignments. However, bringing the self struct in scope seems too permissive, as it also gives access to state variables. In C++  the access is controlled more fine grained.

cmnrd:
I am still not sure what the code generator does with the target code expressions through. As far as I unterstand parameters in the C target, they are fully resolved at compile time and the self structs are intitializes directly for wach instance. If that is a case, an approach like proposed by Omar will not work without significant refactoring.

core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
```diff
@@ -24,10 +27,36 @@ public static String getInitializer(ParameterInstance p) {
...
+    String initializer = null;
 
     CTypes ctypes = CTypes.generateParametersIn(p.getParent().getParent());
     Initializer values = p.getActualValue();
-    return ctypes.getTargetInitializer(values, p.getDefinition().getType());
+    initializer = ctypes.getTargetInitializer(values, p.getDefinition().getType());
+
+    // If Initializer is a code expression
+    // parse for parent parameter values being used inside expressions
+    // Look for <parent-param> encapsulated inside ${ <parent-param> }
```
lhstrh, line 38: This `S{}` syntax for meta variables is something that is very specific to your tooling, so I don't think that this kind of substitution really belongs here. What I would do is find a way to bring the parameters into scope, so that the expressions are valid, and use the `${}` syntax in the initializers of the parameters themselves (not the expressions that reference them). Have you tried this?

OmerMajNition: We're good as far as parent reactor's parameters are in the scope to have valid expressions, whichever syntax you propose. This ${} syntax is just to differentiate between child's bank_index and parent's bank_index. Syntax is your call, this change is not tied up with our tooling, this is purely lingua franca related.

lhstrh:
This is just a note to keep track of this issue: in our meeting today we've identified a better solution that involves defining a "self" variable at the top of the code generated in `CGenerator.generateReactorInstance`.

lhstrh:
Looking good :rocket:

OmerMajNition:
I've also tested it our existing code and looks good, better cleaner solution than the previous one.

lhstrh:
The Gradle exceptions I'm seeing in the "by target" tests look very weird though. I had not seen those before.

OmerMajNition:
@lhstrh,  Added the test file, for your review please.

lhstrh (approved):
Looks great! Thanks, @OmerMajNition!

edwardalee (approved):
Nice!  Looks like a simple fix!

Issue #1856, opened 2023-06-19
Show reaction names in diagrams


lhstrh:
After merging #1853, I suggest that we deprecate the `@label` attribute and instead use the name given in the declaration (if there is one). The code generators can also benefit from this and make the generated code more readable by making use of this.

cmnrd:
The reaction name should also be shown in the diagrams (instead of the current label).

cmnrd:
Since we decided to keep the `@label` attribute (see discussion in https://github.com/lf-lang/lingua-franca/pull/2016), this issue only requires displaying the reaction names in the diagrams in order to be fixed.

PR #2016, opened 2023-09-21
Display reaction names in diagrams and deprecated the `@label` attribute

cmnrd:
Fixes https://github.com/lf-lang/lingua-franca/issues/1856. Note that this does not ensure yet, that the target code generators actually name the reactions as given in the LF code. For targets not supporting bodyless reactions, this is purely cosmetic and not a strict requirement.

### DIFF
--- a/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
+++ b/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
@@ -129,11 +129,7 @@ import org.lflang.generator.RuntimeRange;
 import org.lflang.generator.SendRange;
 import org.lflang.generator.TimerInstance;
 import org.lflang.generator.TriggerInstance;
-import org.lflang.lf.Connection;
-import org.lflang.lf.LfPackage;
-import org.lflang.lf.Model;
-import org.lflang.lf.Reactor;
-import org.lflang.lf.StateVar;
+import org.lflang.lf.*;
 import org.lflang.util.FileUtil;
 
 /**
@@ -220,8 +216,8 @@ public class LinguaFrancaSynthesis extends AbstractDiagramSynthesis<Model> {
   public static final SynthesisOption CYCLE_DETECTION =
       SynthesisOption.createCheckOption("Dependency Cycle Detection", true);
 
-  public static final SynthesisOption SHOW_USER_LABELS =
-      SynthesisOption.createCheckOption("User Labels (@label in JavaDoc)", true)
+  public static final SynthesisOption SHOW_REACTION_NAMES =
+      SynthesisOption.createCheckOption("Label reactions with their names (if given)", true)
           .setCategory(APPEARANCE);
   public static final SynthesisOption SHOW_HYPERLINKS =
       SynthesisOption.createCheckOption("Expand/Collapse Hyperlinks", false)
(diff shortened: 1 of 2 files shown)


lhstrh:
I haven't looked at the code, but let's make sure to keep the `@label` annotations for reactors, because I think they are useful for supplying additional information.

core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
```diff
@@ -129,11 +129,7 @@
 import org.lflang.generator.SendRange;
 import org.lflang.generator.TimerInstance;
 import org.lflang.generator.TriggerInstance;
-import org.lflang.lf.Connection;
-import org.lflang.lf.LfPackage;
-import org.lflang.lf.Model;
-import org.lflang.lf.Reactor;
-import org.lflang.lf.StateVar;
+import org.lflang.lf.*;
```
lhstrh, line 132: The [style guide](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports) forbids these:
```quote
Wildcard imports, static or otherwise, are not used.
```

core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
```diff
@@ -1712,8 +1708,18 @@ private KNode addErrorComment(KNode node, String message) {
...
-    if (getBooleanValue(SHOW_USER_LABELS)) {
-      String commentText = AttributeUtils.getLabel(element);
+    if (getBooleanValue(SHOW_REACTION_NAMES)) {
+      String commentText = null;
+      // If the element is a reaction, try to read its name
+      if (element instanceof Reaction) {
+        commentText = ((Reaction) element).getName();
+      }
+      // If the element does not have a name, check if it has an @label attribute
+      // TODO: the @label attribute is depricaetd and should be removed at release 0.7.0
```
lhstrh, line 1718:
```suggestion
      // TODO: the @label attribute is deprecated and should be removed at release 0.7.0
```

core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
```diff
@@ -1712,8 +1708,18 @@ private KNode addErrorComment(KNode node, String message) {
...
   private Iterable<KNode> createUserComments(EObject element, KNode targetNode) {
-    if (getBooleanValue(SHOW_USER_LABELS)) {
-      String commentText = AttributeUtils.getLabel(element);
+    if (getBooleanValue(SHOW_REACTION_NAMES)) {
+      String commentText = null;
+      // If the element is a reaction, try to read its name
+      if (element instanceof Reaction) {
+        commentText = ((Reaction) element).getName();
+      }
+      // If the element does not have a name, check if it has an @label attribute
```
lhstrh, line 1717: I think we should still be able to write @label("Some description") next to a reactor class and show it. I think that this change should cement that for reactions, either the reaction name (if it is specified) or a numeral is shown inside the chevron, but I'm starting to think that we should still keep the `@label` around (also for reactions) to serve as additional explanation in the diagram, just as we have them for reactor classes.

cmnrd:
> I think we should still be able to write @label("Some description") next to a reactor class and show it. I think that this change should cement that for reactions, either the reaction name (if it is specified) or a numeral is shown inside the chevron, but I'm starting to think that we should still keep the @label around (also for reactions) to serve as additional explanation in the diagram, just as we have them for reactor classes.

Yes, I actually had the same thought. Otherwise, we would be conflating names with annotations. I'll try to find an alternative solution.

cmnrd:
Closing this in favor of https://github.com/lf-lang/lingua-franca/pull/2030

PR #2018, opened 2023-09-22
Stabilized LSP diagram genenation and added default expansion option

cmnrd:
This PR adds a dependency of the language server on `de.cau.cs.kieler.klighd.incremental` as suggested in this comment: https://github.com/lf-lang/vscode-lingua-franca/issues/103#issuecomment-1731023470. This seems to indeed stabilize the diagram generation. 

It further adds a new synthesis options that allows to control if new reactors should be expanded or contracted when added. This setting also influences if all reactors are expanded or contracted when opening a new LF file.

Closes https://github.com/lf-lang/vscode-lingua-franca/issues/103 

Perhaps this change also addresses https://github.com/lf-lang/vscode-lingua-franca/issues/142, but we will need further testing.


lhstrh (approved):
Thanks for fixing this, @cmnrd!

PR #2022, opened 2023-09-23
Static update instr

erlingrj:
This PR updates the PRET VM instructions in the C target. For portability, I argue that the right approach is to fix the data types, this is also what the JVM does. A JVM `long` is always 64 bit, regardless of platform.

I found a very nice way of making the instruction format flexible using `union` in C. Now the C implementation has the following semantics for instruction operands. 

Each operand is either 
1. A _pointer_ to a 64 bit unsigned integer. No requirements on the width of the pointer itself (32bit on NRF52 and 64bit on laptops)
2. An immediate 64 bit unsigned integer. Here the width is fixed. This we need to be 64bit to encode time.

See the companion PR in reactor-c here: https://github.com/lf-lang/reactor-c/pull/277


core/src/main/resources/lib/platform/zephyr/boards/nrf52dk_nrf52832.overlay
```diff
@@ -0,0 +1,3 @@
+&timer1 {
```
lsk567, line 1: Should this change be included in this PR, or another Zephyr-specific PR?

lsk567 (approved):
Overall looks good to me. Thanks!

core/src/main/resources/lib/platform/zephyr/boards/nrf52dk_nrf52832.overlay
erlingrj: This is some changes needed for Zephyr. They are already in master. The best would be to bring this `static-schedule` branch up-to-date with master first. Then I wouldnt need that change

lsk567: Ok, sounds good. I will merge bring the `static-schedule` branch up-to-date.

lsk567: The `static-schedule` branch has been updated.

PR #2023, opened 2023-09-23
Avoidance of random CI failures due to failing Gradle download

lhstrh:
Sometimes CI fails because a job fails to download Gradle (which happens upon the first invocation of `gradlew`). This PR adds a step that repeatedly (3x) attempts the download before giving up.


cmnrd (commented):
I think we should a sleep of 10 to 30 seconds in between the tries

PR #2030, opened 2023-09-27
Show reaction names in diagrams

cmnrd:
This PR adds the "Reaction Names" diagram synthesis option, which shows the name of a reaction instead of its index. If a reaction does not have a name in LF code, its internal name is shown (see code and figure below).
```
main reactor {
  reaction hello(startup)
  reaction(shutdown)
}
```
HelloBodylessWorld

In order to align the index number that we show in the diagrams with the number in the name for anonymous reactions, I also modified the `getName()` method for reaction instances to increment the index by one. This partly addresses https://github.com/lf-lang/lingua-franca/issues/1782, but does not fully resolve the issue as there are many instances in the C code generator, where it does not use the name of the reactor instance, but instead constructs its own name.

Fixes https://github.com/lf-lang/lingua-franca/issues/1856

### DIFF
--- a/core/src/main/java/org/lflang/generator/ReactionInstance.java
+++ b/core/src/main/java/org/lflang/generator/ReactionInstance.java
@@ -29,6 +29,7 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import org.eclipse.xtext.xbase.lib.StringExtensions;
 import org.lflang.TimeValue;
 import org.lflang.ast.ASTUtils;
 import org.lflang.lf.Action;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/lflang/generator/ReactionInstance.java
```diff
@@ -29,6 +29,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import org.eclipse.xtext.xbase.lib.StringExtensions;
```
edwardalee, line 32: Is this import really necessary?  It looks like it is used to just check for a null or empty string. I think that this is a rather heavyweight dependency for such small functionality.

edwardalee (changes requested):
This is a very nice improvement!  One minor request inserted...

core/src/main/java/org/lflang/generator/ReactionInstance.java
lhstrh: We already have this dependency throughout the code base. If we want to remove it, we have quite a bit more work to do...

cmnrd: Xtext is one of our core dependencies. Using its functionality throughout shouldn't do any additional harm.

cmnrd:
@edwardalee could you approve this PR?

lhstrh:
It would probably look better (in the diagram) without the underscore.

cmnrd:
> It would probably look better (in the diagram) without the underscore.

You mean `reaction2` instead of `reaction_2`? That would be an easy change and probably also looks better in the generated C code.

cmnrd:
What d you think about the bold typeface? I kept this from the index numbers, but I am wondering if it might get too much for longer names.

lhstrh:
> What d you think about the bold typeface? I kept this from the index numbers, but I am wondering if it might get too much for longer names.

I meant with a space, but if it has to literally match the identifier then I guess the underscore is fine...

I also prefer normal over bold.

PR #2032, opened 2023-09-28
Fix bug in C decentralized execution with small after delays

petervdonovan:
See https://github.com/lf-lang/reactor-c/pull/280 for details/discussion.


lhstrh (approved):

petervdonovan:
Looks like we got a deadlock in the TypeScript version of `SimpleFederated.lf`. It's conceivable that this deadlock wouldn't happen in `master` since the TypeScript federated execution uses the same RTI (right?) But I do not have the bandwidth to debug the TypeScript generator right now.

erlingrj:
I am merging this because the `reactor-c` counterpart was already merged and this causes other PRs to fail

petervdonovan:
Thanks! Ominously, the reason why this has not been merged is that a test flaked after it was already in the merge queue, but Marten and I basically intended to do what you just did in the hope that this is still a net improvement.

Issue #1677, opened 2023-03-28
Making the src-gen folder more stand-alone


edwardalee:
The src-gen folder currently generated by lfc does not quite contain enough information to build the generated C code. To reproduce:
```
cd test/C
lfc src/Minimal.lf
cd src-gen/Minimal
rm -rf build
mkdir build
cd build
cmake ../
make
```
This fails with
```
/Users/eal/lingua-franca-intellij/test/C/src-gen/Minimal/core/../include/core/platform.h:50:2: error: Must define either LF_UNTHREADED or LF_THREADED runtime
#error Must define either LF_UNTHREADED or LF_THREADED runtime
 ^
1 error generated.
```
In order to get the program to build, you have issue the following incantation:
```
cmake -DLF_REACTION_GRAPH_BREADTH=1 -DLF_THREADED=1 -DNUMBER_OF_WORKERS=0 -DSCHEDULER=NP -DLOG_LEVEL=2 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/Users/eal/lingua-franca-intellij/test/C -DCMAKE_INSTALL_BINDIR=bin ../
```
@erlingrj suggested that a simple solution would be to write the default/provided target properties to both `CMakeLists.txt` and `CompileDefinitions.txt`. We can put the variables behind statements like this:
```
if (NOT DEFINED NUMBER_OF_WORKERS)
  set(NUMBER_OF_WORKERS 42 )
endif()
```
This way we get (1) plug-and-play integration with cmake, (2) easy to extend to other build system because the info is also readily available in CompileDefinitions.txt and (3) change target properties without rebuilding by changing passing arguments to cmake.

erlingrj:
Now that the compiler generates vscodes files I would say that this is less of an issue. But still would be a nice feature. In fact, it might make the generation of vscode files redundant.

PR #2036, opened 2023-10-03
Generate default values for the user-configurable CMake flags

erlingrj:
This adds default values to the CMake flags which currently is passed over the command line when invoking cmake. This removes the need for:
1. Generating a `.vscode/settings.json` file to open and build generated sources in VScode.
2. The need for `lfc` to invoke `cmake` with all the flags
3. The need to generate `CompileDefinitons.txt` with the default values

Closing #1677


erlingrj:
I am asking for your feedback here, @petervdonovan, since I believe you made the CMake integration and you might have some arguments for why we should not put the default flags in the generates CMakeLists.txt and instead pass them on the command line.

petervdonovan:
I am in favor of this! IIRC we have discussed this as a group multiple times and collectively agreed that this is the right choice.

lhstrh (approved):
This looks nice!

PR #2037, opened 2023-10-04
Remove deprecated C schedulers++

### LINKED ISSUE #2009: `LET` not a valid identifier in C target
`LET` is defined in `lf_types.h` and clashes with a program parameter named `LET`, causing it to not compile.

E.g.:
```
main reactor(LET: time = 30ms) {
  ...
``

erlingrj:
This PR does two things:

1. Removes the deprecated C schedulers (as discussed with @edwardalee and @petervdonovan) GEDF_NP_CI and PEDF_NP.
2. Prepends the scheduler selection macro with SCHED. E.g. "SCHED_ADAPTIVE" instead of "ADAPTIVE". This should address Martens problem of wanting to use scheduler names as variables. Btw, we dont even have a LET scheduler, so that problem was easily fixed, but in general I think this change is a good idea

Closes #2009

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -18,5 +18,5 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
     implementation "com.diffplug.spotless:spotless-plugin-gradle:$spotlessVersion"
 
-    implementation "gradle.plugin.com.github.johnrengelman:shadow:$shadowJarVersion"
+    implementation "com.github.johnrengelman:shadow:$shadowJarVersion"
 }
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -1,6 +1,6 @@
 [versions]
-spotlessVersion=6.11.0
+spotlessVersion=6.21.0
 spotlessLibVersion=2.30.0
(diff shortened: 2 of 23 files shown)


edwardalee:
We need to remove or fix GEDF_NP as well.  It drops events with federated execution, possibly because deadlines are not correctly propagated upstream (reaction order dependencies are ignored). Even if this is fixed, it does very little because it only prioritizes deadlines per level.

erlingrj:
Then I propose we remove it and leave it as an open issue to fix it and bring back into main. Currently I am seeing some wierd level errors in CI on the single federated test that hardcodes the usage of the adaptive scheduler...

erlingrj:
@edwardalee I could take a stab at fixing the GEDF scheduler if you can provide a test case where it fails. Even if it "only" prioritizes reactions after deadline this is still an important feature. E.g. your example with a pipeline sense->process->act with a deadline on `act` needs it

edwardalee (approved):
LGTM!  Let's merge this!

lhstrh:
> LGTM!  Let's merge this!

I'm confused. Didn't Erling fix the issue in another PR instead?

erlingrj:
The fixes are already merged into `reactor-c` together with a fix that allows you to use LET as a variable. This PR has corresponding changes in this repo

PR #2040, opened 2023-10-05
Fixed trimming of the recorded test output

cmnrd:
This fixes a bug in our test framework. It seems like `Runtime.getRuntime().freeMemory()` does not indicate the total memory that can still be allocated, but only the free memory that is already allocated by the runtime. Therefore, the recorded output was often cut prematurely. This fix only considers the size of the recorded string buffer. It also makes sure, that the buffer is trimmed after deleting elements.


erlingrj:
Thanks alot, Christian. The PR looks good, I am testing this fix on my feature-branch. And will report back

erlingrj (approved):
Works well. LGTM!

Issue #2026, opened 2023-09-24
Federated with Deadline fails to invoke reaction


edwardalee:
The following test nondeterministically fails to invoke the first reaction of the `Actuate` reactor.  Removing the deadline in the second reaction of `Actuate` eliminates the problem.  The problem occurs whether the deadline is violated or not.  The mere presence of the deadline statement is sufficient to trigger the bug.

<img width="866" alt="image" src="https://github.com/lf-lang/lingua-franca/assets/8513334/f83a69dd-da89-4ccc-910a-29e60cdab147">

```
target C {
  timeout: 5s
}
preamble {=
  #include "platform.h"
=}
reactor Sense(period:time = 500ms) {
  state count:int = 1
  output out:int
  timer t(0, period)
  reaction(t) -> out {=
    lf_set(out, self->count++);
  =}
}
reactor Process {
  input in:int
  output out:int
  reaction(in) -> out {=
//    lf_sleep(MSEC(35));
    lf_set(out, in->value * 2);
  =}
}
reactor Actuate {
  input in1:int
  input in2:int
  state count:int = 0
  reaction(in1) {=
    printf(PRINTF_TIME ": Received %d\n", lf_time_logical_elapsed(), in1->value);
    self->count++;
  =}
  reaction(in2) {=
    printf(PRINTF_TIME ": Second input: %d\n", lf_time_logical_elapsed(), in2->value);
  =} deadline(30 ms) {=
    printf(PRINTF_TIME ": Second input: %d    ", lf_time_logical_elapsed(), in2->value);
    printf("PANIC! Deadline violated!\n");
  =}
  reaction(shutdown) {=
    if (self->count != 11) {
      lf_print_error_and_exit("Expected the first reaction to be invoked 11 times! Got %d", self->count);
    }
  =}
}
federated reactor {
  s1 = new Sense(period = 500ms)
  s2 = new Sense(period = 1s)
  c1 = new Process()
  a = new Actuate()
  s1.out -> c1.in
  c1.out -> a.in1
  s2.out -> a.in2
}
```

edwardalee:
[...]

PR #2045, opened 2023-10-05
Temporarily disable GEDF_NP scheduler

edwardalee:
As reported in #2026, the GEDF_NP scheduler incorrectly executes LF code at least for federated reactors. This PR disables the automatic selection of GEDF_NP when there are deadlines.  You can still set a target property to choose this scheduler, but until we fix this scheduler, it should never be automatically selected.

### DIFF
--- a/core/src/main/java/org/lflang/ast/ASTUtils.java
+++ b/core/src/main/java/org/lflang/ast/ASTUtils.java
@@ -153,7 +153,7 @@ public class ASTUtils {
    * Get the main reactor defined in the given resource.
    *
    * @param resource the resource to extract reactors from
-   * @return An iterable over all reactors found in the resource
+   * @return The top-level main reactor, or null if there is no main or federated reactor or the top-level is federated.
    */
   public static Optional<Reactor> getMainReactor(Resource resource) {
     return StreamSupport.stream(
--- a/core/src/main/java/org/lflang/ast/ASTUtils.java
+++ b/core/src/main/java/org/lflang/ast/ASTUtils.java
@@ -153,8 +153,7 @@ public class ASTUtils {
    * Get the main reactor defined in the given resource.
    *
    * @param resource the resource to extract reactors from
-   * @return The top-level main reactor, or null if there is no main or federated reactor or the
-   *     top-level is federated.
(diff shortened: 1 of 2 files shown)


erlingrj (approved):
LGTM

core/src/main/java/org/lflang/ast/ASTUtils.java
```diff
@@ -153,7 +153,8 @@ public static List<Reactor> getAllReactors(Resource resource) {
    * Get the main reactor defined in the given resource.
    *
    * @param resource the resource to extract reactors from
-   * @return An iterable over all reactors found in the resource
+   * @return The top-level main reactor, or null if there is no main or federated reactor or the
```
lhstrh, line 156: This comment isn't right either. The return type is `Optional<Reactor>`, meaning that if a reactor is found, its presence can be checked using the `isPresent()` method and retrieved using `get()`.

lhstrh (changes requested):

core/src/main/java/org/lflang/ast/ASTUtils.java
edwardalee: Feel free to provide a better comment. I'm no so familiar with this pattern.  Otherwise, I'll just revert to the previous incorrect comment.

lhstrh:
```suggestion
   * @return A top-level main reactor presented as an {@code Optional} that is present if it could be found.
```

core/src/main/java/org/lflang/ast/ASTUtils.java
```diff
@@ -153,7 +153,8 @@ public static List<Reactor> getAllReactors(Resource resource) {
...
+   *     top-level is federated.
```
lhstrh, line 157:
```suggestion
```

lhstrh:
It looks like @erlingrj may have solved the bug; do we want to therefore hold back on merging this?

erlingrj:
Yes, I already merged in a fix in reactor-c. Can one of you review and approve this PR: https://github.com/lf-lang/lingua-franca/pull/2037

edwardalee:
Closing this as @erlingrj fixed the bug with GEDF_NP.

PR #2047, opened 2023-10-06
[C-Target] Tracing

mkhubaibumer:
Allows user to disable system-tracing when tracing feature is enabled for C-Target

User can now specify system trace option with
```
target C {
    tracing: {
        trace-file-name:"sometrace",
        system-trace: true,
    },
    fast: true
}
```


lhstrh:
@mkhubaibumer: I will take care of the conflict that have arisen and then submit a review.

Issue #2048, opened 2023-10-07
Non-deterministic error in decentralized federated program


erlingrj:
`DecentralizedP2PUnbalancedTimeoutPhysical.lf`
```
target C {
  timeout: 1 msec,
  coordination: decentralized
}

reactor Clock(offset: time = 0, period: time = 1 sec) {
  output y: int
  timer t(offset, period)
  state count: int = 0

  reaction(t) -> y {=
    (self->count)++;
    //printf("Reacting at time %ld.\n", lf_time_logical_elapsed());
    lf_set(y, self->count);
  =}

  reaction(shutdown) {=
    lf_print("SUCCESS: the source exited successfully.");
  =}
}

reactor Destination {
  input x: int
  state s: int = 1

  reaction(x) {=
    // printf("%d\n", x->value);
    if (x->value != self->s) {
      lf_print_error_and_exit("Expected %d and got %d.", self->s, x->value);
    }
    self->s++;
  =}

  reaction(shutdown) {=
    lf_print("**** shutdown reaction invoked.");
    lf_print("Approx. time per reaction: %lldns", lf_time_physical_elapsed()/(self->s+1));
  =}
}

federated reactor(period: time = 10 usec) {
  c = new Clock(period=period)
  d = new Destination()
  c.y ~> d.x
}
```

Is failing sometimes:
```
Federate DecentralizedP2PUnbalancedTimeoutPhysical in Federation ID 'f316e441b2522e7bd76d635f7ed5f6f1b3eb34a63f7d542d'
RTI: Federation ID: f316e441b2522e7bd76d635f7ed5f6f1b3eb34a63f7d542d
RTI: Number of federates: 2
RTI: Clock sync: init
RTI: Clock sync exchanges per interval: 10
Starting RTI for 2 federates in federation ID f316e441b2522e7bd76d635f7ed5f6f1b3eb34a63f7d542d.
RTI using TCP port 15045 for federation f316e441b2522e7bd76d635f7ed5f6f1b3eb34a63f7d542d.
RTI: Listening for federates.
#### Launching the federate federate__c.
#### Launching the federate federate__d.
#### Bringing the RTI back to foreground so it can receive Control-C.
RTI -i ${FEDERATION_ID} -n 2 -c init exchanges-per-interval 10
[...]

PR #2049, opened 2023-10-07
Making timing of test more realistic

edwardalee:
The `DecentralizedP2PUnbalancedTimeoutPhysical.lf` has been failing nondeterministically. I believe this may due to the lack of STP offsets and the extremely short timeout (1 ms).  This frequently the destination to shut down before receiving any messages at all and possibly before the source has even been able to get itself fully set up. This PR changes the timing so that the receiving federate has STP offsets. The sending federate sends 100 messages with a spacing of 100us. Because the connection is physical, the receiving federate may not receive all of them, but any that it receives will be in order.  This makes the test more realistic and hopefully less flaky.

### DIFF
--- a/test/C/src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf
+++ b/test/C/src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf
@@ -7,7 +7,7 @@
  * between source and destination.
  */
 target C {
-  timeout: 1 msec,
+  timeout: 10 msec,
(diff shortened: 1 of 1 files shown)


edwardalee:
I am converting this to a draft PR because it could mask rather than fix the bug reported in #2048

edwardalee:
With PR #2118, this test seems to no longer fail with the original (rather aggressive) timing. So I suggest we leave it alone and see whether the problem resurfaces.

PR #2050, opened 2023-10-10
Fixed bug in the C++ reaction dependency analysis

cmnrd:
This pulls in https://github.com/lf-lang/reactor-cpp/pull/54 and adds a couple of testcases.


revol-xut (approved):
LGTM

PR #2052, opened 2023-10-13
Removal of `ulog` from `reactor-ts`, code generator updated to allow this change

axmmisaka:
1. Removed references to `Log.global` (renamed to `Log.globalLogger`); replaced with `Log.error()` and `Log.setLevel()`;
2. Removed references to `ulog`.

This is to accommodate https://github.com/lf-lang/reactor-ts/pull/237

### DIFF
--- a/core/src/main/kotlin/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
+++ b/core/src/main/kotlin/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
@@ -74,7 +74,7 @@ class TSParameterPreambleGenerator(
                 |    if (__processedCLArgs.${parameter.name} !== null) {
                 |        __CL${parameter.name} = __processedCLArgs.${parameter.name};
                 |    } else {
-                |        Log.global.error(__clUsage);
+                |        Log.error(null, () => __clUsage);
                 |        throw new Error("Custom '${parameter.name}' command line argument is malformed.");
                 |    }
                 |}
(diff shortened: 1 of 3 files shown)


axmmisaka:
Why is this failing :cry:

axmmisaka:
Hmm...... Zephyr test is failing?

core/src/main/resources/lib/ts/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "LinguaFrancaDefault",
     "type": "commonjs",
     "dependencies": {
-        "@lf-lang/reactor-ts": "^0.6.0",
+        "@lf-lang/reactor-ts": "git://github.com/lf-lang/reactor-ts.git#use-pinojs-as-logger",
```
lhstrh, line 5:
```suggestion
        "@lf-lang/reactor-ts": "git://github.com/lf-lang/reactor-ts.git#master",
```

core/src/main/kotlin/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
```diff
@@ -202,7 +202,7 @@ class TSParameterPreambleGenerator(
         |    if (__processedCLArgs.id !== null) {
         |        __federationID = __processedCLArgs.id;
         |    } else {
-        |        Log.global.error(__clUsage);
+        |        Log.error(null, () => __clUsage);
```
lhstrh, line 205: This looks a little weird, but I think it's our own API. What is the `null` argument for?

lhstrh (changes requested):
After merging the corresponding `reactor-ts` PR, please apply the suggested change before merging this PR.

core/src/main/kotlin/org/lflang/generator/ts/TSParameterPreambleGenerator.kt
axmmisaka: When `message` (arg1) is called, it is called as if it is a function of `object` (arg0), and `arg0` will/might be its `this`.
If no `this` is referenced, then passing in `null` would suffice.

We might need to refactor this later, since this design is a bit unnecessary.

https://github.com/lf-lang/reactor-ts/blob/9798f65d700d2e801c2a5c242b32704f74e47d72/src/core/util.ts#L70-L84

PR #2057, opened 2023-10-14
Handling of unknown width for multiports

### LINKED ISSUE #1833: Misleading multiport warning and NullPointerException in diagram synthesis
```
target Cpp

public preamble {=
  inline std::size_t foo() { return 4; }
=}

reactor Src {
  output[4] out: void
}

reactor Sink {
  input[{=foo()=}] in: void
}

main reactor Minimal {
  src = new Src()
  sink = new Sink()
  src.out -> sink.in
}
```

The above legal C++ program produces a misleading warning message: "Left width 4 does not match right width 0". This probably comes from a false assumption made somewhere in the instance graph.

Also, the diagram synthesis fails for this program with a NullPointerException.

edwardalee:
This fixes #1833.

### DIFF
--- a/core/src/main/java/org/lflang/ast/ASTUtils.java
+++ b/core/src/main/java/org/lflang/ast/ASTUtils.java
@@ -1479,6 +1479,9 @@ public class ASTUtils {
         } else {
           return -1;
         }
+      } else if (term.getCode() != null) {
+        // The width is given in target language code, so we can't know what it is.
+        return -1;
       } else if (term.getWidth() > 0) {
         result += term.getWidth();
       } else {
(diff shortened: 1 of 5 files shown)


lhstrh (commented):
LGTM! We should probably add the failing code from #1833 as a test case.

edwardalee:
I don't know how to add a test for this.  The NPE occurred during diagram synthesis.  During compilation, you only get a a warning.

Note that fundamentally, because in Cpp the widths of banks and multiports may not be knowable before executing the code, you have no assurance that the diagram is correct.  My fix errs on the side of making more connections rather than fewer.  There are probably examples where you will get false reports of causality loops.

cmnrd:
Thanks for fixing this Edward! I think it is OK to assume more connections (and be conservative in cycle detection). Situations where this can go wrong are rather obscure.

Since we introduced the lfd tool, we also test diagram generation in CI. However, it is not fully functional at the moment, as I didn't figure pit how to draw the giagrams fully expanded yet. So I am not sure if the bug would be triggered. In any case, I think it would be good to at a testcase.

lhstrh (approved):
Approving this because having the fix is more important than not having the test.

PR #2060, opened 2023-10-16
Threading support for the `rp2040` platform

gundralaa:
for tracking compiler modifications to support baremetal threading on the rp2040 platform.
still in draft state

TODO
- [ ] add validator checks to always restrict number of workers to 2

### DIFF
--- a/core/src/main/java/org/lflang/TargetProperty.java
+++ b/core/src/main/java/org/lflang/TargetProperty.java
@@ -1748,7 +1748,7 @@ public enum TargetProperty {
     AUTO,
     ARDUINO,
     NRF52("Nrf52", true),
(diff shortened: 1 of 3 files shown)


lhstrh:
You might want to wait implementing the validator checks because there is a big refactoring going on right now.

lhstrh:
@gundralaa, #2008 has been merged, so now is a good time to have another look at this. There are some conflicts, but they should be easy to resolve, I think. To implement the validation check that you mention in the description, have a look at `PlatformProperty.validate()`.

lhstrh:
This would be helpful for getting this example to work: https://github.com/lf-lang/lf-pico-template/blob/main/src/HelloThread.lf

Issue #2066, opened 2023-10-19
lfc --version should give the commit ID, not just SNAPSHOT


erlingrj:
If you have a nightly lfc installed. I think the `lfc --version` should give the commit from which it is built, not just `SNAPSHOT`. This will be useful for bug-reports.

PR #2069, opened 2023-10-20
CI updated to include commit SHA in version number when creating nightly build

axmmisaka:
This looks like sticking a big piece of duct tape on a giant pile of duct tapes, but I think this is the only way I could think of that could fix https://github.com/lf-lang/lingua-franca/issues/2066.

If there are ways to pass the commit SHA in as gradle/compiler property for the compiler to use, I'm all ears.

Result:
image

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -17,6 +17,10 @@ jobs:
           fetch-depth: 0
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
+      - name: Modify property file to contain the commit SHA
+        shell: bash
+        run: >
+          sed --regexp-extended --in-place "s/^(VERSION = .*)$/\1@$(git rev-parse --short HEAD)/" core/src/main/resources/org/lflang/StringsBundle.properties
       - name: Build and package lf cli tools (nightly build)
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -17,6 +17,10 @@ jobs:
           fetch-depth: 0
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
+      - name: Modify property file to contain the commit SHA
+        shell: bash
+        run: >
+          sed --regexp-extended --in-place "s/^(VERSION = .*)$/\1@$(git rev-parse --short HEAD)/" core/src/main/resources/org/lflang/StringsBundle.properties
```
erlingrj, line 23:
```suggestion
          export VERSION_TIMESTAMP="$(date +'%Y-%m-%d')"
          sed --regexp-extended --in-place "s/^(VERSION = .*)$/\1 ($(git rev-parse --short HEAD) $VERSION_TIMESTAMP)/" core/src/main/resources/org/lflang/StringsBundle.properties 
```

erlingrj (commented):
Cool, thanks for looking at this @axmmisaka. I don't think your approach is hacky at all! I suggest that we format it a little "cleaner" and adds the date of the build. This is also how `rustc --version` looks.

lhstrh (approved):
LGTM!

PR #2071, opened 2023-10-21
Fix memory leak in unfederated execution 

jackyk02:
Similar to the memory leak issue in federated execution, the reference counts for some PyObjects are not being appropriately managed. In the case for unfederated execution, when the reaction that takes an input and schedules an action with the value of that input, it increases the reference count of self->_lf_inp->value, which is unnecessary.

Initialization:
Untitled1

After execution for 20 rounds, the memory usage increased to 10GBs:
Untitled2


edwardalee:
The segfault with `src/DelayArrayWithAfter.lf` indicates that the leak is because of a failure to _decrement_ the reference count at the right point rather than an _extra increment_ of the reference count.  Perhaps the reference count for the value needs to be decremented in `generateForwardBody`?

lhstrh:
Could this be related to #2082?

PR #2075, opened 2023-10-25
Citation file

lhstrh:
Why have this? See: https://citation-file-format.github.io/ TL;DR: to enable other to cite our software.

Summary of the edits provided in this PR:
- Featuring contributors with > 100 commits (in order of size of contribution)
- Affiliation listed in according to time frame during which bulk of the contributions were made

I realize that the above criteria are arbitrary, but we need some objective measure. Happy to revise based on feedback.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,78 @@
+cff-version: 1.2.0
+title: Lingua Franca
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: Marten
+    family-names: Lohstroh
+    email: marten@berkeley.edu
+    affiliation: UC Berkeley
+    orcid: 'https://orcid.org/0000-0001-8833-4117'
+  - given-names: Edward
+    family-names: Lee
+    email: eal@berkeley.edu
+    affiliation: UC Berkeley
+    orcid: 'https://orcid.org/0000-0002-5663-0584'
+  - given-names: Soroush
+    family-names: Bateni
+    email: soroosh129@gmail.com
+    affiliation: UT Dallas
+    orcid: 'https://orcid.org/0000-0002-5448-3664'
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -0,0 +1,78 @@
...
+   - given-names: Anirudh
+    family-names: Rengarajan
+    email: arengarajan@berkeley.edu
+    affiliation: UC Berkeley
+repository-code: 'https://github.com/lf-lang/lingua-franca'
+url: 'https://lf-lang.org/'
+abstract: >-
+  Lingua Franca is a polyglot coordination language for building complex
+  systems in a component-wise fashion with worry-free deterministic
+  concurrency, zero-effort distribution, and time as a first-class citizen.
```
edwardalee, line 69: low-effort?

edwardalee (approved):
LGTM.  One thing we might consider is that contributors that commit primarily to associated repos (reactor-c, reactor-Cpp, etc) might not be counted but perhaps should be.

CITATION.cff
lhstrh, line 69:
```suggestion
  concurrency, low-effort distribution, and time as a first-class citizen.
```

lhstrh:
> LGTM. One thing we might consider is that contributors that commit primarily to associated repos (reactor-c, reactor-Cpp, etc) might not be counted but perhaps should be.

I checked, but there are none currently. But we will of course have to keep updating this document over time...

petervdonovan (approved):

axmmisaka:
> > LGTM. One thing we might consider is that contributors that commit primarily to associated repos (reactor-c, reactor-Cpp, etc) might not be counted but perhaps should be.
> 
> I checked, but there are none currently. But we will of course have to keep updating this document over time...

Maybe (another) GH Action that checks contributors and generate citation file from time to time...

Issue #2081, opened 2023-11-03
Double-freeing of tokens when exposed to frequent aynchronous lf_schedule_int


erlingrj:
I have been trying to debug an elusive bug I found in a federated program which I think is related to race conditions in the token handling. I was not able to exactly reproduce it in an un-federated setting. But I was able to create a very simple program that ends up with double-freeing tokens and sometimes yielding segmentation fault: `SIGSEGV (address boundary error)`. I suspect that this is a symptom of the same underlying problem I am facing with the federated program.

To reproduce:

```
target C {
  timeout: 10 msec
}

reactor WithPhy {
  preamble {=
    #include "platform.h"
    void* ping(void* actionref) {
      for (int i = 0; i<100; i++) {
        lf_schedule_int(actionref, 0, i);
      }
      return NULL;
    }
  =}
  physical action a:int

  reaction(startup) -> a {=
    lf_thread_t thread_id;
    lf_thread_create(&thread_id, ping, a);
  =}

  reaction(a) {= =}
}

main reactor {
  w = new WithPhy()
}
```
 
 This program schedules a 100 physical actions right after one another. Adding sleeps between the schedulings seems to solve the issue. It alternates between working, but throwing a: `WARNING: Token being freed that has already been freed: 0x7f2490000f70` and seg faulting.
 
 This particular warning means that `lf_done_using` is invoked on a token that already has a ref-count of zero.
 
 There are a few token errors which currently are just printed as warnings. I think these should be considered as errors and halt the program and thus any tests with such warnings will fail.

erlingrj:
I found the race condition and have a proposed fix here: https://github.com/lf-lang/lingua-franca/pull/2082

erlingrj:
It solves the federated issue also. This bug might also have caused other federated tests to be flaky. The reason why this is surfacing now is that I recently disabled some throttling algorithm on the sockets. Previously, messages sent between federates where throtteled and spaced out. Sending was blocked until the previous messages was ACK'ed. Now we just send them without waiting for ACK. So we can now have the scenario that the a federate has mulitple messages in his socket that he reads out and schedules one after the other. This triggered the very unlikely thread-ordering that I am reproducing here

lhstrh:
Wow. Amazing detective work! Will have a look at your fix.

PR #2082, opened 2023-11-03
Fix race-condition in _lf_replace_template_token

erlingrj:
This addresses the problem described in #2081 

The problem was that we moved tokens from the trigger structs to the action structs outside a critical section. This created a race condition with external asynchronus contexts trying to schedule events onto the very same action struct.

This simple solution does the moving of tokens inside i critical section. I think another solution might be to re-think the design. Particularily. 
1. Why does `lf_schedule` put the token on BOTH the event queue AND the action struct? The runtime will later move the token from event queue -> trigger struct and then in the reaction preambles, it moves it from the trigger struct to the action struct.
2. Why is the token moved from trigger-struct to action-struct in the code-generated preamble. It would be more natural to do that from wihin the runtime (which is in a critical section)

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CReactionGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CReactionGenerator.java
@@ -336,6 +336,7 @@ public class CReactionGenerator {
         ? String.join(
             "\n",
             DISABLE_REACTION_INITIALIZATION_MARKER,
+            "lf_critical_section_enter(self->base.environment);",
             "self->_lf_"
                 + outputName
                 + ".value = ("
@@ -343,14 +344,13 @@ public class CReactionGenerator {
                 + ")self->_lf__"
                 + actionName
                 + ".tmplt.token->value;",
-            "lf_critical_section_enter(self->base.environment);",
             "_lf_replace_template_token((token_template_t*)&self->_lf_"
                 + outputName
                 + ", (lf_token_t*)self->_lf__"
                 + actionName
                 + ".tmplt.token);",
-            "lf_critical_section_exit(self->base.environment);",
-            "self->_lf_" + outputName + ".is_present = true;")
+            "self->_lf_" + outputName + ".is_present = true;",
+            "lf_critical_section_exit(self->base.environment);")
         : "lf_set(" + outputName + ", " + actionName + "->value);";
   }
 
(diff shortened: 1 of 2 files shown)


erlingrj:
@edwardalee

erlingrj:
I had to include `platform.h` for the reaction bodies in order to code-generate a critical section in the reaction preamble. It suggests that maybe another solution is better. Like having the runtime move the token instead of code-generating it.

lhstrh (approved):
You found a needle in a haystack. Terrific job, @erlingrj.

core/src/main/java/org/lflang/generator/c/CReactionGenerator.java
```diff
@@ -561,11 +563,13 @@ private static String generateActionVariablesInReaction(
                 + " != NULL && "
                 + tokenPointer
                 + "->value != NULL);",
+            "lf_critical_section_enter(self->base.environment);",
```
edwardalee, line 566: I think the critical section needs to include the two operations above, setting is_present and has_value.  Also the setting of the value below.

erlingrj:
@lhstrh we can wait a little with merging. There might be a better way of addressing this

core/src/main/java/org/lflang/generator/c/CReactionGenerator.java
```diff
@@ -343,11 +343,13 @@ public static String generateForwardBody(
                 + ")self->_lf__"
                 + actionName
                 + ".tmplt.token->value;",
+            "lf_critical_section_enter(self->base.environment);",
```
edwardalee, line 346: I think the critical section needs to include the operations above, setting the value.  Also the setting of is_present below.

edwardalee (changes requested):
Good catch, but I think the critical section needs to be a little bigger or the action struct could become inconsistent.

erlingrj:
@edwardalee I expanded the critical sections according to your comments.

erlingrj:
@edwardalee, have you seen this test failure of `ChainWithDelay` before?
```
 Federate 2: FATAL ERROR: Attempted to insert reactions for a trigger that had an intended tag that was in the past. This should not happen under centralized coordination. Intended tag: (2500000, 0). Current tag: (3000000, 0)).
```

erlingrj:
I was able to reproduce it here and I have the trace of federated communication. Should we discuss it in a separate issue? It might be related to fixes you are working on? The RTI erroneously gives a PTAG to a federate

edwardalee:
Yes, I'm seeing these kinds of failures. There is an easy solution, but it will mask a more fundamental problem.  The easy solution is to prohibit the use of PTAG except for zero-delay cycles.  We will want to do that eventually, but first we should fix the handling of PTAG because with this easy fix, it will be hard to produce these errors.

erlingrj:
Separate issue for the ChainWithDelay PTAG problem: https://github.com/lf-lang/lingua-franca/issues/2084

PR #2085, opened 2023-11-05
Checks for uniqueness of members

lhstrh:


core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -139,7 +139,7 @@ public class LFValidator extends BaseLFValidator {
 
   @Check(CheckType.FAST)
   public void checkAction(Action action) {
-    checkName(action.getName(), Literals.VARIABLE__NAME);
+    checkName(action.getName(), Literals.MEMBER__NAME);
```
cmnrd, line 142: Shouldn't this PR also add `checkName` calls to other syntactic elements like method, reaction, parameters and state variables? Or am I missing something here about how this code works?

lhstrh: I created an interface called `Member` that is implemented by all of those.

cmnrd: Yes, but I expected to see a new check `checkMember` added that actually performs the name check. The way it looks now, we only check names for actions and ports. Or does Xtext automatically take care of checking uniqueness here?

lhstrh: Uniqueness is checked automatically based on the `name` attribute. This only specifies which nodes should be checked against one another.

PR #2086, opened 2023-11-06
Minimal changes to support enclaves refactoring

edwardalee:
The goal of this PR is to get @erlingrj 's Enclave work merged in reactor-c while first making minimal changes to lingua-franca. Later we can pull in the lingua-franca side of the Enclaves work.


erlingrj:
See this PR instead #2095

erlingrj:
@edwardalee I closed this PR rather than force pushing to the branch. I will bring your test over to the new PR

PR #2088, opened 2023-11-06
Output path specified in JSON used correctly

lhstrh:
This is to accommodate interactions with `lingo`, which specifies an output directory in JSON.


lhstrh:
After this has been merged, @edwardalee, check out https://github.com/lf-lang/lingo, install it using `cargo install --path .`, then run `cd sandbox` and `lingo build`.

edwardalee (approved):
LGTM.

Issue #1986, opened 2023-09-01
lfc gives no clue how to run TypeScript code


edwardalee:
When compiling a TypeScript target program, lfc used to print something like this, as shown [here](https://www.lf-lang.org/docs/handbook/a-first-reactor?target=ts#structure-of-an-lf-project):

```
#####################################
To run the generated program, use:

    node ...path-to-project.../src-gen/Minimal/dist/Minimal.js

#####################################
```

It no longer prints any such thing, leaving the programmer to guess what to do next.

Better yet, it should generate a shell script in `bin`, as the Python target does currently, to invoke the program.

hokeun:
I'm looking into this.

lhstrh:
Great! Let's opt for generating a script in `bin`, like we do with Python. Should be an easy fix.

PR #2090, opened 2023-11-08
Fix to generate launch script for TS target and print informational message

hokeun:
This fixes https://github.com/lf-lang/lingua-franca/issues/1986.

Also did a minor code cleanup.

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -230,9 +230,7 @@ public class FedLauncherGenerator {
 
     // Write the launcher file.
     File file = fileConfig.binPath.resolve(fileConfig.name).toFile();
-    messageReporter
-        .nowhere()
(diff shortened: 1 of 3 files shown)


edwardalee (approved):
LGTM.

lhstrh (changes requested):
This looks great and it works well, but I have some suggestions for improvement:
 - don't just print the directory but print the full path of the executable
 - adjust the "Generating launcher for federation" message to do the same

hokeun:
> * adjust the "Generating launcher for federation" message to do the same

Could you elaborate on this? I'm not sure what you meant by "do the same."

lhstrh:
> > * adjust the "Generating launcher for federation" message to do the same
> 
> Could you elaborate on this? I'm not sure what you meant by "do the same."

Sorry for not being clear. I just pushed a fix here because it was quicker than trying to clarify :smile: 

See https://github.com/lf-lang/lingua-franca/pull/2090/commits/b6eba5f6b995a407e33e25f17762b544b9859d3e

lhstrh (approved):

PR #2092, opened 2023-11-08
New `--no-source-mapping` switch to disable line directives

### LINKED ISSUE #2093: Add an option to `lfc` to disable line directives

lhstrh:
Fixes #2093.

### DIFF
--- a/core/src/main/java/org/lflang/target/property/NoSourceMappingProperty.java
+++ b/core/src/main/java/org/lflang/target/property/NoSourceMappingProperty.java
@@ -1,0 +1,17 @@
+package org.lflang.target.property;
+
+/** Directive to allow including OpenSSL libraries and process HMAC authentication. */
+public final class NoSourceMappingProperty extends BooleanProperty {
+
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/lflang/target/property/NoSourceMappingProperty.java
```diff
@@ -0,0 +1,17 @@
+package org.lflang.target.property;
+
+/** Directive to allow including OpenSSL libraries and process HMAC authentication. */
```
lhstrh, line 3:
```suggestion
/** Directive to instruct the code generator to not produce line directives. */
```

petervdonovan (approved):
Yep, looks good, let's merge.

I'll admit I was initially surprised to see such a big diff, but this approach is probably safer than a global variable and can't really go wrong as long as the `suppress` parameter comes from the right source, which should be an easy guideline to follow. An alternative would have been to strip out all the `#line` directives at the very end, but that wouldn't necessarily be better.

PR #2099, opened 2023-11-10
Fix protobufs path problem

lhstrh:


test/C/src/serialization/PersonProtocolBuffers.lf
```diff
@@ -9,13 +9,13 @@
...
- *
- * https://github.com/protobuf-c/protobuf-c
- *
- * The code generator assumes that executables are installed within the PATH. On a Mac, this is
- * typically at /usr/local/bin.
+ * (Building protobuf from source is slow, so avoid doing that if possible.)
+ * 
+ * Next install the C plugin for protocol buffers:
+ * 
+ * $ brew install protobuf-c
```
lhstrh, line 16: I'm not sure this is an improvement because this only works for Mac, and the link that is being referred to provides general-purpose installation instructions... We can add the Brew command, but I'd consider that an FYI, not the main message.

test/C/src/serialization/ProtoNoPacking.lf
```diff
@@ -8,13 +8,14 @@
  *
  * $ brew install protobuf
  *
- * Building protobuf from source is slow, so avoid doing that if possible. Next install the C plugin
- * for protocol buffers from
+ * (Building protobuf from source is slow, so avoid doing that if possible.)
+ * 
+ * Next install the C plugin for protocol buffers:
+ * 
+ * $ brew install protobuf-c
```
lhstrh, line 15: Same issue. Works only on Mac.

PR #2101, opened 2023-11-10
Better placement of line directives

petervdonovan:
I believe that this addresses the concerns raised by @KyotoMicrocomputer, except for the command-line flag that suppresses line directives. The flag can be added separately.


lhstrh (approved):
This looks great! I've started a PR for the `--no-source-mapping` flag here: #2092.

PR #2102, opened 2023-11-11
Use more explicit include guards for Arduino

### LINKED ISSUE #2062: Duplicate Structure Names causes Inability to Compile Composed Reactors for Arduino
The following example for Arduino is unable to compile with the arduino-cli toolchain. I am using lf-cli-0.5.2-SNAPSHOT-nightly-20231015050236-Linux-x86_64; this issue exists on lf-cli-0.5.1-Linux-x86_64 as well. I am currently testing on Windows Pro 11 WSL with an Ubuntu 22.04 VM.

dualreactorblink.lf
```
target C {
    platform: {
        name: "arduino",
        board: "arduino:avr:uno"
    }
}

reactor Blinker {
    input in_port:bool

    reaction(startup) {=
        pinMode(LED_BUILTIN, OUTPUT);
    =}

    reaction(in_port) {=
        digitalWrite(LED_BUILTIN, in_port->value ? HIGH : LOW);
    =}
}

reactor Timer {
    timer t1(0, 500 msec)
    state on_off:bool = false
    output out_port:bool

    reaction (t1) -> out_port {=
        self->on_off = !self->on_off;
        lf_set(out_port, self->on_off);
    =}
}

main reactor {
    the_blinker = new Blinker()
    the_timer = new Timer()
    the_timer.out_port -> the_blinker.in_port;
}
```

I get the following errors when I run lfc:
In file included from /xxx/lf/src-gen/dualreactorblink/_dualreactorblink_main.ino:2:0:
/xxx/lf/src-gen/dualreactorblink/include/dualreactorblink/dualreactorblink.h:26:3: error: conflicting declaration 'typedef struct blinker_in_port_t blinker_in_port_t'
 } blinker_in_port_t;
   ^~~~~~~~~~~~~~~~~
In file included from /xxx/lf/src-gen/dualreactorblink/_blinker.ino:2:0:
/xxx/lf/src-gen/dualreactorblink/include/dualreactorblink/Blinker.h:31:3: note: previous declaration as 'typedef struct blinker_in_port_t blinker_in_port_t'
 } blinker_in_port_t;
   ^~~~~~~~~~~~~~~~~
In file included from /xxx/lf/src-gen/dualreactorblink/_timer.ino:2:0:
/xxx/lf/src-gen/dualreactorblink/include/dualreactorblink/Timer.h:32:3: error: conflicting declaration 'typedef struct timer_out_port_t timer_out_port_t'
 } timer_out_port_t;
   ^~~~~~~~~~~~~~~~
In file included from /xxx/lf/src-gen/dualreactorblink/_dualreactorblink_main.ino:2:0:
/xxx/lf/src-gen/dualreactorblink/include/dualreactorblink/dualreactorblink.h:35:3: note: previous declaration as 'typedef struct timer_out_port_t timer_out_port_t'
 } timer_out_port_t;
   ^~~~~~~~~~~~~~~~
   
Looking through the generated files, I observed that struct blinker_in_port_t is declared in both Blinker.h and dualreactorblink.h. The definitions are different, though. The structure in Blinker.h includes some FEDERATED ifdefs. Other than that, the two structures are the same. Tracing through the .h and .ino files, both of the header files end up included for the build, which looks like it causes the conflicting declaration error.

I do not know if this is an error with how I have set up the program, but I followed the instructions here (https://www.lf-lang.org/docs/handbook/arduino?target=c) to set up the arduino compile and the instructions here (https://www.lf-lang.org/docs/handbook/composing-reactors?target=c) to make the example. 

Thank you. Please let me know if there is additional information I can add to help clarify this.

petervdonovan:
I believe that this is only a problem for Arduino because Arduino brings in header files that are not actually `#include`d. This solution is a little bit of a hack because it should not be necessary to add this extra `#ifndef`, but it solves the problem and it should be mostly harmless for users who do not target Arduino.

Closes #2062. Note that the test case originally used in #2062 would not have worked because Arduino Uno does not have enough memory for our runtime.

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CGenerator.java
@@ -948,7 +948,7 @@ public class CGenerator extends GeneratorBase {
     CodeBuilder header = new CodeBuilder();
     CodeBuilder src = new CodeBuilder();
     final String headerName = CUtil.getName(tpr) + ".h";
-    var guardMacro = headerName.toUpperCase().replace(".", "_");
+    var guardMacro = CUtil.internalIncludeGuard(tpr);
     header.pr("#ifndef " + guardMacro);
     header.pr("#define " + guardMacro);
     generateReactorClassHeaders(tpr, headerName, header, src);
--- a/core/src/main/java/org/lflang/generator/c/CReactorHeaderFileGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CReactorHeaderFileGenerator.java
@@ -55,7 +55,7 @@ public class CReactorHeaderFileGenerator {
       GenerateAuxiliaryStructs generator,
       String topLevelPreamble) {
     CodeBuilder builder = new CodeBuilder();
-    appendIncludeGuard(builder, tpr);
+    appendIncludeGuards(builder, tpr);
     builder.pr(topLevelPreamble);
     appendPoundIncludes(builder);
     tpr.doDefines(builder);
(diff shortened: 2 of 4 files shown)


lhstrh:
It's a clever solution! I like it! It might be worth also generating a comment explaining the use of this extra include guard...

lhstrh (approved):
:+1: Very nice!

PR #2105, opened 2023-11-15
Fix mistake in code-generated reaction preamble.

erlingrj:
This is  a race condition that I tried addressing in #2082 but I didn't do it properly. This PR finally makes the critical section include all accesses to the state shared between the reaction body and the async context scheduling a physical action.


edwardalee (approved):
LGTM.

PR #2110, opened 2023-11-20
Fix undetected `scheduler` and `workers` properties

lsk567:
This PR fixes the problem of the `scheduler` and `workers` target properties being ignored by the compiler. Luckily the fix was straightforward.

However, the fact that `concurrent/Workers.lf` does not detect this problem suggests that its implementation is insufficient, but I am not sure how to implement it correctly at the moment.

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CGenerator.java
@@ -1947,6 +1947,15 @@ public class CGenerator extends GeneratorBase {
       // So that each separate compile knows about modal reactors, do this:
       CompileDefinitionsProperty.INSTANCE.update(targetConfig, Map.of("MODAL_REACTORS", "TRUE"));
     }
+    if (!targetConfig.get(
+        SingleThreadedProperty.INSTANCE)) { // FIXME: This logic is duplicated in CMake
+      pickScheduler();
+      // FIXME: this and pickScheduler should be combined.
+      var map = new HashMap<String, String>();
+      map.put("SCHEDULER", targetConfig.get(SchedulerProperty.INSTANCE).getSchedulerCompileDef());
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -1947,6 +1947,15 @@ protected void setUpGeneralParameters() {
       // So that each separate compile knows about modal reactors, do this:
       CompileDefinitionsProperty.INSTANCE.update(targetConfig, Map.of("MODAL_REACTORS", "TRUE"));
     }
+    if (!targetConfig.get(
+        SingleThreadedProperty.INSTANCE)) { // FIXME: This logic is duplicated in CMake
+      pickScheduler();
+      // FIXME: this and pickScheduler should be combined.
```
lhstrh, line 1953: I'm a little worried about two FIXMEs just being copied instead of addressed properly.

lhstrh (changes requested):
~~Could you explain why this fix works? Also, could you have a look at the FIXMEs and see if you can resolve them?~~

core/src/main/java/org/lflang/generator/c/CGenerator.java
lhstrh: I had a look at the FIXMEs and removed them because the logic looks OK. I just made some simplifications.

lhstrh (approved):

core/src/main/java/org/lflang/generator/c/CGenerator.java
lsk567: > I had a look at the FIXMEs and removed them because the logic looks OK. I just made some simplifications.
OK, sounds good!

I wasn't quite sure what duplication the FIXME was referring to...

lhstrh: I checked it and I think we're good. It shouldn't have been left in there in the first place. It was probably my mistake.

PR #2112, opened 2023-11-20
Fix for cycle detection issue

### LINKED ISSUE #2033: Cycle detection erroneously reports cycles with physical connections
Consider this minimal program:
```
target C

reactor Foo {
  input in: void
  output out: void
  reaction(in)  -> out
}

reactor B {
  input in: void
  output out: void
  reaction(in)  -> out
}

main reactor Minimal {
  a = new Foo()
  b = new Foo()

  a.out ~> b.in
  b.out -> a.in
}
```
Notice the physical connection. The compiler (and diagram) reports a dependency cycle nonetheless. Interestingly, the physical connection gets not highlighted as part of the cycle (neither in the compiler output nor in the diagram).

lhstrh:
Fixes #2033.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
@@ -386,7 +386,8 @@ public class FedASTUtils {
       PortInstance p,
       MixedRadixInt index,
       FedConnectionInstance connection) {
-    if (connection.getDefinition().getDelay() != null || connection.getDefinition().isPhysical()) return;
+    if (connection.getDefinition().getDelay() != null || connection.getDefinition().isPhysical())
+      return;
     var a = LfFactory.eINSTANCE.createAttribute();
     a.setAttrName("_tpoLevel");
     var e = LfFactory.eINSTANCE.createAttrParm();
--- a/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
@@ -386,7 +386,7 @@ public class FedASTUtils {
       PortInstance p,
       MixedRadixInt index,
       FedConnectionInstance connection) {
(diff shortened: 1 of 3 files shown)


lhstrh:
Looks like the code I touched is also used for federation to figure out the topology. For cycle detection, a physical connection should break the cycle, which it does now, but it makes federated tests fail. Not sure how to address this. @edwardalee, could you take a look?

edwardalee:
It is not surprising that this change would cause failures. It changes the semantics of the eventualDestinations method. I'm surprised there aren't unfederated failures. I also think that excluding connections with delays is suds. Maybe this is related to federated failures with multiport's and delays?

lhstrh:
To me, it is surprising. This code is used for causality analysis in the validator. Maybe it should not be, but for that use, connections that are physical or have delay should not imply dependencies.

edwardalee:
The `eventualDestinations` method comment says this:
```
   * Return a list of ranges of this port, where each range sends to a list of destination ports
   * that receive data from the range of this port. Each destination port is annotated with the
   * channel range on which it receives data. The ports listed are only ports that are sources for
   * reactions, not relay ports that the data may go through on the way. Also, if there is an
   * "after" delay anywhere along the path, then the destination is not in the resulting list.
```
I think your reasoning is that a physical connection is much like an `after` connection with an unspecified amount of delay. This is reasonable, so changing this method (and its description) is defensible, but it still means making sure that all uses still work.  If I had to guess, there is a place where this method is used to set `num_destinations`, and maybe this is being used in federated execution.

byeong-gil:
@lhstrh I fixed the compile error due to the wrong TPO level assignment and added a test for this PR's change. Could you review it?

petervdonovan (approved):
I'm not aware of any way that these changes could cause problems. I think it's true that physical connections do not need to be addressed in TPO.

lhstrh (commented):
This looks like the right solution to me!

PR #2113, opened 2023-11-20
Changes to Codecov configuration

lhstrh:
These changes are meant to instruct Codecov to wait until CI has finished (and not error out prematurely). It also tightens the threshold of allowable reductions in code coverage to -5% overall (and -50% for the diff) with respect to the current coverage (which is about 75%).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -13,14 +13,14 @@ coverage:
     project: # settings affecting project coverage
       default:
         enabled: true
-        target: auto # auto % coverage target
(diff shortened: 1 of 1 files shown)


lhstrh:
It doesn't seem like it waits for CI to complete:

image

petervdonovan (approved):

petervdonovan:
I think that TypeScript federated linting is broken. I don't think that this is the first time when TypeScript federated linting tests have flaked. Perhaps we should either disable those tests and temporarily retract the feature, or fix it soon.

PR #2118, opened 2023-11-25
Remove absent messages

edwardalee:
This PR removes most uses of absent messages for federates that are not in a zero-delay cycle.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
+++ b/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
@@ -179,7 +179,8 @@ public class FederateInstance {
 
   /**
    * List of networkMessage actions corresponding to network input ports whose upstream federates
-   * are in zero-delay cycles. This should be a subset of the networkMessageActions.
+   * are in zero-delay cycles and the connection has no after delay.
+   * This should be a subset of the networkMessageActions.
    */
   public List<Action> zeroDelayCycleNetworkMessageActions = new ArrayList<>();
 
--- a/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
+++ b/core/src/main/java/org/lflang/federated/generator/FederateInstance.java
@@ -607,23 +607,23 @@ public class FederateInstance {
   }
 
   /**
-   * Return true if there is a downstream federate that is in a zero-delay cycle.
-   * This is needed because such a federate may be called upon to generate absent
-   * messages even if it is not itself in a zero-delay cycle.  An example is the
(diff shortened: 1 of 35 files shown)


core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -559,6 +559,76 @@ private void indexExcludedTopLevelReactions(Reactor federatedReactor) {
...
+  }
+
+  /**
+   * Return true if there is a downstream federate that is in a zero-delay cycle. This is needed
+   * because such a federate may be called upon to generate absent messages even if it is not itself
+   * in a zero-delay cycle. An example is the test `test/C/src/federated/FeedbackDelay.lf`. The
+   * absent message may be needed even if there is a delay on the connection to the downstream
+   * federate.
+   */
+  public boolean isUpstreamOfZeroDelayCycle() {
```
byeong-gil, line 616: I don't understand why this check is needed. Because of the below code, the network message action of `reaction_1` of the federate c in `FeedbackDelay.lf` is not in `zeroDelayCycleNetworkMessageActionTable`. So, the federate c doesn't wait for port absent messages from the federate pl. Then, why does the federate pl send ABS messages?
https://github.com/lf-lang/lingua-franca/blob/279b3fd1c2b7f5e7b365e759e60785611ff60d18/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java#L280

image

byeong-gil (changes requested):

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -178,10 +178,10 @@ public Instantiation getInstantiation() {
   public List<Expression> networkMessageActionDelays = new ArrayList<>();
 
   /**
-   * List of networkMessage actions corresponding to zero-delay connections. This should be a subset
-   * of the networkMessageActions.
+   * List of networkMessage actions corresponding to network input ports whose upstream federates
+   * are in zero-delay cycles. This should be a subset of the networkMessageActions.
```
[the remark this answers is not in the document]
edwardalee: Good catch!!

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
edwardalee: Good catch again!!  Fixed in 224356488cef0b60dab569b4c24f7213ea86449e

lhstrh:
Subsumed by #2140.

PR #2123, opened 2023-12-01
Fix ROS2 CI Setup

cmnrd:


lhstrh (approved):
Thanks for the fix, @cmnrd!

PR #2128, opened 2023-12-08
Fix minimum spacing of actions and reference to freed event

edwardalee:
This PR addresses https://github.com/lf-lang/reactor-c/issues/135 and https://github.com/lf-lang/reactor-c/issues/215, fixing the min_spacing option on actions and removing the reference to a possibly freed or recycled event. It has a companion PR in lingua-franca. Thanks @SheaFixstars for prompting this.

This has a [companion PR in reactor-c](https://github.com/lf-lang/reactor-c/pull/318)


lhstrh (approved):
:+1:

PR #2130, opened 2023-12-12
Bugfix for TypeScript in dev mode

lhstrh:
Ensures that when an alternative reactor-ts runtime is pointed to, it is actually used.


hokeun (approved):
This looks good to me!

byeong-gil (approved):
LGTM!

PR #2131, opened 2023-12-13
Fix remote launch and add -l option to log to files

edwardalee:
This modifies the run script for federated programs so that if you give a `-l` argument then stdout and stderr from the RTI and each federate are captured into log files. These go into the cwd and have names `RTI.log` and `federate_name.log`, so they will get overwritten if multiple runs are performed in the same directory.

@lhstrh : can you see what we can do in CI to use this to report all the outputs from federated tests?  I don't think we need persistent artifacts here and in fact would prefer not because we clutter user's disks if they use this and the file names will have to become much more complicated.

The larger part of this PR is to fix the remote launching of federates, which hasn't worked for a while.  It now works.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
@@ -293,7 +293,7 @@ public class FedGenerator {
                 new TargetConfig(
                     subFileConfig.resource, GeneratorArguments.none(), subContextMessageReporter);
             if (targetConfig.get(DockerProperty.INSTANCE).enabled
-                && targetConfig.target.buildsUsingDocker()
+                    && targetConfig.target.buildsUsingDocker()
                 || fed.isRemote) {
               NoCompileProperty.INSTANCE.override(subConfig, true);
             }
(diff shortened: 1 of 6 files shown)


lhstrh:
Maybe this should be put in a directory with the name of the executable (and a timestamp), so that things don't get overwritten?

lhstrh:
> @lhstrh : can you see what we can do in CI to use this to report all the outputs from federated tests? I don't think we need persistent artifacts here and in fact would prefer not because we clutter user's disks if they use this and the file names will have to become much more complicated.

It's unclear to me what you have in mind here: writing to file _will_ create persistent artifacts, so isn't that the goal? GitHub can serve artifacts with a user-specified retention (e.g., 3 days).

Should the launch script check for exit codes and remove the produced files if execution was successful?

edwardalee:
> > @lhstrh : can you see what we can do in CI to use this to report all the outputs from federated tests? I don't think we need persistent artifacts here and in fact would prefer not because we clutter user's disks if they use this and the file names will have to become much more complicated.
> 
> It's unclear to me what you have in mind here: writing to file _will_ create persistent artifacts, so isn't that the goal? GitHub can serve artifacts with a user-specified retention (e.g., 3 days).
> 
> Should the launch script check for exit codes and remove the produced files if execution was successful?


What I have in mind is much simpler, similar to what we do with trace files.  Put them in cwd, don't worry about them being overwritten.  This way, they won't clutter my disk.  In CI, output them on the trace log if and only if a test fails. Otherwise, discard them.

lhstrh:
I think the changes in this PR are included in #2140, right? Can we close this one?

edwardalee:
Yes, these changes are in #2140.

PR #2134, opened 2023-12-17
Remove bogus port search

edwardalee:
This PR removes the port search done by the RTI if it cannot bind to the specified or default port. This never worked and resulted in federates taking a long time to fail as they searched over possible ports. There is a companion PR in reactor-c.


test/C/src/federated/DistributedLoopedActionDecentralized.lf
```diff
@@ -26,22 +26,11 @@ reactor Receiver(take_a_break_after: int = 10, break_interval: time = 400 msec)
 
   reaction(in) {=
     tag_t current_tag = lf_tag();
-    lf_print("At tag (%lld, %u) received value %d with STP violation (%lld, %u).",
-      current_tag.time - lf_time_start(),
-      current_tag.microstep,
-      in->value,
-      current_tag.time - in->intended_tag.time,
-      current_tag.microstep - in->intended_tag.microstep
+    lf_print("Inside, at tag " PRINTF_TAG ", received value %d.",
```
lhstrh, line 29: What does "inside" mean?

lhstrh (approved):
Changes look good to me, although I still have to have a look at the reactor-c PR. Though 25 commits is excessive for 42 lines changed. Seems straightforward to clean up and wrap into just a couple commits.

test/C/src/federated/DistributedLoopedActionDecentralized.lf
edwardalee: Inside the hierarchical reactor, to distinguish it from the reaction at the top level, which also reads the input.

lhstrh:
Subsumed by #2140.

PR #2136, opened 2023-12-18
String literal allowed as property name

oowekyala:
This implements [this solution](https://github.com/lf-lang/lingua-franca/pull/1238#issuecomment-1311553308) to #1194. It would supersede #1238.

This needs an update to the documentation website.

### DIFF
--- a/core/src/test/java/org/lflang/tests/LFParsingTest.java
+++ b/core/src/test/java/org/lflang/tests/LFParsingTest.java
@@ -35,9 +35,9 @@ public class LFParsingTest {
 
   @Test
   public void testParsingTargetPropertyAsString() throws Exception {
-    assertNoParsingErrorsIn("target C { \"a c time reactor\": 2 };");
-    assertNoParsingErrorsIn("target C { \"reactor\": 2 };");
-    expectParsingErrorIn("target C { reactor: 2 };");
+    assertNoParsingErrorsIn("target C { \"a c time reactor\": 2 }; reactor Foo{}");
+    assertNoParsingErrorsIn("target C { \"reactor\": 2 }; reactor Foo{}");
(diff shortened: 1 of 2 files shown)


lhstrh (approved):
This looks great! Thanks, @oowekyala.

lhstrh:
@oowekyala: I still think that it might be nice to be able to use keywords without quotes in the target properties, don't you? Being able to use quotes equally makes sense, but I'm think that we might want to merge both this PR _and_ #1238, actually.

core/src/test/java/org/lflang/tests/LFParsingTest.java
```diff
@@ -33,6 +33,18 @@ public void testLexingEmptyTargetProperties() throws Exception {
     // assertNoParsingErrorsIn("target C {x:[,]};  \nreactor Foo {}");
   }
 
+  @Test
+  public void testParsingTargetPropertyAsString() throws Exception {
+    assertNoParsingErrorsIn("target C { \"a c time reactor\": 2 }; reactor Foo{}");
+    assertNoParsingErrorsIn("target C { \"reactor\": 2 }; reactor Foo{}");
+    expectParsingErrorIn("target C { reactor: 2 }; reactor Foo{}");
+
+    // array elements
```
lhstrh, line 42: This commented-out code seems irrelevant. Remove?

lhstrh (changes requested):

core/src/test/java/org/lflang/tests/LFParsingTest.java
lhstrh: I see they're also there in `testLexingEmptyTargetProperties` :shrug:

Not a big deal, I guess. Let's just merge.

lhstrh (approved):

PR #2137, opened 2023-12-20
Adjustment to test output to show progress

lhstrh:
This turns the progress bar into an explicit percentage and shows which test is currently run. This is useful for situations where a test gets stuck.


edwardalee (approved):
LGTM!

edwardalee:
I merged it because we can't fix master without it.

PR #2138, opened 2023-12-20
Problematic federtated tests excluded from macOS runs

lhstrh:
Note that this branch is based off the `show-progress` branch, let's merge this into `show-progress` and then merge `show-progress` into `master`.


edwardalee (approved):
LGTM, although it looks like the flaky tests are not being flaky today (??).  I would if we should try to merge this with no tests actually in the flaky directory and see how things go.

lhstrh:
The non-deterministic deadlock in macOS test runs was confirmed to occur during compilation, not while running tests, so excluding particular tests for macOS is not a solution.

PR #2140, opened 2023-12-21
Various bugfixes and cleanups in the support for federated programs

edwardalee:
This PR replaces #2118 and #2134 
It removes most uses of absent messages for federates that are not in a zero-delay cycle.
It also improves the port search done by the RTI if it cannot bind to the specified or default port. This never worked and resulted in federates taking a long time to fail as they searched over possible ports. There is a companion PR in reactor-c: https://github.com/lf-lang/reactor-c/pull/323.


lhstrh:
Looks like we're running in to port availability problems again: https://github.com/lf-lang/lingua-franca/actions/runs/7289958865/job/19865706975

This was the problem that was (although imperfectly) addressed by the port searching...

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -466,6 +487,94 @@ private void analyzeFederates(Reactor federation, LFGeneratorContext context) {
...
+    // Create the connections between federates.
+    // Each connection in the AST may represent more than one connection between
+    // federation instances because of banks and multiports. We need to generate communication
+    // for each of these. To do this, we create a ReactorInstance so that we don't have
+    // to duplicate the rather complicated logic in that class. We specify a depth of 1,
+    // so it only creates the reactors immediately within the top level, not reactors
+    // that those contain.
+    ReactorInstance mainInstance = new ReactorInstance(federation, messageReporter);
+
+    new ReactionInstanceGraph(mainInstance); // Constructor has side effects; its result is ignored
```
lhstrh, line 500: I realize that this is copied code, but this looks awkward...

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -466,6 +487,94 @@ private void analyzeFederates(Reactor federation, LFGeneratorContext context) {
...
+      for (PortInstance output : child.outputs) {
+        for (SendRange srcRange : output.getDependentPorts()) {
+          for (RuntimeRange<PortInstance> dstRange : srcRange.destinations) {
+            MixedRadixInt srcID = srcRange.startMR();
+            MixedRadixInt dstID = dstRange.startMR();
+            int dstCount = 0;
+            int srcCount = 0;
+
+            while (dstCount++ < dstRange.width) {
+              // FIXME: If connection is interleaved, these digits will be reversed?
```
lhstrh, line 512: We should either address this or attach it to an issue.

core/src/main/java/org/lflang/generator/ReactorInstance.java
```diff
@@ -191,7 +191,8 @@ public ReactorInstance(Reactor reactor, ReactorInstance parent, MessageReporter
   /**
    * Assign levels to all reactions within the same root as this reactor. The level of a reaction r
    * is equal to the length of the longest chain of reactions that must have the opportunity to
-   * execute before r at each logical tag. This fails and returns false if a causality cycle exists.
+   * execute before r at each logical tag. This returns a non-empty graph if a causality cycle
```
lhstrh, line 194:
```suggestion
   * execute before r at each tag. This returns a non-empty graph if a causality cycle
```

core/src/main/java/org/lflang/generator/c/CCompiler.java
```diff
@@ -370,6 +370,7 @@ private boolean outputContainsKnownCMakeErrors(String CMakeOutput) {
    *     the compile command will produce a binary.
    */
   public LFCommand compileCCommand(String fileToCompile, boolean noBinary) {
+    // FIXME: This is obsolete.
```
lhstrh, line 373: We should address this or tie it to an issue.

test/C/src/federated/DecentralizedP2PComm.lf
```diff
@@ -1,14 +1,18 @@
 target C {
-  timeout: 1 sec,
+  timeout: 5 sec,
   tracing: true,
-  clock-sync: off,
+  clock-sync: off,  // logging: log,
```
lhstrh, line 4:
```suggestion
  clock-sync: off,
```

test/C/src/federated/FeedbackDelay5.lf
```diff
@@ -0,0 +1,53 @@
+target C {
+  timeout: 900 ms
```
lhstrh, line 2: Why not just 1 second like the other tests?

lhstrh (commented):
I only have some minor comments/feedback; haven't yet had a look the reactor-c changes and will look at those next.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -382,7 +364,11 @@ private String getLaunchCode(String rtiLaunchCode) {
         "# The RTI will be brought back to foreground",
         "# to be responsive to user inputs after all federates",
         "# are launched.",
-        rtiLaunchCode,
+        "if [ \"$1\" = \"-l\" ]; then",
```
lhstrh, line 367: If you use `"""` you don't have to escape and it's easier to read.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -420,54 +406,127 @@ private String getRemoteLaunchCode(
...
+            + " and subdirectories federate_name, bin, and log on host "
             + getUserHost(federate.user, federate.host)
             + "\"",
-        "# The >> syntax appends stdout to a file. The 2>&1 appends stderr to the same file.",
         "ssh " + getUserHost(federate.user, federate.host) + " '\\",
-        "    mkdir -p "
-            + remoteBase.resolve(remoteRelSrcGenPath).resolve("core")
-            + " "
-            + remoteBase.resolve("bin")
+        "    mkdir -p " + binDirectory + " " + logDirectory + "; \\",
```
lhstrh, line 432: If we were to use ```""" ... """.formatted(...)` here, this code would be a lot easier to read.

lhstrh:
The tests are passing! :fireworks:

core/src/main/java/org/lflang/federated/extensions/CExtension.java
```diff
@@ -776,21 +773,20 @@ private String generateCodeToInitializeFederate(FederateInstance federate, RtiCo
               "// Create a socket server to listen to other federates.",
               "// If a port is specified by the user, that will be used",
               "// as the only possibility for the server. If not, the port",
-              "// will start from STARTING_PORT. The function will",
-              "// keep incrementing the port until the number of tries reaches PORT_RANGE_LIMIT.",
-              "create_server(" + federate.port + ");",
-              "// Connect to remote federates for each physical connection.",
+              "// will be selected by the OS (by specifying port 0).",
```
petervdonovan, line 776: Sounds like a big improvement

core/src/main/java/org/lflang/graph/TopologyGraph.java
```diff
@@ -1,176 +0,0 @@
...
-import org.lflang.generator.TriggerInstance;
-import org.lflang.lf.Connection;
-import org.lflang.lf.Variable;
-
-/**
- * A graph with vertices that are ports or reactions and edges that denote dependencies between
- * them.
- *
- * <p>NOTE: This is not used anywhere anymore, but we keep it in case this particular graph
- * structure proves useful in the future.
```
petervdonovan, line 45: Well, this does look like the more efficient cycle detection algorithm, but I'm OK with deleting it like this if that's the most expedient thing to do. We can revive this code later as needed

test/C/src/federated/AfterDelaysWithMultiport.lf
```diff
@@ -0,0 +1,51 @@
+target C {
```
petervdonovan, line 1: A short description of what bug this test catches might help

test/C/src/federated/DecentralizedP2PSTPViolation.lf
```diff
@@ -0,0 +1,74 @@
...
+
+// The STP offset here will force STP violations in a and b.
+reactor Delay(STP_offset: time = 250 ms) {
+  input in: int
+  output out: int
+
+  reaction(in) -> out {=
+    lf_set(out, in->value);
+  =} STP(0) {=
+    // With the 50ms above, STP violations shouldn't occur, but they do
```
petervdonovan, line 62:
```suggestion
    // With the 250ms above, STP violations shouldn't occur, but they do
```

test/C/src/federated/DistributedMultiport.lf
```diff
@@ -23,9 +23,12 @@ reactor Destination(width: int = 3) {
...
     for (int i = 0; i < in_width; i++) {
       if (in[i]->is_present) {
-        lf_print("Received %d at %d.", in[i]->value, i);
+        tag_t now = lf_tag();
+        lf_print("Received %d at channel %d at tag " PRINTF_TAG, in[i]->value, i,
+            now.time - lf_time_start(), now.microstep
+        );
         if (in[i]->value != self->count++) {
-            lf_print_error_and_exit("Expected %d.", self->count - 1);
+            lf_print_error("Expected %d.", self->count - 1);
```
petervdonovan, line 31: Why is this error no longer treated as fatal? It looks like the test can now pass even if this error occurs, but the absence or presence of this error should be completely deterministic because this program uses centralized coordination and does not involve physical actions.

test/C/src/federated/FederatedFileReader.lf
```diff
@@ -41,13 +42,21 @@ reactor Check {
...
+    self->received = true;
     if (strcmp("Hello World", in->value) != 0) {
       lf_print_error_and_exit("Expected 'Hello World'");
     }
   =}
+
+  reaction(shutdown) {=
+    if (!self->received) {
+        lf_print_error_and_exit("No input received.");
+    }
```
petervdonovan, line 58: Great, this looks more robust than before.

test/C/src/federated/FeedbackDelay3.lf
```diff
@@ -0,0 +1,40 @@
+target C {
```
petervdonovan, line 1: It might be nice for these new `FeedbackDelay` tests to have a short comment explaining what makes them different from each other or what different bugs they catch. This is because `FeedbackDelay` is one of the most complicated tests, so it can be hard for me to see what is going on just by looking at it.

test/C/src/multiport/InterleavedHierarchy.lf
```diff
@@ -0,0 +1,32 @@
...
+  reaction(t) -> out {=
+    for(int i = 0; i < 2; i++) {
+      lf_set(out[i], self->count++);
+    }
+  =}
+}
+
+reactor B(bank_index: int = 0) {
+  input[3] in: int
+  state count: int = 0  // FIXME
```
petervdonovan, line 19: This looks unfinished.

But this test definitely looks useful, I'm glad it is being added.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
edwardalee: This doesn't really have anything to do with this PR.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
edwardalee: Added a test verifying correctness and removed comment in commit #3d6de027d.

core/src/main/java/org/lflang/generator/c/CCompiler.java
edwardalee: This again has nothing to do with this PR, but I've removed the obsolete function in commit #8fbdd58cd.

test/C/src/federated/FeedbackDelay5.lf
edwardalee: This is so that PhysicalPlant can be imported unmodified. It expects 10 inputs, and if we set to timeout to 1s, it will get 11 and flag an error.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
edwardalee: Not going to do this now. This should be done throughout this file.

core/src/main/java/org/lflang/TimeValue.java
```diff
@@ -209,4 +209,35 @@ public TimeValue add(TimeValue b) {
...
+  /**
+   * Return this time value minus the specified time value but no less than 0.
+   *
+   * <p>The unit of the returned TimeValue will be the minimum of the units of both operands except
+   * if only one of the units is TimeUnit.NONE. In that case, the unit of the other input is used.
+   *
+   * @param b The right operand
+   * @return A new TimeValue (the current value will not be affected)
+   */
+  public TimeValue subtract(TimeValue b) {
```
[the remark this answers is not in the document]
edwardalee: This is a good suggestion, but not going to do this now.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -559,6 +560,53 @@ private void indexExcludedTopLevelReactions(Reactor federatedReactor) {
...
+   */
+  public boolean isInZeroDelayCycle() {
+    if (_isInZeroDelayCycleCalculated) return _isInZeroDelayCycle;
+    _isInZeroDelayCycleCalculated = true;
+    var visited = new HashSet<FederateInstance>();
+    return _isInZeroDelayCycleInternal(this, this, visited);
+  }
+
+  /** Internal helper function for isInZeroDelayCycle(). */
+  private boolean _isInZeroDelayCycleInternal(
```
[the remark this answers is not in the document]
edwardalee: I think the caching approach has the same effect. It only gets done once.  The only real difference with the caching approach right now is that it doesn't get done until it is needed. In the future, when we have dynamically changing federations, however, the caching approach will be better.

test/C/src/federated/DistributedMultiport.lf
edwardalee: Good catch!  Fixed in commit #6c2ffa110.

test/C/src/federated/FeedbackDelay3.lf
edwardalee: Good suggestion.  Comments added in commit #7408a89d5.

test/C/src/federated/AfterDelaysWithMultiport.lf
edwardalee: Good suggestion.  Resolved in commit #8100c8420.

lhstrh:
:star_struck:

:heavy_check_mark: :heavy_check_mark: :heavy_check_mark:

lhstrh (approved):

petervdonovan (approved):

PR #2143, opened 2023-12-22
Adjustments to launch script

lhstrh:


core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -197,26 +197,18 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
...
                 "\n",
-                "echo \"RTI has exited. Wait for federates to exit.\"",
-                "# Wait for launched processes to finish.",
-                "# The errors are handled separately via trap.",
+                "EXITED_SUCCESSFULLY=true",
+                "wait $RTI || EXITED_SUCCESSFULLY=false",
                 "for pid in \"${pids[@]}\"",
                 "do",
-                "    wait $pid || exit $?",
+                "    wait $pid || EXITED_SUCCESSFULLY=false",
```
edwardalee, line 209: According to the bash man page, wait will return 127 for a non-existent process. Doesn't this mean that if a federate exits before you get to this wait (which seems likely, since the RTI has exited), that you will get 127 and then EXITED_SUCCESSFULLY will be set to false?

lhstrh: I pushed a fix that catches this, but it should be noted that, this way, we could be reporting success if a federate exits with a failure if the process disappears before we start the wait. Maybe this is better than reporting failure upon success, but neither sound like a good situation to me...

lhstrh: It seems to me that we can't really manage this in a bash script. Maybe the federates should report to the RTI, and then the RTI can report success or failure. In other words, the RTI would have to exit last, not first.

edwardalee: Yes, this is a good point. There is one simplification, however, which is that a timeout is always failure. If the RTI gets a SIG_INT, it's because the test failed to finish in time, so this should always report failure.

edwardalee: Actually, in normal termination, the RTI does exit last.  It exits after it gets a RESIGN message from all federates.  I think, however, that there isn't right now a mechanism for a federate to resign with failure. I will look into adding this, but it doesn't help with the deadlocks.

petervdonovan (approved):
Tentatively this looks OK, and I approve merging it, but this bash script has been through so many iterations, without necessarily monotonically decreasing bugginess, that I'm not sure I trust my own judgment anymore. I do not have a particularly strong mastery of advanced bash scripting techniques, and at some point it is possible that we should just throw up our hands and port it to Python.

PR #2145, opened 2024-01-05
Parallel compilation disabled to avoid triggering deadlock in concurrent invocations of CMake

edwardalee:
In CI, macOS tests have been nondeterministically failing because the federated tests lock up and get cancelled after a long time (originally 6 hours, recently modified to 2 hours).  It turns out that these failures were due to the compiler locking up, not the generated federated programs.  This PR disables parallel compilation of the federated programs, which apparently can cause such lock ups.


petervdonovan (approved):
Wow, this is not what I was expecting and I have no idea how this could cause deadlocks. Congratulations on finally fixing this -- until I have the time to find out exactly how the deadlocks happened, I am totally happy to keep parallel compilation disabled.

lhstrh:
Glad we finally found the source of the problem! Disabling for now is the right thing to do, and we can revisit and optimize later.

PR #2147, opened 2024-01-10
Fix for #2087 

### LINKED ISSUE #2087: Distinct labels flagged as duplicate
This bug has been found by @lsk567. The following code yields the error `Duplicate AttrParm 'value'`:

```
reactor Source {
  output out: int
  timer t(1 nsec, 10 msec)
  state s: int = 0

  @label(value="Foo")
  reaction(startup) {= lf_print("Starting Source"); =}

  @label(value="Bar")
  reaction(t) -> out {=
    lf_set(out, self->s++);
    lf_print("Inside source reaction_0");
  =}
}
```

AFAIK, this should be equivalent to the following code, which does not yield errors:

```
reactor Source {
  output out: int
  timer t(1 nsec, 10 msec)
  state s: int = 0

  @label("Foo")
  reaction(startup) {= lf_print("Starting Source"); =}

  @label("Bar")
  reaction(t) -> out {=
    lf_set(out, self->s++);
    lf_print("Inside source reaction_0");
  =}
}
```
The problem appears to me that uniqueness checks are done at the level of the `eContainer`, and while the attributes are associated with the syntactic element they precede, which are different, they in the same container. There must be a way to override this behavior but I'm not sure how. Perhaps @a-sr knows?

a-sr:
Fixes incorrect scope for detection of duplicate IDs in attribute parameters. Now only duplicate parameter IDs in the same attribute raise and error.
Closes #2087.

### DIFF
--- a/core/src/main/java/org/lflang/validation/LFNamesAreUniqueValidationHelper.java
+++ b/core/src/main/java/org/lflang/validation/LFNamesAreUniqueValidationHelper.java
@@ -1,7 +1,13 @@
 package org.lflang.validation;
 
+import java.util.Map;
 import org.eclipse.emf.ecore.EClass;
+import org.eclipse.xtext.naming.QualifiedName;
+import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.validation.NamesAreUniqueValidationHelper;
+import org.eclipse.xtext.validation.ValidationMessageAcceptor;
+import org.lflang.lf.AttrParm;
+import org.lflang.lf.Attribute;
 import org.lflang.lf.LfPackage;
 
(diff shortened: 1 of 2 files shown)


lhstrh (changes requested):
Looks good to me! This should be covered by a unit test, however; one that expects no errors when the same attribute name is used in two different locations and one that expects an error for code that indeed has duplicate attributes.

lhstrh (approved):

a-sr:
Thank you for adding the test. I was quite busy the last days.

lhstrh:
No worries. I realized, however, that we still have some work to do in terms of validation, but that can be done later (see #2162).

PR #2151, opened 2024-01-15
Python docker parent image fixed at version `3.10`

petervdonovan:
Fixes a version mismatch that causes a failure when compiling with CMake because we do not get support Python versions later than 3.10.


lhstrh (approved):
:+1:

PR #2158, opened 2024-01-21
Removal of image-specific commands in case base image is specified

### LINKED ISSUE #2152: Docker Python support should allow the user to install dependencies
Currently the base image in the generated dockerfile appears to be hardcoded and the final const `defaultBaseImage` is used unconditionally rather than being overridden. This should be an easy fix and appears to always have been the plan.

Since it is apparently common to use pip to fix a `requirements.txt` file and to commit `requirements.txt` for reproducibility, it is possible that in the long run we should provide special support for that in addition to the all-purpose solution of just overriding the base image. This is low-priority, if we choose to do it at all.

lhstrh:
Fixes #2152 and addresses the problem that different base images support different commands for installing dependencies.

### DIFF
--- a/core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
+++ b/core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
@@ -70,7 +70,7 @@ public class CDockerGenerator extends DockerGenerator {
     if (baseImage().equals(defaultImage())) {
       return """
           # Install build dependencies
-          RUN set -ex && apk add --no-cache %s musl-dev cmake make,
+          RUN set -ex && apk add --no-cache %s musl-dev cmake make
           """
           .formatted(compiler);
     } else {
--- a/core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
+++ b/core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
@@ -4,7 +4,6 @@ import org.eclipse.xtext.xbase.lib.IterableExtensions;
 import org.lflang.generator.LFGeneratorContext;
 import org.lflang.target.Target;
 import org.lflang.target.property.BuildCommandsProperty;
-import org.lflang.target.property.DockerProperty.DockerOptions;
 import org.lflang.util.StringUtil;
 
 /**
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
```diff
@@ -63,13 +63,32 @@ protected String generateDockerFileContent() {
...
+    if (baseImage().equals(defaultImage())) {
+      return """
+          # Install build dependencies
+          RUN set -ex && apk add --no-cache %s musl-dev cmake make,
+          """
+          .formatted(compiler);
+    } else {
+      return """
+          # Check for build dependencies
+          RUN which make && which cmake && which %s,
```
petervdonovan, line 79:
```suggestion
          RUN which make && which cmake && which %s
```

core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
```diff
@@ -63,13 +63,32 @@ protected String generateDockerFileContent() {
...
   }
 
+  @Override
+  protected String generateRunForBuildDependencies() {
+    var config = context.getTargetConfig();
+    var compiler = config.target == Target.CCPP ? "g++" : "gcc";
+    if (baseImage().equals(defaultImage())) {
+      return """
+          # Install build dependencies
+          RUN set -ex && apk add --no-cache %s musl-dev cmake make,
```
petervdonovan, line 73:
```suggestion
          RUN set -ex && apk add --no-cache %s musl-dev cmake make
```

core/src/main/java/org/lflang/generator/docker/RtiDockerGenerator.java
```diff
@@ -17,19 +17,20 @@ public RtiDockerGenerator(LFGeneratorContext context) {
...
             cd build && \\
             cmake ../ && \\
             make && \\
             make install
 
         # Use ENTRYPOINT not CMD so that command-line arguments go through
         ENTRYPOINT ["RTI"]
-        """;
+        """
+        .formatted(baseImage(), generateRunForBuildDependencies());
```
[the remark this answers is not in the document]
lhstrh: You're completely right.

petervdonovan (approved):

PR #2160, opened 2024-01-22
Fix some zephyr build issues

erlingrj:
1. The user-provided `prj.conf` was not included.
2. We generated CMake code that linked against some Zephyr kernel library this generated a cyclic-dependency. I am trying to remove this linkage and lets see if all the tests still pass

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
@@ -144,9 +144,12 @@ public class CCmakeGenerator {
     // Setup the project header for different platforms
     switch (platformOptions.platform()) {
       case ZEPHYR:
-        cMakeCode.pr("# Set default configuration file. To add custom configurations,");
-        cMakeCode.pr("# pass -- -DOVERLAY_CONFIG=my_config.prj to either cmake or west");
+        cMakeCode.pr("# Include default lf conf-file.");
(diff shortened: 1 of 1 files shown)


erlingrj:
Unfortunately, I am not able to find a good solution to this. It is strangely complicated to just link the following three things without creating a CMake target cycle or other linker issues:
1. Reactor-c library
2. LF application
3. Zephyr kernel

Im afraid I will have to come back to this in March

erlingrj:
I am updating this PR to just fix a tiny mistake in the code-generated CMakeLists for Zephyr. Now we  check for any `prj.conf` file in the src-gen directory and includes it. This is used by our custom west command.

erlingrj:
@lhstrh, all tests pass except the label-test which doesn't pass when I re-run it, even though there now is a label. We don't need to rerun the entire test suite to get it all green I think...

lhstrh:
The reason for the test failure is clear from the output:

> Error: Label error. Requires at least 1 of: feature, enhancement, fix, bug, bugfix, cleanup, refactoring, documentation, docs, test, tests, testing, upgrade, bump, dependency, version, misc, miscellaneous, ignore, exclude, trivial, duplicate, question, invalid, wontfix, ci, automated, experimental. Found: zephyr

I added the label `bugfix`.

lhstrh (approved):
:+1:

PR #2167, opened 2024-01-26
Use Zephyr's CMake extension to correctly build and link reactor-c

erlingrj:
We had experienced some issues with linking both reactor-c and the user-generated app to the Zephyr project. It is resolved by using Zephyrs cmake extension to create the reactor-c lib.

Also to avoid a name collision I have renamed the library from core to reactor-c.


edwardalee (approved):
LGTM!

PR #2169, opened 2024-01-26
Use login shell so ~/.bash_profile is sourced

edwardalee:
In resurrecting the MQTT examples in the playground, I realized that the shell script that we generate to launch a federate does not source `~/.bash_profile`, which makes it much more difficult to use external libraries with federated programs.  The federates can be launched manually, but the shell script to launch them will not work. In the case of MQTT, it needs to find the Paho MQTT dynamically linked libraries, and without this change, there is no way to tell the script where to find them.


edwardalee:
I don't think this two-byte change really requires a review, so rather than waste people's time, I'm putting this in the merge queue.

Issue #2150, opened 2024-01-10
Watchdogs sometimes fail on program termination


edwardalee:
The one and only test for watchdogs is in `test/C/src/concurrent/failing/Watchdog.lf`.  This test sometimes fails in CI for what seems to be a very simple reason: The background watchdog thread that is waiting for the timeout can time out after the program has terminated and has completed at least part of its termination sequence, which includes freeing memory devoted to the event queue. This thread then tries to schedule something on the event queue, resulting in a segmentation fault or worse.

lhstrh:
@edwardalee, has this been resolved, or are there more changes necessary?

edwardalee:
This was resolved by @erlingrj in [PR 341 of reactor-c](https://github.com/lf-lang/reactor-c/pull/341).

PR #2172, opened 2024-01-28
Add watchdogs to environment struct

erlingrj:
These changes move the watchdogs to the environment struct and thus make the watchdog compatible with the future enclave support. The corresponding changes in reactor-c also address the termination issue for the watchdogs as described in #2150 

See the corresponding reactor-c PR : https://github.com/lf-lang/reactor-c/pull/341


core/src/main/java/org/lflang/generator/c/CWatchdogGenerator.java
```diff
@@ -83,8 +92,7 @@ protected static int generateInitializeWatchdogs(CodeBuilder code, ReactorInstan
     if (foundOne) {
       code.pr(temp.toString());
     }
-    code.pr("SUPPRESS_UNUSED_WARNING(_lf_watchdog_count);");
-    return watchdogCount;
+    enclaveInfo.numWatchdogs += watchdogCount;
```
edwardalee, line 95: This doesn't look quite right.  Won't `watchdogCount` always be 0?

edwardalee (changes requested):
Looks good to me except one possible flaw identified. Is it a flaw?

core/src/main/java/org/lflang/generator/c/CWatchdogGenerator.java
erlingrj: I think you overlooked the problem the place where it is incremented. Right above

edwardalee (approved):
LGTM!  The LSP test failure is probably unrelated (these tests are randomized) and probably due to the code generator generating blank lines, which the Python validator in the LSP doesn't like but the Python interpreter apparently has no problem with.

PR #2173, opened 2024-01-31
Use local rti in DistributedStopDecentralizedContainerized.lf

lhstrh:


test/C/src/docker/federated/DistributedStopDecentralizedContainerized.lf
```diff
@@ -6,7 +6,7 @@
 target C {
   coordination: decentralized,
   docker: {
-    rti-image: "rti:latest"
+    rti-image: "rti:local"
```
edwardalee, line 9: What does `ti:local` mean?  I couldn't find any documentation for this.  There is no mention here: https://www.lf-lang.org/docs/reference/containerized-execution

Note that searching for "docker" on the lf-lang website give this:

image

Similarly, a search for rti gives no results.

edwardalee (changes requested):
Not clear what this means.

test/C/src/docker/federated/DistributedStopDecentralizedContainerized.lf
lhstrh: We index the documentation weekly, so the description isn't indexed yet. The relevant PR is https://github.com/lf-lang/lf-lang.github.io/pull/227.

For Docker users, `rti:local` is very clear. The naming convention for image names is `<user-name>/<image-name>:<tag-name>`. So I simply reserved the `local` tag to refer to a locally built container (i.e., not pulled from Docker Hub).

lhstrh: Actually, I would expect the "Docker" keyword to produce hits in Algolia, but it doesn't. So it seems we have a search bug.

lhstrh: See https://github.com/lf-lang/lf-lang.github.io/issues/229 and https://github.com/lf-lang/lf-lang.github.io/pull/230.

PR #2175, opened 2024-01-31
CI for macOS and Windows only ran periodically, not continuously

lhstrh:
~~Soon as we find a fix for the mysterious random hanging in the compiler on macOS we should revert this, but in the interested of not holding up everything else we're doing, I propose we simply disable the macOS tests for the time being.~~

~~I changed this PR from disabling the macOS tests to switching to much faster M1 runners. This does not address the bottleneck problem that motivates my idea to move macOS tests to the weekend, so we could still do this, but it makes the problem less severe. For reference, the `build-toolchain` job took 15 minutes on a regular macOS runner, and only 3 minutes on the M1.~~

~~This likely also makes the dreaded concurrency bug go away, although, of course, it will still lurk somewhere. I think we are better off spending our time on other things; we can attempt to fix the problem once we have a bug report from a user.~~

The experiments in this PR show that we can speed up CI quite a bit by using fancier runners, but that the financial costs are significant. This brings me back to the initial proposal: do not run macOS/Windows tests that clog up the CI pipeline all the time, but only periodically (at least until we have a solution that addresses the bottleneck).

### DIFF
--- a/.github/workflows/all-misc.yml
+++ b/.github/workflows/all-misc.yml
@@ -2,6 +2,9 @@
 name: CI (misc)
 
 on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 8 * * 6'
   workflow_dispatch:
   push:
     branches:
@@ -27,7 +30,7 @@ jobs:
     needs: check-diff
     uses: ./.github/workflows/build.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
     if: ${{ needs.check-diff.outputs.run_build == 'true' }}
 
   # Run tests for the standalone compiler.
@@ -36,7 +39,7 @@ jobs:
     needs: check-diff
     uses: ./.github/workflows/cli-tests.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
 
   # Run language server tests.
   lsp:
@@ -44,7 +47,7 @@ jobs:
     needs: check-diff
     uses: ./.github/workflows/lsp-tests.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
 
   check-labels:
     uses: ./.github/workflows/check-labels.yml
--- a/.github/workflows/all-targets.yml
+++ b/.github/workflows/all-targets.yml
@@ -2,6 +2,9 @@
 name: CI (by target)
 
 on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 8 * * 6'
   workflow_dispatch:
   push:
     branches:
--- a/.github/workflows/non-target-specific-extended.yml
+++ b/.github/workflows/non-target-specific-extended.yml
@@ -22,18 +22,18 @@ jobs:
     if: ${{ inputs.all || github.event.pull_request.draft }}
     uses: ./.github/workflows/build.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
 
   # Run tests for the standalone compiler.
   cli:
     if: ${{ inputs.all || github.event.pull_request.draft }}
     uses: ./.github/workflows/cli-tests.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft  }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
 
   # Run language server tests.
   lsp:
     if: ${{ inputs.all || github.event.pull_request.draft }}
     uses: ./.github/workflows/lsp-tests.yml
     with:
-      all-platforms: ${{ !github.event.pull_request.draft  }}
+      all-platforms: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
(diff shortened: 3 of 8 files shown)


lhstrh:
The macOS tests have been a pain in general because we only get 5 runners. Perhaps a better (and permanent) solution would be to run CI for macOS only once a week.

edwardalee:
It's not just macOS issues.  I'm getting failures in LSP tests (a Python parse error!) and code coverage failures.
I don't think it's a good idea to suspend the Mac tests.  The failures are not repeatable (I ran lfc-dev 3000 times on my Mac for an example that hung in CI without a hang).  We need to keep trying until where lfc is deadlocking.

lhstrh:
> It's not just macOS issues. I'm getting failures in LSP tests (a Python parse error!) and code coverage failures. I don't think it's a good idea to suspend the Mac tests. The failures are not repeatable (I ran lfc-dev 3000 times on my Mac for an example that hung in CI without a hang). We need to keep trying until where lfc is deadlocking.

Yes, there are other issues, but let's not throw everything on one big pile. The problem I have with the current approach is that it wastes tremendous resources and is a huge drag on development velocity. For each time something is thrown off the merge queue, we waste 3 hours of CI time (a lot more if you consider concurrent execution of jobs). For each time that a single job fails, this demands someones attention, pulling them away from useful work. 

Let me address the issues you bring up separately and then explain why I think it's a good idea remove macOS and Windows tests from our critical path.

1) LSP tests fail sometimes in unexpected ways that have nothing to do with changes. This is because these tests are randomized. This may be annoying, but the incidence of these kinds of failures is low.

2) CovCov sometimes errors out and lets CI fail. This is annoying, and outside of our control, unfortunately. We can remove code coverage or report it elsewhere. Our attempts at making it more robust have not been fruitful.

3) We have a concurrency bug in our compiler than is extremely difficult to reproduce locally but surfaces reliably in CI but with no context that helps us address the issue. If we want to find this bug, we should use better tools. This has also been pointed out by @petervdonovan, and I think he's right. Keeping things "as-is" with macOS CI runs is not going to help us solve this problem.

In our toolchain, we rely on several cross-platform abstractions: JVM for the compiler and POSIX for the C runtime; NodeJS for the TypeScript target, but also Bash, and I'm probably overlooking a few things. All of these promise that if things run on Linux, they should also run on macOS. The same can be said for Windows WSL. Nonetheless, we are being conservative and testing across platforms, anyway, and indeed sometimes we catch problems that way. But most of the time we don't. Yet, our entire CI pipeline is dominated by the cost of doing this, because we only get 5 macOS runners (vs 60 Linux runners, which also happen to execute faster). This typically means that we have tons of Linux runners idling, ready to pick up work, while macOS runners are always busy working on a backlog of jobs that's waiting for them. What I'm saying is: I don't think the cost of this approach is worth it.

Therefore, what I think we should do is be "Linux first" and only test Windows and macOS periodically, on weekends, when CI usage is lower.

lhstrh:
Regarding the CodeCov issues, perhaps #2176 helps.

lhstrh:
Note: if we accept this solution, we need to drop several tests from the "required" list.

cmnrd:
Is there a way to only run Windows/MacOs test if a certain label is set? Or only on the merge queue? I am all for reducing the CI workload, but I am reluctant to skip tests before merging. The consequence will be that we merge breaking changes, and I rather be sure that at least all the tests pass before the fact, then having to resolve issues quickly after the fact.

lhstrh:
The labeling approach is a good suggestion, but that doesn't address the problem of getting thrown off the merge queue due to flaky macOS failures. I can also fold this PR and just use forced merges until the macOS issues are dealt with.

cmnrd:
Could you point me to a couple of those problematic runs (or do we have a collection somewhere)?

cmnrd:
The alternative path would be to simply drop MacOS support... Apple's restrictive virtualization policies and expensive licensing don't make it particularly attractive for open-source projects to support MacOS. Maybe we shouldn't and instead provide easy to set up VMs and/or docker images as a Linux-like development environment on top of MacOS.

lhstrh:
@cmnrd: Here's the latest example: https://github.com/lf-lang/reactor-c/actions/runs/7764998996/job/21193284538?pr=342

It just gets stuck -- no output.

edwardalee:
One issue is that we don't know for sure whether it is lfc that is hanging or cmake.  When I was able to once reproduce it on my Mac, it was cmake.  We disabled parallel uses of cmake, and this made the freeze less frequent.  It also made the freeze happen on non-federated tests.

Could we update the output to not just tell which test is running, but also tell us when lfc completes and cmake starts?

edwardalee:
Also, when I've seen a freeze on a particular test, I've put my Mac in a loop repeatedly recompiling the test.  I have had it run for hours for thousands of cycles and have not been able to reproduce the freeze.  It took a few thousand runs before to reproduce the cmake freeze, so I guess this isn't surprising.

lhstrh:
> One issue is that we don't know for sure whether it is lfc that is hanging or cmake. When I was able to once reproduce it on my Mac, it was cmake. We disabled parallel uses of cmake, and this made the freeze less frequent. It also made the freeze happen on non-federated tests.
> 
> Could we update the output to not just tell which test is running, but also tell us when lfc completes and cmake starts?

CMake is ran as a subprocess of `lfc`. We could augment the output to be less binary and show stages like "Generating Code...", "Building...", etc.

lhstrh:
Having discussed this PR in more detail with @edwardalee, @elgeeko1, and @petervdonovan, we have agreed to merge this PR in the interests of reducing pressure on CI quickly. The consequences are as follows:
 - we no longer run macOS or Windows workflows automatically upon pushing, merging, or in the merge queue
 - to gain confidence that these platforms work prior to merging [manual workflow dispatch](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) can be used to trigger these platform-specific tests
 - it is the responsibility of the contributors and reviewers to pay attention that when changes are made to macOS or Windows platform support, the appropriate tests are run manually
 - whatever problems aren't caught through manual runs will be discovered through an automatic a weekly run
 - before releasing, manual CI runs will be done on the `release` branch to assure that all tests are passing
 - if at any point we conclude that the cost of maintaining Windows and macOS support is still too high, we can decide to formally drop support for these platforms

Note: the macOS/Windows tests will henceforth also no longer be marked as "required to pass before merging," which is configured separately, through GitHub's settings of this repository.

PR #2178, opened 2024-02-01
Rp2040 multithreaded target support 

sberkun:
This PR adds multithreaded support from the RP2040 platform. Builds off of the work in #1831.

Corresponding PR in reactor-c: https://github.com/lf-lang/reactor-c/pull/344. This PR will be updated to reactor-c's main branch once the corresponding PR is merged.


core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -504,9 +505,9 @@ private static String setUpMainTargetRp2040(
     code.pr("pico_sdk_init()");
     code.newLine();
     code.pr("add_subdirectory(core)");
-    code.pr("target_link_libraries(core PUBLIC pico_stdlib)");
-    code.pr("target_link_libraries(core PUBLIC pico_multicore)");
-    code.pr("target_link_libraries(core PUBLIC pico_sync)");
+    code.pr("target_link_libraries(reactor-c PUBLIC pico_stdlib)");
+    code.pr("target_link_libraries(reactor-c PUBLIC pico_multicore)");
+    code.pr("target_link_libraries(reactor-c PUBLIC pico_sync)");
```
sberkun, line 510: The library in `core` was renamed from `core` to `reactor-c`, breaking single-threaded rp2040 support. This is bugfix.

There might be a cleaner way to link the pico libraries to the platform files (perhaps linking from within core/CMakeLists.txt), but that would probably be better to handle in a separate PR.

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -75,10 +75,10 @@ protected PlatformOptions fromString(String string, MessageReporter reporter) {
   @Override
   public void validate(TargetConfig config, MessageReporter reporter) {
     var singleThreaded = config.get(SingleThreadedProperty.INSTANCE);
-    if (!singleThreaded && config.get(PlatformProperty.INSTANCE).platform == Platform.RP2040) {
+    if (!singleThreaded && !config.get(PlatformProperty.INSTANCE).platform.isMultiThreaded()) {
```
sberkun, line 78: I'm guessing that this check was meant to be more general than just the RP2040, but was introduced before the `platform.isMultiThreaded` attribute. Currently all platforms are multithreaded so this check isn't very useful yet.

edwardalee (approved):
LGTM!

PR #2180, opened 2024-02-02
Reduced scope for conflicting main reactors check

lhstrh:
We do a check to find conflicting main reactors in the package. We should limit the scope of the search to the `src`directory because there could be other sources present in the package outside of the `src` directory that should not be labeled as conflicting.

This fix was prompted by the use of the `lf-lang/action-check-lf-files` which puts a checkout of `lingua-franca` in the workspace.

~~Required for https://github.com/lf-lang/lf-pico-template/pull/2.~~


edwardalee (approved):
LGTM.

PR #2182, opened 2024-02-03
Have colorized outputs from C/Cpp compilers as default

erlingrj:
This simply adds set(CMAKE_COLOR_DIAGNOSTICS ON) to the CMakeLists.txt generated for C and Cpp programs. This will colorize the compiler output so it is much easier to separate errors from warnings etc.


lhstrh (approved):
Great idea! Thanks for suggesting this!

PR #2183, opened 2024-02-05
Fixed cmake deprecation warnings in the C target

### LINKED ISSUE #2120: CMake warnings
In many C tests, at least, I always get the following warning:
```
Multiple markers at this line
- Warning in imported file: /Users/edwardlee/LinguaFranca/test/C/fed-gen/FeedbackDelay/src/federate__p.lf
- CMake Deprecation Warning at CMakeLists.txt:26 (cmake_policy):

    The OLD behavior for policy CMP0023 will be removed from a
   future version
  of CMake.

  The cmake-policies(7) manual explains
   that the OLD behaviors of all
  policies are deprecated and
   that a policy should be set to OLD only under
  specific short-term
   circumstances.  Projects should be ported to the NEW
  behavior
   and not rely on setting a policy to OLD.
```
For example, opening `src/federated/FeedbackDelay.lf` in Epoch results in this warning being highlighted at the top of the file.

cmnrd:
The C code generator set the [CMP0023 Policy](https://cmake.org/cmake/help/latest/policy/CMP0023.html) to `OLD` for an inexplicable reason. This PR simply removes that code. 

Likely the old policy setting was needed as some kind of workaround. If some of the tests fail, we will need to find an actual solution to the underlying problem.

Fixes #2120

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
@@ -231,17 +231,6 @@ public class CCmakeGenerator {
     cMakeCode.pr("set(CMAKE_CXX_STANDARD 17)");
     cMakeCode.pr("set(CMAKE_CXX_STANDARD_REQUIRED ON)");
     cMakeCode.newLine();
-    if (!targetConfig.getOrDefault(CmakeIncludeProperty.INSTANCE).isEmpty()) {
-      // The user might be using the non-keyword form of
-      // target_link_libraries. Ideally we would detect whether they are
-      // doing that, but it is easier to just always have a deprecation
-      // warning.
-      cMakeCode.pr(
-          """
-                cmake_policy(SET CMP0023 OLD)  # This causes deprecation warnings
-
-                """);
-    }
 
     // Set the build type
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
Less code! Yay! :rocket:

petervdonovan:
> The C code generator sets the [CMP0023 Policy](https://cmake.org/cmake/help/latest/policy/CMP0023.html) to OLD for an inexplicable reason.

I'm pretty sure I'm the one who did that.

IIRC this was for compatibility with ROS (specifically, with `ament`). I don't remember which test prompted this change (it might even have been something in the `examples` repo), but it could have been one of the C serialization tests.

cmnrd:
> > The C code generator sets the [CMP0023 Policy](https://cmake.org/cmake/help/latest/policy/CMP0023.html) to OLD for an inexplicable reason.
> 
> I'm pretty sure I'm the one who did that.
> 
> IIRC this was for compatibility with ROS (specifically, with `ament`). I don't remember which test prompted this change (it might even have been something in the `examples` repo), but it could have been one of the C serialization tests.

Yes, see the second file that changed. There was an inconsistent use of ament in the code generator. Adding `PUBLIC` there fixed the problem.

Issue #2154, opened 2024-01-18
CMake builds should support use of CMAKE_TOOLCHAIN_FILE for cross-compilation


petervdonovan:
Currently, it isn't possible to specify the toolchain file, e.g. using the `cmake-include` target property, because the files specified using that target property do not get plugged into the main generated CMake file until after many commands have been run, at which point it is too late to set the `CMAKE_TOOLCHAIN_FILE` variable.

We should fix this, either by allowing general user-supplied CMake variables to be globally set in the beginning of the generated CMake file, or (as a more targeted solution) by allowing the user to specify a toolchain file. If we do this, it should in principle be possible to get fairly general support for cross-compilation, without having to ship `lfc` (or even `lingo`) with tightly integrated support for specific individual toolchains.

Ideally, we would test such cross-compilation support using an emulator such as QEMU. I suppose that testing in CI would be the hard part of this feature request.

erlingrj:
I was just about to make this very issue. I like the general option better because I could use that for the Zephyr-support. The hack I am currently using as a workaround is passing a script to the `build` target property:
```
build.sh
#!/bin/bash
cmake -Bbuild -DCMAKE_TOOLCHAIN_FILE=/home/erling/dev/rpidev/toolchain_rpi.cmake $LF_SOURCE_GEN_DIRECTORY
make -C build
```

PR #2185, opened 2024-02-06
redid the Stm32 support. last one was having rebase issues

naichenzhao:


core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
...
+      // By default, we set it to STM32f446RE
+      code.pr("set(MCU_FAMILY STM32F4xx)");
+      code.pr("set(MCU_MODEL STM32F446xx)");
+    } else {
+      code.pr("set(MCU_FAMILY STM32" + boardProperties[0].substring(0, 2) + "xx)");
+      code.pr("set(MCU_MODEL STM32" + boardProperties[0].substring(0, 4) + "xx)");
+    }
+
+
+    code.pr("set(CPU_PARAMETERS -mcpu=cortex-m4 -mthumb -mfpu=fpv4-sp-d16 -mfloat-abi=softfp)");
```
lhstrh, line 574: It seems that these parameters should depend on the chosen platform.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
...
+    code.newLine();
+
+    // Define linker and startup scropts
+    if (boardProperties.length < 1 || boardProperties[0].equals("")) {
+      // By default, we set it to STM32f446RE
+      code.pr("set(STARTUP_SCRIPT ${STM_DIR}/startup_stm32f446xx.s)");
```
lhstrh, line 580: Also platform board dependent.

lhstrh:
@erlingrj and @tanneberger: it seems to me that a better place to implement this would be in `lingo`, but @naichenzhao would need help with that. Luckily, this isn't a big PR, so it might be worth bringing it to `lingo` right away, and not merge it into `lingua-franca` first. WDYT? Would you be able to help? We do need this functionality soon though, because @naichenzhao and the people in his lab will be using it extensively for their project...

core/src/main/resources/lib/platform/stm32/arm-none-eabi-gcc.cmake
```diff
@@ -0,0 +1,23 @@
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_PROCESSOR arm)
+
+# Some default GCC settings
+set(TOOLCHAIN_PREFIX arm-none-eabi-)
+set(FLAGS
```
erlingrj, line 6: Does this really belong here? E.g. `--specs=nano.specs` instructs it to link with newlib nano. Maybe the user wants full newlib, or another library altogether?

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -223,6 +228,13 @@ CodeBuilder generateCMakeCode(
     cMakeCode.pr("endif()");
 
     cMakeCode.pr("# Require C11");
+    switch (platformOptions.platform()) {
```
erlingrj, line 231: Move this so that it doesnt appear under that comment?

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
     code.newLine();
     return code.toString();
   }
-}
+
+
+  private static String setUpMainTargetStm32(
+          boolean hasMain, String executableName, Stream<String> cSources, String[] boardProperties) {
+    var code = new CodeBuilder();
+    code.pr("\n# ########################################################");
```
erlingrj, line 548: I think we should use consistent formatting on the comments generated in CMakeLists.txt

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
...
+    code.newLine();
+    code.newLine();
+
+    code.pr("add_subdirectory(core)");
+    code.newLine();
+    code.pr("set(LF_MAIN_TARGET " + executableName + ")");
+
+    // Declaration preamble
+    code.pr("set(PROJECT_DIR ${CMAKE_CURRENT_SOURCE_DIR})");
+    code.pr("set(STM_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../STM_sdk)");
```
erlingrj, line 560: This seems to require that you have a directory called `STM_sdk` in some relative position wrt to src-gen directory? I think we want the user to download the STM SDK themselves and define an environment variable pointing to it

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
erlingrj: These might actually be commen for all STM32s

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
...
+            "        -Wuseless-cast\n" +
+            "        -Wsuggest-override>\n" +
+            "    $<$<CONFIG:Debug>:-Og -g3 -ggdb>\n" +
+            "    $<$<CONFIG:Release>:-Og -g0>)\n");
+    code.newLine();
+    code.pr("target_link_options(${LF_MAIN_TARGET} PRIVATE\n" +
+            "    -T${MCU_LINKER_SCRIPT}\n" +
+            "    ${CPU_PARAMETERS}\n" +
+            "    -Wl,-Map=${CMAKE_PROJECT_NAME}.map\n" +
+            "    --specs=nosys.specs\n" +
```
erlingrj, line 646: The FLAGS part of `arm-none-eabi-gcc.cmake` should be moved here

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -519,4 +540,134 @@ private static String setUpMainTargetRp2040(
...
+            "    COMMAND ${CMAKE_OBJCOPY} -O ihex $<TARGET_FILE:${LF_MAIN_TARGET}>\n" +
+            "    ${LF_MAIN_TARGET}.hex\n" +
+            "    COMMAND ${CMAKE_OBJCOPY} -O binary $<TARGET_FILE:${LF_MAIN_TARGET}>\n" +
+            "    ${LF_MAIN_TARGET}.bin)");
+    code.newLine();
+
+
+    code.pr("\n# ######################################################");
+    code.pr("\n# ############## [ End STM32 main target] ##############");
+    code.pr("\n# ######################################################");
```
erlingrj, line 668: Comment formatting not consistent with the rest of the CMakeList.txt

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -912,6 +912,12 @@ protected void copyTargetFiles() throws IOException {
           // VS Code configurations
           FileUtil.copyFileFromClassPath("/lib/platform/rp2040/launch.json", vscodePath, true);
         }
+        case STM32 -> {
+          // Copy over STM32 library (Currently hard-coded)
+          FileUtil.copyFileFromClassPath(
+              "/lib/platform/stm32/arm-none-eabi-gcc.cmake", fileConfig.getSrcGenPath(), true);
```
erlingrj, line 918: Will this work when lfc is just distributed as a JAR?

core/src/main/java/org/lflang/target/property/type/PlatformType.java
```diff
@@ -18,6 +18,7 @@ public enum Platform {
     LINUX("Linux", true),
     MAC("Darwin", true),
     ZEPHYR("Zephyr", true),
+    STM32("Stm32", false),
```
erlingrj, line 21: Should we call it STM32? I dont think it is commonly written Stm32

erlingrj (changes requested):
This looks promising @naichenzhao! 

On a higher level. I worry that we will just pile on code and complexity for each target we want to support. It would be a huge improvement if STM32/Pico/Zephyr/++ support could be captured in a single CMake file that is not code-generated but just embedded in the JAR. @petervdonovan discuss this a little her: https://github.com/lf-lang/lingua-franca/issues/2154. 

I think we can merge in the STM32 stuff. But after this we should think a little about how we can make this scale better.

core/src/main/java/org/lflang/generator/c/CGenerator.java
petervdonovan: Yes, I believe everything on the classpath should be bundled in the JAR

lhstrh:
@naichenzhao, do you think you'll find an opportunity to work on this anytime soon? This efforts needs to be revived...

core/src/main/resources/lib/platform/stm32/arm-none-eabi-gcc.cmake
naichenzhao: this im not super sure about, I essentially copied this over from the generic STM32 toolchain so I assumed it should be correct

erlingrj: OK, then it is probably fine to keep it in the toolchain fiile

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
erlingrj: Any update here? I think we should require an environmental variable pointing to where STM_sdk is located and the user can then put it anywhere

naichenzhao: This currently requires a hard-coded location. I was planning on keeping it like this for now, since we also have a sample skeleton that also gives the STM_sdk

erlingrj: I dont think this is a very scalable solution. I think we should require an environmental variable called `STM_SDK_PATH` to be set. The generated CMakeLists can verify that it is set and produce an error if not.

core/src/main/java/org/lflang/generator/c/CMainFunctionGenerator.java
```diff
@@ -80,6 +80,10 @@ private String generateMainFunction() {
       case RP2040 -> {
         return String.join("\n", "int main(void) {", "   return lf_reactor_c_main(0, NULL);", "}");
       }
+
+      case STM32 -> {
+        return String.join("\n", "int main(void) {", "   return lf_reactor_c_main(0, NULL);", "}");
+      }
```
erlingrj, line 86: Identical to RP2040, so merge it into the same case?

naichenzhao: I personally feel like its easier to read when we have the separation for STM32 and rp2040 though

erlingrj:
This PR depends on #2432 to be merged first.

Issue #2179, opened 2024-02-01
Generic Reactors does not sanitize characters for filesystem


SheaFixstars:
Creating a generic reactor that has a more complex type including `:,<,>` chars causes files to be written in the generated output. The symbols while allowed within the filesystem cause the make toolchain to stop functioning with the following message.

The build fails with the given output for the test
```
CMakeFiles/generic_char_sanatize.dir/build.make:101: *** target pattern contains no '%'.  Stop.
```

The test LF file to make the issue occur. EDIT: Updated the example to properly work if it does compile.
```
target CCpp;

preamble 
{=
    #include <memory>
=}

reactor TestGeneric<T>
{
    input test_input:T;

    reaction(test_input) 
    {=
        if (test_input->is_present)
        {
            lf_print("Got the test input");
        }
    =}
}

main reactor
{
    test = new TestGeneric<std::shared_ptr<std::string>>();

    reaction(startup) -> test.test_input
    {=
        lf_set(test.test_input, std::make_shared<std::string>(*(new std::string("test"))));
    =}
}
```

Below is a dir listing for the src-gen showing the issue. 
```
shea@FSS-000070-Shea:~/lf_tests$ ls -1 src-gen/generic_char_sanatize/
CMakeLists.txt
_generic_char_sanatize_main.cpp
_generic_char_sanatize_main.h
_testgenericstd::shared_ptr.cpp
_testgenericstd::shared_ptr.h
core
generic_char_sanatize.cpp
include
lib
```

I would assume this would just be a matter of sanitizing the invalid chars but an unsure of all the locations in code this would touch

SheaFixstars:
I forgot to mention this is with LF 0.6.0.

cmnrd:
Does it work if you escape the type with `{= =}`? E.g. `test = new TestGeneric<{=std::shared_ptr<std::string>=}>()`

petervdonovan:
IIRC we already have some mechanism to sanitize certain characters, but we just forgot to handle these. Should be an easy fix.

SheaFixstars:
I edited the main example as I didn't test properly if it actually works with the problem solved. So it should work now :) when the issue is resolved.

PR #2186, opened 2024-02-06
Escape special characters in type arguments.

petervdonovan:
This resolves multiple FIXMEs in the code as well as issue #2179.


lhstrh (approved):
Thanks for the quick fix, @petervdonovan!

PR #2190, opened 2024-02-10
Replace use of deprecated APIs in C tests

edwardalee:
This PR is a companion to [353 in reactor-c](https://github.com/lf-lang/reactor-c/pull/353), which removes deprecated API macros originally used in reaction bodies in the C target.


edwardalee:
This PR is ready for review.  There is a [companion PR in reactor-c](https://github.com/lf-lang/reactor-c/pull/353) that requires a [further companion review in the benchmarks repo](https://github.com/lf-lang/benchmarks-lingua-franca/pull/59) in order to get all tests to pass.

lhstrh:
Looks like the right thing to do. It implies that the next version should be `1.0.0`, not `0.7.0`...

lhstrh:
@arengarajan99 worked on this a while ago but hit roadblocks because the code itself wasn't well organized. That situation has improved a lot since then, so I'm curious whether his attempt could be reinvigorated. 

On the versioning: I think that we should plan a 1.0 version more carefully and release it only once we can confidently consider it stable... We can choose to strictly follow semantic versioning rules from v1.0 and on...

PR #2192, opened 2024-02-13
Plugin API for tracing

petervdonovan:
[See the companion PR in reactor-c for discussion.](https://github.com/lf-lang/reactor-c/pull/342) The changes that were required here to make the reactor-c changes work are relatively minor.


core/src/main/java/org/lflang/generator/CodeBuilder.java
```diff
@@ -126,7 +126,8 @@ public void prSourceLineNumber(EObject eObject, boolean suppress) {
         }
       }
       // Extract the filename from eResource, an astonishingly difficult thing to do.
-      String filePath = CommonPlugin.resolve(eObject.eResource().getURI()).path();
+      String filePath =
+          CommonPlugin.resolve(eObject.eResource().getURI()).path().replace("\\", "\\\\");
```
lhstrh, line 130: Curious. Why is this necessary?

core/src/main/java/org/lflang/generator/CodeBuilder.java
```diff
@@ -558,7 +559,11 @@ public CodeMap writeToFile(String path) throws IOException {
     for (var line : (Iterable<String>) () -> s.lines().iterator()) {
       lineNumber++;
       if (line.contains(END_SOURCE_LINE_NUMBER_TAG) && !path.endsWith(".ino")) {
-        out.append("#line ").append(lineNumber).append(" \"").append(path).append("\"");
+        out.append("#line ")
+            .append(lineNumber)
+            .append(" \"")
+            .append(path.replace("\\", "\\\\"))
```
lhstrh, line 565: Idem

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -928,6 +928,7 @@ protected void copyTargetFiles() throws IOException {
       FileUtil.copyDirectoryContents(coreLib, dest, true);
     } else {
       FileUtil.copyFromClassPath("/lib/c/reactor-c/core", dest, true, false);
+      FileUtil.copyFromClassPath("/lib/c/reactor-c/plugin-defaults", dest, true, false);
```
lhstrh, line 931: Perhaps these should be called modules, not plugins. A module can be part of a core, but "plugin" almost always suggests some peripheral function.

core/src/main/java/org/lflang/generator/CodeBuilder.java
petervdonovan: I do not know why this is a problem now but not before. Windows tests were failing that stopped failing after I did this. The path separator needed an extra level of escaping in order to not cause syntax errors in the generated C files.

core/src/main/java/org/lflang/generator/CodeBuilder.java
petervdonovan: idem

core/src/main/java/org/lflang/generator/CodeBuilder.java
lhstrh: OK :man_shrugging:

core/src/main/java/org/lflang/generator/c/CGenerator.java
petervdonovan: I am mostly agnostic to the naming choices that we use here. Note that in this case we actually do want to support the use of tracing as a peripheral function that users can swap out.

One consideration is that at least early on, we may want some fluidity between what is a plugin vs. what is an internal API, and many things may start out as internal APIs and turn into plugins. If it were up to me, I would wait for a compelling motivation based on a specific, justifiably problematic confusion before spending too much time introducing additional distinctions or discussing terminologies.

petervdonovan: I should also add (before I reference this thread in another discussion) that if it were up to me, pretty much all of our code would consist, at least aspirationally, of "plugins". For me whether something is a plugin is less about how important or "peripheral" its functionality is and more about whether third parties can implement it and "plug it in" to everything else. I see pretty much entity (research group, grad student, company, etc.) in our little LF world as being kind of like a third party in the sense of having their own independent goals, use cases, quality standards, stability requirements, etc.

petervdonovan:
This branch now works when called like so:
```
lfc-dev --tracing --json '{"src": "test/C/src/HelloWorld.lf", "properties": { "trace-plugin": "/path/to/tracing-client-api/target/debug/libmock_lf_trace_plugin.a" }}'
```

(The trace implementation provided in this example (from the `tracing-client-api` repo) is not open-sourced currently.)

Both this and its companion PR are finally ready to merge.

core/src/test/java/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
```diff
@@ -78,6 +79,14 @@
 @InjectWith(LFInjectorProvider.class)
 public class LinguaFrancaValidationTest {
 
+  /**
+   * This is a list of tests that have no representation in LF syntax, possibly because we are
+   * moving away from keeping these in LF files and instead are moving toward keeping them in
+   * configuration files.
+   */
+  private static final List<TargetProperty<?, ?>> propertiesWithoutLfSyntaxRepresentation =
```
lhstrh, line 87: This looks like unnecessary bookkeeping to me. If the `fromAst` method throws an exception, then that exception can just be caught and dealt with, no?

lhstrh (changes requested):
This overall looks good and I'm happy that the handling of the target property is so clean. I left a minor comment that I think should be addressed before merging this.

core/src/test/java/org/lflang/tests/compiler/LinguaFrancaValidationTest.java
petervdonovan: But the test framework needs to know whether that exception indicates a test failure, or if it is the intended behavior.

petervdonovan: In the end I realized that I could not use an exception here because if this test sees the exception, then that means the validator crashes instead of working normally and producing an error message on the correct line.

However, the change that I did make was in the spirit of the suggestion. I report an error message on the correct line, and in the test I check that the error message is there.

lhstrh (approved):
LGTM! :rocket:

PR #2194, opened 2024-02-14
Parameterized default branch in check-diff script

lhstrh:
This script is used in `reactor-c` now, but the main branch in `reactor-c` is `main`, not `master`.


petervdonovan (approved):
Yep, this makes sense. I have no problem with this

PR #2196, opened 2024-02-14
Deflake test

edwardalee:
This changes a decentralized federated test so that it should no longer be flaky.  The test is still useful because it checks that data arrives in order and reliably.

### DIFF
--- a/test/C/src/federated/SmallDelayDecentralized.lf
+++ b/test/C/src/federated/SmallDelayDecentralized.lf
@@ -44,13 +44,18 @@ reactor Print {
     }
     self->c++;
   =} STP(1 sec) {=
-    lf_print_error_and_exit("STP violation at tag " PRINTF_TAG
-        ". This should not happen because the STP offset is large.",
+    lf_print_warning("STP violation at tag " PRINTF_TAG
+        ". This should not happen because the STP offset is large. Checking value anyway.",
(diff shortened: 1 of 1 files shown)


test/C/src/federated/SmallDelayDecentralized.lf
```diff
@@ -44,13 +44,18 @@ reactor Print {
     }
     self->c++;
   =} STP(1 sec) {=
-    lf_print_error_and_exit("STP violation at tag " PRINTF_TAG
-        ". This should not happen because the STP offset is large.",
+    lf_print_warning("STP violation at tag " PRINTF_TAG
```
lhstrh, line 47: This ensures that the test never fails, but the warning will never be shown when the test it executed in our testing framework. So what's the point of still having this test?

edwardalee:
The test verifies that messages that arrive arrive in order with none missing.  This is the best we can do with decentralized execution without introducing flakiness.

test/C/src/federated/SmallDelayDecentralized.lf
lhstrh: Ah, there is indeed another way the test can fail, which I hadn't noticed. Thanks for the fix!

lhstrh (approved):

PR #2197, opened 2024-02-14
Mark two federated tests as failing

petervdonovan:
EnclaveFederatedRequestStop has timed out on macOS and SmallDelayDecentralized flakes with error message "FATAL ERROR: STP violation at tag (0, 3). This should not happen because the STP offset is large"

renamed:    EnclaveFederatedRequestStop.lf -> failing/EnclaveFederatedRequestStop.lf
renamed:    SmallDelayDecentralized.lf -> failing/SmallDelayDecentralized.lf

See the discussion following https://github.com/lf-lang/reactor-c/pull/342#issuecomment-1944240135 for context.

Obviously I would be happy to close this PR in favor of a competing PR that actually fixes the issues. However, I am submitting this for now in case we prefer to take the shortest path toward not having a broken `master`.


edwardalee:
In PR #2196, which was merged today, I changed SmallDelayDecentralized to not be flaky.  The test failure you are quoting predates this change.  So that test should not be moved.

EnclaveFederatedRequestStop should also not be flaky now, so there is a legitimate bug here. I think I may have found a mechanism.  When a federate receives a `MSG_TYPE_STOP_GRANTED` message, it wakes up sleeping enclaves, and those enclaves check the event queue to see whether to send a NET.  I think I see a race condition where an enclave may not see that the stop time has been set...

The protocol for distributed stopping of federates is complicated, involving three different signal types, and enclaves further complicate it.  So it's not surprising that this will take some tuning to get right.  Another good candidate for verification...

petervdonovan:
I have updated this PR to reflect the fact that the issue in `SmallDelayDecentralized` has been fixed. This PR is once again ready for review.

edwardalee:
I ssh'd into the Mac runner and compiled and ran EnclaveFederatedRequestStop repeatedly for two hours (at which point, the runner cancelled the run because it does that if a test takes more than two hours).  The test ran 2509 times without hanging.  On my laptop, I ran the test more than 5000 times without seeing it hang.  I have also gone over the code implementing the distributed request_stop protocol and could not find any race conditions.  So I'm not convinced this test is actually flaky.  Is it possible that it was the compile that hung rather than the test running?  We do still have a problem where lfc sometimes hangs in CI and never returns, but this seems to happen on random tests, not specific ones.

petervdonovan:
OK, I'm happy to be pragmatic and keep this test running in CI unless/until this test genuinely becomes a time waster again. I agree that it's unclear what the root cause of this problem is and I also think that since we stopped running macOS tests in every PR, this test will probably cease to be an obstacle to progress in practice.

But to answer your question, in case it becomes important later:

> Is it possible that it was the compile that hung rather than the test running?

No, I don't think that's possible. In the output when the test timed out, I see the following output which shows that the federation did get started:

> Federate EnclaveFederatedRequestStop in Federation ID '4e9c6ffbcfe450133c6ed8922d70da00bdca8e3eca2823a8'
    #### Launching the federate federate__f1.
    #### Launching the federate federate__f2.
    #### Bringing the RTI back to foreground so it can receive Control-C.
    RTI -i ${FEDERATION_ID} -n 2 -c init exchanges-per-interval 10 \ 
    Federation ID for executable /Users/runner/work/reactor-c/reactor-c/test/C/fed-gen/EnclaveFederatedRequestStop/bin/federate__f2: 4e9c6ffbcfe450133c6ed8922d70da00bdca8e3eca2823a8
    ---- System clock resolution: 1000 nsec
    ---- Start execution at time Wed Feb 14 01:16:30 2024
    ---- plus 482549000 nanoseconds
    Fed 1 (f2): Connected to RTI at localhost:15045.
    Fed 1 (f2): Environment 0: ---- Intializing start tag
    Fed 1 (f2): Starting timestamp is: 1707873390584628000.
    Fed 1 (f2): Environment 0: ---- Spawning 2 workers.
    Fed 1 (f2): Environment 1: ---- Intializing start tag
    Fed 1 (f2): Environment 1: ---- Spawning 1 workers.
    Fed 1 (f2): Environment 2: ---- Intializing start tag
    Fed 1 (f2): Environment 2: ---- Spawning 1 workers.
    Fed 1 (f2): Stopped at tag (50000000, 1)
    Fed 1 (f2): Stopped at tag (50000000, 1)
    Fed 1 (f2): Connection to the RTI closed with an EOF.
    ---- Elapsed logical time (in nsec): 50,000,000
    ---- Elapsed physical time (in nsec): 66,073,000
    ---- Elapsed logical time (in nsec): 50,000,000
    ---- Elapsed physical time (in nsec): 66,074,000
    ---- Elapsed logical time (in nsec): 50,000,000
    ---- Elapsed physical time (in nsec): 66,075,000

The federates say that they reached the stop time within about 66 milliseconds in physical time, and yet the test fails to complete within 5 minutes (at which point it times out).

For all I know this could even be a bug in the blocking macOS platform-specific functions. It could even be nothing more than a remarkable quality-of-service issue on GitHub's end. Or it could indeed be a subtle problem in the shutdown protocol. The bottom line is that it sounds like we've spent enough time on this for now.

PR #2198, opened 2024-02-15
Cleanup of build logic in CGenerator

lhstrh:


edwardalee:
Can you explain what these changes do?

lhstrh:
Sure, but it's not ready for review yet, so I will do that later.

core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
```diff
@@ -116,7 +100,53 @@ public void writeDockerComposeFile(List<DockerData> services, String networkName
...
+    var fileConfig = context.getFileConfig();
+    var packageRoot = fileConfig.srcPkgPath;
+    var srcGenPath = fileConfig.getSrcGenPath();
+    var file = fileConfig.binPath.resolve(fileConfig.name).toFile();
+    var script =
+        """
+        #!/bin/sh
+        set -euo pipefail
+        cd $(dirname "$0")
+        cd ..
```
petervdonovan, line 134:
```suggestion
```
Do we need these two lines? `packageRoot` is an absolute path, right?

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -575,18 +522,47 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
...
+            // If compilation failed, remove any bin files that may have been created.
+            CUtil.deleteBinFiles(fileConfig);
+            context.unsuccessfulFinish();
+          } else {
+            context.finish(GeneratorResult.Status.COMPILED, null);
+          }
+          cleanCode.writeToFile(targetFile);
+        } catch (IOException e) {
+          messageReporter.nowhere().warning("Generated code may still contain line directives.");
+        }
```
petervdonovan, line 541: I'm not sure I'm comfortable handling all errors that occur in the above `try` block using this error message. What if the IOException comes from `runCCompiler` and not `cleanCode.writeToFile(...)`? It's possible that `runCCompiler` and `cleanCode.writeToFile` should go in separate `try` blocks so that we can handle the possible errors in a more specific/granular way.

Previously, the behavior was to show the user a stack trace, I think. The rule of thumb that I am applying here is that it is better to fail ungracefully and show the user an ugly stack trace than to provide a message that may be inaccurate and show the user nothing else.

core/src/main/java/org/lflang/generator/docker/RtiDockerGenerator.java
```diff
@@ -22,7 +22,7 @@ protected String generateDockerFileContent() {
         COPY include /reactor-c/include
         WORKDIR /reactor-c/core/federated/RTI
         %s
-        RUN mkdir build && \\
+        RUN mkdir -p build && \\
```
petervdonovan, line 25: I assume this is intended to address the problem where build directories that are not tracked by Git end up in the classpath of the JAR. I'm not sure about this solution because when we potentially use untracked files in a packaged LFC, that could cause reproducibility problems where a JAR built from a certain commit behaves differently from another JAR built from the same commit. For example, I think the CMakeCache is persisted in the `build` directory across runs of CMake, and it will affect the compilation options.

Probably the long-term solution is to use `git` to obtain copies of the runtimes that do not have any untracked/gitignored files in them (like, `git clean -X` for example). This might take a day to take care of so I won't demand that it be done now. A quick alternative for now would be to `rm -rf` the build directory instead of `mkdir -p` ing it.

core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
lhstrh: We do need these because the used path `packageRoot.relativize(srcGenPath)`. Using a relative path is better because things won't break when you move the stuff to another location (e.g., within the filesystem or between machines).

core/src/main/java/org/lflang/generator/c/CGenerator.java
lhstrh: Previously, a `sneakyThrow` was being used, which I think is worse than printing an at least somewhat helpful error message...

core/src/main/java/org/lflang/generator/docker/RtiDockerGenerator.java
lhstrh: I made this change because building locally was failing because the `build` directory already existed. I don't think this has anything to do with the class path... I'm happy to change `mkdir -p` to `rm -rf` -- I agree that that gives us behavior that is more reproducible.

core/src/main/java/org/lflang/generator/c/CGenerator.java
petervdonovan: But above I described an easy way to make the error message more helpful: use a separate `try` block for `runCCompiler`. Even if you think this is better than printing a stack trace, at least we can agree that if the error came from `runCCompiler`, then the error message should be different.

lhstrh: OK, yeah, we can do more `try` `catches`. Will push a fix.

core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
```diff
@@ -116,7 +100,53 @@ public void writeDockerComposeFile(List<DockerData> services, String networkName
...
+
+  /** Create a launcher script that invokes Docker. */
+  public void createLauncher() {
+    var fileConfig = context.getFileConfig();
+    var packageRoot = fileConfig.srcPkgPath;
+    var srcGenPath = fileConfig.getSrcGenPath();
+    var file = fileConfig.binPath.resolve(fileConfig.name).toFile();
+    var script =
+        """
+        #!/bin/sh
```
petervdonovan, line 131:
```suggestion
        #!/bin/bash
```

`sh` will fail with the error `set: Illegal option -o pipefail`

core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
petervdonovan: You're right about this and I'm OK with this as it is currently. One thing to be aware of however is that although the relative path has the advantage of allowing the directory to be moved, it has the disadvantage of not working when someone creates a symlink to the bin directory.

lhstrh: I hadn't thought about that, but that seems like it might be an odd corner case? Let's not worry about it for now...

core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
lhstrh: Nice catch!

core/src/main/java/org/lflang/generator/c/CGenerator.java
lhstrh: Check again? 8b5e98ac3da1706e92a78f939a36edc30b9d43f4

petervdonovan (approved):

Issue #2189, opened 2024-02-09
Address warnings produced by clang in generated C++ code


cmnrd:
Most LF-Cpp programs produce warnings when compiled with clang >15.0.0. See https://github.com/lf-lang/playground-lingua-franca/issues/99 for a collection of warning messages. These warnings should be easy to address by introducing `[[maybe_unused]]` attributes in the code generator.

PR #2201, opened 2024-02-16
Fixed clang warnings in generated C++ code

cmnrd:
This avoids the warnings reported in https://github.com/lf-lang/playground-lingua-franca/issues/99.

Fixes https://github.com/lf-lang/lingua-franca/issues/2189

### DIFF
--- a/core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
+++ b/core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
@@ -197,7 +197,7 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
 
     private fun Connection.getConnectionLambda(portType: String): String {
         return """
-            [this]($portType left, $portType right) {
+            [&]($portType left, $portType right) {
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
```diff
@@ -197,7 +197,7 @@ class CppAssembleMethodGenerator(private val reactor: Reactor) {
 
     private fun Connection.getConnectionLambda(portType: String): String {
         return """
-            [this]($portType left, $portType right) {
+            [&]($portType left, $portType right) {
```
lhstrh, line 200: I'm unfamiliar with this syntax. What does this change do?

lhstrh (approved):
Looks great! Thanks for the fix, @cmnrd.

core/src/main/kotlin/org/lflang/generator/cpp/CppAssembleMethodGenerator.kt
lhstrh: Nvm, it's explained here: https://en.cppreference.com/w/cpp/language/lambda

PR #2204, opened 2024-02-16
Directory definitions corrected and made to work for federated.

edwardalee:
This PR makes sure that `LF_SOURCE_GEN_DIRECTORY` and `LF_TARGET_FILES_DIRECTORY` are properly defined using CMake.  The latter in particular is needed for user code to be able to access files that are copied using the `files` target directive. This works for both unfederated and federated, even for remote federates.

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -124,7 +124,7 @@ public class FedLauncherGenerator {
       target = user + "@" + host;
     }
 
-    shCode.append("#### Host is ").append(host);
+    shCode.append("#### Host is ").append(host).append("\n");
 
     // Launch the RTI in the foreground.
     if (host.equals("localhost") || host.equals("0.0.0.0")) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -124,7 +124,7 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
       target = user + "@" + host;
     }
 
-    shCode.append("#### Host is ").append(host);
+    shCode.append("#### Host is ").append(host).append("\n");
```
lhstrh, line 127: " " followed by newline looks strange and was probably not the intent?

core/src/main/resources/lib/c/reactor-c
lhstrh, line 1: Are there updates in `reactor-c` that are required by this PR?

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
edwardalee: ??? It's `host` followed by newline.  Previously the code generator was producing:
```
#### Host is fooecho something or other
```

core/src/main/resources/lib/c/reactor-c
edwardalee: No, but reactor-c was (as usual) not aligned with main, and that's where we want it to be for the tests to mean something.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
lhstrh: You're right; I missed the first append.

lhstrh (approved):

PR #2205, opened 2024-02-17
Set CI to allow ssh

edwardalee:
This PR is not meant to be ever merged.  It's here so we can run CI on it and then ssh into a macOS runner.


lhstrh:
To trigger CI, go [here](https://github.com/lf-lang/lingua-franca/actions/workflows/all-targets.yml) and click on "Run workflow." Be sure to select `ssh-macos` as the branch.

PR #2211, opened 2024-02-19
Enabled C federated tests with Rust RTI

chanijjani:
- At first, [thirty-five C federated tests](https://github.com/hokeun/lf-rust-rti/issues/30#issuecomment-1951404245) are evaluated with Rust RTI.


chanijjani:
I want to set a label, but I can't configure it. Could anyone check for me?

lhstrh:
Terrific to see progress on this front, @chanijjani! It's encouraging that you've been able to get things to work, but the way managed to do it, by introducing a new target and replicating tests, is not quite the right approach. 

But perhaps we should not focus on that and instead review your RTI implementation? We can help with the integration at a later point...

chanijjani:
@lhstrh Thank you so much for adding labels to this PR.
I agree that the granularity of the new target needs to be revised and balanced with existing targets. The primary purposes of this PR are to see whether it passes all existing CI tests since I experienced just one failure in [my forked repository](https://github.com/chanijjani/lingua-franca/pull/1) and to get some comments to set the future directions of Rust RTI tests correctly. The merging of this PR is optional to me. We may merge this content (C federated tests with Rust RTI) if I make a new PR based on the correct way of integration later.

PR #2213, opened 2024-02-20
Add support of transient federates in C and Python targets for centralized and decentralized coordination.

ChadliaJerad:
This PR replaces https://github.com/lf-lang/lingua-franca/pull/1695. 
The companion PR in `reactor-c` is https://github.com/lf-lang/reactor-c/pull/358. 

It implements the transient feature in the federation execution. Details of the implementation are documented in Discussion https://github.com/lf-lang/lingua-franca/discussions/2212.

---
## Merged #2609
This PR builds on the centralized transient support and extends the LF federated runtime to support transient federates under decentralized coordination, where all connections are P2P and no RTI message
forwarding occurs. It also adds support for physical connections involving transient federates under centralized coordination, which are P2P connections.

Companion PR in `reactor-c`: [#574](https://github.com/lf-lang/reactor-c/pull/574)

# LF compiler (`lingua-franca`)

## Code generation (`CExtension.java`)

- `lf_connect_to_federate()` now takes an additional `is_transient` parameter.  The difference in behavior is documented in the companion PR. 
- The compiler generates per-federate  `inbound_p2p_connection_is_transient[NUMBER_OF_FEDERATES]` and   `outbound_p2p_connection_is_transient[NUMBER_OF_FEDERATES]`, so the runtime can track which outbound peers are transient.

# Tests

All transient tests are under `test/C/src/federated/transient/`. New tests:

- `DecentralizedTransientDownstreamWithTimer.lf` — transient upstream,  persistent outbound, decentralized coordination.
- `DecentralizedTransientDownstreamWithTwoUpstream.lf` — transient with two inbound federates (one persistent, one transient), authenticated.
- `DecentralizedTransientStatePersistence.lf` — verifies that state is correctly carried across re-joins.
- `DecentralizedTransientHotSwap.lf` — hot-swap of a transient federate under decentralized coordination.
- `DecentralizedTransientWithPhysicalConnection.lf`
- `TransientWithPhysicalConnection.lf`

# Possible future improvements

- Add an icon in the diagram for transient federates.


core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -534,6 +534,21 @@ public void checkInstantiation(Instantiation instantiation) {
         error("Variable-width banks are not supported.", Literals.INSTANTIATION__WIDTH_SPEC);
       }
     }
+
+    // If the Instantiation is annotated as '@transient', then:
+    // - The container has to be a federated reactor,
+    // - The coordination is centralized,
+    // - And the target is C.
+    // FIXME: Conditions 2 and 3 need to be checked
+    // FIXME: Add support of transients in decentralized coordination.
```
edwardalee, line 543: I suggest removing these FIXMEs and instead open an issue.

test/C/src/federated/transient/TransientDownstreamWithTimer.lf
```diff
@@ -0,0 +1,158 @@
...
+  timeout: 2 s
+}
+
+preamble {=
+  #include <stdlib.h>
+  #include <string.h>
+=}
+
+/** Persistent federate that is responsible for lauching the transient federate */
+reactor TransientExec(offset: time = 0, period: time = 0, fed_instance_name: char* = "instance") {
```
edwardalee, line 19: Perhaps change `offset` to `launch_time` and remove the `period` parameter, unless you think this will really be useful to launch the transient multiple times.

test/C/src/federated/transient/TransientDownstreamWithTwoUpstream.lf
```diff
@@ -0,0 +1,125 @@
...
+import Up from "TransientDownstreamWithTimer.lf"
+import Middle from "TransientDownstreamWithTimer.lf"
+
+preamble {=
+  #include <stdlib.h>
+  #include <string.h>
+=}
+
+/** Persistent federate that is responsible for lauching the transient federate */
+reactor TransientExec(offset: time = 0, period: time = 0, fed_instance_name: char* = "instance") {
```
edwardalee, line 21: Same comment about `offset` being renamed `launch_time` and `period` being removed.

test/C/src/federated/transient/TransientHotSwap.lf
```diff
@@ -0,0 +1,153 @@
+/**
+ * This LF program is a variant of TransientDownstreamWithTimer that tests the Hot Swap mechanism.
+ * For this, it tests if the transient's downstream executes as expected. In this test:
+ * - the transient federate DO NOT spontaneously leave the federation.
```
edwardalee, line 4:
```suggestion
 * - the transient federate DOES NOT spontaneously leave the federation.
```

test/C/src/federated/transient/TransientHotSwap.lf
```diff
@@ -0,0 +1,153 @@
...
+  timeout: 2 s
+}
+
+preamble {=
+  #include <stdlib.h>
+  #include <string.h>
+=}
+
+/** Persistent federate that is responsible for lauching the transient federate */
+reactor TransientExec(offset: time = 0, period: time = 0, fed_instance_name: char* = "instance") {
```
edwardalee, line 18: Same comment about `offset` and `period`.  These seem identical across files. Wouldn't it be better to import them from just one file?

edwardalee (approved):
Looks great!  The tests are quite nice. I wonder whether they will prove flaky, since there is always a risk of the launched federates taking an indeterminate amount of time to launch.  Have you seen any flaky failures?

cmnrd (dismissed):
I have significant concerns about this approach, and I don't think that we should merge this lightheartedly without being clear about the benefits and the costs. See [my comment](https://github.com/lf-lang/lingua-franca/discussions/2212#discussioncomment-8554025) in the discussion thread for an explanation.

ChadliaJerad:
> Looks great! The tests are quite nice. I wonder whether they will prove flaky, since there is always a risk of the launched federates taking an indeterminate amount of time to launch. Have you seen any flaky failures?

I agree that there are chances of having the tests prove flaky... But I have not encountered such a case so far on my machine. As a proactive workaround, I increased the timeout in this commit [3503ee1](https://github.com/lf-lang/lingua-franca/pull/2213/commits/3503ee13cc1c72cd370363b36ee5e998f59f6cca).

test/C/src/federated/transient/TransientHotSwap.lf
ChadliaJerad: When `TransientExec` is imported, the execution fails with a segmentation fault when launching `mid`.
Having a `char *` as a parameter (`fed_instance_name`) when instantiating an imported federate causes this. 
Any idea about how to fix this?

test/C/src/federated/transient/TransientHotSwap.lf
```diff
@@ -0,0 +1,153 @@
+/**
+ * This LF program is a variant of TransientDownstreamWithTimer that tests the Hot Swap mechanism.
+ * For this, it tests if the transient's downstream executes as expected. In this test:
+ * - the transient federate DO NOT spontaneously leave the federation.
+ * - the downstream of the transient federate has only one transient as upstream.
+ * - A persistent federate `HotSwapMiddle` executes `mid` after 1s to launch the hot swap.
```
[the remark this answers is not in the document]
ChadliaJerad: The difference is that `Middle` in `TransientHotSwap` does not call `lf_stop()` in the reaction to `in` as in `TransientDownstreamWithTimer`.

And indeed, the documentation in `TransientHotSwap` is fixed ([c7e1bef](https://github.com/lf-lang/lingua-franca/pull/2213/commits/c7e1befcbe7d5670af6b7e76430d0efe5660c609)). Thanks!

ChadliaJerad:
> I have significant concerns about this approach, and I don't think that we should merge this lightheartedly without being clear about the benefits and the costs. See [my comment](https://github.com/lf-lang/lingua-franca/discussions/2212#discussioncomment-8554025) in the discussion thread for an explanation.

Thanks for the feedback, Christian!
Further discussions are in the discussion.

erlingrj:
I have not reviewed the code, but tried a simple example that failed:

```
target C {
  coordination: decentralized
}


reactor R(id: int = 0)  {
  input in: int
  output out: int

  reaction(startup) -> out {=
    printf("Handling startup!\n");
    lf_set(out, self->id);
  =}

  reaction(in) {=
    printf("Got %d\n", in->value);
  =}
}

federated reactor {
  r1 = new R()

  @transient
  r2 = new R(id=2)

  @transient
  r3 = new R(id=3)

  r2.out -> r3.in
  r3.out -> r2.in
}
```

Try starting it in the order:
1. `RTI -n 3 -nt 2`
2. r1
3. r2
4. r3

As expected when r1 is started, he connects to RTI and gets a start tag and executes the startup reaction. 

But when r2 is started, it does not start, it is blocked, presumably waiting for r3 (but that's a federate). When r3 is started it crashes at once with a segfault, when run in debug mode an assertion fails because it has decoded a nonsense port. Only after r3 starts (and crashes) does r2 finally start

edwardalee:
I don't think there is support yet for decentralized coordination.

erlingrj:
Then a compilation error should be issued, a couple of other observations:
- No compiler error/warning if a program with only transients are compiled (we need at least on non-transient right?)
- If a transient federate tries to connect to the RTI before any non-transient has connected, the RTI crashes

ChadliaJerad:
@EhsanKhodadad, could you help investigate why the Patmos CI is hanging until timeout? This started happening recently, and I haven't been able to identify the root cause yet. Thanks!

core/src/main/java/org/lflang/federated/extensions/CExtension.java
```diff
@@ -794,6 +837,7 @@ else if (globalSTP instanceof CodeExprImpl)
             "for (int i = 0; i < NUMBER_OF_FEDERATES; i++) {",
             "    _fed.net_for_outbound_p2p_connections[i] = NULL;",
             "}"));
```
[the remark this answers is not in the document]
ChadliaJerad: The initialization happens later in this same function.

edwardalee (approved):
I think this PR is ready to be merged. The reactor-c companion should be merged first and the submodule updated.

PR #2215, opened 2024-02-22
Fix in CDockerGenerator.java

petervdonovan:
This is just a tiny patch that I told Marten about but kept forgetting to submit.


petervdonovan:
If the base image is overridden then `apk` probably will not work.

lhstrh (approved):
Thanks!

PR #2217, opened 2024-02-22
Report unsupported target properties as errors

cmnrd:
Currently, if the target declaration uses a property that is not supported by the target, this is reported as a warning. This is problematic, as we and potentially users don't notice if a target property is not supported anymore. For instance, we had several uses of the `flags` property in the C tests although it was renamed. Also the benchmarks used properties that we removed (See https://github.com/lf-lang/benchmarks-lingua-franca/pull/61).

This PR converts the warning message into an error, adjusts the unit tests, and fixes some of the C tests. I opted for simply dropping the `flags` property, as apparently it wasn't required and not used for a while. In particular, some tests attempted to avoid optimization, which is not needed as we compile the tests in Debug mode (without optimization) anyway.

### DIFF
--- a/core/src/main/java/org/lflang/target/TargetConfig.java
+++ b/core/src/main/java/org/lflang/target/TargetConfig.java
@@ -173,10 +173,9 @@ public class TargetConfig {
    * @param stage2 The second stage an the error reporter through which to report the warning.
    */
   public void reportUnsupportedTargetProperty(String name, MessageReporter.Stage2 stage2) {
-    stage2.warning(
+    stage2.error(
         String.format(
-            "The target property '%s' is not supported by the %s target and is thus ignored.",
-            name, this.target));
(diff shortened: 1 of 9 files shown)


cmnrd:
@erlingrj The Zephyr test fails because it uses the old `threading` property. I don't understand where the LF files that the test tries to compile originate from. Could you point me to the right place?

lhstrh:
There are changes in `reactor-c` in this PR, but I couldn't find a corresponding PR in that repository.

lhstrh (commented):
These changes look good, but why are there changes to reactor-c?

cmnrd:
Thanks for catching this. reactor-c got staged accidentally.

cmnrd:
Fixed it.

lhstrh (approved):
LGTM!

PR #2222, opened 2024-02-29
 Fixed code generation for connections involving enclave banks with multiports

julianrobledom:
This PR fixes the code generation for the corner case where an enclave is a bank and has a multiport. It also adds test cases.


cmnrd (approved):
Thanks @julianrobledom for contributing the test cases!

PR #2231, opened 2024-03-04
Fix failing Zephyr tests in CI

erlingrj:
This PR tries to address the recent failing Zephyr tests. Unfortunately it was not clear exactly what was happening and I was not able to reproduce the failure, not locally and not when ssh'ing into Github Actions and running the gradle script manually.

The Zephyr CI stuff is complicated since we must launch a QEMU emulation and match the output of the QEMU emulation to look for an FATAL ERROR or a termination. When either is detected we must kill the process running the QEMU emulation. Also we must have a timeout.

I think the error was because a QEMU process was not successfully killed. I added checks to verify that we indeed were able to kill it, and then it started working again.


lhstrh (approved):
LGTM! Thanks for these fixes, @erlingrj!

Issue #2083, opened 2023-11-05
Handling of target properties in the C code generator is problematic


lhstrh:
The C code generator creates a separate target config for each imported file and uses it specifically to handle the `cmake-include` and `files` target property, but ignores everything else. See `CGenerator.inspectReactorEResource` and `GeneratorUtils.getLFResource`.

Moreover, `FederateGenerator` implements a redundant but similarly ad-hoc handling of target properties from imported files (see #2080).

Instead, we should build a single `TargetConfig` and let the `TargetProperty` classes specify what the update behavior should be (e.g., override, append, etc.)

PR #2232, opened 2024-03-04
Bugfixes in handling of target properties across imports

### LINKED ISSUE #2149: Timeout should not be inherited
It appears that if you import a reactor from another file, you inherit the timeout target property from that file. This should not be the case.

### LINKED ISSUE #2286: Target property not imported if we extend the reported reactor
If the imported reactor is instantiated it works as expected, but if it is extended it does not.

Example:

In ImportMe.lf
```
target C {
  cmake-include: "Include.cmake"
}
reactor R {
  reaction(startup) {= =}
}
```
```
target C 
import R from "./ImportMe.lf"
reactor R2 extends R { }
main reactor {
  r = new R2()
}
```

byeonggiljun:
This PR fixes bugs in the handling of target properties across imports. It addresses #2083; fixes #2149; fixes #2286.

- [x] #2149
- [x] #2340
- [x] #2333

## Summary by CodeRabbit

- **New Features**
  - Introduced methods to handle loading behavior for properties from imported files, federates, and federations.
  - Added capability to copy user files in code generation.

- **Refactor**
  - Transitioned resource collections from `List` to `Set` for improved handling.
  - Restructured logic for merging and updating configurations based on target properties.
  
- **Bug Fixes**
  - Improved handling of conflicting connections in modal reactors.
  
- **Documentation**
  - Updated internal documentation to reflect new methods and parameter changes.

- **Tests**
  - Added new test cases for imported CMake includes and federated property loading behavior.


byeong-gil:
@lhstrh I also tried to fix https://github.com/lf-lang/lingua-franca/issues/2083 in this PR. I made `TargetProperty` specify the behavior of being imported. Now, it contains an enum of `OVERRIDE`, `APPEND`, and `IGNORE`. Currently, only `FileListProperty` has `APPEND` strategy. However, I'm not sure if this is what you wanted and we can use this to address the below fixme.
https://github.com/lf-lang/lingua-franca/blob/5997ca5af0321cee5eff7db022706e523fecb563/core/src/main/java/org/lflang/generator/c/CGenerator.java#L725-L734

I'd appreciate it if you give me a more detailed direction. What validation should be done?

lhstrh:
Thanks for taking a look at this, @byeong-gil. The issue is a bit tricky.
We essentially have several different ways of setting a target config, each of which might have a different expected outcome:
 1. initialize a configuration (i.e., override the defaults)
 2. update an already initialized configuration based on a config from an imported file (i.e., main inherits from import)
 3. initialize the configuration of a federate on the basis of a configuration of the federation (i.e., federate inherits from federation)
 4. update federate's config on the basis of an imported file that it was defined in (i.e., federate inherits from import)

In a way, each target property already specifies what its update behavior is, through its implementation of `update`. So it seems that what we need is just a way to specify whether a particular property should be regarded at all in scenarios 2, 3, and 4. If it is to be considered, we just use `update`. Do you agree? Or did I miss something?

lhstrh:
I had a look, but it doesn't seem we're actually dealing with target properties from imported files in unfederated programs (it's only addressed in `FederateTargetConfig`). That doesn't seem right to me...

lhstrh:
Oh, I see. The reason why "it works" is because of the stuff mentioned in https://github.com/lf-lang/lingua-franca/issues/2083. A better way to do this would be to create general a mechanism for getting updates from imported files.

The way I would probably go about doing this would be to create some functions in `TargetConfig` that have a default implementation that works for most target properties. Stuff like:
 - `loadFromImport`
 - `loadFromFederation`
 - `loadFromFederate`

I would guess that `loadFromImport` by default returns `false`, `loadFromFederation` and `loadFromFederate` should probably return `true`. For those properties that have a different behavior, just override those functions.

lhstrh:
In order to honor `loadFromImport`, we'd have some extra work to do, but it shouldn't be hard, and it would fix #2083.

byeong-gil:
> I had a look, but it doesn't seem we're actually dealing with target properties from imported files in unfederated programs (it's only addressed in `FederateTargetConfig`). That doesn't seem right to me...

Yes, I didn't deal with unfederated programs because I wasn't sure whether I was in the right way.

Now I understand what you want to do. Thanks, @lhstrh! To clarify, I wanna ask one more question.

For what situation `loadFromFederate` being used? `loadFromImport` is used when a reactor is imported from another lf file (scenarios 2 and 4) and `loadFromFederation` is used when each federate file is generated by `FedGenerator` (scenario 3) I guess.

lhstrh:
> For what situation `loadFromFederate` being used? 

My understanding it should be used to determine which properties to load from the federate itself (as opposed to the federation). Specifically, in the constructor of `FederateTargetconfig`, the following should be filtered by `loadFromFederation`:
```
    // Load properties from the main file
    load(federationResource, reporter);
```
while the following should be filtered by `loadFromFederate`:
```
    // Load properties from the federate file
    mergeImportedConfig(federateResource, federationResource, reporter);
```

Does this make sense?

byeong-gil:
Oh, so it's the federate version of `loadFromImport`, right?

lhstrh:
I think that a federate itself can also have imports, which would also need to be handled...

byeong-gil:
Hmm... Then I don't understand yet. Actually, the below code handles imported federate. 
```
// Load properties from the federate file
mergeImportedConfig(federateResource, federationResource, reporter);

private void mergeImportedConfig(
      Resource federateResource, Resource mainResource, MessageReporter messageReporter) {
    // If the federate is imported, then update the configuration based on target properties
    // in the imported file.
```

Regarding the below file, the timeout property from the federation is `5 sec`. And the timeout from the imported file is `10 sec`. I can't imagine how a federate itself can have a different property from the federation.

In `DistributedCount.lf`
```
target C {
  timeout: 5 sec
}

import Count from "../lib/Count.lf"

reactor Print {
  input in: int
  state c: int = 1

  reaction(in) {=
    //
  =}
}

federated reactor DistributedCount(offset: time = 200 msec) {
  c = new Count()
  p = new Print()
  c.out -> p.in after offset
}
```

In `Count.lf`
```
target C {
  timeout: 10 sec
}
...
```

lhstrh:
> Regarding the below file, the timeout property from the federation is `5 sec`. And the timeout from the imported file is `10 sec`. I can't imagine how a federate itself can have a different property from the federation.

Right, so that suggests that the `loadFromFederate` function in `TimeOutProperty` should return `false`.

lhstrh:
By the way, I would look in the `fed-gen` folder to inspect the code generated for each federate. That will be the code that is worked on, not the original LF source.

byeong-gil:
Ok, I'll analyze the code more and fix it. Thanks for your help!

byeong-gil:
@lhstrh I attempted to fix this issue again. Now, `TargetProperty` classes have methods named `loadFromImport`, `loadFromFederation`, and `loadFromFederate`. And in `CGenerator`, the generator walks through every property from the imported file and decides the behavior based on `loadFromImport`.

I have two questions.
1. I think we don't need the `loadFromFederation` method as a federate always inherits the properties from its federation. What do you think?
2. For now, the default value of `loadFromImport`' and `loadFromFederate` is `false` except for `CmakeIncludeProperty`. Because general properties should not be inherited from the imported file. Are there any other properties that should return `true` instead? I'm not familiar with every property being used... In my opinion, `FileListProperty` (superclass of `CmakeIncludeProperty`, `ProtobufsProperty`, and `FilesProperty`), `CompilerProperty`, and `CompilerFlagsProperty`, are possible candidates to be expanded from an imported file.

lhstrh:
> @lhstrh I attempted to fix this issue again. Now, `TargetProperty` classes have methods named `loadFromImport`, `loadFromFederation`, and `loadFromFederate`. And in `CGenerator`, the generator walks through every property from the imported file and decides the behavior based on `loadFromImport`.
> 
> I have two questions.

This looks great, @byeong-gil. Let me make a pass over this and push changes that will help me answer your question. I believe you got us 90% there already.

lhstrh:
This is taking a little longer than expected. One problem I see is that some things are handled in `CGenerator` that should be dealt with in `GeneratorBase`. Ideally, we'd gather all the target properties and then do things like copy files, but the problem is that the paths are stored as strings, not as fully resolved paths, and resolving files has to be done relative to the file in which the target property is found.

core/src/main/java/org/lflang/target/TargetConfig.java
```diff
@@ -257,6 +347,14 @@ public String listOfRegisteredProperties() {
         .collect(Collectors.joining(", "));
   }
 
+  /** Return the target properties that are have been assigned a value. */
```
cmnrd, line 350:
```suggestion
  /** Return the target properties that have been assigned a value. */
```

core/src/main/java/org/lflang/target/TargetConfig.java
```diff
@@ -148,6 +173,71 @@ public TargetConfig(Resource resource, GeneratorArguments args, MessageReporter
...
+      Path relativePath,
+      Predicate<TargetProperty> loadOrNot,
+      MessageReporter err) {
+    pairs.forEach(
+        pair -> {
+          var p = config.forName(pair.getName());
+          if (p.isPresent()) {
+            var value = pair.getValue();
+            var property = p.get();
+            if (property instanceof FileListProperty fileListProperty) {
```
cmnrd, line 227: Why do we need to handle FileListProperty specially? Is this to update paths that are relative to an imported resource?

core/src/main/java/org/lflang/target/TargetConfig.java
```diff
@@ -148,6 +173,71 @@ public TargetConfig(Resource resource, GeneratorArguments args, MessageReporter
     validate(reporter);
   }
 
+  /**
+   * If the federate that target configuration applies to is imported, merge target properties
```
cmnrd, line 177:
```suggestion
   * If the federate that the target configuration applies to is imported, merge target properties
```

cmnrd (approved):
This looks like a considerable improvement! I left a few small comments.

core/src/main/java/org/lflang/target/TargetConfig.java
lhstrh: Exactly.

PR #2233, opened 2024-03-06
Removed the `compiler-flags` target property

cmnrd:
This removes the `compiler-flags` property. This property is not used in any of our examples or tests at the moment. Also, when used, we show a warning that it shouldn't be used because it interferes with cmake. So I think it's best to simply remove this property.


core/src/main/java/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -110,9 +108,6 @@ public PythonGenerator(LFGeneratorContext context) {
   private PythonGenerator(
       LFGeneratorContext context, PythonTypes types, CCmakeGenerator cmakeGenerator) {
     super(context, false, types, cmakeGenerator, new PythonDelayBodyGenerator(types));
-    // Add the C target properties because they are used in the C code generator.
-    CompilerProperty.INSTANCE.override(this.targetConfig, "gcc"); // FIXME: why?
```
erlingrj, line 114: This suggests that the Python target only works with GCC? Let's see if all tests pass...

erlingrj (approved):
Looks good. Compiler-flags can easily be added with `cmake-include` so no need for extra generality!

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
cmnrd: I don't know why it was there. I figured I just try to remove it. Seems to work fine on Mac where clang is the default.

lhstrh: :laughing: at the `FIXME` (which is probably mine)

lhstrh (approved):
Good idea!

PR #2234, opened 2024-03-07
Improved support for Docker

lhstrh:
At times, it is necessary to run a command (e.g., to set environment variables) prior to building or when hitting the entry point. We add target properties to point to scripts for this purpose. We also add an option to skip the build step, in case integration into a more sophisticated build process is needed.

Related to #2232 in the sense that we need to check the handling across imported files etc.

- [x] Add `pre-build-script`, `post-build-script`, and `pre-run-script` entries to `docker` target property dictionary
- [x] Add `builder-base` and `runner-base` entries to `docker` target property dictionary to replace `FROM`
- [x] Ensure that scripts are copied into the docker image
- [x] Add `no-build` option (`false` by default) to skip the docker build step
- [x] Fix paths referenced in compile definitions
- [x] Perform builds for TypeScript target in docker container (not on host)
- [x] Fix https://github.com/lf-lang/lingua-franca/issues/2278
- [x] https://github.com/lf-lang/lingua-franca/pull/2322
- [x] Let `runner-base` default to `builder-base` if specified
- [x] Ensure tracing plugins work
- [x] Ensure that `pre-build-script` is composed with app using `&&`
- [x] https://github.com/lf-lang/lf-lang.github.io/pull/267
- [x] Add tests


cmnrd:
Does this PR address https://github.com/lf-lang/lingua-franca/issues/2286 or https://github.com/lf-lang/lingua-franca/issues/2278? If not, should it include fixes?

petervdonovan:
Marking sub-task "Ensure tracing plugins work" as completed by https://github.com/xronos-inc/tracing-example/pull/2.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
```diff
@@ -24,31 +32,174 @@ public DockerGenerator(LFGeneratorContext context) {
...
+  /** Return the default base image. */
+  public abstract String defaultImage();
+
+  /** Return the base image to be used during the building stage. */
+  protected String builderBase() {
+    return baseImage(context.getTargetConfig().get(DockerProperty.INSTANCE).builderBase());
+  }
+
+  /** Return the base image to be used during the running stage. */
+  protected String runnerBase() {
```
petervdonovan, line 193: Would it be helpful for this to default to builderBase? I assume that cross-compiling is a sufficiently uncommon use case that having the two be the same by default would be the least surprising behavior for users.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
```diff
@@ -24,31 +32,174 @@ public DockerGenerator(LFGeneratorContext context) {
...
+                .flatMap(java.util.Collection::stream)
+                .collect(Collectors.toList()),
+            " \\\n\t&& ");
+  }
+
+  /** Return the ENTRYPOINT command. */
+  protected String generateEntryPoint() {
+    return "ENTRYPOINT ["
+        + getEntryPointCommands().stream()
+            .map(cmd -> "\"" + cmd + "\"")
```
petervdonovan, line 129: How robust we choose to make this depends on how long-term this code will be used. If this is going to be used long-term, then maybe we should escape nested quotes in a way that is compatible with the syntax of POSIX-compliant shells. Doing this properly could be a bit of work.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
```diff
@@ -24,31 +32,174 @@ public DockerGenerator(LFGeneratorContext context) {
...
+            .collect(Collectors.joining(","))
+        + "]";
+  }
+
+  /** Return a COPY command to copy the executable from the builder to the runner. */
+  protected String generateCopyOfExecutable() {
+    var lfModuleName = context.getFileConfig().name;
+    return "COPY --from=builder /lingua-franca/%s/bin/%s ./bin/%s"
```
petervdonovan, line 137:
```suggestion
    return "COPY --from=builder /lingua-franca/%s/bin/%s ./bin/%s"  // safe because context.getFileConfig().name never contains spaces
```

Alternatively, we could just wrap the argument in quotes.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
```diff
@@ -24,31 +32,174 @@ public DockerGenerator(LFGeneratorContext context) {
...
-  public String userRunCommand() {
-    var runCmd = context.getTargetConfig().get(DockerProperty.INSTANCE).run();
-    if (runCmd != null && !runCmd.isEmpty()) {
-      return "RUN " + runCmd;
+  /** Return a COPY command to copy the scripts from the builder to the runner. */
+  protected String generateCopyOfScript() {
+    var script = context.getTargetConfig().get(DockerProperty.INSTANCE).preRunScript();
+    if (!script.isEmpty()) {
+      return "COPY --from=builder /lingua-franca/%s/src-gen/%s ./scripts/%s"
+          .formatted(context.getFileConfig().name, script, script);
```
petervdonovan, line 146: The script name actually could contain spaces, so we probably should account for that. The easiest way to handle that would be to crash with a runtime error and a helpful error message if the script name contains spaces.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
```diff
@@ -24,31 +32,174 @@ public DockerGenerator(LFGeneratorContext context) {
...
+    }
+    return "# (No pre-run script provided.)";
+  }
+
+  /**
+   * Return a list of strings used to construct and entrypoint. If this is done for a federate, then
+   * also include additional parameters to pass in the federate ID.
```
petervdonovan, line 153:
```suggestion
   * also include additional parameters to pass in the federation ID.
```

At first I was surprised that the federation ID is hardcoded to 1, but now I see that it actually makes sense because use of distinct Docker networks serves the same purpose as using distinct federation IDs. I'm happy with this implementation choice.

core/src/main/java/org/lflang/generator/GeneratorBase.java
```diff
@@ -627,6 +629,24 @@ protected void cleanIfNeeded(LFGeneratorContext context) {
...
+  /** Create Dockerfiles and docker-compose.yml, build, and create a launcher. */
+  protected boolean buildUsingDocker() {
+    // Create docker file.
+    var dockerCompose = new DockerComposeGenerator(context);
+    var dockerData = getDockerGenerator(context).generateDockerData();
+    try {
+      dockerData.writeDockerFile();
+      dockerCompose.writeDockerComposeFile(List.of(dockerData));
+    } catch (IOException e) {
+      context.getErrorReporter().nowhere().error("Error while writing Docker files");
```
petervdonovan, line 644: Something like this might improve debuggability:
```suggestion
      context.getErrorReporter().nowhere().error("Error while writing Docker files: " + (e.getMessage() == null ? "No cause given" : e.getMessage()));
```

test/C/src/docker/federated/DockerOptions.lf
```diff
@@ -0,0 +1,23 @@
...
+  logging: DEBUG,
+  timeout: 1 s,
+  coordination: centralized,
+  docker: {
+    rti-image: "rti:local",
+    pre-build-script: "foo.sh",
+    pre-run-script: "bar.sh",
+    post-build-script: "baz.sh",
+    no-build: false
+  }
```
petervdonovan, line 11: AFAICT this does not actually function as a test because it will not cause the test suite to fail if the provided scripts are not executed. I am OK with this if this solution is not necessarily long-term.

petervdonovan (approved):
All of my comments and suggestions are optional to address depending on priorities. This PR looks good to merge.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
lhstrh: That's a good point. But I imagine there is a library that we can use to do this?

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
lhstrh: Yes, I thought about this too. Should be an easy fix, so I'll do that.

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
lhstrh: I found one here: https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html#escapeXSI-java.lang.String-

core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
```diff
@@ -30,79 +32,47 @@ public String defaultImage() {
...
+                  Path.of("/lingua-franca", context.getFileConfig().name),
+                  false),
+              context.getErrorReporter(),
+              context.getTargetConfig().target == Target.CCPP);
+      return List.of(
+          "mkdir -p bin",
+          String.format(
+              "%s -DCMAKE_INSTALL_BINDIR=./bin -S src-gen -B bin", ccompile.compileCmakeCommand()),
+          "cd bin",
+          "make all",
```
petervdonovan, line 68: The usual name for this directory is `build`, not `bin`, and people usually keep their `build` and `bin` directories separate so that the `bin` directory only contains executables. Is there a particular reason why it was done in this way?

lhstrh: Not that I know of. It says `BINDIR` though. Doesn't that suggest `bin`?

petervdonovan: Sorry, I meant where it says `-B bin` and `mkdir -p bin`. The `BINDIR` probably should be `bin`, but what I thought was odd was that it was the same as build build directory that was created with `mkdir -p` and passed to `cmake` using the `-B` flag.

lhstrh: IDK, maybe @cmnrd could weigh in on this?

test/C/src/docker/federated/DockerOptions.lf
lhstrh: An error would result if the scripts weren't accessible in the container, but it would be better if these scripts were to do something like set an environment variable that would be checked. Could you take a stab at turning this into a more meaningful test, @petervdonovan?

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
lhstrh: Addressed this in https://github.com/lf-lang/lingua-franca/pull/2234/commits/151d540279ffe8d2980936c780dec9f223c8f0f8, please check.

lhstrh:
@a-sr: The Epoch build fails because I introduced a dependency. Could you please push a fix to `epoch`? This is the dependency: https://commons.apache.org/proper/commons-text/dependency-info.html

core/src/main/java/org/lflang/generator/docker/DockerGenerator.java
petervdonovan: I think this change is good, except that I removed the unnecessary uses of escaping for the `ENTRYPOINT` command, as we discussed on Zulip. They were causing the `DockerOptions` test to fail.

test/C/src/docker/federated/DockerOptions.lf
petervdonovan: Done

lhstrh:
> @a-sr: The Epoch build fails because I introduced a dependency. Could you please push a fix to `epoch`? This is the dependency: https://commons.apache.org/proper/commons-text/dependency-info.html

I made a start and tried https://github.com/lf-lang/epoch/pull/74 but it doesn't seem to work

gradle.properties
```diff
@@ -22,6 +22,7 @@ freehepVersion=2.4
 swtVersion=3.124.0
 spotbugsToolVersion=4.7.3
 jcipVersion=1.0
+commonsTextVersion=1.12.0
```
lhstrh, line 25:
```suggestion
commonsTextVersion=1.11.0
```
Because Epoch can't access `1.12.0` for some reason. See https://github.com/lf-lang/epoch/pull/75.

lhstrh:
> > @a-sr: The Epoch build fails because I introduced a dependency. Could you please push a fix to `epoch`? This is the dependency: https://commons.apache.org/proper/commons-text/dependency-info.html
> 
> I made a start and tried [lf-lang/epoch#74](https://github.com/lf-lang/epoch/pull/74) but it doesn't seem to work

Thanks for coming to the rescue, @a-sr!

core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
cmnrd: Don't know either. Choosing `bin` for the build directory name is indeed strange. In particular, since we also install in `bin` and thus the separation between build and bin directory seems to be gone.

test/C/src/docker/federated/DockerOptions.lf
lhstrh: [279b00e](https://github.com/lf-lang/lingua-franca/pull/2234/commits/279b00ea468dd214212aaf86d66fdee728820f74)

core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
lhstrh: :man_shrugging: I don't see anything actionable here unless there is a concrete alternative proposed.

cmnrd: The alternative is using:
```
"%s -DCMAKE_INSTALL_BINDIR=./bin -S src-gen -B build", ...
cd build,
```
which seems more sane.

lhstrh: @petervdonovan would you want to address this? If so, please push a fix and hit resolve. If no, please just resolve this conversation to unblock the merge.

Issue #2202, opened 2024-02-16
Remove the deprecated list syntax


cmnrd:
We decided to deprecate the interpretation of a comma-separated list of expressions enclosed in `(` and `)` as a list, in favor of using the corresponding expressions and syntax rules in the target language. E.g., we would use `{0, 1, 2, 3}` in C/C++ or `[0, 1, 2, 3]` in Python. I think it is time to drop this old rule, also because it may cause confusion (see https://github.com/lf-lang/lingua-franca/issues/2056).

In particular, we would remove this syntax for `AssignmentInitializer` from the grammar:
```
    | '='? parens?='(' (exprs+=Expression (','  exprs+=Expression)* (trailingComma?=',')?)? ')'
```

lhstrh:
Yes, this seems like the right thing to do to me.

Issue #2056, opened 2023-10-14
Spurious warning about deprecated syntax


edwardalee:
The Python test [ModalCycleBreaker.lf](https://github.com/lf-lang/lingua-franca/blob/7af2e84953ddfe9cfeea606ddbaf7762f0faf5c1/test/Python/src/modal_models/ModalCycleBreaker.lf#L58) contains the following syntax in the main reactor:

```
  test = new TraceTesting(events_size = 1, trace = (  // keep-format
    0,1,0,
    100000000,1,1,
    100000000,1,2,
    100000000,1,3,
    100000000,1,4,
    200000000,1,6,
    100000000,1,7,
    100000000,1,8,
    100000000,1,9
  ), training = False)
```

This generates the following warning:

```
lfc: warning: This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
 --> src/modal_models/ModalCycleBreaker.lf:58:52
   |
57 |   modal = new Modal()
58 |   test = new TraceTesting(events_size = 1, trace = (  // keep-format
   |                                                    ^ This syntax is deprecated in the Python target, use an equal sign instead of parentheses for assignment.
   |
59 |     0,1,0,

lfc: info: Code generation finished.
```

However, the syntax is correct and up-to-date.

cmnrd:
> However, the syntax is correct and up-to-date.

The warning is actually correct. We used to interpret `(...)` as a list. And if you check the generated code, it converts the `(...)`expression to a list using Python's `[...]` syntax. We decided to deprecate this syntax as it is largely misleading. In Python, `(...)` denotes a tuple, not a list.

I think we should actually follow up and remove this old syntax rule for good.

PR #2235, opened 2024-03-07
Remove list types and list initialization syntax

cmnrd:
This PR implements another major step towards completing https://github.com/lf-lang/lingua-franca/issues/1492. In addition to the built-in type `time`, LF would handle fixed-sized and dynamically sized lists/arrays specially. This required additional logic in our "type system" as well as dedicated syntax for initializing such list types. For instance, it was possible to write the following (in the C target):
```
state foo: int[3](1, 2, 3)
state bar time[](1 ms, 2 ms, 3ms)
```
We also implemented this syntax for other targets, but it never worked all that well and it clashed with principles of the other target languages. Therefore, we decided in https://github.com/lf-lang/lingua-franca/issues/1492 to drop this style of initialization. In addition, https://github.com/lf-lang/lingua-franca/pull/1580 deprecated the `(...)` initializer syntax in favor of target language expressions. So in C, we can write now:
```
state foo: int[3] = {1, 2, 3}
```

This PR follows up and removes the deprecated syntax and performs additional cleanup. Concretely, it makes the following changes:
- Simplify/update the grammar rules.
- Disable parenthesis-based initialization for all targets except C++
- Drop LF-native support for lists and arrays. We fully rely on the target type system now.
- Due to the previous point, we cannot support time lists anymore. Thus, the rule for the time array type `time[]` is removed. However, this can be easily replaced with the target time type, e.g., in C: `state foo: interval_t[] = { 1ms, 2ms, 3ms}`.
- Rename the array spec syntax rule to clearly describe it as C-style. This is to allow later addition of other array types, like for Rust which uses `[i32, 4]`.
- Drop support for the c-style array notation from the C++ and the Rust target. We simply use the native types in these languages instead (e.g., std::vector<>, std::array<> for C++).
- Update test cases that still used the old initializer syntax.

Fixes https://github.com/lf-lang/lingua-franca/issues/1675
Fixes https://github.com/lf-lang/lingua-franca/issues/2056
Fixes https://github.com/lf-lang/lingua-franca/issues/2202


lhstrh (approved):
This looks great! It sure has simplified a few things.

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -321,6 +312,8 @@ Expression:
     | {CodeExpr} code=Code
     | BracedListExpression
     | BracketListExpression
+    // FIXME: Should disable this rule until a later release
```
cmnrd, line 315: Reminder to self: convert this into an issue and remove the FIXME

PR #2236, opened 2024-03-08
Add -f flag to rm between zephyr test runs

erlingrj:
Ran into a race condition with deletion of a temporary file in between zephyr test runs. Should be solved by adding -f to the rm command


lhstrh (approved):
:+1:

PR #2240, opened 2024-03-11
Python tests to use an rti that matches the reactor-c checkout

Jakio815:
The python test's yml file is in wrong order.

The RTI is installed before updating the reactor-c to the latest version.

I changed the order of the Install RTI below checkout to the newest reactor-c version.

Found this bug during debugging lf-lang/reactor-c#392

### DIFF
--- a/.github/workflows/py-tests.yml
+++ b/.github/workflows/py-tests.yml
@@ -33,16 +33,6 @@ jobs:
           fetch-depth: 0
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
-      - name: Check out specific ref of reactor-c
-        uses: actions/checkout@v3
-        with:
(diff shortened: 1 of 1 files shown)


lhstrh:
FYI: Just changed the label so that this gets indexed properly in our release notes.

Jakio815:
> The fixes to `py-tests.yml` look right to me. Thanks for the fix, @Jakio815! What about the `reactor-c` submodule update? Did you mean to check that in?

@lhstrh I think I'm pointing `main`, am I wrong? I have nothing to change with the `reactor-c` repo, just the `.yml` file.

Jakio815:
Oh I just saw that the `master` repo is not pointing `main`. I'll just change it back.

lhstrh (approved):
Perfect. Thanks!

PR #2245, opened 2024-03-13
Added newline after trailing slash

edwardalee:
This fixes a bug in the command to remotely launch an RTI by inserting a newline after the trailing slash.

There is still a problem, however, which is that `ssh` with a command does not source any bash profile files, so the PATH never gets set, so the RTI will likely not be found on the remote machine.  The "right" fix, IMHO, is that we should code generate the RTI just like a federate and launch it just like a federate.  But short of that, it would be better if FedLauncherGenerator created a bash script in a local file, then copied it to the remote machine and launched it there.

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -401,7 +401,7 @@ public class FedLauncherGenerator {
         "        echo \"The source code can be found in org.lflang/src/lib/core/federated/RTI\"",
         "        exit 1",
         "    fi",
-        "    " + rtiLaunchString + " 2>&1 | tee -a " + logFileName + "' &",
+        "    " + rtiLaunchString + "\n 2>&1 | tee -a " + logFileName + "' &",
         "# Store the PID of the channel to RTI",
         "RTI=$!",
         "# Wait for the RTI to boot up before",


vinzbarbuto (commented):
That could fix the problem

vinzbarbuto (approved):

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -401,7 +401,7 @@ private String getRemoteLaunchCode(
         "        echo \"The source code can be found in org.lflang/src/lib/core/federated/RTI\"",
         "        exit 1",
         "    fi",
-        "    " + rtiLaunchString + " 2>&1 | tee -a " + logFileName + "' &",
+        "    " + rtiLaunchString + "\n 2>&1 | tee -a " + logFileName + "' &",
```
lhstrh, line 404: This looks a little weird to me. Why is it reasonable to assume that the argument `rtiLaunchString` would have a trailing slash? Seems more like a problem that needs to be solved at the call site(s) of `getRemoteLaunchCode`. That said, I don't think this code is very critical, so let's just merge and move on...

vinzbarbuto (changes requested):
The bug is actually present on line 394. I suggest we replace `"\n\"` with `"\n"\` as the sequence `\"` seems to be causing the error.

lhstrh:
> The bug is actually present on line 394. I suggest we replace `"\n\"` with `"\n"\` as the sequence `\"` seems to be causing the error.

You're welcome to push a fix to this branch...

vinzbarbuto:
Sure, Marten. I'll attend to it promptly.

Il giorno ven 15 mar 2024 alle ore 18:51 Marten Lohstroh <
***@***.***> ha scritto:

vinzbarbuto:
Edward and Marten, prior to making any modifications, I conducted some tests, and it seems the code generator is now functioning correctly. It appears that the issue has been resolved in one of the previous commits. Therefore, I believe we can proceed to close the pull request.

lhstrh:
I'm confused. Which version were you using when you reported the problem?

vinzbarbuto:
I'm quite confused as well. Last week, it wasn't working, so I had to manually delete the `\` separator. However, after the building phase, the issue I was facing seems to have vanished. My hunch is that I didn't update all the git submodules while dealing with the lfc-dev compiler

lhstrh:
I never really understood the issue this was trying to solve, and the fix didn't look good to me. Should we just close this?

Issue #2171, opened 2024-01-27
Federated compilation with the clean flag does not work correctly


erlingrj:
If I do a clean compile of a federated program ( `lfc -c`) the generated binary from the first federate will be cleaned when this command invokes `lfc` on the second federate and so on.  So the generated bash script will not be successful since it will not find all binaries it needs in `fed-gen/bin`

lhstrh:
This is a known bug and we'll fix it soon. I think there may be a duplicate issue about this already.

PR #2247, opened 2024-03-19
Fixed clean building federated programs

cmnrd:
Fixes https://github.com/lf-lang/lingua-franca/issues/2171


lhstrh (approved):
Looks like a clean solution to me! No pun intended..

PR #2249, opened 2024-03-24
Fix in error handling for Docker builds

petervdonovan:
We were always returning false for whether errors occurred in federated docker builds, but we should be checking the context to see whether an error was reported.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedGenerator.java
@@ -222,7 +222,10 @@ public class FedGenerator {
         context.getErrorReporter().nowhere().error("Docker build failed.");
       }
     } catch (IOException e) {
-      context.getErrorReporter().nowhere().error("Docker build failed due to invalid file system state.");
+      context
+          .getErrorReporter()
+          .nowhere()
+          .error("Docker build failed due to invalid file system state.");
     }
   }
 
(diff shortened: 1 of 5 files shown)


petervdonovan:
By the way, there is also a problem in `DOCKER_RUN_SCRIPT` in `TestBase`. It completed with exit code 0 even though the build failed with a nonzero exit code. I do not want to debug this because I do not want to have to correctly do async fork-join-style programming in `bash`. Furthermore, the basic idea behind how the script does error handling looks brittle to me because it assumes that the test passes by default, and only fails if a special string is read on stdout. It is preferable to assume that tests fail by default and only mark them as passing if they print a special string, because occasional false positives are preferable to occasional false negatives.

Technically things are OK now because the test framework does not mark failing tests as passing anymore, but maybe I should open an issue about this.

lhstrh (approved):
These changes look great! Thanks for these much-needed fixes, @petervdonovan. Is there a PR in `reactor-c` that corresponds to the changes included in https://github.com/lf-lang/reactor-c/compare/7427d987d725111141139819c5341921edd0fef1...2a3c2575be15f62b1229f8d8967b03d7fb16887a?

lhstrh:
It looks like there are still unaddressed failures also: https://github.com/lf-lang/lingua-franca/actions/runs/8406281187/job/23019999735?pr=2249

PR #2251, opened 2024-04-03
Docker compose also create bin directory if not present

erlingrj:
This solves an issue where the docker compose generator fails to create the launcher because the bin directory was cleaned and is not present. I might only have surfaced now due to Peter making sure we checked the return value from the docker tests. (I am not sure because I did not look at those changes though)

### DIFF
--- a/core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
+++ b/core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
@@ -144,7 +144,9 @@ public class DockerComposeGenerator {
       writer.write(script);
       writer.close();
     } catch (IOException e) {
-      messageReporter.nowhere().warning("Unable to write launcher to: " + file.getAbsolutePath()+ " with error: " + e);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/lflang/generator/docker/DockerComposeGenerator.java
```diff
@@ -142,7 +144,9 @@ public void createLauncher() {
       writer.write(script);
       writer.close();
     } catch (IOException e) {
-      messageReporter.nowhere().warning("Unable to write launcher to: " + file.getAbsolutePath());
+      messageReporter
+          .nowhere()
+          .warning("Unable to write launcher to: " + file.getAbsolutePath() + " with error: " + e);
```
lhstrh, line 149:
```suggestion
          .warning("Unable to write launcher to " + file.getAbsolutePath() + " with error: " + e);
```

lhstrh (approved):
Good catch. Thanks, @erlingrj!

PR #2253, opened 2024-04-04
Replacement of `javax.inject` with `com.google.inject`

soerendomroes:
Javax will soon be incompatible and is also being removed from KLighd and ELK.
Hence, I suggest to also remove it here (and I need a commit hash to refer to when updating the target platform for Epoch).


lhstrh (approved):
Thanks for being proactive with these fixes! Have you confirmed that these are all of them?

lhstrh (changes requested):
The code formatting errors are easily addressed by running `./gradlew spotlessApply`.
I'm puzzled by the Docker errors, but it seems like `docker-compose` isn't found. Maybe we should use `docker compose` instead.

soerendomroes:
I have an Eclipse setup that successfully installs and works with a checked out version of Klighd and ELK. But since the ELK release on Maven has problems, we have not yet released Klighd to test this with a release, only with the main.

So it works for me, but I will come back to this before/after the Klighd 3.0.0 release.

lhstrh (approved):

soerendomroes:
@lhstrh Thanks for merging and making the PR spotless. I seem to have missed this during travel preparations.

lhstrh:
No problem! Easy fix 😄

PR #2255, opened 2024-04-08
Fix rust setup action in CI

cmnrd:
The action ATiltedTree/setup-rust disappeared overnight. This PR switches to another action.


cmnrd:
Merging this without review to fix CI asap.

PR #2260, opened 2024-04-15
IntelliJ run config to start the LS

soerendomroes:
I think it might be nice to have a run configuration to start the language server from Intellij since many people seem to be using that as a development environment.

The idea is to start the LS on port 7670 and to have a run configuration for the vscode-lingua-franca repository that also connects to this port.

If the messages itself should be debugged, I occasionally use wireshark (tcp port 7670) to monitor the traffic (but this should normally not be necessary).


edwardalee (approved):
This looks good to me.  Can we add a section to this page:

https://www.lf-lang.org/docs/developer/developer-intellij-setup

on how to actually use this?

cmnrd (approved):
I like the idea and second the request for usage instructions.

PR #2262, opened 2024-04-21
Support for FlexPRET platform

magnmaeh:
This PR adds support for the FlexPRET platform. In addition, it adds an `Option<T>` on the PlatformOptions type so we know whether a specific key/value pair was set by the user or just had its default value.

Depends on https://github.com/lf-lang/reactor-c/pull/412


magnmaeh:
@lhstrh / @erlingrj I'm having an issue with the baudrate target property. Specifically this line, which sets the default baudrate for a platform if the user does not specify any: https://github.com/lf-lang/lingua-franca/blob/f840daae1414bc32a267fc09fea111300fe01fce/core/src/main/java/org/lflang/target/property/PlatformProperty.java#L43. The reason for this is: "9600 is a standard amongst systems like Arduino, so it's the default value."

When there is a default baudrate, I'm not sure if I'll be able to determine whether the user tried to set the property or if it is just a default value. (If there is a way to do this, please let me know; I haven't programmed in Java until now, so I'm struggling to understand everything going on.) What I want to do for FlexPRET is not allow the user to set it, because this is something the FlexPRET SDK controls entirely. 

Maybe it would be possible to provide different default values depending on what platform is used? Or just set the default value of baudrate to 0 and have Arduino set it to 9600 somewhere else.

lhstrh:
> @lhstrh / @erlingrj I'm having an issue with the baudrate target property. Specifically this line, which sets the default baudrate for a platform if the user does not specify any:
> 
> https://github.com/lf-lang/lingua-franca/blob/f840daae1414bc32a267fc09fea111300fe01fce/core/src/main/java/org/lflang/target/property/PlatformProperty.java#L43
> . The reason for this is: "9600 is a standard amongst systems like Arduino, so it's the default value."
> 
> When there is a default baudrate, I'm not sure if I'll be able to determine whether the user tried to set the property or if it is just a default value. (If there is a way to do this, please let me know; I haven't programmed in Java until now, so I'm struggling to understand everything going on.) What I want to do for FlexPRET is not allow the user to set it, because this is something the FlexPRET SDK controls entirely.
> 
> Maybe it would be possible to provide different default values depending on what platform is used? Or just set the default value of baudrate to 0 and have Arduino set it to 9600 somewhere else.

The way it's currently set up, you can detect whether a property as a whole has been set using `isSet` on `TargetConfig`, but that doesn't allow you to check which individual dictionary items (such as `baudrate`) were set. We could make the default platform specific, but would that solve your problem?

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -74,11 +78,51 @@ protected PlatformOptions fromString(String string, MessageReporter reporter) {
...
+        if (platform.port() != null) {
+          reporter
+            .at(config.lookup(this), Literals.KEY_VALUE_PAIR__VALUE)
+            .warning("Port property ignored for emulator");
+        }
+        if (platform.flash() == true) {
+          reporter
+            .at(config.lookup(this), Literals.KEY_VALUE_PAIR__VALUE)
+            .warning("Flash property ignored for emulator");
+        } // TODO: Add check on baudrate not set
```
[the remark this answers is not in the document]
lhstrh: Yes, I see the problem. Perhaps we could make `flash` a record with an `isSet` and `value` field. You would then be able to write:
```
if (platform.flash().isSet() && platform.flash().value() == true)
```
WDYT?

magnmaeh:
> The way it's currently set up, you can detect whether a property as a whole has been set using `isSet` on `TargetConfig`, but that doesn't allow you to check which individual dictionary items (such as `baudrate`) were set. We could make the default platform specific, but would that solve your problem?

I guess it would solve my problem if the default is 9600 for Arduino and 0 for all other platforms, because 0 would in that case mean "not set". I guess the more robust solution would be having the `isSet` method for all dictionary items, though. If I end up implementing it for the `flash` property it hopefully won't be hard to extend to the other fields.

core/src/main/java/org/lflang/target/property/PlatformProperty.java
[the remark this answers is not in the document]
magnmaeh: I went with this suggestion, which worked quite well. I only implemented it for the `flash` and `baudRate` variables though, which probably looks a little inconsistent when looking at the other variables (but the strings have `null` which means the same here).

core/src/main/java/org/lflang/util/FlexPRETUtil.java
```diff
@@ -0,0 +1,50 @@
+package org.lflang.util;
```
magnmaeh, line 1: I now have this file, which I created heavily inspired by the ArduinoUtil.java file.

My question is, do you know any good candidates for where I should call my `flashTarget` function? I must admit I struggle to get a grasp of what happens where, hehe

core/src/main/java/org/lflang/target/property/PlatformProperty.java
```diff
@@ -111,6 +180,34 @@ public String name() {
     return "platform";
   }
 
+  /** 
+   * The purpose of this record is to keep track of whether the flash option
+   * was set by the user or is a default value. This is useful for producing
+   * warnings or errors.
+   */
+  public record FlashOption(
```
lhstrh, line 188: How about something like:
```
record Option<T>(boolean setByUser, T value) {}
```

core/src/main/java/org/lflang/util/FlexPRETUtil.java
lhstrh: Good question. It would have to happen sometime after code generation. Probably somewhere in this branch:
https://github.com/lf-lang/lingua-franca/blob/f840daae1414bc32a267fc09fea111300fe01fce/core/src/main/java/org/lflang/generator/c/CGenerator.java#L515

What happens here should be factored out in a bunch of smaller functions, but that's a different issue...

core/src/main/java/org/lflang/target/property/PlatformProperty.java
magnmaeh: Good idea, I went with that! :)

magnmaeh:
Currently I'm struggling a bit with code size issues. When I compile a test program, e.g., `TimeoutZeroThreaded`, I get code sizes that already are too big to fit inside FlexPRET realized on an FPGA (since FlexPRET uses scratchpad memories (SPM), for ROM and RAM it has very little of it).

```
Memory region         Used Size  Region Size  %age Used
             ROM:      159108 B       256 KB     60.69%
             RAM:       31968 B       256 KB     12.19%
```

(This is with FlexPRET using 256 kB for instruction SPM and data SPM, which is unrealistic but I'm using it to be able to compile the programs. A more realistic choice is 64 kB for both.)

I have checked what takes up code size with `nm <program> --size-sort`, and see that e.g., `_vfprintf_r` takes up 10 kB. I have tried to track down what code pulls this function in, because it does not appear in the FlexPRET standalone applications - only the LF applications. But I have not been able to figure this out. 

FlexPRET has its own standalone `printf` implementation, so this is just an artifact from `newlib` probably. So a natural solution would be to remove these newlib artifacts. That is kind of scary though, because it may have unexpected consequences down the road. The test yields the same outputs before and after removing it, though.

I have found two possible methods to remove these functions.
1. Add `-nostdlib` to compile code and manually link in object files from `newlib` that we actually want. This seems possible, but there might be a lot of object files to link in. And if a user in the future wants to use an additional `newlib` feature, they would have to link in that object file manually too. (https://stackoverflow.com/questions/38609303/how-to-add-prebuilt-object-files-to-executable-in-cmake) 
2. Use the `linker`'s `--wrap` option to remove certain functions. Essentially, you can write `--wrap=_vfprintf_r`, which will replace calls to `_vfprintf_r` with `__wrap__vfprintf_r` and rename `_vfprintf_r` to `__real__vfprintf_r`. Then we can implement `__wrap__vfprintf_r` as just nothing, which will not link in `_vfprintf_r`. (https://sourceware.org/binutils/docs-2.40/ld.html#index-_002d_002dwrap_003dsymbol)

The best option, of course, would be to figure out where in the LF source code the function gets pulled in from and see if we can filter this out for embedded systems. I found [this method](https://stackoverflow.com/a/31091776/16358883), but I ended up just tracking within newlib.

I tried to do 2. on the same program, and got promising results:

```
Memory region         Used Size  Region Size  %age Used
             ROM:      113628 B       256 KB     43.35%
             RAM:       30144 B       256 KB     11.50%
```

I'm not quite sure why this reduces code size by more than 10 kB (the size of the function), but probably other code becomes redundant and does not get linked in either, I guess.

I think both these options are kind of bad because they might cause headaches down the road. But they might be necessary. Do you have any thoughts @erlingrj?

erlingrj:
I dont fully understand. Do we have our own, fully-featured printf implementation for FlexPRET? Then we should make sure that this is what is being used for the functions in `util.c` in reactor-c. But probably, it does not have all those features needed there. Another option would be to look in `logging.h` and `util.c` in reactor-c, and provide our own implementations of them. In fact, it would be great if that was a general possibility. I am sure other embedded targets would benefit from that also

magnmaeh:
I think the problem is that `newlib` is so intertwined that when you call one of its functions, it will include references to a lot more than just that function. An example is `tzset`, which references `_tzset_unlocked_r`, which references `_malloc_r` and `siscanf`. Malloc is already quite big. `siscanf` references `__ssvfiscanf_r`, which is the second biggest function I have found (6.7 kB). `__ssvfiscanf_r` might not ever be called, but it needs to be included in the binary because it might be. So if we "know" we never use it, we can just replace it with a stub to save a lot of code space.

I assume `_vfprintf_r` comes from a similar chain of references. And there are probably many `newlib` functions that ultimately reference `_vfprintf_r` too. I have not found a standardized method to find the "top-level" functions (like `tzset`). But if we can find such a method we can maybe try to sanitize reactor-c from these `newlib` functions.

erlingrj:
The tzset probably comes from the call to `ctime` in the very beginning of `main`. I think if this was behind a macro, and likewise all the print-functions could optionally be implemented for the target platform then the code size problem would be more managable

magnmaeh:
Ah ok, that's where `tzset` came from!

But for this PR it would maybe be a bit much to sanitize all of reactor-c from these function calls? Because I assume it's a big task to find all functions like `ctime` and replace them with a lightweight equivalent that still works. Is it ok to instead do the `--wrap` method I mentioned for now and create an issue on reactor-c on this very topic?

erlingrj:
Yes, this would only affect the FlexPRET target. It would be great if you made a Github issue for this problem and linked to it in a comment where you apply the wrap workaround.

magnmaeh:
Is this a fair description of the issue? https://github.com/lf-lang/reactor-c/issues/418#issue-2276625443

magnmaeh:
I have a few questions: (@erlingrj / @lhstrh)
1. I replaced all fields of `PlatformOptions` with the `Option<T>` @lhstrh suggested, except for the `Platform` field. This is to know whether the value has been set by the user or not (as discussed here: https://github.com/lf-lang/lingua-franca/pull/2262#issuecomment-2073632803). However, many fields already had an implicit way of checking whether it had been set by the user. E.g, since `board` is a String it was checked against `null`. I have replaced everywhere I found `if (board == null)` with `if (!board.setByUser())` and similarly for the other fields, but I am not 100% sure I have caught everything. I'm not sure how thorough CI is; maybe we can trust it to catch it if I did not fix everything?
2. I added a few build tests by just copying the RP2040 and Zephyr tests. I do not fully understand the API - so if the tests I've created do not make sense or there should be more of them, please let me know ([core/src/integrationTest/java/org/lflang/tests/runtime/CFlexPRETTest.java](https://github.com/lf-lang/lingua-franca/pull/2262/files#diff-1f87fee5b2da3031d0ae4f826f406a7c4a7788075fc78d040a6a7a7cc1336ff6)). But is there a way to execute the tests for (emulated) embedded systems? To me it seems we can only test building for embedded platforms?
3. I'm having an issue with getting continuous standard output from an `LFCommand`. My command just runs a bash script, which under the hood runs a FlexPRET's emulator with the compiled LF program. The LF program normally prints out something every second, but when running it with an LFCommand, the output is only printed out once the entire program is finished. I've tried setting the `verbose` option on, but that does not seem to solve it. Do you have any suggestions? The relevant file (https://github.com/lf-lang/lingua-franca/pull/2262/files#diff-e74d4e3e1fc8d88fcdd4cc009c4932227bc85b0efd0e1e9bef9b3c03e87d0e16)

Once I've solved these issues I think what's left is only:
1. Verifying everything works on FPGA
2. CI for tests

erlingrj:
> I have a few questions: (@erlingrj / @lhstrh)
> 
>     1. I replaced all fields of `PlatformOptions` with the `Option<T>` @lhstrh suggested, except for the `Platform` field. This is to know whether the value has been set by the user or not (as discussed here: [Add support for FlexPRET #2262 (comment)](https://github.com/lf-lang/lingua-franca/pull/2262#issuecomment-2073632803)). However, many fields already had an implicit way of checking whether it had been set by the user. E.g, since `board` is a String it was checked against `null`. I have replaced everywhere I found `if (board == null)` with `if (!board.setByUser())` and similarly for the other fields, but I am not 100% sure I have caught everything. I'm not sure how thorough CI is; maybe we can trust it to catch it if I did not fix everything?
> 

I would say that if CI passes and you have made an effort to search it down, then we should merge it and fix any errors if they arise later on.

>     2. I added a few build tests by just copying the RP2040 and Zephyr tests. I do not fully understand the API - so if the tests I've created do not make sense or there should be more of them, please let me know ([core/src/integrationTest/java/org/lflang/tests/runtime/CFlexPRETTest.java](https://github.com/lf-lang/lingua-franca/pull/2262/files#diff-1f87fee5b2da3031d0ae4f826f406a7c4a7788075fc78d040a6a7a7cc1336ff6)). But is there a way to execute the tests for (emulated) embedded systems? To me it seems we can only test building for embedded platforms?
> 

Since we have an emulator of FlexPRET we could run CI on it also. However, I don't think those tests belong here in the lingua-franca repo. I think we should just build a couple of HelloWorlds, maybe maybe run a few of them on the emulator. But it is not trivial, see next comment.

>     3. I'm having an issue with getting continuous standard output from an `LFCommand`. My command just runs a bash script, which under the hood runs a FlexPRET's emulator with the compiled LF program. The LF program normally prints out something every second, but when running it with an LFCommand, the output is only printed out once the entire program is finished. I've tried setting the `verbose` option on, but that does not seem to solve it. Do you have any suggestions? The relevant file (https://github.com/lf-lang/lingua-franca/pull/2262/files#diff-e74d4e3e1fc8d88fcdd4cc009c4932227bc85b0efd0e1e9bef9b3c03e87d0e16)
> 

I don't think you can continuously monitor the output of commands, you only get it once the command terminates. I did a workaround for this in a bash script for running Zephyr tests on a QEMU emulator, since the QEMU emulator does not terminate. You will find it under .github/scripts I believe. It is a bash script that monitors the output and terminates if certain patterns are found. I don't think it is worth to do this for FlexPRET. The reason is that it is a little brittle, and FlexPRET has a big stack of dependencies and if some test for some reason fails, it is more likely to be a problem in the CPU design or the SDK than in LFC or reactor-c. 

> 
> Once I've solved these issues I think what's left is only:
> 
>     1. Verifying everything works on FPGA
> 
>     2. CI for tests

Great!

magnmaeh:
Ok, I've verified it still works on FPGA and added some code for CI. (Though the check does not appear here)

I think this PR is ready for review too :)

lhstrh:
@magnmaeh, you'd have to resolve the conflict on `reactor-c` in order to get the tests to run.

magnmaeh:
> @magnmaeh, you'd have to resolve the conflict on `reactor-c` in order to get the tests to run.

Ah, then I might have to wait for https://github.com/lf-lang/reactor-c/pull/412 to get merged first. Because I assume lf-lang repo does not see my branch in my forked reactor-c repo and that's why it complains.

erlingrj:
To solve the merge conflict, then check out your  PR branch locally and merge `lf-lang:master` into it. This will yield a merge for the reactor-c submodule. To resolve it, make sure that it is checked out at your reactor-c branch also and add the submodule

lhstrh:
I'm a little confused as to why not all tests are running...

lhstrh:
> I'm a little confused as to why not all tests are running...

Found the problem:
``` 
Invalid workflow file: .github/workflows/all-targets.yml#L1
too many workflows are referenced, total: 21, limit: 20
```

lhstrh:
It looks like `only-c.yml` has too many entries. Maybe we have a separate top-level `c-embedded.yml` where we list all the embedded platforms.

magnmaeh:
> It looks like `only-c.yml` has too many entries. Maybe we have a separate top-level `c-embedded.yml` where we list all the embedded platforms.

I'm not that familiar with CI, so I may have misunderstood something. I tried making a `c-embedded.yml` file, but since that is just included it doesn't reduce the number of reusable workflows called. Did you mean we should make an `all-embedded.yml` file like `all-misc.yml` and `all-targets.yml`?

magnmaeh:
Ok, I got the FlexPRET tests to run, but code coverage report fails. I kind of just copied that part from Arduino and Zephyr. Any immediate ideas as to what could be wrong here?

```
      - name: Run FlexPRET smoke tests
        run: |
          cd "$FP_DIR" && source env.bash && cd -
          export RISCV_TOOL_PATH_PREFIX=$FP_RISCV_COMPILER
          ./gradlew core:integrationTest \
            --tests org.lflang.tests.runtime.CFlexPRETTest.* \
            core:integrationTestCodeCoverageReport
          rm -rf test/C/src-gen
      - name: Report to CodeCov
        with:
          files: core/build/reports/jcoco/integrationTestCodeCoverageReport/integrationTestCodeCoverageReport.xml
        if: ${{ github.repository == 'lf-lang/lingua-franca' }}
```

.github/workflows/c-flexpret-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+          cd "$FP_DIR" && source env.bash && cd -
+          export RISCV_TOOL_PATH_PREFIX=$FP_RISCV_COMPILER
+          ./gradlew core:integrationTest \
+            --tests org.lflang.tests.runtime.CFlexPRETTest.* \
+            core:integrationTestCodeCoverageReport
+          rm -rf test/C/src-gen
+      - name: Report to CodeCov
+        uses: ./.github/actions/report-code-coverage
+        with:
+          files: core/build/reports/jcoco/integrationTestCodeCoverageReport/integrationTestCodeCoverageReport.xml
```
lhstrh, line 57:
```suggestion
          files: core/build/reports/jacoco/integrationTestCodeCoverageReport/integrationTestCodeCoverageReport.xml
```

.github/workflows/all-embedded.yml
```diff
@@ -0,0 +1,33 @@
+# Tests for embedded platforms
+name: CI (by embedded platforms)
```
lhstrh, line 2:
```suggestion
name: CI (embedded platforms)
```

lhstrh:
> Ok, I got the FlexPRET tests to run, but code coverage report fails. I kind of just copied that part from Arduino and Zephyr. Any immediate ideas as to what could be wrong here?
> 
> ```
>       - name: Run FlexPRET smoke tests
>         run: |
>           cd "$FP_DIR" && source env.bash && cd -
>           export RISCV_TOOL_PATH_PREFIX=$FP_RISCV_COMPILER
>           ./gradlew core:integrationTest \
>             --tests org.lflang.tests.runtime.CFlexPRETTest.* \
>             core:integrationTestCodeCoverageReport
>           rm -rf test/C/src-gen
>       - name: Report to CodeCov
>         uses: ./.github/actions/report-code-coverage
>         with:
>           files: core/build/reports/jcoco/integrationTestCodeCoverageReport/integrationTestCodeCoverageReport.xml
>         if: ${{ github.repository == 'lf-lang/lingua-franca' }}
> ```
> 
> (I'm sorry for asking so much about this stuff. It's just that if it's a simple fix then it might save me a few hours of diving into this.)

No worries. Looks like just a typo in the `files` path. Pushed a fix -- let's see what happens.

lhstrh:
> > It looks like `only-c.yml` has too many entries. Maybe we have a separate top-level `c-embedded.yml` where we list all the embedded platforms.
> 
> I'm not that familiar with CI, so I may have misunderstood something. I tried making a `c-embedded.yml` file, but since that is just included in `only-c.yml` it doesn't reduce the number of reusable workflows called. Did you mean we should make an `all-embedded.yml` file like `all-misc.yml` and `all-targets.yml`?

Looks like you figured it out :rocket:

lhstrh:
This looks good now, with the only caveat that there are some tests that are "expected to pass" as part of the merge requirements but have now been moved. We can bypass the branch protection once we're ready to merge and then adjust the GitHub configuration.

magnmaeh:
Thank you for fixing my typo! :)

https://github.com/lf-lang/lf-lang.github.io/pull/254 is now ready for review as well :) I think the only things left are:
1. Transferring ownership of https://github.com/magnmaeh/lf-flexpret-template to lf-lang
2. Fixing link in docs to point to the lf-lang lf-flexpret-template (I assume it's going to be  https://github.com/lf-langlf-flexpret-template, though)
3. Remove lingua-franca submodule in lf-flexpret-template (one this PR is merged)

.github/workflows/all-embedded.yml
```diff
@@ -0,0 +1,33 @@
+# Tests for embedded platforms
+name: CI (embedded platforms)
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 8 * * 6'
```
erlingrj, line 7: What is this? Are you setting up a cronjob to regularly run this CI job?

.github/actions/setup-flexpret/action.yml
```diff
@@ -0,0 +1,48 @@
...
+
+        # This rest is copied directly from FlexPRET's `azure-pipelines.yml`
+
+        # Ubuntu 20.04 only has Verilator 4.028 but we neeed a more modern version
+        # so we do not use 'sudo apt-get install -y -qq verilator' here.
+        wget -q https://github.com/sifive/verilator/releases/download/4.036-0sifive2/verilator_4.036-0sifive2_amd64.deb -O verilator.deb
+        sudo dpkg -i verilator.deb
+
+        # Install riscv compiler
+        wget -q https://github.com/xpack-dev-tools/riscv-none-embed-gcc-xpack/releases/download/v8.3.0-2.3/xpack-riscv-none-embed-gcc-8.3.0-2.3-linux-x64.tar.gz -O riscv.tgz
```
erlingrj, line 19: We can instead directly fetch the correct compiler, see this workflow: https://github.com/lf-lang/interpret/blob/main/.github/workflows/ci.yml

.github/workflows/c-embedded.yml
```diff
@@ -0,0 +1,28 @@
+name: C embedded
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+env:
+  # 2020.11
+  vcpkgGitRef: 0bf3923f9fab4001c00f0f429682a0853b5749e0
```
erlingrj, line 9: What does this mean?

core/src/main/java/org/lflang/util/FlexPRETUtil.java
```diff
@@ -0,0 +1,81 @@
...
+      LFCommand flash = commandFactory.createCommand(cmd, args);
+
+      /**
+       * Check if environment variable FP_SDK_FPGA_INTERFACE_PROGRAM is set. If it is set, we should
+       * let the user know it is not supported by LF.
+       *
+       * <p>We cannot support it because FP_SDK_FPGA_INTERFACE_PROGRAM will lead to a [Y/n] prompt,
+       * and the LFCommands do not have a way of getting the prompt response.
+       *
+       * <p>When FP_SDK_FPGA_INTERFACE_PROGRAM is not set, there is no prompt.
```
erlingrj, line 45: This was a little unclear to me. If the user has set an environmental variable FP_SDK_FPGA_INTERFACE_PROGRAM then we cannot flash the executable. So the default behavior is that the variable is not set and we can flash? Under what circumstances will that variable be set? Manually be the user somehow?

core/src/main/java/org/lflang/util/FlexPRETUtil.java
```diff
@@ -0,0 +1,81 @@
...
+              && platform.flash().value() == true;
+
+      if (interfaceProgramRelevant && envInterfaceProgram != null) {
+        messageReporter
+            .nowhere()
+            .info("FP_SDK_FPGA_INTERFACE_PROGRAM not supported by LF. It will have no effect.");
+      }
+
+      if (flash != null) {
+        /** Set the variable to nothing for the flash command, which will avoid the [Y/n] prompt */
```
erlingrj, line 66: THis was also not clear. Setting FP_SDK_FPGA_INTERFACE_PROGRAM to an empty string avoids a [Y/n] prompt? From where?

erlingrj (changes requested):
I think this looks very good. Just a few minor suggestions

.github/workflows/all-embedded.yml
magnmaeh: I just copied this from the `all-misc.yml` and `all-targets.yml`, which are the two other top-level CI jobs. I can remove it if you'd like.

.github/workflows/c-embedded.yml
magnmaeh: Like the other stuff: Copied from `all-misc.yml` and `all-targets.yml`.

core/src/main/java/org/lflang/util/FlexPRETUtil.java
magnmaeh: Hmm no, we can always flash the executable but we cannot support the [Y/n] prompt from this part of the `fpga-app.sh`: (because the LF command does not support stdin)

```
# Check if `xclip` exists
if ! command -v xclip > /dev/null 2 >& 1
then
    # Let the user know they can install `xclip` to automatically get command
    # copied to clipboard
    echo "Install \`xclip\` to automatically copy command to clipboard."
else
    # We ask to copy command; do not want to overwrite clipboard if user
    # has something important there
    echo "Copy command to clipboard? [Y/n]"
    read answer

    # Default response is yes
    if [ ! "$answer" != "${answer#[Nn]}" ] ; then 
        # Copy command to clipboard
        echo "$interface_cmd" | xclip -selection clipboard
        echo "Command copied to clipboard. Use CTRL + V to paste."
    fi
fi
```

The prompt only appears if the user has set the FP_SDK_FPGA_INTERFACE_PROGRAM environment variable. So in that case, we let the user know it is not supported.

core/src/main/java/org/lflang/util/FlexPRETUtil.java
magnmaeh: That is from this part of `fpga-app.sh.in`:

```
# Exits if not set
if [[ -z "${FP_SDK_FPGA_INTERFACE_PROGRAM}" ]]; then
    exit 0
fi
# Source the user selected program to interface with the FPGA, e.g., `picocom.source`
# This sets the `interface_cmd` variable
source "$interface_cmd_sourcefile"

# Print out info on how to interface with software
echo "To interface with the FPGA, run:"
echo "$interface_cmd"

# Check if `xclip` exists
if ! command -v xclip > /dev/null 2 >& 1
then
    # Let the user know they can install `xclip` to automatically get command
    # copied to clipboard
    echo "Install \`xclip\` to automatically copy command to clipboard."
else
    # We ask to copy command; do not want to overwrite clipboard if user
    # has something important there
    echo "Copy command to clipboard? [Y/n]"
    read answer

    # Default response is yes
    if [ ! "$answer" != "${answer#[Nn]}" ] ; then 
        # Copy command to clipboard
        echo "$interface_cmd" | xclip -selection clipboard
        echo "Command copied to clipboard. Use CTRL + V to paste."
    fi
fi
```

magnmaeh: But yes, I can see how it is not at all obvious from the Java code...

.github/actions/setup-flexpret/action.yml
magnmaeh: Fixed :) Thanks for the tip!

core/src/main/java/org/lflang/util/FlexPRETUtil.java
erlingrj: Why not just launch the FP_SDK_FPGA_INTERFACE_PROGRAM from the shell script instead of copying the launch command into the clipboard?

core/src/main/java/org/lflang/util/FlexPRETUtil.java
erlingrj: So if FP_SDK_FPGA_INTERFACE_PROGRAM is empty or not set, it returns early and then what? Is it flashed with some default program? If so, couldn't we just require the user to install that program to avoid this logic?

core/src/main/java/org/lflang/util/FlexPRETUtil.java
magnmaeh: Good question, hehe. I guess I considered doing this to be a little more versatile - with this approach the user can run commands between this script and actually running the "interface_cmd". But I realize now that use case is a bit narrow and probably not very useful. Maybe it's better to just run it automatically.

core/src/main/java/org/lflang/util/FlexPRETUtil.java
magnmaeh: Maybe it's better to just add the entire file here. As you can see, flashing is done independently of whether `FP_SDK_FPGA_INTERFACE_PROGRAM` is set or not (with `send_uart.py`). The `FP_SDK_FPGA_INTERFACE_PROGRAM` refers to what program the user wants to capture standard output with - e.g., `picocom`, `putty`, or some other serial listener program. 

```
# !/bin/bash
#
# This file is auto generated by CMake and should not be modified. 
#
# This script will attempt to upload your compiled software to an FPGA, 
# assuming correct hardware setup and that it runs the bootloader found in 
# `$FP_SDK_PATH/bootloader`.
#
# You got this script because the CMake variable `TARGET` was set to
# `fpga`. Either specify `emulator` or nothing to get a script that runs
# the emulator instead.

set -e

if [[ -z "${FP_PATH}" ]]; then
    echo "Environment variable FP_PATH not set!"
    exit 1
fi

if [[ -z "${FP_SDK_PATH}" ]]; then
    echo "Environment variable FP_SDK_PATH not set!"
    exit 1
fi

flashdev=@FP_FLASH_DEVICE@
baudrate=@UART_BAUDRATE@
imem_path=@IMEM_LOCATION@/@IMEM_NAME@
imem_serialized_path=/tmp/@IMEM_NAME@.serialized
interface_cmd_sourcefile="${FP_SDK_PATH}/cmake/utils/fpga-interface/@FP_INTERFACE_PROGRAM@.source"

echo "Serializing app and placing its result in /tmp"
python3 ${FP_PATH}/scripts/serialize_app.py ${imem_path} ${imem_serialized_path}

echo "Flashing serialized app on device ${flashdev} with baudrate ${baudrate}"

# Script takes usb_path, baudrate, and file
python3 ${FP_PATH}/scripts/send_uart.py ${flashdev} ${baudrate} ${imem_serialized_path}

if [[ -z "${FP_SDK_FPGA_INTERFACE_PROGRAM}" ]]; then
    exit 0
fi
# Source the user selected program to interface with the FPGA, e.g., `picocom.source`
# This sets the `interface_cmd` variable
source "$interface_cmd_sourcefile"

# Print out info on how to interface with software
echo "To interface with the FPGA, run:"
echo "$interface_cmd"

# Check if `xclip` exists
if ! command -v xclip > /dev/null 2 >& 1
then
    # Let the user know they can install `xclip` to automatically get command
    # copied to clipboard
    echo "Install \`xclip\` to automatically copy command to clipboard."
else
    # We ask to copy command; do not want to overwrite clipboard if user
    # has something important there
    echo "Copy command to clipboard? [Y/n]"
    read answer

    # Default response is yes
    if [ ! "$answer" != "${answer#[Nn]}" ] ; then 
        # Copy command to clipboard
        echo "$interface_cmd" | xclip -selection clipboard
        echo "Command copied to clipboard. Use CTRL + V to paste."
    fi
fi

```

magnmaeh:
@lhstrh I tried to transfer ownership of https://github.com/magnmaeh/lf-flexpret-template but I need access to create public repos on lf-lang to do that.

Screenshot from 2024-05-20 13-15-33

lhstrh:
@magnmaeh: sent you an invite to join the org!

.github/workflows/all-embedded.yml
lhstrh: We use CRON to do weekly (at 08:00 on Saturday), to run tests on macOS runners. We don't need it for the embedded platforms, so we should remove this.

.github/workflows/all-embedded.yml
```diff
@@ -0,0 +1,33 @@
+# Tests for embedded platforms
+name: CI (embedded platforms)
+
+on:
+  schedule:
```
lhstrh, line 5:
```suggestion
```

.github/workflows/all-embedded.yml
```diff
@@ -0,0 +1,33 @@
...
+    # * is a special character in YAML so you have to quote this string
```
lhstrh, line 6:
```suggestion
```

.github/workflows/all-embedded.yml
lhstrh, line 7:
```suggestion
```

.github/workflows/c-embedded.yml
lhstrh: Used by Windows runners only.

.github/workflows/c-embedded.yml
```diff
@@ -0,0 +1,28 @@
+name: C embedded
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+env:
```
lhstrh, line 7:
```suggestion
```

.github/workflows/c-embedded.yml
```diff
@@ -0,0 +1,28 @@
...
+  # 2020.11
```
lhstrh, line 8:
```suggestion
```

.github/workflows/c-embedded.yml
lhstrh, line 9:
```suggestion
```

core/src/main/java/org/lflang/util/FlexPRETUtil.java
```diff
@@ -0,0 +1,81 @@
+package org.lflang.util;
+
+import java.io.IOException;
+import java.util.Collections;
+import org.eclipse.xtext.xbase.lib.Exceptions;
```
lhstrh, line 5: Do not use

lhstrh (commented):
LGMT, modulo some nitpicks.

.github/workflows/all-embedded.yml
erlingrj: Aha, sorry, I just hadnt seen this before in the CI. Makes sense

erlingrj: But, should be removed from embedded targets yes

core/src/main/java/org/lflang/util/FlexPRETUtil.java
```diff
@@ -0,0 +1,81 @@
...
+        flash.setEnvironmentVariable("FP_SDK_FPGA_INTERFACE_PROGRAM", "");
+
+        int ret = flash.run(context.getCancelIndicator());
+        if (ret != 0) {
+          messageReporter.nowhere().error("Command " + cmd + " failed with exit code: " + ret);
+          throw new IOException("Could not flash");
+        }
+      }
+    } catch (IOException e) {
+      Exceptions.sneakyThrow(e);
```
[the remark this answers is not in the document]
magnmaeh: The exceptions weren't really necessary either way, so I just removed them and used the messageReporter instead :)

core/src/main/java/org/lflang/util/FlexPRETUtil.java
erlingrj: I think this would be wise. What program (minicom/picocom/putty) that is used to write over the application is a detail that hardly matters for the user

magnmaeh: Fixed in https://github.com/pretis/flexpret/pull/96! :)

erlingrj:
The failure in TS is tracked here #2293, I think we should merge even with this failure and this must be solved independently. 

The last issue is that  zephyr and arduino tests appear as required but are not run:
image

I don't really see why, those workflows are not references anywhere except in the c-embedded.yml (which runs) and in the reactor-c CI

lhstrh:
> The failure in TS is tracked here #2293, I think we should merge even with this failure and this must be solved independently.
> 
> The last issue is that zephyr and arduino tests appear as required but are not run: image
> 
> I don't really see why, those workflows are not references anywhere except in the c-embedded.yml (which runs) and in the reactor-c CI

It's configured in the GitHub settings. This needs to be changed because the workflows have moved around. I'll take care of this.

erlingrj (approved):

PR #2263, opened 2024-04-21
Klighd upgraded to v3.0.1

soerendomroes:


soerendomroes:
This fails since the way of initializing and using Klighd changed (beginning with the 3.0.X release the functionality previously provided by `klighd.standalone` is provided in a different plugin). Maybe, @a-sr can point to the corresponding commit in kieler/semantics or @NiklasRentzCAU has prepared a guide to adapt the release.

NiklasRentzCAU:
Sure, I can provide you some links and insight on what needs to be changed here.
Generally, the [release notes of KLighD 3.0.1](https://github.com/kieler/KLighD/releases/tag/release-2024-04-3.0.1) state what needs to be changed according to the breaking changes in language server applications: 
1. Replace any `de.cau.cs.kieler.klighd.standalone` with `de.cau.cs.kieler.klighd.setup` (might not even occur in the files, as we changed that dependency downstream in klighd.lsp) 
2. Replace any occurence of `de.cau.cs.kieler.klighd.standalone.KlighdStandaloneSetup` in the source code to `de.cau.cs.kieler.klighd.setup.KlighdStandaloneSetup`.

3. (recommended, but not strictly neccessary) remove the klighd.incremental dependency from your lsp/build.gradle: the animations now work even without this, this dependency is now only required in Eclipse-based applications, so not your language server.

If you do any manual injection stuff with KLighD, you also might need to update parts of that from the old javax.inject to jakarta.inject or com.google.inject with Google Guice >=7.0.0 

The commit in semantics for the first two steps is [here.](https://github.com/kieler/semantics/commit/8c78256ccc505adb33b1ae6085094c40e71fc94e)

Neccessary changes that I see are [here](https://github.com/soerendomroes/lingua-franca/blob/7b74832a4e827c8ce9ec0f04b2b3a880409e4657/cli/lfd/src/main/java/org/lflang/cli/Lfd.java#L5)(change to .setup), and [here](https://github.com/soerendomroes/lingua-franca/blob/7b74832a4e827c8ce9ec0f04b2b3a880409e4657/lsp/build.gradle#L14-L19) (remove)

lhstrh:
> Epoch still seems to have problems with that. This should be the case since it uses an old version of ELK (and Klighd) in its target platform, which does not support the MODEL_ORDER option for separate connected components packing.
> 
> @lhstrh I did some things to get Epoch working with the new release but I do not know all dependencies by heart so https://github.com/lf-lang/epoch/pull/65 should take a little longer to fix.

This is the part where @a-sr usually saves the day 😆

lsp/build.gradle
```diff
@@ -10,13 +10,6 @@ dependencies {
         exclude group: 'org.eclipse.platform', module: 'org.eclipse.swt.*'
         exclude group: 'org.eclipse.platform', module: 'org.eclipse.swt'
     }
-
-    // This dependency ensures correct animations and bookkeeping during updates
-    // See https://github.com/lf-lang/vscode-lingua-franca/issues/103#issuecomment-1731023470
-    implementation ("de.cau.cs.kieler.klighd:de.cau.cs.kieler.klighd.incremental:$klighdVersion") {
```
lhstrh, line 16: @soerendomroes: are you sure his is safe to remove?

lhstrh (approved):

lsp/build.gradle
lhstrh: Since you commented on https://github.com/lf-lang/vscode-lingua-franca/issues/103#issuecomment-1731023470 in the past, I assume that you concluded it was safe to remove. Let me resolve this for now so that we can merge.

NiklasRentzCAU: It was safe to remove, see my last comment's number 3.

PR #2264, opened 2024-04-21
Xtext bumped to 3.34.0

lhstrh:
For release notes, see: https://eclipse.dev/Xtext/releasenotes.html#/releasenotes/2024/02/27/version-2-34-0.


soerendomroes:
@lhstrh Maybe this fails since a few tests still use javax and not com.google.gson for injection. I guess beginning with 2.33 Xtext might also drop support for javax.infject, hence this might fail.

lhstrh:
OK, let's see: 729d42bdc

soerendomroes:
spotless seems to be missing but it seems to solve the problem.

PR #2267, opened 2024-04-25
Set annotated layout options for modes.

soerendomroes:
@gbazydlo here is your solution for this.


soerendomroes:
@a-sr Maybe you can comment whether this is the way to do it for the Modal Model synthesis.

lhstrh (commented):
This looks fine to me, but what does it do?

a-sr (approved):
This is fine. It applies layout annotation defined at the reactor definition to the invisible parent node of the modes.

PR #2270, opened 2024-04-27
Fix NPE bug for instance in mode

### LINKED ISSUE #2269: Importing a reactor with instance in a mode triggers a NPE
File `lib/InstanceInMode.lf` instantiates a reactor inside a mode:
```
target C
reactor U {
  reaction(startup) {==}
}
reactor B {
  reaction(startup) {==}
  initial mode A {
    u = new U()   
  }
}
```
File `ImportInstanceInMode.lf` imports the outer reactor:
```
target C {
  timeout: 0s
}
import B from "lib/InstanceInMode.lf"
main reactor {
  b = new B()
}
```
Compiling this program triggers a null-pointer exception.

edwardalee:
This PR fixes #2269.


core/src/main/java/org/lflang/graph/InstantiationGraph.java
```diff
@@ -31,10 +31,7 @@
 import java.util.Set;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.lflang.ast.ASTUtils;
-import org.lflang.lf.Instantiation;
-import org.lflang.lf.Model;
-import org.lflang.lf.Reactor;
-import org.lflang.lf.ReactorDecl;
+import org.lflang.lf.*;
```
lhstrh, line 34: The [style guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) prohibits wildcard imports, so I think this particular change should be reverted.

lhstrh (changes requested):
Besides the wildcard imports, looks good to me!

core/src/main/java/org/lflang/graph/InstantiationGraph.java
lhstrh:
```suggestion
import org.lflang.lf.Instantiation;
import org.lflang.lf.Model;
import org.lflang.lf.Reactor;
import org.lflang.lf.ReactorDecl;
```

lhstrh (approved):

PR #2272, opened 2024-05-01
Fix for validation of target properties dictionaries

### LINKED ISSUE #2266: No validation on PlatformOption keys
Currently, the following Platform dictionary compiles with no errors:

```
    platform: {
        name: "FlexPRET",
        board: "emulator",
        port: "/dev/ttyUSB0",
        flash: false,
        baud-rate: 38400,
        hello: 77
    }
```

This is unfortunate, because `hello` is not an accepted key and should report an error. It also causes confusion if the user writes `baudrate` instead of `baud-rate`. Since lfc does not report anything, the user naturally could think `baudrate: 38400` was accepted, when in fact it had the same effect as writing `hello: 77`.

magnmaeh:
Fixes #2266.

The bug was just that `return` was inside the for-loop, so only the first key/value pair would be validated.

### DIFF
--- a/core/src/main/java/org/lflang/target/property/type/DictionaryType.java
+++ b/core/src/main/java/org/lflang/target/property/type/DictionaryType.java
@@ -76,8 +76,8 @@ public enum DictionaryType implements TargetPropertyType {
                       + keysList());
           valid = false;
         }
-        return valid;
       }
+      return valid;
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
Nice catch, @magnmaeh!

PR #2275, opened 2024-05-07
Extensions to the trace-plugin property

erlingrj:
This PR addresses a couple of issues:
- Updating the trace API + several other fixes by Peder
- code-generating trace plugin option to CMake
- Allowing to set a RUN cmd to the generated Docker
- Parsing the tracing and trace-plugin options correctly
- 

There is also a PR in reactor-c with corresponding fixes here: https://github.com/lf-lang/reactor-c/pull/420


lhstrh:
Looks like https://github.com/lf-lang/reactor-c/pull/420 should be merged first and then pulled into this PR with a submodule update.

test/C/src/ActionDelay.lf
```diff
@@ -1,5 +1,7 @@
 // Test logical action with delay.
-target C
+target C {
+  platform: Zephyr
```
petervdonovan, line 3: Why is this change included in this PR? This comment also applies to the other changes to the C tests. It looks like the changes sneaked in with the "[Have TracePluginProperty simple inherit from StringProperty](https://github.com/lf-lang/lingua-franca/pull/2275/commits/aaf29395e5b4409b5a0892209fa0deb4a77c698d)" commit.

core/src/main/java/org/lflang/target/property/DockerProperty.java
```diff
@@ -45,20 +46,26 @@ public DockerOptions fromAst(Element node, MessageReporter reporter) {
...
       for (KeyValuePair entry : node.getKeyvalue().getPairs()) {
         DockerOption option = (DockerOption) DictionaryType.DOCKER_DICT.forName(entry.getName());
-        if (option != null && option.equals(DockerOption.FROM)) {
-          from = ASTUtils.elementToSingleString(entry.getValue());
-        }
-        if (option != null && option.equals(DockerOption.RTI_IMAGE)) {
-          rti = ASTUtils.elementToSingleString(entry.getValue());
+        switch (option) {
+          case FROM -> from = ASTUtils.elementToSingleString(entry.getValue());
+          case RUN -> run = ASTUtils.elementToSingleString(entry.getValue());
```
petervdonovan, line 51: I'm OK with adding this feature, but I'd like to draw attention to it as something to keep in mind if and when we later decide to consider a different approach to Dockerfile generation. We have already discussed that (e.g. having a user write a template for the docker file and supplying values to plug into their template), so this point should not be a blocker for merging this PR.

petervdonovan (approved):
Looks good, except maybe some stray edits to the tests.

test/C/src/ActionDelay.lf
erlingrj: Yes, thank you! This is a mistake

core/src/main/java/org/lflang/target/property/DockerProperty.java
erlingrj: Yeah, I agree. I was also thinking in those directions. As we get more experience with containerized deployment of LF programs it might get clearer how we best expose the right knobs to the user. This was added to allow using `FROM: ubuntu` and then also specify `RUN apt update && apt install cmake` which is needed.

Issue #2276, opened 2024-05-08
Federate with imported reactor in a mode does not work


erlingrj:
```
target C;
import T1 from "TimerTest.lf"
reactor R {
  initial mode {
    r = new T1()
  }
}
federated reactor {
  fed = new R()
}
```

Fails with:
```
Cleaning /home/erling/dev/lf-ex/bin
Cleaning /home/erling/dev/lf-ex/src-gen
lfc: info: ##### Generating code for federate federate__fed in directory /home/erling/dev/lf-ex/fed-gen/FedsWithModes/src
lfc: info: ******** Using 1 threads to compile the program.
lfc: info: Generating code for: file:/home/erling/dev/lf-ex/fed-gen/FedsWithModes/src/federate__fed.lf
lfc: info: Generation mode: STANDALONE
lfc: info: Generating sources into: /home/erling/dev/lf-ex/fed-gen/FedsWithModes/src-gen/federate__fed
lfc: info: Target Configuration:
      - keepalive: true
      - single-threaded: false
      - cmake-include: [include/federate__fed_extension.cmake]
      - compile-definitions: {EXECUTABLE_PREAMBLE=, FEDERATED=, FEDERATED_CENTRALIZED=, NUMBER_OF_FEDERATES=1, FEDERATE_ID=0}
      - docker: DockerOptions[enabled=false, from=, rti=lflang/rti:rti]
      - _fed_setup: include/_federate__fed_preamble.h
lfc: info: Copied 'include/federate__fed_extension.cmake' from the file system.
lfc: info: Script for launching the federation: /home/erling/dev/lf-ex/bin/FedsWithModes
lfc: info: Code generation finished.
Exception in thread "pool-1-thread-1" java.lang.NullPointerException: Cannot invoke "org.lflang.lf.Reactor.getTypeParms()" because "definition" is null
        at org.lflang.generator.c.TypeParameterizedReactor.<init>(TypeParameterizedReactor.java:83)
        at org.lflang.generator.c.TypeParameterizedReactor.<init>(TypeParameterizedReactor.java:34)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:820)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:882)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:882)
        at org.lflang.generator.ReactorInstance.<init>(ReactorInstance.java:94)
[...]

PR #2277, opened 2024-05-08
Search for imported reactors to include reactors instantiated in modes

erlingrj:
Solves #2276


lhstrh (changes requested):
This looks great, but please add a test modeled after your example that fails without these changes.

lhstrh (approved):
:+1:

Issue #2282, opened 2024-05-13
Compilation of programs for the Pololu 3pi+ 2040 robot broke with the 0.7.0 release


NiklasRentzCAU:
With the newest release of Lingua Franca 0.7.0, the compilation of programs for the Pololu 3pi 2040 robot throws errors and does not compile correctly anymore. I tested one of the programs of the [Embedded Systems Lab](https://www.lf-lang.org/embedded-lab/Sensors.html), specifically the [AccelerometerDisplay.lf](https://github.com/lf-lang/lf-3pi-template/blob/main/src/AccelerometerDisplay.lf) example with both the 0.6.0 and 0.7.0 versions of the VS Code extension and the lfc tool. Both old versions 0.6.0 compile fine, while the 0.7.0 versions show this in the output and the error log (taken from the VS Code extension):

[CMakeError.log](https://github.com/lf-lang/lingua-franca/files/15296141/CMakeError.log)
[CMakeOutput.log](https://github.com/lf-lang/lingua-franca/files/15296142/CMakeOutput.log)

```
Cleaning /home/nre/git/lf-3pi/include
Cleaning /home/nre/git/lf-3pi/src-gen/AccelerometerDisplay/build
--- Current working directory: /home/nre/git/lf-3pi/src-gen/AccelerometerDisplay/build
--- Executing command: cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/home/nre/git/lf-3pi -DCMAKE_INSTALL_BINDIR=bin -DLF_FILE_SEPARATOR="/" -DLF_SOURCE_DIRECTORY="/home/nre/git/lf-3pi/src" -DLF_PACKAGE_DIRECTORY="/home/nre/git/lf-3pi" -DLF_SOURCE_GEN_DIRECTORY="/home/nre/git/lf-3pi/src-gen/AccelerometerDisplay" /home/nre/git/lf-3pi/src-gen/AccelerometerDisplay
-- The C compiler identification is GNU 10.3.1
-- The CXX compiler identification is GNU 10.3.1
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
-- Detecting C compiler ABI info
Using PICO_SDK_PATH from environment ('/home/nre/git/lf-3pi/pico-sdk')
PICO_SDK_PATH is /home/nre/git/lf-3pi/pico-sdk
[...]

PR #2283, opened 2024-05-13
Add RPi pico support for v0.7

erlingrj:
The RPi pico support needs updates to support v0.7. In this PR I make a few minor changes needed. I have not setup the pico toolchain locally so haven't tested it. See: https://github.com/lf-lang/lingua-franca/issues/2282 Maybe @sberkun can follow up on it? 

Would be great with a simple CI test that at least builds a hello world to avoid this in the future. Do you have any estimate of how hard that would be Samuel?

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
@@ -493,9 +493,9 @@ public class CCmakeGenerator {
     code.pr("pico_sdk_init()");
     code.newLine();
     code.pr("add_subdirectory(core)");
-    code.pr("target_link_libraries(core PUBLIC pico_stdlib)");
-    code.pr("target_link_libraries(core PUBLIC pico_multicore)");
-    code.pr("target_link_libraries(core PUBLIC pico_sync)");
+    code.pr("target_link_libraries(reactor-c PUBLIC pico_stdlib)");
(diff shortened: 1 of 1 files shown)


sberkun:
Sorry for the delay! Following some recent changes (splitting the runtime into many CMake modules), the libraries need to be added to `lf-low-level-platform-api` as well as `reactor-c`. https://github.com/lf-lang/reactor-c/pull/344 addresses `lf-low-level-platform-api` and this PR (or #2178) addresses `reactor-c`.

Setting up a compilation test should be doable, the main steps are:
 - install GCC cross compiler (see [raspberrypi documentation](https://www.raspberrypi.com/documentation/microcontrollers/c_sdk.html#quick-start-your-own-project))
 - clone the sdk
 - set PICO_SDK_PATH environment variable
 - compile (but not run) a hello-world test
I'm not familiar with how the automated tests are set up, but I imagine this could be accomplished by copying the Ubuntu tests and adding a few lines.

sberkun (approved):

edwardalee:
Looks like this change is also included in #2178, so I think this can be safely be closed.

lhstrh:
> Looks like this change is also included in #2178, so I think this can be safely be closed.

@sberkun, could you confirm this?

lhstrh:
Given the lack of response I'm assuming it's safe to close this.

PR #2294, opened 2024-05-22
Fix deadline inference, test GEDF, and remove chain ID

edwardalee:
This PR fixes deadline propagation. Previously, a reaction would inherit the deadline of a downstream reaction only if there was a data path through ports to the downstream reaction.  This corrects this so that a reaction that is lexically earlier than another reaction in a reactor inherits the deadline of that later reactor.

This PR also adds the test [PR 426 of reactor-c](https://github.com/lf-lang/reactor-c/issues/426) that checks for correct behavior of the GEDF scheduler.  It also updates reactor-c to include the corrected GEDF scheduler.

This PR also gets rid of the chain_id, which never worked.


petervdonovan (approved):
I think this PR is good to merge. The inferred deadline algorithm looks right to me, and I am happy that `PeriodicTask` checks execution order using physical time instead of using a global variable.

lhstrh:
@edwardalee: After https://github.com/lf-lang/reactor-c/issues/426 has been merged, I'll update reactor-c with an interactive rebase and clean up the history.

lhstrh:
I wasn't able to do a full rebase without going through too much trouble because merges with master had already happened along the way, so I just did a small few fixups... But `reactor-c` points to `main` now.

Issue #2280, opened 2024-05-09
port.set(0) in Cpp target generates compile error "ambiguous call of overloaded function"


elgeeko1:
Issue: Setting an output port with integral type to the value 0 produces a compile error, unable to disambiguate the call to `set`.

LF version: 0.7.0 stable

Workaround 1: Statically cast `.set(static_cast<uint32_t>(0))`

Workaround 2: Comment out the line `void set(std::nullptr_t) = delete;` from `port.hh` and run `make`.

Steps to recreate:

```c++
target Cpp;
reactor A {
    output a: uint32_t;

    reaction(startup) -> a {=
        a.set(75);
    =}

    reaction(shutdown) -> a {=
        a.set(0);
    =}
}
```

The line `a.set(0)` produces the following compile error:

```shell
A.cc: In member function ‘void A::Inner::reaction_2(const reactor::ShutdownTrigger&, reactor::Output<unsigned int>&)’:
A.cc:86:14: error: call of overloaded ‘set(int)’ is ambiguous
   86 |   a.set(0);
      |   ~~~~~~~~~~~^~~
In file included from src-gen/reactor-cpp-default/include/reactor-cpp/connection.hh:18,
                 from src-gen/reactor-cpp-default/include/reactor-cpp/reactor-cpp.hh:14,
                 from A.hh:9,
                 from A.cc:7:
src-gen/reactor-cpp-default/include/reactor-cpp/port.hh:134:8: note: candidate: ‘void reactor::Port<T>::set(const T&) [with T = unsigned int]’
  134 |   void set(const T& value) { set(make_immutable_value<T>(value)); }
      |        ^~~
src-gen/reactor-cpp-default/include/reactor-cpp/port.hh:135:8: note: candidate: ‘void reactor::Port<T>::set(T&&) [with T = unsigned int]’
  135 |   void set(T&& value) { set(make_immutable_value<T>(std::forward<T>(value))); }
      |        ^~~
src-gen/reactor-cpp-default/include/reactor-cpp/port.hh:137:8: note: candidate: ‘void reactor::Port<T>::set(std::nullptr_t) [with T = unsigned int; std::nullptr_t = std::nullptr_t]’ (deleted)
  137 |   void set(std::nullptr_t) = delete;
      |        ^~~
make[2]: *** [App/CMakeFiles/App.dir/build.make:76: App/CMakeFiles/App.dir/App/A.cc.o] Error 1
[...]

PR #2302, opened 2024-05-30
Disambiguate set(0) for C++ ports

cmnrd:
This pulls in https://github.com/lf-lang/reactor-cpp/pull/57 and adds a new test case.

Fixes https://github.com/lf-lang/lingua-franca/issues/2280


lhstrh (approved):
Nice fix!

PR #2303, opened 2024-05-31
Fixed CMake generator for pico platform

edwardalee:
This PR fixes the CMake generator so that the RP2040 platform works with the new layout of reactor-c.  To be able to build RP2040 code, we also need the [companion PR](https://github.com/lf-lang/reactor-c/pull/441).


cmnrd (commented):
Why do we need to configure the std lib that reactor-c links to in the generated CmakeLists.txt? Shouldn't this be part of the platform support in reactor-c?

lhstrh (approved):

Issue #2310, opened 2024-06-04
Issue regarding target python 


chauncygu:
When I finished generated code and ran Python code with `python /Users/shangding/Documents/mycode/lingua_franca/test_program/src-gen/first_reactor/first_reactor.py`, It happens errors:
`Traceback (most recent call last):
  File "/Users/shangding/Documents/mycode/lingua_franca/test_program/src-gen/first_reactor/first_reactor.py", line 7, in <module>
    from LinguaFrancafirst_reactor import (
SystemError: initialization of LinguaFrancafirst_reactor did not return an extension module` 

The source code within `first_reactor.lf` is as follows:

`target Python

main reactor {
  reaction(startup) {=
    print("Hello World.")
  =}
}`

On my laptop, C can work well, but Python has the above issue.

edwardalee:
This is the same issue as #2298.  We really need to revert the change referenced there.

Issue #2298, opened 2024-05-29
Wrong Python version invoked


edwardalee:
Between lfc 0.7.0 and the current version (lfc 0.7.3-SNAPSHOT), something changed that results in the wrong Python being used.
In version 0.7.0, lfc creates a script in `bin` that looks like this for me:
```
        /Users/edwardlee/.pyenv/shims/python3.10 /Users/edwardlee/git/playground-lingua-franca/examples/Python/src-gen/lib/Plotters/Plotters.py "$@"
```
This is the version of Python that is in my path:
```
~/git/playground-lingua-franca/examples/Python % which python
/Users/edwardlee/.pyenv/shims/python
~/git/playground-lingua-franca/examples/Python % python --version
Python 3.10.13
```
But in the current version, the script uses the built-in Mac version of Python:
```
        /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/bin/python3.9 /Users/edwardlee/git/playground-lingua-franca/examples/Python/src-gen/lib/Plotters/Plotters.py "$@"
```
What changed and why?  This breaks almost everything in the Python target.

edwardalee:
The offending (merged) PR is this one #2292.  It is a one-line change:
```
-            find_package(Python 3.10.0...<3.11.0 REQUIRED COMPONENTS Interpreter Development)
+           find_package(Python 3.9.0...<3.11.0 REQUIRED COMPONENTS Interpreter Development)
```
Unfortunately, this change causes CMake to reference the built-in Python on Mac, not the one in the path, which means virtual environments don't work anymore and you are stuck using the built-in Python.  This change was put in to support the no-GIL Python, but I thought we had other reasons for requiring Python 3.10?  @jackyk02 ? Suggestions?

lhstrh:
> I thought we had other reasons for requiring Python 3.10? @jackyk02 ? Suggestions?

Are these reasons documented anywhere? One thing we could do is add a workflow that runs the tests for Python 3.10 and see what happens. If the tests don't pass, we should revert https://github.com/lf-lang/lingua-franca/pull/2292.

The problem you're running into is triggered by the change, but it's really because you apparently have Python 3.10 installed on your machine and CMake finds it, which is a different issue. Following your suggestion to not let CMake determine which version of Python to use, I created https://github.com/lf-lang/lingua-franca/issues/2299.

edwardalee:
[...]

PR #2312, opened 2024-06-05
No more support for Python 3.9

lhstrh:
This reverts commit 69c44d5076d99cb61e9b81b7cd06b55dd56ff106.

Fixes https://github.com/lf-lang/lingua-franca/issues/2310 and https://github.com/lf-lang/lingua-franca/issues/2298.

This is not a real fix, however, because the issue due to the interaction between CMake and Python, not because of anything particular to Python 3.9.


lhstrh:
Issue already exists.

PR #2315, opened 2024-06-07
Support for Patmos platform on static-schedule-single-thread branch by direct Makefile

EhsanKhodadad:


schoeberl:
In general, as it is mostly Patmos-specific changes, it looks fine. However, 2 points: (1) should a change in core/threaded/scheduler_static.c really be part of this commit? @lsk567 needs to approve those changes. (2) is test/C the best place for Patmos-specific README and Makefile?

lsk567:
@EhsanKhodadad Thanks for working on the Patmos support and get them to work with the static scheduler. However, I don't think these files belong in this [lingua-franca](https://github.com/lf-lang/lingua-franca) repository. They seem to belong to the [reactor-c](https://github.com/lf-lang/reactor-c) repository.

Suggestions for a proper integration:
- `lf_patmos_support.c` and `lf_patmos_support.h` belong to `reactor-c/low_level_platform`. Here is a [FlexPRET example](https://github.com/lf-lang/reactor-c/blob/main/low_level_platform/impl/src/lf_flexpret_support.c).
- The Makefile and README should go into a separate template repo, such as [this Pico template](https://github.com/lf-lang/lf-pico-template).

But given that this branch for single-threaded execution is not a stable solution (code from the `threaded` folder is leaking into single-threaded code), I suggest we _not_ merge `static-schedule-single-threaded` into `static-schedule` or `master`. All the changes can park on this branch or a separate repo for this upcoming paper. 

Then, we can work on the proper Patmos integration when we can print static schedules into a `main()` function. Therefore, I suggest we come back to this PR later.

lhstrh:
Is PR still relevant considering there is #2383? Can it be closed?

Issue #2319, opened 2024-06-14
Reactor naming mechanism in C target


OmerMajNition:
```
target C;

reactor A (bank_index:int = 0, name:string = "A", log_level:int = 0) {
    reaction (startup) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Started\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}

    reaction (shutdown) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Terminated\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}
}

reactor B (bank_index:int = 0, name:string = "B", log_level:int = 0) {
    a = new [3] A();
    reaction (startup) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Started\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}

    reaction (shutdown) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Terminated\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}
}

reactor C (bank_index:int = 0, name:string = "C", log_level:int = 0) {
    b = new [2] B();
    reaction (startup) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Started\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}

    reaction (shutdown) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
                    "%s [%d] Terminated\n",
                    lf_time_logical_elapsed(), lf_tag().microstep, lf_time_physical_elapsed(),
                    self->name, self->bank_index
        );
    =}
}

reactor D (bank_index:int = 0, name:string = "D", log_level:int = 0) {
    a = new [2] A(name = "Another");
    reaction (startup) {=
        fprintf (   stderr, "(%lld, %u) physical_time:%lld "
[...]

PR #2320, opened 2024-06-14
naming banked reactors

OmerMajNition:
https://github.com/lf-lang/lingua-franca/issues/2319 has the description of the problem addressed with this PR

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CGenerator.java
@@ -1923,8 +1923,42 @@ public class CGenerator extends GeneratorBase {
                     + ";",
                 selfRef + "->" + parameter.getName() + " = " + temporaryVariableName + ";"));
       } else {
-        initializeTriggerObjects.pr(
-            selfRef + "->" + parameter.getName() + " = " + initializer + ";");
+        if (parameter.getName().equals("name")) {
+          ReactorInstance parent = instance.getParent();
+          boolean parent_name_found = false;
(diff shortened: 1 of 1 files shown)


cmnrd:
I am not entirely sure what this tries to achieve. Is this about getting the fully qualified names of reactors? I think we already provide functionality for that, also in the C target.

cmnrd:
@petervdonovan Could you comment on this? Are there FQNs available in C, and how are they exposed?

petervdonovan:
> Are there FQNs available in C, and how are they exposed?

If they are available in the C target, I am not aware of them, and after looking briefly I do not see them in the source code (in particular, there is no `name` field in the `self_base_t` struct defined [here](https://github.com/lf-lang/reactor-c/blob/76408229db1e096d586f96eb7ea5259de6405220/include/core/lf_types.h#L276).

It looks like this PR adds FQNs without requiring any extension of `self_base_t` when the `name` field is not used, so I think we might as well merge it. CI is failing, but I think that should be easily fixed.

cmnrd:
Ok. I still don't understand what the PR does, though. @OmerMajNition could you add some explanation? Is the idea to introduce a magic parameter `name` that gets automatically assigned the FQN?

OmerMajNition:
@cmnrd , I've added further description in GH-Issue: https://github.com/lf-lang/lingua-franca/issues/2319. Please let me know if we need further clarification.

edwardalee:
I believe this is fixed in a better way with #2450 .

PR #2321, opened 2024-06-16
Stop removing GCC on MacOS in CI

petervdonovan:
The LSP tests are error prone because they test interactions with external tools. They broke because correct behavior depends on how the external tools are made available on the system.

A robust solution would involve either removing the feature that the LSP tests are designed to test, or dockerizing the LSP tests. However, this PR provides a quick stopgap solution instead so that we can avoid having to force an immediate long-term decision.


petervdonovan:
My latest commit message explains why I decided to give up on this aspect of the workflow run. I can try harder if we think it is important to do so.

lhstrh (approved):
Works for me -- thanks for fix and the explanation!

PR #2322, opened 2024-06-17
Docker support for C++ target

lhstrh:


core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+            "install",
+            "--config",
+            CppStandaloneGenerator.Companion.buildTypeToCmakeConfig(
+                context.getTargetConfig().get(BuildTypeProperty.INSTANCE)));
+    return Stream.of(mkdirCommand, cmakeCommand, makeCommand, installCommand)
+        .map(CppDockerGenerator::argListToCommand)
+        .toList();
+  }
+
+  static String argListToCommand(List<String> args) {
```
lhstrh, line 95: @petervdonovan: Good idea to factor this out, but this probably belongs in `DockerGenerator`

core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt
```diff
@@ -94,10 +96,34 @@ class CppGenerator(
             context.reportProgress(
                 "Code generation complete. Compiling...", IntegratedBuilder.GENERATED_PERCENT_PROGRESS
             )
-            if (platformGenerator.doCompile(context)) {
-                context.finish(GeneratorResult.Status.COMPILED, codeMaps)
+            if (targetConfig.get(DockerProperty.INSTANCE).enabled) {
+                FileUtil.deleteDirectory(context.fileConfig.srcGenPath.resolve("src-gen"))
```
lhstrh, line 100: Is there an elegant way to factor this out into a function?

lhstrh:
@oowekyala are you interested in pulling together the same ability for Rust?

core/src/main/java/org/lflang/generator/docker/CDockerGenerator.java
```diff
@@ -40,7 +40,7 @@ public List<String> defaultEntryPoint() {
   @Override
   protected String generateRunForInstallingDeps() {
     var config = context.getTargetConfig();
-    var compiler = config.target == Target.CCPP ? "g++" : "gcc";
+    var compiler = config.target == Target.CCPP || config.target == Target.CPP ? "g++" : "gcc";
```
lhstrh, line 43:
```suggestion
    var compiler = config.target == Target.CCPP ? "g++" : "gcc";
```
(because `CppDockerGenerator` no longer extends `CDockerGenerator`

core/src/main/kotlin/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt
```diff
@@ -153,6 +153,8 @@ class CppStandaloneCmakeGenerator(private val targetConfig: TargetConfig, privat
                 |cmake_minimum_required(VERSION 3.5)
                 |project(${fileConfig.name} VERSION 0.0.0 LANGUAGES CXX)
                 |
+                |option(REACTOR_CPP_LINK_EXECINFO "Link against execinfo" OFF)
```
cmnrd, line 156: Why do we need to link to execinfo?

core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+    cmakeCommand.addAll(
+        List.of(
+            "-DCMAKE_INSTALL_BINDIR=bin",
+            "-DCMAKE_INSTALL_PREFIX=.",
+            "-DREACTOR_CPP_LINK_EXECINFO=ON",
+            "-S",
+            "src-gen",
+            "-B",
+            "build"));
+    var makeCommand =
```
cmnrd, line 69: Duplicating the complete commands is quite unfortunate, as the commands used here and in the normal build will get out of sync. Also, this does not for the ROS2 support of the C++ target. I think a better approach would be to extend the platform generator interface, adding a method for querying the commands that are used there. Alternatively, we could also specialize the docker generator for the standalone and the ros2 platform and have a `getDockerGenerator` method as part of the platform generator interface.

cmnrd (changes requested):
This looks like a good start, but unfortunately, this does not work on my machine. I get:
```
WARN[0000] a network with name lf exists but was not created for project "helloworldcontainerized".
Set `external: true` to use an existing network
[+] Running 1/0
 ✔ Container HelloWorldContainerized  Created                                                                                                                                                                       0.0s
Attaching to HelloWorldContainerized
HelloWorldContainerized  | Error loading shared library libreactor-cpp-default.so.1: No such file or directory (needed by ./bin/HelloWorldContainerized)
HelloWorldContainerized  | Error loading shared library libexecinfo.so.1: No such file or directory (needed by ./bin/HelloWorldContainerized)
HelloWorldContainerized  | Error loading shared library libstdc++.so.6: No such file or directory (needed by ./bin/HelloWorldContainerized)
HelloWorldContainerized  | Error loading shared library libgcc_s.so.1: No such file or directory (needed by ./bin/HelloWorldContainerized)
HelloWorldContainerized  | Error relocating ./bin/HelloWorldContainerized: _ZN7reactor14ReactorElementC2ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS0_4TypeEPNS_7ReactorE: symbol not found
HelloWorldContainerized  | Error relocating ./bin/HelloWorldContainerized: _ZSt9use_facetISt5ctypeIcEERKT_RKSt6locale: symbol not found
HelloWorldContainerized  | Error relocating ./bin/HelloWorldContainerized: __cxa_free_exception: symbol not found
HelloWorldContainerized  | Error relocating ./bin/HelloWorldContainerized: __cxa_begin_catch: symbol not found
HelloWorldContainerized  | Error relocating ./bin/HelloWorldContainerized: _ZNKSt9basic_iosIcSt11char_traitsIcEE4failEv: symbol not found
...

File: core/src/main/kotlin/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt (hunk 4 above)
Comment (petervdonovan) [reply core/src/main/kotlin/org/lflang/generator/cpp/CppStandaloneCmakeGenerator.kt:156]: It is necessary to do this explicitly on Alpine. There is some discussion [here](https://github.com/awslabs/aws-lambda-cpp/issues/124), in particular [this comment](https://github.com/awslabs/aws-lambda-cpp/issues/124#issuecomment-974645808).

File: core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java (hunk 5 above)
Comment (petervdonovan) [reply core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java:69]: OK, this sounds like it should work.

Comment (petervdonovan) [reply core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java]: I ran out of time before I got the build to work on ROS. What needs to be done there is that there needs to be a special Configurator for ROS 2 support + Docker support in the test framework, and the ROS 2 bash script needs to somehow be copied into the Docker image. I could do that tomorrow if we think that it is important.

Comment (lhstrh) [reply core/src/main/java/org/lflang/generator/docker/CppDockerGenerator.java]: Great effort, @petervdonovan. This brings us really close.

File: core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt [hunk 6]
```diff
@@ -94,14 +94,48 @@ class CppGenerator(
...
+                    context.unsuccessfulFinish()
+                }
             }
         }
     }
 
+    /**
+     * Copy the contents of the entire src-gen directory to a nested src-gen directory next to the generated Dockerfile.
+     */
+    private fun copySrcGenBaseDirIntoDockerDir() {
```
cmnrd, line 113: This is a very hacky solution. I wish there would be an easy way to avoid the copying. The fact that the copying is needed is symptomatic of a more profound issue in the way the normal code generator and the docker generator interact. If the docker generator needs sources to be in a specific location, it should pass an according file config to the C++ generator, and it will do that. However, this doesn't work in the current architecture, as everything goes through GeneratorBase and the target code generator ends up calling the docker generator, and not the other way around.

cmnrd (approved):
I pushed a quite significant overhaul. Most importantly, the docker generators are now inner classes of the platform-specific generators (`CppStandalonGenerator` and `CppRos2Generator`). 

Docker generation for ROS2 works now, too. This is particularly useful as it allows building and running ROS nodes without a ROS installation.

core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt
lhstrh: I agree with you...

lhstrh (commented):
@cmnrd, this looks very nice!!

PR #2328, opened 2024-06-20
Fixed docker testing

cmnrd:
For some reason, we were not testing the docker generation using the launch scripts that we place in `bin`, but instead the test framework would generate a custom docker compose file and run that. This, however, did not work at all. I don't understand the precise reasons, but for the docker tests it would simply do nothing. Thus, a couple of problems in the docker generation were masked, as we never really executed the containers or tested the generated launch script.

This PR simply removes the generation of custom docker compose files in the test framework and executes the tests using the generated launch script. This PR also includes a series of fixes for bugs that were uncovered by this change.

## Summary by CodeRabbit

- **New Features**
  - Added Docker support across multiple components for enhanced containerization capabilities.
  - Introduced a Docker parameter in the TypeScript file configuration, which can be set dynamically.

- **Refactor**
  - Refined file path handling and command creation to improve script generation and Docker integration.
  - Simplified Docker-related functionality in test setups.

- **Bug Fixes**
  - Adjusted method and command logic to ensure compatibility with Docker environments.

These changes collectively improve Docker support, streamline file path handling, and refine script generation logic.

### DIFF
--- a/core/src/main/java/org/lflang/generator/LFGenerator.java
+++ b/core/src/main/java/org/lflang/generator/LFGenerator.java
@@ -60,7 +60,7 @@ public class LFGenerator extends AbstractGenerator {
         case Python -> new PyFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case CPP -> new CppFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case Rust -> new RustFileConfig(resource, srcGenBasePath, useHierarchicalBin);
-        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin, null);
       };
     } catch (IOException e) {
       throw new RuntimeException(
--- a/core/src/main/java/org/lflang/generator/LFGenerator.java
+++ b/core/src/main/java/org/lflang/generator/LFGenerator.java
@@ -60,9 +60,7 @@ public class LFGenerator extends AbstractGenerator {
         case Python -> new PyFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case CPP -> new CppFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case Rust -> new RustFileConfig(resource, srcGenBasePath, useHierarchicalBin);
-          // null is passed to the docker argument because it is up to the TS Generator to tell the
-          // FileConfig whether Docker is enabled
-        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin, null);
+        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin);
       };
     } catch (IOException e) {
       throw new RuntimeException(
--- a/core/src/main/java/org/lflang/generator/LFGenerator.java
+++ b/core/src/main/java/org/lflang/generator/LFGenerator.java
@@ -60,6 +60,8 @@ public class LFGenerator extends AbstractGenerator {
         case Python -> new PyFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case CPP -> new CppFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case Rust -> new RustFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+          // null is passed to the docker argument because it is up to the TS Generator to tell the
+          // FileConfig whether Docker is enabled
         case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin, null);
       };
     } catch (IOException e) {
--- a/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
+++ b/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
@@ -42,7 +42,7 @@ import java.nio.file.Path
  *  @author Hokeun Kim
  */
 class TSFileConfig(
-    resource: Resource, srcGenBasePath: Path, useHierarchicalBin: Boolean, var docker: Boolean? = null
+    resource: Resource, srcGenBasePath: Path, useHierarchicalBin: Boolean
 ) : FileConfig(resource, srcGenBasePath, useHierarchicalBin) {
 
     /**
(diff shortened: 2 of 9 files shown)


cmnrd:
Turns out that the TypeScript docker test is failing. I don't understand why, though. It works fine if I compile it with lfc, even if I manually force `hierarchicalBin=true`. @petervdonovan is that something you could help with?

lhstrh (commented):
This looks like some serious cruft in the testing framework you got rid of. I wasn't even aware that it was there...

petervdonovan:
> This, however, did not work at all. I don't understand the precise reasons, but for the docker tests it would simply do nothing.

I'm not sure I understand because it seems to me like the tests have been doing something. For example, [this comment](https://github.com/lf-lang/lingua-franca/pull/2322#pullrequestreview-2126146353) that you posted earlier @cmnrd shows a runtime error (during dynamic linking) while executing a Docker test with `docker compose`, presumably via the Bash script that was in the test framework.

petervdonovan (dismissed):
With the changes in this PR, `test/C/src/docker/FilesPropertyContainerized` does not result in a test failure even when it finishes with a nonzero exit code. You can verify this by adding `exit(1)` to the startup reaction and running the C Docker tests:
```c
target C {
  files: "../include/hello.h",
  docker: true
}

preamble {=
  #include "hello.h"
=}

main reactor {
  reaction(startup) {=
    hello_t hello;
    exit(1);
    printf("SUCCESS\n");
  =}
}
```

But in `master`, this modified version of the test fails with a helpful error message, as it should:
```
====================================================================
ERROR: One or multiple containers exited with a non-zero exit code.
       See the log above for details. The following containers failed:
FilesPropertyContainerized exited with code 1

+---------------------------------------------------------------------------+
```

core/src/main/java/org/lflang/generator/LFGenerator.java
```diff
@@ -60,7 +60,7 @@ public static FileConfig createFileConfig(
         case Python -> new PyFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case CPP -> new CppFileConfig(resource, srcGenBasePath, useHierarchicalBin);
         case Rust -> new RustFileConfig(resource, srcGenBasePath, useHierarchicalBin);
-        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin);
+        case TS -> new TSFileConfig(resource, srcGenBasePath, useHierarchicalBin, null);
```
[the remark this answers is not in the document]
petervdonovan: Yes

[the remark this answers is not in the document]
lhstrh: I'm actually not sure I understand why we're doing this...

core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
```diff
@@ -54,13 +54,23 @@ class TSFileConfig(
...
+    }
+
     override fun getCommand(): LFCommand {
-        return LFCommand.get(
-            "node",
-            listOf(srcPkgPath.relativize(executable).toString()),
-            true,
-            srcPkgPath
-        )
+        if (docker == true) {
```
lhstrh, line 62: Why should the command be different depending on whether we use Docker or not?

petervdonovan: In this PR, we change to using the command in order to execute the program when we are running tests. In Docker mode, the command is no longer `node`; instead, the command that needs to be run is a command that executes the generated launcher script, which, in turn, executes `docker compose`.

lhstrh: Why don't we just unify these use cases and execute a generated script in all cases? That would make most sense to me... In fact, this might already be what we're doing, except the test framework didn't catch up?

petervdonovan: In the test framework, that's what we were doing before we had this PR. But you're right, that's not what we've been doing in general. For most purposes we've been paying attention to the executables that the different targets say they produce; they say what executable they produce by overriding this method.

TypeScript does override the method, which caused a problem now that we changed the Docker tests to use the output of that method. One way to solve this is to essentially un-override by calling the superclass method when in Docker mode. It might be better to ignore the executables that the different targets say they produce and to assume that they all use a generated launcher script that is located in the same place, but it isn't obvious to me that that is better, and it wasn't important enough to me for me to think about it too hard. I'll be happy with whichever way we do it.

lhstrh: I think all targets should either produce a binary or a script that is the entry point. That gives a uniform user experience and is the easiest to deal with in the rest of the toolchain.

petervdonovan: That makes sense, but the only downside is that it increases the chance that we will have issues related to running Bash scripts on Windows.

lhstrh: I'd be OK with that. Most Windows users will use WSL if they are to use our tools from the command line.

petervdonovan: That is OK with me as well, but it is not consistent with the approach that we have been taking toward testing, so we might need to change some things to make that pass on Windows.

It also is not really in the same spirit as the kind of Windows support that we have claimed to have, since if the intended way to use LF on Windows is via WSL, we do not really need Windows support at all, and can probably just get by with Linux support. That is OK with me, too -- actually it is better than OK, because it means less work for everyone -- but it is a change that we might need people to be on board with.

lhstrh (approved):
:rocket:

cmnrd:
> I'm not sure I understand because it seems to me like the tests have been doing something. For example, [this comment](https://github.com/lf-lang/lingua-franca/pull/2322#pullrequestreview-2126146353) that you posted earlier @cmnrd shows a runtime error (during dynamic linking) while executing a Docker test with `docker compose`, presumably via the Bash script that was in the test framework.

I don't fully understand either. It appears that our tests were good at picking up errors where our generated binary fails, but not where the container execution itself fails. The error message that you link to did not show up in testing, but only when I executed the program myself. Similarly, I was struggling with an issue where the entry point in the dockerfile was wrong, but it did not show up in the tests (The output from the execute step was simply empty). 

> With the changes in this PR, test/C/src/docker/FilesPropertyContainerized does not result in a test failure even when it finishes with a nonzero exit code. You can verify this by adding exit(1) to the startup reaction and running the C Docker tests:

That is an excellent catch! I did not notice that `docker compose up` would mask those errors. I addressed this in https://github.com/lf-lang/lingua-franca/pull/2330

> Pretty sure this has been addressed

Nope...

Issue #2299, opened 2024-05-29
Ability to specify Python version


lhstrh:
It would be great if we could *explicitly* specify which version of to use.

PR #2332, opened 2024-06-26
New target property for specifying Python version

jackyk02:
This PR introduces a new target property `python-version` that allows users to explicitly specify the Python version to be used. 

Related Issues and PR:
[Issue #2298](https://github.com/lf-lang/lingua-franca/issues/2298)
[Issue #2299](https://github.com/lf-lang/lingua-franca/issues/2299)
[PR #2292](https://github.com/lf-lang/lingua-franca/pull/2292)

Example:
```
target Python {
  python-version: "3.9.10"
}
```

If `python-version` is not specified, the system will default to using Python 3.10 or newer.

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a specific Python version when generating Python code, defaulting to Python 3.10 or newer if no version is specified.

- **Improvements**
  - Updated the Python code generation process to allow more precise control over the Python interpreter version used, enhancing compatibility and stability.

### DIFF
--- a/core/src/main/java/org/lflang/generator/python/PythonGenerator.java
+++ b/core/src/main/java/org/lflang/generator/python/PythonGenerator.java
@@ -278,7 +278,6 @@ public class PythonGenerator extends CGenerator {
 
   @Override
   protected void handleProtoFiles() {
-    pythonVersion = targetConfig.get(PythonVersionProperty.INSTANCE);
     for (String name : targetConfig.get(ProtobufsProperty.INSTANCE)) {
       this.processProtoFile(name);
       int dotIndex = name.lastIndexOf(".");
--- a/core/src/main/java/org/lflang/generator/python/PythonGenerator.java
+++ b/core/src/main/java/org/lflang/generator/python/PythonGenerator.java
@@ -60,6 +60,7 @@ import org.lflang.lf.Port;
 import org.lflang.lf.Reaction;
 import org.lflang.lf.Reactor;
 import org.lflang.target.Target;
+import org.lflang.target.TargetConfig;
 import org.lflang.target.property.DockerProperty;
 import org.lflang.target.property.ProtobufsProperty;
 import org.lflang.target.property.PythonVersionProperty;
@@ -90,7 +91,6 @@ public class PythonGenerator extends CGenerator {
   private final List<String> pythonRequiredModules = new ArrayList<>();
 
   private final PythonTypes types;
-  private static String pythonVersion;
 
   public PythonGenerator(LFGeneratorContext context) {
     this(
@@ -364,7 +364,6 @@ public class PythonGenerator extends CGenerator {
    */
   @Override
   public void doGenerate(Resource resource, LFGeneratorContext context) {
-    setUpPythonVersion();
     int cGeneratedPercentProgress = (IntegratedBuilder.VALIDATED_PERCENT_PROGRESS + 100) / 2;
     code.pr(
         PythonPreambleGenerator.generateCIncludeStatements(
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/lflang/target/property/PythonVersionProperty.java
```diff
@@ -0,0 +1,17 @@
+package org.lflang.target.property;
+
+/** Directive for specifying a specific version of the reactor runtime library. */
+public final class PythonVersionProperty extends StringProperty {
```
lhstrh, line 4: We probably need some custom implementation of `validate` in this class, and check that no unsupported versions are selected.

lhstrh (changes requested):

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -556,6 +559,12 @@ protected void additionalPostProcessingForModes() {
     PythonModeGenerator.generateResetReactionsIfNeeded(reactors);
   }
 
+  /** Initializes the Python version based on the user's configuration. */
+  protected void setUpPythonVersion() {
```
lhstrh, line 563: Just return a string here instead of setting a static variable.

lhstrh (changes requested):
Thanks for addressing the feedback! Looks much better now. Please still get rid of the static variable.

lhstrh:
Note also [this comment](https://github.com/lf-lang/lingua-franca/issues/2298#issuecomment-2256104512).

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
lhstrh: Fixed it for you.

core/src/main/java/org/lflang/generator/python/PythonGenerator.java
```diff
@@ -88,6 +89,7 @@ public class PythonGenerator extends CGenerator {
   private final List<String> pythonRequiredModules = new ArrayList<>();
 
   private final PythonTypes types;
+  private static String pythonVersion;
```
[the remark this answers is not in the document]
lhstrh: Fixed.

lhstrh (commented):
Looks good now, but I'd like to discuss this feature in the developer meeting because I'd like to better understand how this would interact with virtual environments.

lhstrh (approved):

cmnrd:
Looks like [`FindPython`](https://cmake.org/cmake/help/latest/module/FindPython.html) has a variable, which when set to `FIRST` favors the currently active virtual environment.

core/src/main/java/org/lflang/target/property/PythonVersionProperty.java
```diff
@@ -0,0 +1,35 @@
...
+    return "python-version";
+  }
+
+  @Override
+  public void validate(TargetConfig config, MessageReporter reporter) {
+    String version = config.get(PythonVersionProperty.INSTANCE);
+    if (!version.isEmpty() && !version.contains("3.10")) {
+      reporter
+          .at(config.lookup(this), Literals.KEY_VALUE_PAIR__NAME)
+          .warning(
```
lhstrh, line 28: @cmnrd: note that we issue a warning if anything other than `3.10` is selected.

jackyk02:
I've conducted a series of tests to evaluate Python version detection in various environments:

**1. Python 3.10 Virtual Environment**
   - Created with: `conda create -n py310 python=3.10`
   - Activated with: `conda activate py310`
   - Target Python configuration:
     ```
     target Python {
       python-version: "3.10"
     }
     ```
   - Result: Correct Python interpreter found regardless of whether `python-version` was specified.

**2. Python 3.11 Virtual Environment**
   - Created with: `conda create -n py311 python=3.11`
   - Activated with: `conda activate py311`
   - Target Python configuration:
     ```
     target Python {
       python-version: "3.11"
     }
     ```
   - Result: Correct Python interpreter found when `python-version` was specified.
   - Note: If `python-version` is not specified (defaults to 3.10), the following error occurs:
     ```
     CMake Error at /usr/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
       Could NOT find Python (missing: Python_INCLUDE_DIRS Python_LIBRARIES
       Development Development.Module Development.Embed) (found suitable version
       "3.10.14", required range is "3.10.0...<3.11.0")
     ```

**3. Global Python Environment**
   - Assumes Python 3.10 is the interpreter determined by the `$PATH` environment variable
   - Deactivated conda environment: `conda deactivate`
   - Target Python configuration:
     ```
     target Python {
       python-version: "3.10"
     }
     ```
Result: Correct Python interpreter found 

Based on these tests, it appears that `FindPython` already prioritizes the currently active virtual environment. Activating a conda virtual environment essentially applies settings to the shell to use a specific Python interpreter along with its associated modules and libraries (e.g. Python_INCLUDE_DIRS and Python_LIBRARIES).

Our current approach allows users to manage their Python environments and this PR provides an option for them to set the Python version used by LF.

cmnrd:
I can confirm this also for venv with cmake 3.30 on Linux. When the virtual environment is disabled, cmake uses the system's interpreter. When it is active, it uses the interpreter from the virtual environment.

I also found out that `Python_FIND_VIRTUALENV` is not specified by default. I don't know what the default behavior is, perhaps it changed, but I don't think so as this is usually communicated in the cmake documentation.

edwardalee:
The python-version option has been implemented in #2441 .

PR #2333, opened 2024-06-26
Improved handling of `files` target property

byeonggiljun:
The changes in this PR ensure that all files are properly collected from imported reactors before copying them.


core/src/main/java/org/lflang/generator/GeneratorBase.java
```diff
@@ -219,8 +219,6 @@ public void doGenerate(Resource resource, LFGeneratorContext context) {
     }
 
     // Process target files. Copy each of them into the src-gen dir.
-    // FIXME: Should we do this here? This doesn't make sense for federates the way it is
-    // done here.
     copyUserFiles(this.targetConfig, context.getFileConfig());
```
lhstrh, line 222:
```suggestion
```
I think we should remove the call to `copyUserFiles` because not all targets support the `files` property...

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -710,7 +712,8 @@ private void inspectReactorEResource(ReactorDecl reactor) {
       if (lfResource != null) {
         var config = lfResource.getTargetConfig();
         // FIXME: this should not happen here, but once, after collecting all the files.
-        copyUserFiles(config, lfResource.getFileConfig());
+        // copyUserFiles(config, lfResource.getFileConfig());
```
lhstrh, line 715:
```suggestion
```

core/src/main/java/org/lflang/generator/c/CGenerator.java
```diff
@@ -719,7 +722,22 @@ private void inspectReactorEResource(ReactorDecl reactor) {
               if (p.isPresent()) {
                 var property = p.get();
                 if (property.loadFromImport()) {
-                  property.update(this.targetConfig, pair, messageReporter);
+                  if (property instanceof FileListProperty) {
```
lhstrh, line 725: I see why you did this (the base path is not available in `FileListProperty`), but it's kind of unclean to have this special case for handling `FileListProperty` here. Should we consider passing in a `GeneratorContext` in the `update` function instead, so we gain access to both the target config _and_ the file config?

core/src/main/java/org/lflang/generator/GeneratorBase.java
byeonggiljun: I did this in [57f29cc](https://github.com/lf-lang/lingua-franca/pull/2333/commits/57f29cc1d51a4105db723fc2c78bca6aee892bf7). Also, I found that `TypeScript` target also doesn't support `files` property. Let's see if this doesn't break any other tests...

core/src/main/java/org/lflang/generator/c/CGenerator.java
byeonggiljun: This seems to have already been resolved.

core/src/main/java/org/lflang/generator/c/CGenerator.java
byeonggiljun: I also wanna do this. However, I see one problem here. Consider [DistributedCMakeIncludeSeparateCompile.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/target/DistributedCMakeIncludeSeparateCompile.lf) looks like below.

```
target C {
  timeout: 0 sec
}

import Foo from "CMakeInclude.lf"
...
```
`CMakeInclude.lf`(https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/target/CMakeInclude.lf) has the `cmake-include` property like below.

```
target C {
  cmake-include: [
      "../include/mlib-cmake-extension.cmake",
      "../include/foo-cmake-compile-definition.txt"],
  timeout: 0 sec
}
```

When we call `inspectReactorEResource` to gather `CMakeIinclude.lf`'s target property, we need to update the **parent reactor** (`DistributedCMakeIncludeSeparateCompile.lf`)'s `TargetConfig` using the **child reactor** (`CMakeInclude.lf`)'s `FileConfig`. Thus, solely passing one `GeneratorContext` doesn't resolve this problem. Do you think passing both parent _and_ child's `GeneratorContext` to the function `update` makes sense?

lhstrh: Yep, you're right. I realized this too... Let me noodle on it a little bit. I have a long flight ahead, so I'll work on this stuff a bunch.

lhstrh: I think we need an interface `ImportableTargetProperty` with an additional `update` method that carries an extra `sourceContext`. Only few implementations of `TargetProperty` will need to implement it.

lhstrh: There already was a mechanism in place for federations that we could simply reuse, so I did that. I think this is good to merge now.

PR #2335, opened 2024-06-28
Fixed docker generation for images that use /bin/sh

cmnrd:
Some images, like debian slim, use the `/bin/sh` which does not provide a `source` command. This PR replaces `source` with `.`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved script execution by updating command paths to use `.` instead of `source`.


lhstrh (approved):
Looks good!

PR #2340, opened 2024-06-30
Fix for handling of target properties in imported files accessed through reactor class inheritance

### LINKED ISSUE #2286: Target property not imported if we extend the reported reactor

lhstrh:
Fixes #2286.

## Summary by CodeRabbit

- **New Features**
  - Introduced `ImportedCMakeInclude.lf` to support importing and extending reactor definitions with a timeout mechanism.

- **Improvements**
  - Enhanced resource management by switching from `List` to `Set`, ensuring unique resources and improving efficiency in `GeneratorBase` and `GeneratorUtils`.

### DIFF
--- a/core/src/main/java/org/lflang/generator/GeneratorBase.java
+++ b/core/src/main/java/org/lflang/generator/GeneratorBase.java
@@ -34,6 +34,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


cmnrd (approved):
Looks good!

PR #2347, opened 2024-07-02
Fixes in adaptive scheduler for federated execution

### LINKED ISSUE #2271: CycleDetection.lf deadlocks with adaptive scheduler
I can reproduce this on `master` with my laptop. It succeeds with the other schedulers.

petervdonovan:
Closes #2271.

There are still unfreed memory warnings when using `build-type: DEBUG`, but I think these are a separate bug because they also appear with the NP scheduler.


cmnrd (approved):
Thanks!

PR #2351, opened 2024-07-02
Add an option to capture fed outputs

edwardalee:
When running a federated program on a single machine, stdout from all federates is combined into one unreadable stream.  This PR adds a command-line option `-l` which causes stdout to be redirected to log files.  This makes debugging easier.


edwardalee:
Oh, it seems somehow these changes got merged into master already. Closing this.

PR #2352, opened 2024-07-04
Updated hyperlinks in README

maurerle:
sets correct URL for documentation and research results


lhstrh (commented):
~~We will want to fix this for earlier versions as well...~~

lhstrh:
~~> We will want to fix this for earlier versions as well...~~

Nevermind; this is in `lingua-franca`, not the website, where we freeze/duplicate docs and sometimes have to replicate fixes.

PR #2353, opened 2024-07-07
Define self variable so it can be used in instantiations

edwardalee:
In the Python target, when setting parameters of an instance, it is not possible in native initializer code to reference the parameters of the container. This PR fixes that and adds a test.

### DIFF
--- a/test/Python/src/BankIndex.lf
+++ b/test/Python/src/BankIndex.lf
@@ -1,0 +1,14 @@
+target Python;
+reactor A(bank_index = 0, value = 0) {
+  reaction (startup) {=
+    print("bank_index: {:d}, value: {:d}".format(self.bank_index, self.value))
+    if (self.value != 4 - self.bank_index):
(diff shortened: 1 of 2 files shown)


test/Python/src/BankIndex.lf
```diff
@@ -0,0 +1,14 @@
+target Python
+
+reactor A(bank_index=0, value=0) {
+  reaction(startup) {=
+    print("bank_index: {:d}, value: {:d}".format(self.bank_index, self.value))
+    if (self.value != 4 - self.bank_index):
+      sys.stderr.write("ERROR: Expected value to be 3 - bank_index.\n")
```
lhstrh, line 7: Off by one?
```suggestion
      sys.stderr.write("ERROR: Expected value to be 4 - bank_index.\n")
```

lhstrh (changes requested):
Looks good, modulo nitpick.

lhstrh (approved):
:+1:

PR #2356, opened 2024-07-08
Diagram support for watchdogs

edwardalee:
This PR adds diagram support for watchdogs. They are shown as squares with a "W".

### DIFF
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -218,6 +218,7 @@ Deadline:
     'deadline' '(' delay=Expression ')' code=Code;
 
 Watchdog:
+    (attributes+=Attribute)*
     'watchdog' name=ID '(' timeout=Expression ')'
     ('->' effects+=VarRefOrModeTransition (',' effects+=VarRefOrModeTransition)*)?
     code=Code;
--- a/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
+++ b/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
@@ -1269,8 +1269,7 @@ public class LinguaFrancaSynthesis extends AbstractDiagramSynthesis<Model> {
       _utilityExtensions.setID(node, watchdog.uniqueID());
       nodes.add(node);
       setLayoutOption(node, CoreOptions.PORT_CONSTRAINTS, PortConstraints.FIXED_SIDE);
-      Pair<KPort, KPort> ports =
-          _linguaFrancaShapeExtensions.addWatchdogFigureAndPorts(node);
+      Pair<KPort, KPort> ports = _linguaFrancaShapeExtensions.addWatchdogFigureAndPorts(node);
       if (watchdog.getTimeout() != null) {
         _kLabelExtensions.addOutsideBottomCenteredNodeLabel(
             node, String.format("timeout: %s", watchdog.getTimeout().toString()), 7);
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
```diff
@@ -1251,6 +1258,33 @@ private Collection<KNode> transformReactorNetwork(
...
+      }
+      // connect source
+      for (KPort source : watchdogSources.get(watchdog)) {
+        connect(createDelayEdge(watchdog), source, ports.getKey());
+      }
+
+      // connect targets
+      for (KPort target : watchdogDestinations.get(watchdog)) {
+        connect(createDelayEdge(watchdog), ports.getValue(), target);
+      }
```
soerendomroes, line 1285: You might want to enable layout annotation for watchdogs, as done [here](https://github.com/lf-lang/lingua-franca/blob/4bb000e4dc88988e40e681592a2fc16f89908a2b/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java#L1058) for timers.
The layout annotations should be applied after everything else that sets layout options finishes so after `addWatchdogFigureAndPorts` was called.

Adding a method to configure layout for watchdogs in LayoutPostProcessing, as done [here](https://github.com/lf-lang/lingua-franca/blob/4bb000e4dc88988e40e681592a2fc16f89908a2b/core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java#L1057) might not be necessary and I will add this once a new ELK release may add more interesting ordering related layout options.

core/src/main/java/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.java
```diff
@@ -839,53 +879,6 @@ public KText addTextButton(KContainerRendering container, String text) {
     return textToAdd;
   }
 
-  /** Creates the triangular line decorator with text. */
-  public KPolygon addActionDecorator(KPolyline line, String text) {
```
soerendomroes, line 843: I suppose this might no longer be necessary? If this is the case, please resolve this comment.

core/src/main/java/org/lflang/diagram/synthesis/LinguaFrancaSynthesis.java
edwardalee: Done in 521a50245

core/src/main/java/org/lflang/diagram/synthesis/styles/LinguaFrancaShapeExtensions.java
edwardalee: This was not called anywhere, so I deleted it.

soerendomroes (approved):

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -218,6 +218,7 @@ Deadline:
     'deadline' '(' delay=Expression ')' code=Code;
 
 Watchdog:
+    (attributes+=Attribute)*
```
lhstrh, line 221: I'm surprised to see a grammar change in a PR about diagram support. Why this change?

edwardalee: So that you can add annotations, like labels and layout annotations.

lhstrh: :+1:

lhstrh (commented):
Should this PR include an update to `reactor-c`?

soerendomroes:
The watchdog seems to be missing [here](https://github.com/lf-lang/lingua-franca/blob/4865086bf50862f82fe5d476471688cc5fbe59f1/core/src/main/java/org/lflang/AttributeUtils.java#L67-L86) to be able to get its attributes.

PR #2357, opened 2024-07-11
Fixed build script support in C++ docker generation

cmnrd:
This fix implements proper support for pre and post build scripts in the C++ targets and ensures that the src directory is available in the container

### DIFF
--- a/core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt
+++ b/core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt
@@ -96,6 +96,7 @@ class CppGenerator(
             )
             if (targetConfig.get(DockerProperty.INSTANCE).enabled) {
                 copySrcGenBaseDirIntoDockerDir()
+                FileUtil.copyDirectoryContents(context.fileConfig.srcPkgPath.resolve("src"), context.fileConfig.srcGenPath.resolve("src"), true)
                 buildUsingDocker()
             } else {
                 if (platformGenerator.doCompile(context)) {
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/lflang/generator/cpp/CppRos2Generator.kt
```diff
@@ -104,12 +105,27 @@ class CppRos2Generator(generator: CppGenerator) : CppPlatformGenerator(generator
...
             )
             return commands.map { argListToCommand(it) }
         }
+
+        override fun getPreBuildCommand(): MutableList<String> {
+            val script = context.targetConfig.get(DockerProperty.INSTANCE).preBuildScript
+            if (script.isNotEmpty()) {
+                return mutableListOf(". src/" + StringEscapeUtils.escapeXSI(script))
+            }
+            return mutableListOf(".", "/opt/ros/rolling/setup.sh")
```
cmnrd, line 119:
```suggestion
            return mutableListOf(". /opt/ros/rolling/setup.sh")
```

core/src/main/kotlin/org/lflang/generator/cpp/CppGenerator.kt
```diff
@@ -96,6 +96,7 @@ class CppGenerator(
             )
             if (targetConfig.get(DockerProperty.INSTANCE).enabled) {
                 copySrcGenBaseDirIntoDockerDir()
+                FileUtil.copyDirectoryContents(context.fileConfig.srcPkgPath.resolve("src"), context.fileConfig.srcGenPath.resolve("src"), true)
```
lhstrh, line 99: This line looks a bit sus. Why are we copying `src` into `src-gen/src`?

cmnrd: `context.fileConfig.srcGenPath` points to the directory that we put the Dockerfile in. `src` is copied for the same reason as we copy `src-gen`. In the C++ target's LF files may refer to files in the `src` directory, and hence it needs to be available in the docker image.

lhstrh (approved):

PR #2359, opened 2024-07-12
Effects made accessible in watchdog handlers

lhstrh:
The grammar allows effects to be specified for a watchdog, but the compiler wasn't resolving effects correctly and the code generator needed to be updated to ensure that effects are brought into scope. This PR also updates the diagram synthesis to render watchdog effects correctly.

### DIFF
--- a/core/src/main/java/org/lflang/diagram/synthesis/util/ReactorIcons.java
+++ b/core/src/main/java/org/lflang/diagram/synthesis/util/ReactorIcons.java
@@ -106,7 +106,7 @@ public class ReactorIcons extends AbstractSynthesisExtensions {
         //                if (error != null) {
         //                    var errorText = _kContainerRenderingExtensions.addText(rendering,
         // "Icon not found!\n"+error);
-        //                    _kRenderingExtensions.setForeground(errorText, Colors.RED);
+        //                    _kRenderingExtensions.setForeground(errorText, Colors.BLUE);
         //                    _kRenderingExtensions.setFontBold(errorText, true);
         //                    _kRenderingExtensions.setSurroundingSpaceGrid(errorText, 8, 0);
         //                }
--- a/core/src/main/java/org/lflang/diagram/synthesis/util/ReactorIcons.java
+++ b/core/src/main/java/org/lflang/diagram/synthesis/util/ReactorIcons.java
@@ -106,7 +106,7 @@ public class ReactorIcons extends AbstractSynthesisExtensions {
         //                if (error != null) {
         //                    var errorText = _kContainerRenderingExtensions.addText(rendering,
         // "Icon not found!\n"+error);
-        //                    _kRenderingExtensions.setForeground(errorText, Colors.BLUE);
+        //                    _kRenderingExtensions.setForeground(errorText, Colors.RED);
         //                    _kRenderingExtensions.setFontBold(errorText, true);
         //                    _kRenderingExtensions.setSurroundingSpaceGrid(errorText, 8, 0);
         //                }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/lflang/generator/c/CWatchdogGenerator.java
```diff
@@ -9,16 +9,12 @@
...
 import org.lflang.ast.ASTUtils;
 import org.lflang.generator.CodeBuilder;
 import org.lflang.generator.ReactorInstance;
-import org.lflang.lf.Mode;
-import org.lflang.lf.ModeTransition;
-import org.lflang.lf.Reactor;
-import org.lflang.lf.VarRef;
-import org.lflang.lf.Variable;
-import org.lflang.lf.Watchdog;
+import org.lflang.lf.*;
```
lhstrh, line 17: The style guide does not allow wildcard imports.

test/C/src/concurrent/WatchdogAction.lf
```diff
@@ -0,0 +1,74 @@
...
+ * watchdog expirations. This version uses an action instead of a reaction to the watchdog.
+ * @author Benjamin Asch
+ * @author Edward A. Lee
+ */
+target C {
+  timeout: 11000 ms
+}
+
+reactor Watcher(timeout: time = 1500 ms) {
+  // Offset ameliorates startup time.
```
lhstrh, line 13: What startup time is this referring to?

core/src/main/java/org/lflang/generator/c/CWatchdogGenerator.java
edwardalee: Intellij does this automatically.  Here are the places where this appears, most of which have nothing to do with this PR:

image

test/C/src/concurrent/WatchdogAction.lf
edwardalee: The time it takes the program to start up.

edwardalee:
I've pushed support here for the diagrams, so that when a watchdog declares an action as an effect, we get this:

image

test/C/src/concurrent/WatchdogAction.lf
lhstrh: I think "ameliorates" is a bit odd in this sentence. Did you mean "amortizes"?

core/src/main/java/org/lflang/generator/c/CWatchdogGenerator.java
lhstrh: It must have slipped past reviews before, but that doesn't change the fact that the style guide prohibits it... Doesn't IntelliJ ask for confirmation before it carries out such automated refactoring?

core/src/main/java/org/lflang/diagram/synthesis/util/ReactorIcons.java
```diff
@@ -106,7 +106,7 @@ public void handleIcon(KContainerRendering rendering, ReactorDecl reactor, boole
         //                if (error != null) {
         //                    var errorText = _kContainerRenderingExtensions.addText(rendering,
         // "Icon not found!\n"+error);
-        //                    _kRenderingExtensions.setForeground(errorText, Colors.RED);
+        //                    _kRenderingExtensions.setForeground(errorText, Colors.BLUE);
```
lhstrh, line 109: Seems like an accidental change?

edwardalee: Oops. Reversed with 938bed159.

test/C/src/concurrent/WatchdogAction.lf
edwardalee: Replaced with 
```
  // Offset may reduce the likelihood of flakiness if long startup times occur.
```

edwardalee (approved):
I think this is ready to merge.  Watchdogs still do not work with modes, but I've opened issue #2360.

PR #2369, opened 2024-07-16
Ability to use of target code expressions for time values in C++

cmnrd:


test/Cpp/src/target/TimeExpression.lf
```diff
@@ -0,0 +1,53 @@
...
+=}
+
+reactor Foo(foo: time = {= get_delay() =}) {
+  timer t(0, foo)
+
+  state timer_triggered: bool = false
+
+  reaction(t) {=
+    timer_triggered = true;
+    if(get_elapsed_logical_time() % 200ms != reactor::Duration::zero()) {
```
lhstrh, line 16: Probably need to run the formatter.

lhstrh:
```suggestion
    if (get_elapsed_logical_time() % 200ms != reactor::Duration::zero()) {
```

test/Cpp/src/target/TimeExpression.lf
```diff
@@ -0,0 +1,53 @@
...
+main reactor {
+  foo = new Foo()
+
+  state timer_triggered: bool = false
+
+  timer t(0, {= get_delay() + 200ms =})
+
+  reaction(t) {=
+    timer_triggered = true;
+    if(get_elapsed_logical_time() % 400ms != reactor::Duration::zero()) {
```
cmnrd, line 39:
```suggestion
    if (get_elapsed_logical_time() % 400ms != reactor::Duration::zero()) {
```

edwardalee:
Why was this change made only for the Cpp target?
Seems that all targets could benefit from this.

cmnrd:
Because, to me, it is unclear if this would work reliably in other targets. It is easy enough to enable it, but afaik the C based targets rely on the assumption that time values are resolvable statically.

PR #2370, opened 2024-07-16
Changes in the Lingua Franca Language Server to support improvements in the VSCode extension

vinzbarbuto:
This code includes the extensions made to the Lingua Franca Language Server to support the latest improvements for the [VSCode extension](https://github.com/lf-lang/vscode-lingua-franca/pull/156), specifically those related to the [Lingua Franca Package Explorer](https://github.com/lf-lang/vscode-lingua-franca/blob/extending/LF_PACKAGE_EXPLORER.md).


lhstrh:
Also in this branch, conflicts remain.

soerendomroes (approved):
Maybe one could rename the `Tree` classes to something more speaking since this is a very specific tree, but otherwise it looks good to me.

vinzbarbuto:
> Maybe one could rename the `Tree` classes to something more speaking since this is a very specific tree, but otherwise it looks good to me.

Perhaps I can rename it as `LibraryFileTree ` while nodes can be `LibraryFileTreeNode` instead of `TreeNode`

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
```diff
@@ -84,6 +100,82 @@ public CompletableFuture<String> build(BuildArgs args) {
...
+            // Return a list of reactors within the file at the specific uri
+            return parseLibraryReactors(uri);
+          } catch (Exception e) {
+            return null;
+          }
+        });
+  }
+
+  /**
+   * Retrieves the target position specified in the LF program file at the given path.
```
edwardalee, line 126:
```suggestion
   * Retrieve the target position specified in the LF program file at the given path.
```

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
```diff
@@ -84,6 +100,82 @@ public CompletableFuture<String> build(BuildArgs args) {
...
+            targetPosition = new NodePosition(node.getStartLine(), node.getEndLine());
+            return targetPosition;
+          } catch (Exception e) {
+            return null;
+          }
+        });
+  }
+
+  /**
+   * Parses a library of reactors specified by the provided URI and constructs a hierarchical
```
edwardalee, line 153:
```suggestion
   * Parse a library of reactors specified by the provided URI and construct a hierarchical
```

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
```diff
@@ -84,6 +100,82 @@ public CompletableFuture<String> build(BuildArgs args) {
...
+      Stream<Reactor> reactors =
+          m.getReactors().stream().filter(r -> r.getName() != null && !r.getName().isEmpty());
+      reactors.forEach(
+          r -> {
+            INode node = NodeModelUtils.getNode(r);
+            NodePosition nodePosition = new NodePosition(node.getStartLine(), node.getEndLine());
+            res.addChild(new LibraryFileTreeNode(r.getName(), res.getUri(), nodePosition));
+          });
+    } catch (Exception e) {
+      return null;
```
edwardalee, line 174: Should this print a warning?  I'm not sure how VS code would handle this output, but it seems possibly better than silently failing.

lsp/src/main/java/org/lflang/diagram/lsp/NodePosition.java
```diff
@@ -0,0 +1,36 @@
+package org.lflang.diagram.lsp;
+
+/** Represents the position of a node within a text document or code file. */
```
edwardalee, line 3:
```suggestion
/**
  * @brief Position of a node within a text document or code file.
  *
  * A position is a pair of integers representing the starting and ending line numbers.
  */
```

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTreeNode.java
```diff
@@ -0,0 +1,51 @@
+package org.lflang.diagram.lsp;
+
+/**
+ * Represents a node in a tree structure, typically used to store information about elements in a
+ * hierarchical context.
+ */
```
edwardalee, line 6:
```suggestion
/**
 * @brief A node in a tree of items, each with a label, a URI, and a position within the URI.
 * @see LibraryFileTree
 */
```

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTree.java
edwardalee, line 1: This class is misnamed. It's not actually a tree.  It's just a list. There is no hierarchy.  I think it would be better if this were actually a tree.  One way to do that would be would be if the `children` element was a list of `LibraryFileTree` rather than `LibraryFileTreeNode`.  This would require adding a `position` private field.  For leaf nodes, the `children` list would be empty (or null).  For non-leaf nodes, the `position` element would be null.

edwardalee (dismissed):
This looks close.  Some suggestions included. We should merge with master and convert to a regular PR rather than a draft.

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTree.java
vinzbarbuto: The hierarchy I envision is as follows: A `LibraryFileTree` object represents an LF file, such as `Input.lf`. It contains information about the file's name, its URI, and a list of children of type `LibraryFileTreeNode`. The latter represents the reactors included in the file. Note that a `LibraryFileTreeNode` does not contain other “children”; it only includes the information needed to display details about the reactors in the extension. Therefore, I used different objects (`LibraryFileTree` for the file and `LibraryFileTreeNode` for the reactors) to clearly distinguish their roles.

The result will be a `LibraryFileTree` object for each library file. Each of these objects will contain a list of `LibraryFileTreeNode` objects, which represent the reactors included in that file.

```mermaid
---
title: Overview
---
classDiagram
    LibraryFileTree --o "0..n" LibraryFileTreeNode : childrens
    LibraryFileTreeNode -- "1" NodePosition : position
    class LibraryFileTree{
        - label
        - uri
    }
    class LibraryFileTreeNode{
        - label
        - uri
    }
   class NodePosition{
        - start
        - end
    }

```

I'm not a fan of collapsing both objects into one, as this could lead to an increased number of checks on the frontend to avoid null fields (such as `position` in the case of `LibraryFileTree` and `children` in the case of `LibraryFileTreeNode`). 

A possible solution to this misnaming issue could be to rename `LibraryFileTree` to `LibraryFileRoot`, as it actually acts as the root of a data structure, and to change `LibraryFileTreeNode` to `LibraryFileNode`. Would this solve the problem?

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
vinzbarbuto: You're right, I could add a warning message and display it as a popup in VSCode. This should work.

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTree.java
edwardalee: Renaming seems necessary to me since this class is not a tree structure.  It is actually a list.  In fact, it could be designed as a subclass of ArrayList that just adds a label and uri.  It would then inherit all the API of ArrayList, which could be useful.

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
soerendomroes: Have you considered not having a pop-up and logging the warning instead?
https://code.visualstudio.com/api/ux-guidelines/notifications

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTree.java
vinzbarbuto: So your idea is to combine both objects into a single class that extends `ArrayList`? However, I still don't see it as a list since the root object has children. It might not be a tree hierarchy, but there is still a hierarchical relationship between the root (LF file) and the nodes (Reactors)

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
vinzbarbuto: Thank you, @soerendomroes, I'll check it out

lsp/src/main/java/org/lflang/diagram/lsp/LibraryFileTree.java
edwardalee: A List is a container for an ordered set of objects.  Isn’t that what this is?

lsp/src/main/java/org/lflang/diagram/lsp/LFLanguageServerExtension.java
vinzbarbuto: @soerendomroes I've handled this situation in the following way:

1. I send the request from the client (VSCode extension).
2. I receive and process the request in the Language Server.
3. If an error occurs, I'll return `null` from the Language Server.
4. If the object request received by the client is `null`, I show an error message by executing the following code:

    ```typescript
    vscode.window.showErrorMessage('Error retrieving data from the Language Server', 
    ...['Try again', 'Cancel']).then(selection => {
        if (selection === 'Try again') {
            this.refreshTree(); // try again
        }
    });
    ```

Does this sound like a good solution?

vinzbarbuto:
## To-Do
- [ ] Rename the `LibraryFileTree` object and modify it to extend the `List<LibraryFileNode>` class, as it functions more like a list data structure rather than a tree.

lhstrh (approved):

PR #2371, opened 2024-07-16
Docker compose override

Depetrol:
Added `docker-config-file` option to docker target options. This allows the user to provide a custom docker-compose file that will add and override the generated docker-compose file using [this feature](https://docs.docker.com/reference/cli/docker/compose/) provided by docker.

The design of the `docker-config-file` is similar to `env-file`. The user specifies a `docker-config-file` parameter in the docker target as the path to a custom docker compose yaml file. 
```
docker: {
    docker-config-file: "./DockerComposeConfig.docker.yml"
}
```
The user-specified yaml file will be copied as `docker-compose-override.yml` under `src-gen` folder. If `docker-compose-override.yml` exists under the `src-gen` folder, the script under `bin` will use `docker-compose.yml` and `docker-compose-override.yml` to start the docker containers: `docker compose -f docker-compose.yml -f docker-compose-override.yml up --abort-on-container-failure
`.
The specified yaml file should be the same structure as a normal `docker-compose.yml` file that has each top level federate as a service. To accommodate this syntax, the `federate__` prefix has been removed from the generated `.lf` files for each federate in `src` folder.


core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -100,9 +100,9 @@ public FederateInstance(
     // If the instantiation is in a bank, then we have to append
     // the bank index to the name.
     if (instantiation.getWidthSpec() != null) {
-      this.name = "federate__" + instantiation.getName() + "__" + bankIndex;
+      this.name = instantiation.getName() + "__" + bankIndex;
```
lhstrh, line 103: This should be fine because instance names are already guaranteed to be unique within a reactor (the top level, in this case).

core/src/main/java/org/lflang/target/property/DockerProperty.java
```diff
@@ -76,7 +79,8 @@ public DockerOptions fromAst(Element node, MessageReporter reporter) {
         preBuildScript,
         postBuildScript,
         runScript,
-        envFile);
+        envFile,
+        dockerConfigFIle);
```
lhstrh, line 83:
```suggestion
        dockerConfigFile);
```

core/src/main/java/org/lflang/target/property/DockerProperty.java
```diff
@@ -43,6 +43,7 @@ public DockerOptions fromAst(Element node, MessageReporter reporter) {
     var postBuildScript = "";
     var runScript = "";
     var envFile = "";
+    var dockerConfigFIle = "";
```
lhstrh, line 46:
```suggestion
    var dockerConfigFile = "";
```

core/src/main/java/org/lflang/target/property/DockerProperty.java
```diff
@@ -168,7 +174,8 @@ public enum DockerOption implements DictionaryElement {
     RTI_IMAGE("rti-image", PrimitiveType.STRING),
     PRE_BUILD_SCRIPT("pre-build-script", PrimitiveType.STRING),
     PRE_RUN_SCRIPT("pre-run-script", PrimitiveType.STRING),
-    POST_BUILD_SCRIPT("post-build-script", PrimitiveType.STRING);
+    POST_BUILD_SCRIPT("post-build-script", PrimitiveType.STRING),
+    DOCKER_CONFIG_FILE("docker-config-file", PrimitiveType.STRING);
```
lhstrh, line 178: Why not call it `docker-compose-override`? I think that's more specific.
```suggestion
    DOCKER_CONFIG_FILE("docker-compose-override", PrimitiveType.STRING);
```

lhstrh: Could even call it `compose-override` because it's already in the `docker` dictionary...

Depetrol: Makes sense. Should I change the key to "compose-override-file" then? Or perhaps just "compose-file" might suffice too because from the user perspective this is nearly identical to writing a docker-compose file for the reactors.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
Depetrol: This seems to be breaking some tests in the C target because some `.lft` files have a "federate__" prefix. Alternative could be when copying the user specified `yml` file, add the prefix `federate__` to each of the service names. However this would introduce another dependency `snakeyaml`.

core/src/main/java/org/lflang/target/property/DockerProperty.java
lhstrh: Not sure. Let me tag @cmnrd and @elgeeko1 to ask for their opinion.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
lhstrh: I'm surprised that a change to service names for Docker Compose would break tracing. Isn't it just a test that needs to be updated? Tagging @edwardalee as he might be able to comment on this.

Depetrol: There seems to be a test trace with name `a_0.lft` and another trace with `federate__a_0.lft` so simply removing the "federate__" from test files would not work. This change also changes the names of the generated `.lf` files for each federate, so this might cause the generated trace to have different names too. I think adding a prefix to each service in the `yml` file when copying should be a relatively simpler fix.

lhstrh: Why not not just change `TargetConfigTests.java:87` to:
```
URI.createURI(tempDir.resolve("src/Federation.lf").toUri().toString()),
```

Depetrol: I added this change and all tests have passed!

Depetrol:
CI is failing because of #2376

elgeeko1:
Can you explain more about this statement?

> To accommodate this syntax, the federate__ prefix has been removed from the generated .lf files for each federate in src folder.

There are other places where we depend on the `federate__` prefix, such as our use of k3s or similar that expect the code generator to prefix the federate names. Can we accomplish this without this change?

Depetrol:
There were some previous [discussion](https://github.com/lf-lang/lingua-franca/pull/2371#discussion_r1680096836) about this --

> This seems to be breaking some tests in the C target because some .lft files have a "federate__" prefix. Alternative could be when copying the user specified yml file, add the prefix federate__ to each of the service names. However this would introduce another dependency snakeyaml.

The C tests has been fixed. But if there are other places that depend on the prefix, this change can be accomplished without removing the `federate__ ` prefix.

lhstrh:
> There were some previous [discussion](https://github.com/lf-lang/lingua-franca/pull/2371#discussion_r1680096836) about this --
> 
> 
> 
> > This seems to be breaking some tests in the C target because some .lft files have a "federate__" prefix. Alternative could be when copying the user specified yml file, add the prefix federate__ to each of the service names. However this would introduce another dependency snakeyaml.
> 
> 
> 
> The C tests has been fixed. But if there are other places that depend on the prefix, this change can be accomplished without removing the `federate__ ` prefix. 

I don't think we should have the prefix. Is there any reason anyone can cite for needing it?

Depetrol:
I have no problem with removing the prefix, both implementations are fine on my side. @elgeeko1 said he has some places that uses the `federate__` prefix.

> There are other places where we depend on the federate__ prefix, such as our use of k3s or similar that expect the code generator to prefix the federate names. Can we accomplish this without this change?

test/Python/src/docker/DockerComposeConfig.docker.yml
elgeeko1, line 1: Do we need the huggingface references, or the shm size to be set for this example? Doesn't look like they would be needed, so a simpler more generic override may be more useful.

elgeeko1:
> To accommodate this syntax, the federate__ prefix has been removed from the generated .lf files for each federate in src folder.

It's not clear from your comment where this prefix is being removed. Which files are you proposing to change?

For better or worse, the LF code generator uses `federate__` as a prefix for the binary names, folder names, and docker compose services names for all files it generates. Consequently there are downstream users who expect this.

Why is this change necessary? Does this mean that the names of the services would be different depending on whether I used an override file or not? If so then it's problematic that my service names would change between the two use cases.

test/Python/src/docker/DockerComposeConfig.docker.yml
Depetrol: It's just an example of the docker compose override file. the references and shm size are just examples of configurations.

Depetrol:
I just had a zoom call with @elgeeko1 and he has projects that depend on the `federate__` prefix, and removing the `federate__` prefix would result in breaking them. He also suggested that adding the `federate__` prefix in the code generator could cause problems by being inconsistent between the user provided service name and the actual running service names. The conclusion was to add the `federate__` prefix in the user provided override file, and the file would be copies as `docker-compose-override.yml` in `src-gen`.

elgeeko1:
Shulu and I had a chance to chat, here's my understanding:

A docker compose override file overrides the behavior of the parent docker compose file by overriding properties of a given service. The service is identified by its service name as defined in the parent docker compose file. The proposal currently in this PR is to have the user-provided docker compose override file reference service names by their LF filename and hence absent the `federate__` prefix, which may be more user-friendly. LF compiler will then parse the user provided override file, parse the service names, and produce a new override file in `fed-gen` with the `federate__` prefix so that the service names match the docker compose file currently generated by LF.

My concerns with this are:
- This introduces a new feature -- it's a small one, but it is a feature, that we use friendlier names in our override file, and have code that parses this file and generates a modified version into `fed-gen`. This would need to be tested and maintained. It adds a new error pathway, for example, if the YAML parsing fails (which will otherwise happen when the docker compose build is run).
- It is confusing to the user that the names of the services in the override file do not match what actually gets created in the containers. The service in my override file `myreactor` should override the `myreactor` service in the original docker compose file, and the container name and hostname of the container will default to `myreactor`. However, in this case, the service name `myreactor` in the override file actually overrides a service `federate__myreactor` and its container name and hostname will be `federate__myreactor`, which is confusing and disagrees with the docker specification.
- An alternate solution is to modify the generated docker compose file in LF to omit `federate__` but this will likely cause breaking changes elsewhere.

I agree that the service names without `federate__` are easier to read and write, and this fits with many other usability improvements that could be made. For now, I think we should stick with the simplest solution that is consistent with the docker standard and the docker compose file currently generated by `lfc`.

That's my only concern with this PR. Outside of this concern, this is a very helpful feature that I will plan to use!

lhstrh:
I'm happy as long as we don't introduce an extra parsing step for the same reasons @elgeeko1  already pointed out.

Depetrol:
I added a command to build RTI docker image in the CI testing environment. Otherwise a docker RTI image version mismatch would cause the test to fail. I think we can merge this then?

lhstrh:
> I added a command to build RTI docker image in the CI testing environment. Otherwise a docker RTI image version mismatch would cause the test to fail. I think we can merge this then?

Wouldn't it suffice to use the option `docker: {rti-image: "rti:local"}`? See docs [here](https://www.lf-lang.org/docs/reference/docker-support/#option-rti-image).

Depetrol:
I think using `docker: {rti-image: "rti:local"}` in a test is undesirable -- we would have to add `docker: {rti-image: "rti:local"}` to all docker related test files. Changing the RTI image in the testing environment is more general and allows the test to better simulate user behavior.

lhstrh:
> I think using `docker: {rti-image: "rti:local"}` in a test is undesirable -- we would have to add `docker: {rti-image: "rti:local"}` to all docker related test files. Changing the RTI image in the testing environment is more general and allows the test to better simulate user behavior.

We already have...

Depetrol:
I'll use `docker: {rti-image: "rti:local"}` then.

lhstrh (approved):
This looks good to go. One last thing: please also update the website to describe the feature.

Depetrol:
I created a PR https://github.com/lf-lang/lf-lang.github.io/pull/281 for the updated documentation on docker compose override. (Also could you merge this current PR? I can see that it's been approved but I don't have merge access)

elgeeko1:
Great work!

PR #2372, opened 2024-07-17
Do not require libexecinfo in C++ docker images

cmnrd:
Alpine has dropped support for libexecinfo since version 3.17. This PR pulls in https://github.com/lf-lang/reactor-cpp/pull/59 which makes the compilation of reactor-cpp more robust, so that it can handle if backtrace support is not available or provided by an alternative implementation. Consequently, we can remove the current workarounds in the LF code generator for docker builds.


petervdonovan (approved):
Thanks!

elgeeko1 (approved):

PR #2375, opened 2024-07-19
Custom Serialization in Python Target

Depetrol:
This PR enables a new feature for using arbitrary python package names that follows a specific API as the connection serializer in Python target federated execution. 

Syntax:
```
<source_port_reference> -> <destination_port_reference> after <time_value> serializer "<serializer_name>"
```

This feature is backwards compatible. If `<serializer_name>` is any of `[NATIVE, ROS2, PROTO]`, it will use these serialization methods instead of finding a python package. 

The custom Python package would have to export a `Serializer` class that has the following API:
```python
class Serializer():
    def serialize(self, obj)->bytes:
        return <bytes_object>
    def deserialize(self, message:bytes):
        return <deserialized_object>
```
The `Serializer` class will be initialized at the first send and first receive on both ends of the connection. It is persisted in each `networksender` and `networkreceiver` as a state variable. Python objects are serialized with `serialize()`, sent over the socket and deserialized with `deserialize()`.

This feature allows for user-defined custom serialization for any strategy. For example, the user could define a serialization method that enables fast inter-process communication with shared memory. Potentially ROS2 and PROTO serialization could be implemented with custom python packages too. It also supports different serialization methods for different connections.


Depetrol:
CI is failing because of #2376

core/src/main/java/org/lflang/federated/generator/FedUtils.java
```diff
@@ -14,7 +14,17 @@ public static SupportedSerializers getSerializer(
...
+      for (SupportedSerializers method : SupportedSerializers.values()) {
+        if (method.name().equalsIgnoreCase(connection.getSerializer().getType())) {
+          serializer =
+              SupportedSerializers.valueOf(connection.getSerializer().getType().toUpperCase());
+          isCustomSerializer = false;
+        }
+        if (isCustomSerializer) {
+          serializer = SupportedSerializers.fromCustomString(connection.getSerializer().getType());
+        }
+      }
```
edwardalee, line 27: The logic here doesn't look quite right.  If `connection.getSerializer()` returns something that doesn't match the first `method`, then you assume it is custom.  But shouldn't it not match any `method`?  Also, what if there is no serializer specified?  Does `connection.getSerializer()` return null? If so, this will give a null-pointer exception.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
```diff
@@ -0,0 +1,165 @@
...
+  String customSerializerPackage;
+
+  public FedCustomPythonSerialization(String customSerializerPackage) {
+    this.customSerializerPackage = customSerializerPackage;
+  }
+
+  @Override
+  public boolean isCompatible(GeneratorBase generator) {
+    if (generator.getTarget() != Target.Python) {
+      throw new UnsupportedOperationException("This class only support Python serialization.");
```
edwardalee, line 48:
```suggestion
      throw new UnsupportedOperationException("The FedCustomPythonSerialization class only supports the Python target.");
```

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
```diff
@@ -0,0 +1,165 @@
...
+        + "if (SerializerClass == NULL) lf_print_error_and_exit(\"Could not find class 'Serializer' in module '%s'.\");\n"
+            .formatted(customSerializerPackage)
+        + "Py_DECREF(pModule);\n"
+        + "self->custom_serializer = PyObject_CallObject(SerializerClass, NULL);\n"
+        + "Py_DECREF(SerializerClass);\n"
+        + "if (self->custom_serializer == NULL) lf_print_error_and_exit(\"Could not instantiate class 'Serializer' in module '%s'.\");\n"
+            .formatted(customSerializerPackage)
+        + "lf_print_log(\"Loaded custom serializer module %s.\\n\");\n"
+            .formatted(customSerializerPackage)
+        + "}\n";
```
edwardalee, line 82: Perhaps it would be better to provide this as a utility function in one of the .h/.c files, e.g. in the `Python Helper Functions` section of `pythontarget.h/c`?  Also, are you sure about the `Py_DECREF(pModule)` and `Py_DECREF(SerializerClass)`?  Is it possible that garbage collection will free these and make the module unusable?  I am not at all sure about that...

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
```diff
@@ -0,0 +1,165 @@
...
+        .append(".buf, &")
+        .append(serializedVarName)
+        .append(".len);\n");
+    serializerCode.append("Py_XDECREF(serializer_serialize);\n");
+    serializerCode.append("Py_XDECREF(args);\n");
+    // Error check
+    serializerCode.append("if (returnValue == -1) {\n");
+    serializerCode.append("    if (PyErr_Occurred()) PyErr_Print();\n");
+    serializerCode.append(
+        "    lf_print_error_and_exit(\"Could not serialize " + serializedVarName + ".\");\n");
```
edwardalee, line 118: Again, this looks like an opportunity to define a utility C function.  The code generator could then just a generate a call to that function.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
```diff
@@ -0,0 +1,165 @@
...
+    deserializerCode.append("if (" + deserializedVarName + " == NULL) {\n");
+    deserializerCode.append("    if (PyErr_Occurred()) PyErr_Print();\n");
+    deserializerCode.append(
+        "    lf_print_error_and_exit(\"Could not deserialize " + deserializedVarName + ".\");\n");
+    deserializerCode.append("}\n");
+
+    // Decrement the reference count
+    deserializerCode.append("Py_XDECREF(serializer_deserialize);\n");
+    deserializerCode.append("Py_XDECREF(message_byte_array);\n");
+    deserializerCode.append("Py_XDECREF(args);\n");
```
edwardalee, line 152: Again, I think this should use a helper C function.

test/Python/src/serialization/CustomSerializer.lf
edwardalee, line 1: Shouldn't this test also check the received values to make sure the serialization and deserialization is actually working?

edwardalee (changes requested):
This looks like a great start!  I've made some suggestions, however.

core/src/main/java/org/lflang/federated/generator/FedUtils.java
Depetrol: Ah I made a mistake... The `if (isCustomSerializer)` should be outside the `for` loop. If there is no serializer specified, it will default to `NATIVE` because there is a null check before line 17 in the original code.
```java
SupportedSerializers serializer = SupportedSerializers.NATIVE;
    if (connection.getSerializer() != null) {
      boolean isCustomSerializer = true;
      for (SupportedSerializers method : SupportedSerializers.values()) {
        if (method.name().equalsIgnoreCase(connection.getSerializer().getType())) {
          serializer =
              SupportedSerializers.valueOf(connection.getSerializer().getType().toUpperCase());
          isCustomSerializer = false;
        }
      }
      if (isCustomSerializer) {
        serializer = SupportedSerializers.fromCustomString(connection.getSerializer().getType());
      }
    }
```

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
Depetrol: Yes that makes sense. I'll move create utility functions in `pythontarget.h/c`. `Py_DECREF()` should be save because `PyImport_Import`, `PyObject_GetAttrString`, `PyObject_CallObject` all increase the reference count so the reference count will not be zero after `Py_DECREF`, hence objects are not garbage collected. However since we don't ever want them to be garbage collected until program termination, it might be better to not do `Py_DECREF`. I'll remove this.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
Depetrol: Yes this is a great idea. It simplifies the code a lot and increases readability.

core/src/main/java/org/lflang/federated/generator/FedUtils.java
edwardalee: OK, this looks better.  I suggest adding a `break` statement after `isCustomSerializer = false`.

test/Python/src/serialization/CustomSerializer.lf
Depetrol: I added the test to check that the received value matches the intended value, and `exit(1)` if otherwise.

core/src/main/java/org/lflang/federated/generator/FedUtils.java
Depetrol: Added!

cmnrd (dismissed):

.github/workflows/py-tests.yml
```diff
@@ -45,6 +45,8 @@ jobs:
         run: pip3 install LinguaFrancaBase
       - name: Install Google API Python Client
         run: pip3 install --upgrade google-api-python-client
+      - name: Install custom serializer test
+        run: pip3 install -e ./test/Python/src/serialization/pickle_serializer
```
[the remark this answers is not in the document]
Depetrol: Yes I can move the install command into the preamble. I'll use `lf.source_directory()` to find the package location. Though it shouldn't be too hard to run test locally, just run the pip install command before running the tests.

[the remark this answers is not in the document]
Depetrol: Because of #2378, installing the custom serializer in the preamble is not feasible with a relative path. If we install the package with absolute path, it would be even harder to run the test locally, because the absolute path is sure to be different in the local environment. 

I think the solution here is to install the custom package first. To run the test locally, run `pip3 install -e ./test/Python/src/serialization/pickle_serializer` in the project root directory. I'll add a comment in the test file.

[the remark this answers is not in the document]
cmnrd: Why does the package need to be installed? Can we not simply put the external Python file into the `files` property and then use a relative import to import the module?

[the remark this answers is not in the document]
Depetrol: I moved the install to the preamble so that the tests are self-contained.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public StringBuilder generateNetworkSerializerCode(String varName, String originalType) {
+    StringBuilder serializerCode = new StringBuilder();
+    // Initialize self->custom_serializer if null
+    serializerCode.append(this.initializeCustomSerializer());
+    // Define the serialized PyObject
+    serializerCode
+        .append(
```
lhstrh, line 76: This would look better with `"""`-quoted block strings.

Depetrol: Indeed -- I've updated the code to use quoted block strings.

cmnrd:
I still cannot say that I like installing something into the user's system as a side effect of the preamble in a test file. I also think that requiring the serializer to be a standalone Python package that is installed separately is not a practical design from a usability perspective. However, I have made those points before and will not continue arguing for them. I cannot give this my approval, but if someone else does, I will not block it.

Depetrol:
I agree that installing something into the user's system as a side effect of the preamble in a test file is not a perfect solution, but it doesn't require developers to install it manually in order to run the test, and makes the test self-contained. 

However I would argue that installing a standalone Python package is more user friendly compared with using relative imports. There are several reasons:

- There are not many ways to implement a serializer, and most likely the user will use a written custom serializer provided by others(most likely perhaps by us). In this case, installing it as a custom python package is the easiest way for the package to be distributed.
- For a generic serializer, it will likely be used multiple times in multiple locations. It would be a hassle for the users to import it every time that it is used.
- If the serializer has to be imported every time that it is used, it would be almost equivalent to provide a "transparent" serializer that assumes the object being sent over the connection is a bytes object, and send the bytes object as-is. The user could do their serialization before passing the object to the "transparent" serializer. For example, `out_port.set(custom_serialize_func(<object>))`.
- If the user references the serializer using relative import without installing it in the system environment, this will break because the generated code is in a different location as the `.lf` code where it is written. The user would also have to specify a `files` parameter every time the serializer is used to copy the serializer into the generated directory.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
edwardalee: I don't see my suggested change in the code.  The reason for the suggestion:
* "This class" is a meaningless message to a user.  What class?
* Grammatical error.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
edwardalee: Again, I don't see this suggestion addressed in the PR... ???

edwardalee (changes requested):
Maybe you haven't pushed the changes that resolve the issues?

lhstrh:
@edwardalee sorry I re-requested a review but I hadn't noticed that you just provided feedback already

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
Depetrol: I think this is perhaps because the changes I made are not reflected in this code snippet, because the code snippet here is referencing the code at the time you made the suggestion. There is an "Outdated" tag on the top right of the code snippet indicating this. The 'Files changed' tab on the top shows the total changes of the PR, I believe you can see the changes I made in there.

core/src/main/java/org/lflang/federated/serialization/FedCustomPythonSerialization.java
Depetrol: I agree, this error message is an oversight on my side. I have made your suggested change [here](https://github.com/lf-lang/lingua-franca/pull/2375/files).

edwardalee (approved):
This looks good to me.  It depends on [this PR](https://github.com/lf-lang/reactor-c/pull/471)

PR #2384, opened 2024-08-01
RTI Docker Hub Continuous Deployment with Multiplatform Support

Depetrol:
This PR partially resolves https://github.com/lf-lang/reactor-c/issues/470 by continuously deploying RTI images on Docker Hub with `nightly` and `master` tag.

- RTI image with `nightly` tag is deployed to Docker Hub along publishing the nightly release.
- RTI image with `master` tag is deployed to Docker Hub when a new commit is pushed to the `master` branch. This happens when a PR is merged.

To facilitate the continuous deployment process, two GitHub secrets need to be added:

- DOCKERHUB_USERNAME: Docker Hub username, in this case it should be an account with push access to `lflang/rti`
- DOCKERHUB_TOKEN: Corresponding [Docker Hub Access Token](https://docs.docker.com/security/for-developers/access-tokens/).

The RTI image is built for the following platforms:

- linux/amd64
- linux/arm64
- linux/arm/v7
- linux/riscv64

### DIFF
--- a/.github/workflows/rti-docker.yml
+++ b/.github/workflows/rti-docker.yml
@@ -1,0 +1,23 @@
+name: Push RTI Image to Docker Hub
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    name: Build and push RTI to Docker Hub
+    steps:
+      - name: Check out lingua-franca repository
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          fetch-depth: 0
+      - name: Build and push RTI to Docker Hub
+        uses: ./.github/actions/push-rti-docker
+        with:
+          tag: master
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/rti-docker.yml
+++ b/.github/workflows/rti-docker.yml
@@ -15,9 +15,16 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+      - name: Check Lingua Franca Version
+        id: check_version
+        run: |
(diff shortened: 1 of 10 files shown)


Depetrol:
Reopening with build for multiple platforms.

.github/workflows/rti-docker.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Check out lingua-franca repository
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          fetch-depth: 0
+      - name: Check Lingua Franca Version
+        id: check_version
+        run: |
+          export LF_VERSION=$(cat core/src/main/resources/org/lflang/StringsBundle.properties | sed 's/VERSION = \([0-9]*\.[0-9]*\.[0-9]*\)-SNAPSHOT/\1/')
```
lhstrh, line 21: I think it might make sense to keep a suffix that indicates that it's a snapshot version and not a release version.

Depetrol: Yes. I changed the version to `$(cat core/src/main/resources/org/lflang/StringsBundle.properties | sed -n 's/.*VERSION = \(.*\)/\1/p' | tr '[:upper:]' '[:lower:]')`

.github/workflows/rti-docker.yml
```diff
@@ -0,0 +1,30 @@
...
+  build-and-push:
+    runs-on: ubuntu-latest
+    name: Build and push RTI to Docker Hub
+    steps:
+      - name: Check out lingua-franca repository
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          fetch-depth: 0
+      - name: Check Lingua Franca Version
```
lhstrh, line 18:
```suggestion
      - name: Look up the current version and export as environment variable
```

.github/workflows/rti-docker.yml
```diff
@@ -0,0 +1,30 @@
...
+        id: check_version
```
lhstrh, line 19: Not referenced, so this line can be safely removed.
```suggestion
```

lhstrh (approved):
Looks great! Left some minor suggestions...

lhstrh (changes requested):
There is one change that needs to be added in order to make this work; the code generator should use the version number as well (instead of `latest`).

lhstrh (approved):
This looks good to me, but it appears our merge queue/CI is broken due to a failure in the Zehpyr tests...

PR #2391, opened 2024-08-02
Update latest rti image for released versions

lhstrh:
Upon release, a push to master happens with a version number that does not end with "-snapshot". When this occurs, also update the image associated with the "latest" tag.

### DIFF
--- a/.github/workflows/rti-docker.yml
+++ b/.github/workflows/rti-docker.yml
@@ -32,4 +32,4 @@ jobs:
           tag: latest
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
-        if: ${{ env.lf_version != *-snapshot }}
+        if: ${{ !endsWith(env.lf_version, '-snapshot') }}
--- a/.github/workflows/rti-docker.yml
+++ b/.github/workflows/rti-docker.yml
@@ -26,3 +26,10 @@ jobs:
           tag: ${{ env.lf_version }}
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Update latest (released versions only)
+        uses: ./.github/actions/push-rti-docker
+        with:
+          tag: latest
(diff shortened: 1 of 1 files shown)


lhstrh:
Note sure if `${{ env.lf_version != *-snapshot }}` works. @Depetrol, can you confirm?

.github/workflows/rti-docker.yml
```diff
@@ -26,3 +26,10 @@ jobs:
           tag: ${{ env.lf_version }}
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Update latest (released versions only)
+        uses: ./.github/actions/push-rti-docker
+        with:
+          tag: latest
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+        if: ${{ env.lf_version != *-snapshot }}
```
Depetrol, line 35:
```suggestion
        if: ${{ ! endsWith( env.lf_version, '-snapshot' ) }}
```

lhstrh: Looks good to me, but how do we confirm that it works?

Depetrol: I think [this](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#endswith) might be what you are looking for.

Depetrol: I think the only way is to do a fork and merge to the master branch in the fork

Depetrol (approved):

.github/workflows/rti-docker.yml
lhstrh: Right. Could you perhaps carry out this experiment?

lhstrh: We could also just publish another release and see if it works. If it doesn't, we could update `latest` manually.

Depetrol: It seems to be working in my fork in [this](https://github.com/Depetrol/lingua-franca/actions/runs/10222853768/job/28288178683) run for [this](https://github.com/Depetrol/lingua-franca/pull/22) PR. An image with `latest` tag has been pushed to [my dockerhub](https://hub.docker.com/r/depetrol/rti/tags).

lhstrh: Perfect!

PR #2392, opened 2024-08-03
Allow rti-docker workflow to be triggered on workflow_call

Depetrol:


.github/workflows/rti-docker.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches:
       - master
   workflow_dispatch:
+  workflow_call:
+
```
lhstrh, line 9: I think that for this to work, the `checkout` action needs to be given a  `repository: lf-lang/lingua-franca` parameter (or it will checkout the calling repo).

Depetrol: Updated!

lhstrh (approved):

PR #2393, opened 2024-08-03
Matching image digest between `latest` tag and latest release tag of `rti` image on DockerHub

Depetrol:
Match RTI docker image digest by building only once and push to Docker Hub as two tags when doing release

### DIFF
--- a/.github/actions/push-rti-docker/action.yml
+++ b/.github/actions/push-rti-docker/action.yml
@@ -11,6 +11,9 @@ inputs:
   DOCKERHUB_TOKEN:
     description: 'The token to log in to Docker Hub'
     required: true
+  latest:
+    description: 'Also push as latest tag if true'
+    default: 'false'
 runs:
   using: "composite"
   steps:
@@ -34,3 +37,13 @@ runs:
         platforms: linux/amd64, linux/arm64, linux/arm/v7, linux/riscv64
         push: true
         tags: lflang/rti:${{ inputs.tag }}
+      if: ${{ inputs.latest == 'false' }}
+    - name: Build and push as latest
+      uses: docker/build-push-action@v6
+      with:
+        file: ./core/src/main/resources/lib/c/reactor-c/core/federated/RTI/rti.Dockerfile
+        context: ./core/src/main/resources/lib/c/reactor-c
+        platforms: linux/amd64, linux/arm64, linux/arm/v7, linux/riscv64
+        push: true
+        tags: lflang/rti:${{ inputs.tag }}, lflang/rti:latest
+      if: ${{ inputs.latest == 'true' }}
--- a/.github/workflows/rti-docker.yml
+++ b/.github/workflows/rti-docker.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
   workflow_dispatch:
+  workflow_call:
+
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
@@ -13,6 +15,7 @@ jobs:
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
+          repository: lf-lang/lingua-franca
           submodules: recursive
           fetch-depth: 0
       - name: Look up the current version and export as environment variable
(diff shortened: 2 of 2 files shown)


.github/actions/push-rti-docker/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
   DOCKERHUB_TOKEN:
     description: 'The token to log in to Docker Hub'
     required: true
+  latest:
+    description: 'Also push as latest tag or not, set to true of false'
```
lhstrh, line 15:
```suggestion
    description: 'Also push as latest tag if true'
```

.github/actions/push-rti-docker/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
...
+    required: true
```
lhstrh, line 16: Not necessary if a default has been provided...
```suggestion
```

.github/actions/push-rti-docker/action.yml
```diff
@@ -11,6 +11,10 @@ inputs:
...
+    default: 'false'
```
lhstrh, line 17: I don't think bools need to be quoted?
```suggestion
    default: false
```

.github/actions/push-rti-docker/action.yml
```diff
@@ -27,10 +31,20 @@ runs:
         submodules: recursive
         fetch-depth: 0
     - name: Build and push
+      if: ${{ inputs.latest == 'false' }}
```
lhstrh, line 34: Just as a matter of style, I prefer to have the `if` at the end of the block.

.github/actions/push-rti-docker/action.yml
```diff
@@ -27,10 +31,20 @@ runs:
...
       uses: docker/build-push-action@v6
       with:
         file: ./core/src/main/resources/lib/c/reactor-c/core/federated/RTI/rti.Dockerfile
         context: ./core/src/main/resources/lib/c/reactor-c
         platforms: linux/amd64, linux/arm64, linux/arm/v7, linux/riscv64
         push: true
         tags: lflang/rti:${{ inputs.tag }}
+    - name: Build and push as latest
+      if: ${{ inputs.latest == 'true' }}
```
lhstrh, line 43: Just as a matter of style, I prefer to have the `if` at the end of the block.

.github/workflows/rti-docker.yml
```diff
@@ -24,15 +24,19 @@ jobs:
           echo "lf_version=$LF_VERSION"
           echo "lf_version=$LF_VERSION" >> $GITHUB_ENV
       - name: Build and push RTI to Docker Hub
+        if: ${{ endsWith(env.lf_version, '-snapshot') }}
```
lhstrh, line 27: Just as a matter of style, I prefer to have the `if` at the end of the block.

.github/workflows/rti-docker.yml
```diff
@@ -24,15 +24,19 @@ jobs:
...
         uses: ./.github/actions/push-rti-docker
         with:
           tag: ${{ env.lf_version }}
+          latest: false
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+
       - name: Update latest (released versions only)
+        if: ${{ !endsWith(env.lf_version, '-snapshot') }}
```
lhstrh, line 36: Just as a matter of style, I prefer to have the `if` at the end of the block.

lhstrh (approved):
This looks great! It looks like this should work. My comments are only cosmetic, but please apply the requested changes and this should be ready to merge.

.github/actions/push-rti-docker/action.yml
Depetrol: IntelliJ idea give an error without quotes -- I think this field has to be string

lhstrh:
Perfect, I you could now do an interactive rebase to clean up the history and get rid of superfluous commits, we can merge.

PR #2394, opened 2024-08-03
Improvements in decentralized coordination

edwardalee:
This PR is a companion to https://github.com/lf-lang/reactor-c/pull/476.
Changes include:
* Allowing use of STA instead of STP_offset and STAA instead of STP.
* Allowing code blocks for setting the STA in the C target (was already allowed for Python).
* Add tests and fix one Python test to not hang on shutdown.


test/C/src/federated/DecentralizedP2PUnbalancedTimeoutPhysical.lf
```diff
@@ -27,7 +27,9 @@ reactor Clock(offset: time = 0, period: time = 1 sec) {
   =}
 }
 
-reactor Destination {
+// This needs an STA offset because otherwise it may advance to the stop time while the
+// physical connection is in the process of assigning a timestamp.
+reactor Destination(STA: time = 1ms) {
```
Depetrol, line 32:
```suggestion
reactor Destination(STA: time = 1 ms) {
```

Depetrol: This should make the code spotless.

Depetrol (approved):

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -224,7 +224,7 @@ Watchdog:
     code=Code;
 
 STP:
-    'STP' '(' value=Expression ')' code=Code;
+    ('STP' | 'STAA') '(' value=Expression ')' code=Code;
```
lhstrh, line 227: Suggestion for improvement: a deprecation notice for `STP` in the validator would make sense.

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -1036,15 +1035,6 @@ public void checkState(StateVar stateVar) {
     }
   }
 
-  @Check(CheckType.FAST)
```
lhstrh, line 1039: This would be the location to issue a warning if `STP` is used instead of `STA`.

lhstrh (commented):
These changes look like good improvements. The usability difficulties and terminology issues remain, but those should be addressed in a separate PR... I do think it's a good idea to prioritize syntax improvements.

core/src/main/java/org/lflang/LinguaFranca.xtext
edwardalee: I suggest we deprecate both syntaxes together when we come up with a better syntax and for now just allow either STP or STAA.  Otherwise, we need to change all the tests, playground examples, and documentation that use STP and STP_offset now, only to change them all again later.

lhstrh (approved):

PR #2400, opened 2024-08-13
A target property for handling `downstream next event tag` (`DNET`) signal

byeonggiljun:
This PR adds a target property for the signal `downstream next event tag`. When the property is true, the federate launcher turns on the option DNET of the RTI.

This PR is a companion PR of [reactor-c#349](https://github.com/lf-lang/reactor-c/pull/349).

There is another companion PR, https://github.com/lf-lang/reactor-ts/pull/296.


.github/workflows/ts-tests.yml
```diff
@@ -38,8 +38,7 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform TypeScript tests
         run: |
-          ./gradlew targetTest -Ptarget=TypeScript
-        # -Druntime="git://github.com/lf-lang/reactor-ts.git#master"
+          ./gradlew targetTest -Ptarget=TypeScript -Druntime="git://github.com/lf-lang/reactor-ts.git#dnet-ad-hoc"
```
edwardalee, line 41: Can we just update the submodule instead?  Let's get the TS changes merged first and remove this change.

core/src/main/java/org/lflang/target/property/DNETProperty.java
```diff
@@ -0,0 +1,25 @@
+package org.lflang.target.property;
+
+/**
+ * If true, the RTI sends DNET signals during the execution of federation. The default is ture (for
+ * now, for testing).
```
edwardalee, line 5:
```suggestion
 * @brief Target property turning on or off the DNET signal optimization.
 * 
 * If this target property is true, the RTI sends DNET (downstream next event tag) signals to an upstream federate
 * to tell the federate that sending LTC and NET signals with tags less than the specified value is unnecessary.
 * The default is true.
```
Note that if we make the default value `false`, this feature will not be adequately tested. Is there any reason it should not be `true` by default?

edwardalee (approved):
Modulo the open questions, this looks good to me, but merging should wait for the corresponding reactor-c PR. I will review that next.

.github/workflows/ts-tests.yml
byeonggiljun: Since the PR [reactor-ts](https://github.com/lf-lang/reactor-ts/pull/296) has been merged, I made the GitHub action use the `master` branch in [83bd58f](https://github.com/lf-lang/lingua-franca/pull/2400/commits/83bd58f7eeb5a8a0e5234ab299186f6a9ebf16b0).

core/src/main/java/org/lflang/target/property/DNETProperty.java
byeonggiljun: I updated the comment in [83bd58f](https://github.com/lf-lang/lingua-franca/pull/2400/commits/83bd58f7eeb5a8a0e5234ab299186f6a9ebf16b0). 

I'm concerned that if every reaction sends a message every time it's triggered, the DNET may be useless. I'm trying to run some applications to see whether the impact is negligible, but I'm struggling to generalize the cases.

I'll make sure to show you a convincing result shortly.

PR #2404, opened 2024-09-09
Angular bracket imports for reusable reactor modules

vinzbarbuto:
This PR proposes a new import statement syntax to distinguish between user-defined reactors import and reusable reactors imports. The new syntax for importing reusable reactors will use angular brackets, as shown: `import Reactor from <path/to/lib/Reusables.lf>`.


lhstrh (changes requested):
Looks like a good start! I left some questions. The major concern I have is that there are not tests. Please add one.

core/src/main/java/org/lflang/ast/ToSExpr.java
```diff
@@ -217,7 +217,8 @@ public SExpr caseImport(Import object) {
     // reactorClasses+=ImportedReactor)* 'from' importURI=STRING ';'?;
     return sList(
         "import",
-        new SAtom<>(object.getImportURI()),
+        // TODO: Check if angular brackets (< >) are required
```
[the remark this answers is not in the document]
lhstrh:
```suggestion
```

The code as is should be fine. We don't really need to reflect the kind of import in the S-Expression.

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -431,6 +431,8 @@ terminal STRING:   '"' ( '\\' . | !('\\' | '"' | '\t' | '\r' | '\n') )* '"'
                  | '"""' -> '"""'
                  ;
 
+terminal LIB_PATH:   '<' ( '\\' . | !('\\' | '"' | '\t' | '\r' | '\n') )* '>';
```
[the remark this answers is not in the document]
vinzbarbuto: I believe we should keep it for `LIB_PATH` as well, as removing this part could prevent us from escaping characters within the path. For example, without it, paths with backslashes like `<C:\Users\Documents\file.txt>` would become invalid.

core/src/main/java/org/lflang/ast/ToLf.java
```diff
@@ -398,7 +398,8 @@ public MalleableString caseImport(Import object) {
         // TODO: This is a place where we can use conditional parentheses.
         .append(list(", ", "", "", false, true, true, object.getReactorClasses()))
         .append(" from \"")
-        .append(object.getImportURI())
+        // TODO: Check if angular brackets (< >) are required
```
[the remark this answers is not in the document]
vinzbarbuto: I checked, and we actually need to switch from the AST node format to the LF file format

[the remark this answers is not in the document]
lhstrh: Wdym?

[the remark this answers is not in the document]
vinzbarbuto: When we execute `lff-dev` with the new import statement, omitting `importPackage` when `importURI` is `null`, we encounter the following error: `lff: fatal error: The formatter failed to produce output that is semantically equivalent to its input`. Therefore, we need to ensure that the formatter can generate output that is semantically equivalent to the input.

test/Python/src/FederatedImportLibrary.lf
```diff
@@ -0,0 +1,20 @@
+# Test the new import statement, with the import path enclosed in angle brackets, in federated LF programs
+target Python {
+  timeout: 2 sec
+}
+
+import Count from <lib/Count.lf>
```
lhstrh, line 6: @tanneberger, tagging you because it still appears to be unclear what should be inside the angular brackets. I thought the notation was reserved for packages, not local paths?

core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -64,7 +64,7 @@ Model:
 /**
  * Import declaration.
  */
-Import: 'import' reactorClasses+=ImportedReactor (',' reactorClasses+=ImportedReactor)* 'from' importURI=STRING ';'?;
+Import: 'import' reactorClasses+=ImportedReactor (',' reactorClasses+=ImportedReactor)* 'from' (importURI=STRING | '<' importPackage=Path '>') ';'?;
```
lhstrh, line 67: :+1:

test/Python/src/FederatedImportLibrary.lf
vinzbarbuto: This is the old version. I still need to implement what we discussed last time. My focus here was to fix the formatting errors causing the check to fail. I'm currently working on a new version where we shorten the paths inside the angular brackets to `<root/file.lf>` for downloaded Lingo packages and `lib/file.lf` for locally defined libraries

lhstrh: Do you mean `"lib/file.lf"` for locally defined libraries? My assumption had been that angular brackets are only for published packages, not locally defined ones.

vinzbarbuto: Exactly, use angular brackets for published packages and quotation marks for locally defined libraries. In my previous comment, I was referring to `"lib/file.lf"`, but I mistakenly omitted the quotation marks

core/src/main/java/org/lflang/ast/IsEqual.java
```diff
@@ -99,6 +99,7 @@ public Boolean caseModel(Model object) {
   public Boolean caseImport(Import object) {
     return new ComparisonMachine<>(object, Import.class)
         .equalAsObjects(Import::getImportURI)
+        .equalAsObjects(Import::getImportPackage)
```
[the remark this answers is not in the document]
lhstrh: Does this effectively function like "equal to importURI OR equal to importPackage"?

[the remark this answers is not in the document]
vinzbarbuto: @lhstrh Do you know how I can debug this function to fully understand its behavior?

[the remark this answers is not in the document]
lhstrh: The comparison machine first checks whether the "other object" is an instance of the given class (see constructor). In order to give a definitive answer about equality, the other object and the given object are also compared in terms of what their getters return. If any of those comparisons yields `false`, the two objects must not be equal. You are correct to add this line. If you hadn't `caseImport` would have returned false positives (because it didn't look at the `getImportURI` getter).

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+   * @return The constructed package URI as a string.
+   * @throws IllegalArgumentException if the URI string or source path is null, empty, or does not
+   *     contain both the library name and file name.
+   */
+  public static String buildPackageURIfromSrc(String uriStr, String src) {
+    if (uriStr == null || src == null || uriStr.trim().isEmpty() || src.trim().isEmpty()) {
+      throw new IllegalArgumentException("URI string and source path must not be null or empty.");
+    }
+
+    String[] uriParts = uriStr.trim().split("/");
```
lhstrh, line 73: Please use Java's `Path` class instead.
```suggestion
    Path path = Paths.get(uriSr);
```

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+    if (uriParts.length < 2) {
```
lhstrh, line 75:
```suggestion
    if (path.getNameCount() < 2) {
```

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Builds a package URI based on the provided URI string and source path. This method works
+   * similarly to the `buildPackageURI`, but it accepts a direct source path instead of a resource.
+   * It traverses upwards to locate the "src/" directory and then constructs the URI pointing to the
+   * library file.
+   *
+   * @param uriStr A string representing the URI of the file. It must contain both the library name
+   *     and file name, separated by a '/'.
+   * @param src The source path from which the URI resolution should start.
```
lhstrh, line 63: A better name for this would be "base" or "root."

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+    while (!rootPath.endsWith("src")) {
+      rootPath = rootPath.getParent();
+      if (rootPath == null) {
+        throw new IllegalArgumentException("The 'src' directory was not found in the given path.");
+      }
+    }
+
+    Path finalPath =
+        rootPath
+            .resolveSibling("target/lfc_include/")
```
lhstrh, line 92: I think this is safer:
```suggestion
            .resolveSibling("target")
            .resolve("lfc_include")
```

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+            .resolve(uriParts[0].trim()) // library name
+            .resolve("src/lib")
```
lhstrh, line 94: Idem:
```suggestion
            .resolve("src")
            .resolve("lib")
```

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+            .resolve(uriParts[1].trim()); // file name
```
lhstrh, line 95:
```suggestion
            .resolve(path.getFileName().toString());
```

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+    return finalPath.toString();
```
lhstrh, line 97: Why convert back to a string? It seems you convert it back to a `Path` again at the call site. Just let the method return `Path`?

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+   *     and file name, separated by a '/'.
+   * @param resource The resource from which the URI resolution should start.
+   * @return The constructed package URI as a string.
+   * @throws IllegalArgumentException if the URI string does not contain both library and file
+   *     names.
+   */
+  public static String buildPackageURI(String uriStr, Resource resource) {
+    URI currentURI = URI.createURI(".");
+    URI rootURI = currentURI.resolve(resource.getURI());
+    StringBuilder pathBuilder = new StringBuilder();
```
lhstrh, line 31: Don't use a string builder to construct paths.

https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html

core/src/main/java/org/lflang/util/ImportUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+    String[] uriParts = uriStr.split("/");
+
+    if (uriParts.length < 2) {
```
lhstrh, line 35: See `getNameCount` in `Path`.

lhstrh (changes requested):
The way you're dealing with paths is too brittle and needs work. Please see my comments.

core/src/main/java/org/lflang/util/ImportUtil.java
vinzbarbuto: The method is invoked [here](https://github.com/lf-lang/lingua-franca/blob/0b94151d072c392b44b3c4819ef6d3895ad990ed/core/src/main/java/org/lflang/federated/generator/FedImportEmitter.java#L35C15-L43C41), where the returned string serves as an argument for the `resolve` function. While there is an overloaded `resolve` method that accepts `Path` objects, I opted to return a string because of the following code:

```java
Path importPath =
    fileConfig
        .srcPath
        .resolve(
            i.getImportURI() != null
                ? i.getImportURI()
                : ImportUtil.buildPackageURIfromSrc(
                    i.getImportPackage(), fileConfig.srcPath.toString()))
        .toAbsolutePath();
```

Since `i.getImportURI()` returns a string, I chose to return a string as well. Alternatively, if we want to return a `Path` and use the `resolve` method that takes `Path` objects, I would need to convert `i.getImportURI()` to a `Path` object.

lhstrh: Yes, converting a importURI to a Path is trivial: Paths.get(), so that seems like a good solution to me.

test/Python/target/lfc_include/library-test/src/lib/Import.lf
```diff
@@ -0,0 +1,12 @@
+target Python
```
tanneberger, line 1: the path `s/target/build`

tanneberger (changes requested):

tanneberger (approved):

lhstrh (approved):

Issue #2300, opened 2024-05-30
Presence of logical action is not reset in C target


a-sr:
After being present once the `is_present` always returns `true` even if the logical action did not trigger.

**Example**:
```
target C {
    timeout: 7 msecs,
    fast: true
}

main reactor {
    logical action a;
    logical action b;
        
    reaction(startup) -> a {=
        lf_schedule(a, MSEC(1));
    =}

    reaction(a, b) -> a, b {=
        if (a->is_present) {
            printf("A");
            lf_schedule(b, MSEC(2));
        }
        if (b->is_present) {
            printf("B");
            lf_schedule(a, MSEC(1));
        }
        printf(" at %d msecs with triggers (%d,%d)\n", lf_time_logical_elapsed() / MSEC(1), a->is_present, b->is_present);
    =}
}
```
**Output** (VS Code v0.7.3):
```
A at 1 msecs with triggers (1,0)
AB at 3 msecs with triggers (1,1)
AB at 4 msecs with triggers (1,1)
AB at 5 msecs with triggers (1,1)
AB at 6 msecs with triggers (1,1)
AB at 7 msecs with triggers (1,1)
```
**Expected behavior**:
`A` and `B` should alternate with different offsets.

cmnrd:
This is a duplicate of https://github.com/lf-lang/lingua-franca/issues/2291

lhstrh:
Fixed via #2409.

Issue #2291, opened 2024-05-21
Actions remain "present" after they're scheduled for the first time (C-Target)


AneesHl:
In the C-Target, once an action is scheduled for the first time then `action_id->is_present` always returns `true`. As an example, the following program 
```c
target C { timeout: 1 sec, keepalive: true }

main reactor {
  state ms_250 : time = 250 msec
  logical action _logical
  physical action _physical : float
  timer t(0 msec, 250 msec);

  reaction(startup) -> _logical, _physical {=
    printf("\n--- Startup Reaction ---\n");
    
    lf_schedule(_logical, self->ms_250);
    lf_schedule(_physical, self->ms_250 * 2);
  =}
  
  reaction(_logical, _physical, t) {=
    printf("\n--- Reaction 2 ---\n");
    
    if (_logical->is_present) {
      printf("_logical->is_present = true\n");
    }
    if (_physical->is_present) {
      printf("_physical->is_present = true\n");
    }
    
    printf("~> elapsed logical time  = " PRINTF_TIME "\n", lf_time_logical_elapsed());
  =}
}
```
it produces the following result:
```
...
--- Startup Reaction ---

--- Reaction 2 ---
~> elapsed logical time  = 0

--- Reaction 2 ---
_logical->is_present = true
~> elapsed logical time  = 250000000

--- Reaction 2 ---
_logical->is_present = true
~> elapsed logical time  = 500000000

--- Reaction 2 ---
_logical->is_present = true
_physical->is_present = true
~> elapsed logical time  = 500410000

--- Reaction 2 ---
_logical->is_present = true
_physical->is_present = true
~> elapsed logical time  = 750000000

--- Reaction 2 ---
_logical->is_present = true
_physical->is_present = true
~> elapsed logical time  = 1000000000
...
```

lhstrh:
Fixed via #2409.

PR #2409, opened 2024-09-24
Fix action is_present field not being reset 

erlingrj:
To correctly reset the `is_present` field of actions we must add a pointer to the `status` field of the associated `trigger_t` to the `is_present_fields` array. This is because, for actions (unlike input ports), the `is_present` field is always overwritten by the `status` field of the `trigger_t` in the reaction preambles.  

UPDATE:
This exposed also an issue with the code-generated setup of the `is_present_fields` array for banks and multiport which is also fixed in this PR

Corresponding reactor-c PR: https://github.com/lf-lang/reactor-c/pull/482

Closing #2300 #2291

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CGenerator.java
@@ -1417,7 +1417,8 @@ public class CGenerator extends GeneratorBase {
             temp.startScopedBankChannelIteration(port, null);
           } else {
             // This branch should not occur because if the port's parent is instance and the port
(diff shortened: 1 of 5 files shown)


erlingrj:
@edwardalee, I believe it is correct now

lhstrh (commented):
Kudos for getting this fixed but I'll leave review to @edwardalee -- I can't quite follow the magic in the code generator...

edwardalee (approved):
This looks right to me now.  I pushed a couple of simplifications and elaborated one of the tests to cover more cases.  If the tests pass, this is ready to merge.

Issue #2410, opened 2024-09-26
Bank of federates always have bank_index of 0 in python target


Depetrol:
When running the below program in the Python target
```python
target Python;
reactor A(bank_index = 0) {
  reaction (startup) {=
    print("bank_index: {:d}".format(self.bank_index))
  =}
}
federated reactor {
  a = new[4] A()
}
```
The output is 
```
bank_index: 0
bank_index: 0
bank_index: 0
bank_index: 0
```
Which is not  expected.

Related issue: https://github.com/lf-lang/lingua-franca/issues/1962

This doesn't happen in the C target:
```c
target C;
reactor A(bank_index:int = 0) {
  reaction (startup) {=
    // print("bank_index: {:d}".format(self.bank_index))
    printf("bankindex: %d\n", self->bank_index);
  =}
}
federated reactor {
  a = new[4] A()
}
```
The result is
```
bankindex: 2
bankindex: 0
bankindex: 3
bankindex: 1
```

Depetrol:
The cause of this issue is in the parameter override of `bank_index`.

```Python
target Python
reactor A(bank_index=0) {
    reaction(startup) {=
        print("bank_index: {:d}".format(self.bank_index))
    =}
}
main reactor  {
    a = new A(bank_index=2)
}
```
outputs 
```
bank_index: 0
```
However in the C target 
```C
target C
reactor A(bank_index:int=0) {
    reaction(startup) {=
        printf("bank_index: %d\n", self->bank_index);
    =}
}
main reactor  {
    a = new A(bank_index=2)
}
```
outputs
```
bank_index: 2
```

lhstrh:
What is the expected behavior, and why? If bank index is a "magic" parameter, and it is possible to override it, then one could argue that the Python behavior makes sense and the C behavior doesn't. It might be better to deal with bank indexes via an API function...

PR #2411, opened 2024-09-26
Fix bank_index Parameter Override

Depetrol:
This PR solves #2410 by allowing explicitly set `bank_index` parameter to override in Python target.


test/Cpp/src/ActionWithNoReaction.lf
```diff
@@ -11,7 +11,7 @@ reactor foo {
   logical action a
 
   reaction(x) -> y, a {=
-    y.set(2*(*x.get()));
+    y.set(2*(*x.undefined_name15291838()));
```
edwardalee, line 14: This looks like an unintended and unrelated change?

edwardalee (approved):
Good catch!  LGTM, except that you seem to have an unintended and unrelated change.

test/Cpp/src/ActionWithNoReaction.lf
Depetrol: Yes this is unintended, probably has something to do with the code editor. I'll remove it.

lhstrh (approved):

PR #2412, opened 2024-09-29
Added reactions to test

edwardalee:
A test newly added in merged #2409 had no reactions.  Apparently, on Windows only, this results in running forever.  I have opened an [issue in reactor-c](https://github.com/lf-lang/reactor-c/issues/485) and propose updating the test here to add reactions.


lhstrh:
Note: changed the labels because this isn't a bugfix but a workaround.

PR #2413, opened 2024-09-29
Default Docker image for Python changed to Alpine

### LINKED ISSUE #2379: Move from `python:3.10-slim` to `python:3.10-alpine` default base image
> `python:3.10-slim` [derives from Debian bookworm](https://github.com/docker-library/python/blob/cb3ed4c9ff44fb15611034ae4d579ef82a6d12e5/3.10/slim-bookworm/Dockerfile) and the C image derives from Alpine. Debian will use bash as its shell, alpine will use POSIX Dash. However I'm surprised this would make a difference on something like quoting where the two shells should be interchangeable.

> I think it would be a good idea to have the Python docker images use the same parent image as our C image, so that they have the same shell and package manager. `python:3.10-alpine` is an official image for Python 3.10 that is based on Alpine, so this would accomplish this.

_Originally posted by @elgeeko1 in https://github.com/lf-lang/lingua-franca/issues/2377#issuecomment-2244017949_

lhstrh:
Fixes #2379.

### DIFF
--- a/core/src/main/java/org/lflang/generator/docker/PythonDockerGenerator.java
+++ b/core/src/main/java/org/lflang/generator/docker/PythonDockerGenerator.java
@@ -9,7 +9,7 @@ import org.lflang.generator.LFGeneratorContext;
  * @author Hou Seng Wong
  */
 public class PythonDockerGenerator extends CDockerGenerator {
-  public static final String DEFAULT_BASE_IMAGE = "python:3.10-slim";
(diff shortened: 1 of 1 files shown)


edwardalee (commented):
I have no knowledge about this so can't provide a meaningful review.

Depetrol:
This looks good to me. The meaningful change here is using Dash instead of Bash as the shell in the image, but this change doesn't affect the user interface since all docker image files are generated. Since all tests have passed, I think this is good to go.

Issue #1489, opened 2022-11-22
Reaction is executed without triggering


a-sr:
In the following program, the reaction in `ForwardFailure` is executed without a justified triggering (`failure` is never set).
```
target C
reactor OnlySuccess {
    input start:bool
    output success:bool
    output failure:bool
    
    reaction (start)  -> success, failure {=
        lf_print("Create Success");
        lf_set(success, true);
    =}
}
reactor ForwardFailure {
    input in:bool
    output out:bool
    reaction (in) -> out {=
        lf_print("Forward Failure");
        lf_set(out, true);
    =}
}
reactor Sequence {
    input start:bool
    output success:bool
    output failure:bool
    s = new OnlySuccess();
    f = new ForwardFailure();
    start -> s.start
    s.success -> success
    s.failure -> f.in
    f.out -> failure
}
main reactor {
    s = new Sequence()
    reaction(startup) -> s.start {=
        lf_print("Start");
        lf_set(s.start, true);
    =}
// Uncommenting this reaction will fix the problem 
// ForwardFailure is no longer executed
//    reaction(s.success) {=
//        lf_print("Success");
//    =}
    reaction(s.failure) {=
        lf_print("Failure");
    =}
}
```
![IllegalTriggering](https://user-images.githubusercontent.com/25612037/203279409-0e5fd0ef-833a-49f3-bd02-81a42e58cf72.png)

Produced output:
```
Start
Create Success
Forward Failure
Failure
```

The problem seems to relate to the fact that the output at the `success` port is not consumed. If uncommenting the reaction in the main reactor, the program yields the correct behavior:
```
Start
Create Success
Success
```

This issue may relate to the bug reported in #1433. There, an unused connection seems to prevent the execution of a reaction.

petervdonovan:
To me, this looks the same as #1321, which is still unresolved.

edwardalee:
This is fixed with #2416.

PR #2416, opened 2024-10-01
Fix disconnected port handling

### LINKED ISSUE #1433: Mode transition fails to trigger in some
I found a simple modal model where a transition fails to trigger when it should.  This is checked into the [transition-trigger branch](https://github.com/lf-lang/lingua-franca/tree/transition-trigger) here:

https://github.com/lf-lang/lingua-franca/blob/transition-trigger/test/C/src/modal_models/TransitionTrigger.lf

Even stranger, if you uncomment the empty reaction on [line 24](https://github.com/lf-lang/lingua-franca/blob/transition-trigger/test/C/src/modal_models/TransitionTrigger.lf#L24), then the test passes. This reaction should really have no effect.

edwardalee:
This fixes #1433 and #1489.

### DIFF
--- a/test/C/src/TriggerIssue.lf
+++ b/test/C/src/TriggerIssue.lf
@@ -35,3 +35,4 @@ main reactor {
   dlc = new Destination()
 
   ui.y -> dlc.y
+}
--- a/test/C/src/TriggerIssue.lf
+++ b/test/C/src/TriggerIssue.lf
@@ -1,0 +1,42 @@
+/**
+ * Destination cannot react to y. Only if x is also a trigger (uncomment
+ * reaction) it will work.
+ */
+ target C {
+  logging: debug
+}
+
+reactor Source {
+  output x: bool
+  output y: bool
+
+  reaction(startup) -> x, y {=
+      lf_print("S");
+      lf_set(y, true);
+  =}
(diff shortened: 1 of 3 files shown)


lhstrh:
Is there anything material to the fix in the reactor-c update?

edwardalee:
> Is there anything material to the fix in the reactor-c update?

No, but I'm testing with reactor-c/main, so it needs to be aligned.

test/C/src/TriggerIssue.lf
```diff
@@ -0,0 +1,38 @@
+/** Destination cannot react to y. Only if x is also a trigger (uncomment reaction) it will work. */
```
lhstrh, line 1: This comment seem wrong? What is the test meant to capture? The output `x` is never set, but it's also not connected. An output is produced on `y`, and the downstream reactor has a reaction that is triggered by it. Did this not work prior to the change?

edwardalee: Good catch!  This comment was left over from the issue.  Fixed in d6d6cd6deee29db19ec08812f37cd67628bb5c6f.

lhstrh (approved):
LGTM!

PR #2417, opened 2024-10-01
Minimal code to run LF on FlexPRET with uC runtime

magnmaeh:
Currently a draft only


core/src/main/kotlin/org/lflang/generator/uc/UcReactorGenerator.kt
```diff
@@ -91,11 +91,12 @@ class UcReactorGenerator(private val reactor: Reactor, fileConfig: UcFileConfig,
         """.trimMargin()
     }
 
+    // FIXME: 3rd argument must be replaced with self->parent if not main reactor
     fun generateCtorDefinition() = with(PrependOperator) {
         """
             | void ${reactor.name}_ctor(${reactor.name} *self, Environment *env) {
             |  size_t trigger_idx = 0;
-            |  Reactor_ctor(&self->super, "${reactor.name}", env, ${if (numChildren > 0) "self->_children" else "NULL"}, $numChildren, ${if (reactor.reactions.size > 0) "self->_reactions" else "NULL"}, ${reactor.reactions.size}, ${if (numTriggers() > 0) "self->_triggers" else "NULL"}, ${numTriggers()});
+            |  Reactor_ctor(&self->super, "${reactor.name}", env, NULL, ${if (numChildren > 0) "self->_children" else "NULL"}, $numChildren, ${if (reactor.reactions.size > 0) "self->_reactions" else "NULL"}, ${reactor.reactions.size}, ${if (numTriggers() > 0) "self->_triggers" else "NULL"}, ${numTriggers()});
```
magnmaeh, line 99: @erlingrj I had to add `NULL` here to make it compile, but I assume it should be `self->parent` for all reactors that are not the `main` reactor. This might be needed on the base branch as well?

erlingrj: Yes. Thanks. Unfortunately, I have made some changes to the reactor-uc API that are not reflected yet in the code-generator. I will try to bring everything up-to-date soon

erlingrj:
@magnmaeh, I am in the process of making some bigger changes to the reactor-uc API and will have to make some updates here in the code-generator as well. One key thing to realize is that we want the code generator to just output a CMake/Make project that can be included in another project. I.e. the main CMakeLists.txt lives outside, and also the main function. For the POSIX platform we have added a main function because it is convenient for quick testing. But we envision that for other platforms we generate a subdirectory in a project and include it in the existing build-flow

erlingrj:
@magnmaeh I have now merged in comprehensive changes to `lingua-franca` and `reactor-uc`. Please check them. To support FlexPRET we should require very little changes to the code-generator. Only in the platform-specific CMake. 

I am planning to fork `lingua-franca` and bring it into the `reactor-uc` repo so we just have a simple monorepo while we develop the runtime and code-generator and reach a stable v0.1

magnmaeh:
Great work! I'll have a look :)

magnmaeh:
I think this is ready for review as well. But I guess you don't want the main/CMake and main/main.c files to be part of the PR? I used it just to verify that it worked, which it did. 

After your additions it even magically solved my one issue with it not printing the "Hello world from FlexPRET!". So that was nice :)

erlingrj:
Can you instead add these changes directly to https://github.com/erlingrj/reactor-uc which now contains the code-generator also?

PR #2420, opened 2024-10-03
Fixed parameterized mutable inputs

### LINKED ISSUE #2117: Parameterized mutable multiport inputs does not work
```
target C;
reactor R(width:int=4) {
  mutable input[width] in: int
  reaction(in) {==}
}
main reactor {
  r = new R()
}
```

gives:
```
/home/erling/dev/examples-lingua-franca/src-gen/MutablMultiport/_r.c: In function ‘_rreaction_function_0’:
/home/erling/dev/examples-lingua-franca/src-gen/MutablMultiport/_r.c:11:24: error: ‘width’ undeclared (first use in this function)
   11 |     _r_in_t _lf_tmp_in[width];
      |                        ^~~~~
/home/erling/dev/examples-lingua-franca/src-gen/MutablMultiport/_r.c:11:24: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [CMakeFiles/MutablMultiport.dir/build.make:104: CMakeFiles/MutablMultiport.dir/_r.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:100: CMakeFiles/MutablMultiport.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
lfc: error:  failed with error code 2
```

The problem appears to be that the code-generator just uses the parameter directly, it should be prepended with `self`.

edwardalee:
Fixed #2117 and added tests. There is [companion PR in reactor-c](https://github.com/lf-lang/reactor-c/pull/487).


erlingrj:
I am a little surprised that we have a test for Enclaved shutdown as we don't "officially" support enclaves in the C target and a coordinated shutdown of the enclaves is a subtle topic. I will take a quick look. It probably has to do with the change you made regarding scheduling the very first timer event if it is passed the timeout time

PR #2421, opened 2024-10-05
Native `forever` and `never` time literal

### LINKED ISSUE #2374: Parse target variable STP (STAA)
Currently Lingua Franca syntax allows passing target variable as STP:
```
reaction() {=
      ...
    =} STP({=FOREVER=}) {=
      ...
    =}
```
The above code is currently accepted by `lfc` as valid code, but the generated code is 
```
staa_lst[0]->STAA = SEC(0);
```
This is because the function `findMaxSTP()` in `FedASTUtils.java` has the following code
```java
return STPList.stream()
        .map(ASTUtils::getLiteralTimeValue)
        .filter(Objects::nonNull)
        .reduce(TimeValue.ZERO, TimeValue::max);
```
that filters out null values produced by `getLiteralTimeValue` when the `STP` parameter is target code. 

This cannot be simply fixed by inserting the expression verbatim because the value of `STP` parameter is currently used in compile time to determine the max. A fix could be to move finding the max STP to runtime, but this introduces larger change to the codebase.

Depetrol:
This PR introduces a new `forever` time literal that can be used as STA and STAA in decentralized execution. This PR is related to #2368 and closes #2374.
This PR also introduces a new `never` time literal that can be used to initialize state values.

### DIFF
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -393,7 +393,7 @@ SignedInt:
 ;
 
 Forever:
-    'FOREVER' | 'forever'
+    'forever'
 ;
 
 Literal:
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -393,11 +393,11 @@ SignedInt:
 ;
 
 Forever:
-    'forever' | 'FOREVER'
+    'forever'
 ;
 
 Never:
-    'never' | 'NEVER'
+    'never'
 ;
 
 Literal:
@@ -529,7 +529,7 @@ Token:
     'startup' | 'shutdown' | 'after' | 'deadline' | 'mutation' | 'preamble' |
     'new' | 'federated' | 'at' | 'as' | 'from' | 'widthof' | 'const' | 'method' |
     'interleaved' | 'mode' | 'initial' | 'reset' | 'history' | 'watchdog' |
-    'extends' | 'forever' | 'FOREVER' | 'never' | 'NEVER' |
+    'extends' | 'forever' | 'never' |
 
     // Other terminals
     NEGINT | TRUE | FALSE |
--- a/core/src/main/java/org/lflang/LinguaFranca.xtext
+++ b/core/src/main/java/org/lflang/LinguaFranca.xtext
@@ -339,7 +339,7 @@ ParameterReference:
 ;
 
 Time:
-    (interval=INT unit=TimeUnit)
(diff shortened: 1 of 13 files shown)


lhstrh:
Keywords in LF are all lowercase -- let's stick with that?

Depetrol:
Makes sense! I'll modify the PR.

edwardalee:
It's unclear to me how this is used.  I don't see any changes in the code generator to detect "forever" and then use the C macro `FOREVER`.  How does this get used?

Depetrol:
This implementation does not use the C macro `FOREVER`. Instead, it introduces a new Lingua Franca time literal named `forever` in the Xtext syntax, and when the compiler sees this time literal it interprets this time literal as `TimeValue.MAX_VALUE`. This is similar to how the compiler currently handles the `0` time literal. The benefit of this implementation is that it is target language agnostic.

edwardalee:
> This implementation does not use the C macro `FOREVER`. Instead, it introduces a new Lingua Franca time literal named `forever` in the Xtext syntax, and when the compiler sees this time literal it interprets this time literal as `TimeValue.MAX_VALUE`. This is similar to how the compiler currently handles the `0` time literal. The benefit of this implementation is that it is target language agnostic.

Ah yes, that makes sense.

edwardalee (approved):
This looks good to me.

core/src/main/java/org/lflang/LinguaFranca.xtext
lhstrh, line 1: Note that `'forever'` has to also be included in the `Token` production. If it's not added, anytime a reaction body has the token `forever` in it, the parser will throw an error.

lhstrh (commented):
Is there no practical use case for`never`? If there is, then shouldn't it also be allowed as a time value?

core/src/main/java/org/lflang/LinguaFranca.xtext
Depetrol: Fixed, thanks for pointing it out!

Depetrol:
I haven't yet encountered a use case where `never` is used. However, if there is a use case for it, I would be glad to add it to the syntax as well!

Depetrol:
I was looking into the tests and found a test `LastTimeReplace.lf` that uses the `never` time literal. It is noted that 
```C
reaction(startup) {=
    // Unfortunately, a time state variable cannot be initialized with NEVER.
    // So we do that here.
    self->last = NEVER;
  =}
```
I think introducing the `never` time literal makes sense in this use case. Also on second thought, perhaps we should allow `FOREVER` and `NEVER` as well to align with the target code constants?

lhstrh (approved):
LGTM!

PR #2423, opened 2024-10-11
Fix Unintended Action Override

Depetrol:
This is a companion PR to https://github.com/lf-lang/reactor-c/pull/490 that introduces a test case for concurrent actions.


edwardalee (approved):
Nice test! Pretty aggressive to create a hundred threads that all schedule physical actions.

Depetrol:
Yes, this test with a hundred threads triggers the concurrency bug pretty consistently.

PR #2424, opened 2024-10-12
Fix windows

edwardalee:
For some reason, Windows tests were disabled for a bunch of recent merges.  This is an attempt to get things working again in master.


lhstrh:
You can trigger the Windows/macOS platforms by manually triggering the workflow. We had disabled them because everyone was perpetually watching progress bars.

A manual run revealed there are still problems with Python and TypeScript on Windows. The TypeScript failure is sus and might be a nondeterministic one. @vinzbarbuto, the Python failure has to do with paths:

```
Running: src\lingo_imports\FederatedTestImportPackages.lf (51%)
    0    [Test worker] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
    java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:/D:/a/lingua-franca/lingua-franca/test/Python/src/lingo_imports/FederatedTestImportPackages.lf
```

vinzbarbuto:
> A manual run revealed there are still problems with Python and TypeScript on Windows. The TypeScript failure is sus and might be a nondeterministic one. @vinzbarbuto, the Python failure has to do with paths:
> 
> ```
> Running: src\lingo_imports\FederatedTestImportPackages.lf (51%)
>     0    [Test worker] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
>     java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:/D:/a/lingua-franca/lingua-franca/test/Python/src/lingo_imports/FederatedTestImportPackages.lf
> ```

I’m not sure what’s causing this error. It looks like it’s related to the path resolution, but I’m wondering why it didn’t happen in the other test we ran.

lhstrh:
> > A manual run revealed there are still problems with Python and TypeScript on Windows. The TypeScript failure is sus and might be a nondeterministic one. @vinzbarbuto, the Python failure has to do with paths:
> > ```
> > Running: src\lingo_imports\FederatedTestImportPackages.lf (51%)
> >     0    [Test worker] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
> >     java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:/D:/a/lingua-franca/lingua-franca/test/Python/src/lingo_imports/FederatedTestImportPackages.lf
> > ```
> 
> I’m not sure what’s causing this error. It looks like it’s related to the path resolution, but I’m wondering why it didn’t happen in the other test we ran.

It's the `:` in the path that's the problem.

vinzbarbuto:
> It's the `:` in the path that's the problem.

Yes, it seems to be the Windows path issue. Do you know how I can reproduce the error on macOS?

lhstrh:
> > It's the `:` in the path that's the problem.
> 
> Yes, it seems to be the Windows path issue. Do you know how I can reproduce the error on macOS so I can fix this?

Nope. If you have no access to a Windows machine, your best bet would probably be to guess and debug in CI.

lhstrh:
The path issue has been resolved, but other issues remain (including the one this PR meant to address, it seems). See: https://github.com/lf-lang/lingua-franca/actions/runs/11308047484/job/31450473838

vinzbarbuto:
> The path issue has been resolved, but other issues remain (including the one this PR meant to address, it seems). See: https://github.com/lf-lang/lingua-franca/actions/runs/11308047484/job/31450473838

I believe the problem outlined [here](https://github.com/lf-lang/lingua-franca/actions/runs/11308759290/job/31451930032#step:11:493) is related to federated execution on Windows. To test the new import statement, I added:

- `src/LingoFederatedImport.lf` to test the new import statement for federated programs in the `src` folder
- `src/lingo_imports/FederatedTestImportPackages.lf` to test the new import statement for federated programs in a subfolder of `src`

However, since Edward moved `LingoFederatedImport.lf` to `src/federated/LingoFederatedImport.lf` to skip federated execution on Windows, the latter file is already in a subdirectory of `src`. Therefore, we can remove `src/lingo_imports/FederatedTestImportPackages.lf` since is a redundant test and avoid the error related to federated execution on Windows.

lhstrh:
> > The path issue has been resolved, but other issues remain (including the one this PR meant to address, it seems). See: https://github.com/lf-lang/lingua-franca/actions/runs/11308047484/job/31450473838
> 
> I believe the problem outlined [here](https://github.com/lf-lang/lingua-franca/actions/runs/11308759290/job/31451930032#step:11:493) is related to federated execution on Windows. To test the new import statement, I added:
> 
>     * `src/LingoFederatedImport.lf` to test the new import statement for federated programs in the `src` folder
> 
>     * `src/lingo_imports/FederatedTestImportPackages.lf` to test the new import statement for federated programs in a subfolder of `src`
> 
> 
> However, since Edward moved `LingoFederatedImport.lf` to `src/federated/LingoFederatedImport.lf` to skip federated execution on Windows, the latter file is already in a subdirectory of `src`. Therefore, we can remove `src/lingo_imports/FederatedTestImportPackages.lf` since is a redundant test and avoid the error related to federated execution on Windows.

Sorry, I do not understand what you mean. Does it mean that you can fix it? :smile:

vinzbarbuto:
> Sorry, I do not understand what you mean. Does it mean that you can fix it? 😄

Yes, @edwardalee said:

> Indeed, the Windows tests cannot pass because of this file:
>
> `test/Python/src/LingoFederatedImport.lf`
>
> which fails under Windows. Windows does not support federated execution. This should be in the federated subdirectory so it is excluded by the Windows tests.

Therefore, we should place all federated tests in the `federated` subdirectory to prevent Windows from failing to execute them. Edward moved `LingoFederatedImport.lf` to `federated/LingoFederatedImport.lf`, but that wasn't the only test I added for federated executions. I also added this file: `src/lingo_imports/FederatedTestImportPackages.lf`, which is now failing:
```shell
 Failed: src\lingo_imports\FederatedTestImportPackages.lf in 0.00 seconds
```

The solution could be either to add this test to the `federated` directory or remove it. I suggest removing it since it’s redundant; I initially used it to test two corner cases, but now `federated/LingoFederatedImport.lf` is sufficient.

lhstrh:
OK, please go head :+1:

vinzbarbuto:
> OK, please go head 👍

Done. This should solve at least the Python tests

.github/workflows/ts-tests.yml
```diff
@@ -24,9 +24,9 @@ jobs:
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
       - name: Setup Node.js environment
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
```
lhstrh, line 29:
```suggestion
          node-version: 22
```

erlingrj (approved):
Looks good. But not sure if there is anything we can do about variable length arrays on Windows. This blog post says that even if the support C11 and C17 they will not support variable-length arrays.

>Astute readers will note that VLAs are also not supported. Variable length arrays are generally less efficient than comparable fixed sized arrays, and generally inefficient compared to equivalent malloc(), when safely and securely implemented. VLAs provide attack vectors comparable to those of the infamous gets() — deprecated and destined to removal — for opportunities of “shifting the stack” and other exploits. For these reasons we intend not to support VLAs as an optional feature in C11.
https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

The question has been posed before, but, do we really need Windows-support?

PR #2426, opened 2024-10-16
Fix Windows issue

petervdonovan:
The generated run scripts should now work properly using a POSIX-like shell installed on Windows.


lhstrh:
Wow, excited to see you found a fix! It's surprising that this problem never surfaced earlier...

lhstrh:
Unfortunately, looks like it doesn't work: https://github.com/lf-lang/lingua-franca/actions/runs/11360061459/job/31597193776

petervdonovan:
> Unfortunately, looks like it doesn't work: https://github.com/lf-lang/lingua-franca/actions/runs/11360061459/job/31597193776

You're right. Trying again in a separate PR

PR #2427, opened 2024-10-16
File extension of TypeScript launch script adjusted on Windows

petervdonovan:
In this branch, the tests pass for me locally for TypeScript on Windows.

### DIFF
--- a/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
+++ b/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
@@ -27,6 +27,7 @@ package org.lflang.generator.ts
 
 import org.eclipse.emf.ecore.resource.Resource
 import org.lflang.FileConfig
+import org.lflang.generator.GeneratorUtils
 import org.lflang.util.FileUtil
 import org.lflang.util.LFCommand
 import java.io.IOException
@@ -54,8 +55,7 @@ class TSFileConfig(
         FileUtil.deleteDirectory(srcGenPath)
     }
 
-    override fun getCommand(): LFCommand {
-        val jsPath = srcGenPath.resolve("dist").resolve("${name}.js")
-        return LFCommand.get("node", mutableListOf(jsPath.toString()), true, srcPkgPath);
+    override fun getExecutableExtension(): String {
+        return if (GeneratorUtils.isHostWindows()) ".bat" else ""
     }
 }
--- a/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
+++ b/core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
@@ -53,4 +53,9 @@ class TSFileConfig(
         super.doClean()
         FileUtil.deleteDirectory(srcGenPath)
     }
+
+    override fun getCommand(): LFCommand {
+        val jsPath = srcGenPath.resolve("dist").resolve("${name}.js")
+        return LFCommand.get("node", mutableListOf(jsPath.toString()), true, srcPkgPath);
+    }
 }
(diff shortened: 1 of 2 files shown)


petervdonovan:
It looks like only the Docker tests are failing now.

lhstrh:
While not using the script may be a reasonable workaround, I think it makes the test less complete, and exposes us to problems like the failing Docker tests. @Depetrol had suggested that the issue was that the script didn't have a `.ps1` extension, which seems simple to fix.

Depetrol:
The reason why typescript tests are failing seems to be powershell in `windows-latest` image only being able to run scripts ending with `.bat` or `.ps1` file extensions. An alternative fix(which I'll push to this branch) is to change the generated script in `bin` to end with `.bat` extension(to align with other targets) and modify the run command to run the `.bat` script.

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -451,7 +451,11 @@ class TSGenerator(
             context.unsuccessfulFinish()
         } else {
             context.finish(GeneratorResult.Status.COMPILED, codeMaps)
-            val shScriptPath = fileConfig.binPath.resolve(fileConfig.name)
+            val shScriptPath = if (GeneratorUtils.isHostWindows()) {
```
lhstrh, line 454: Why replicate the logic of `getExecutableExtension` here? Seems like you can just call `fileConfig.getExecutable()`.

core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
```diff
@@ -53,4 +54,8 @@ class TSFileConfig(
         super.doClean()
         FileUtil.deleteDirectory(srcGenPath)
     }
+
+    override fun getExecutableExtension(): String {
+        return if (GeneratorUtils.isHostWindows()) ".bat" else ""
```
lhstrh, line 59: Why would TS use `.bat` and other targets `.exe` (see FileConfig.getExecutableExtension()`).

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
Depetrol: You're right! I wrote it this way to see if it would work and modified the `getExecutableExtension` later. I pushed a new commit to simplify this code.

core/src/main/kotlin/org/lflang/generator/ts/TSFileConfig.kt
Depetrol: Python target also uses `.bat` in windows(see `PyFileConfig.java`). I think this is because C, C++ and Rust compiles code into binary, while python and typescript generates a script to run the generated code.

lhstrh: :+1:

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
```diff
@@ -47,6 +47,7 @@ import org.lflang.util.FileUtil
 import java.nio.file.Files
 import java.nio.file.Path
 import java.util.*
+import kotlin.io.path.extension
```
lhstrh, line 50: Doesn't seem like this is used?
```suggestion
```

lhstrh (approved):

core/src/main/kotlin/org/lflang/generator/ts/TSGenerator.kt
Depetrol: Yes this is unused.

PR #2429, opened 2024-10-17
Fix Unintended Action Override

Depetrol:
This PR is the companion PR for https://github.com/lf-lang/reactor-c/pull/491. It contains a fix in the code generator that is required for the change in `reactor-c` to pass the tests, and introduces a more aggressive test `ConcurrentActionRepeat.lf` that runs the `ConcurrentAction.lf` introduced in https://github.com/lf-lang/lingua-franca/pull/2423 repeatedly to trigger another concurrency bug.


lhstrh (approved):
The test looks good to me, but I haven't looked closely at the reactor-c part of this...

PR #2431, opened 2024-10-29
API for getting fully-qualified name in Cpp target

OmerMajNition:
Cpp runtime has support for fully qualified name but is not available to reaction bodies. This PR adds support for fqn() in reactions.

### DIFF
--- a/core/src/main/resources/lib/cpp/lfutil.hh
+++ b/core/src/main/resources/lib/cpp/lfutil.hh
@@ -61,6 +61,7 @@ public:
   reactor::Duration get_elapsed_physical_time() const { return reactor->get_elapsed_physical_time(); }
   reactor::Environment* environment() const { return reactor->environment(); }
   const std::string& fqn() const { return reactor->fqn(); }
+  const std::string& name() const { return reactor->name(); }
   void request_stop() const { return environment()->sync_shutdown(); }
 };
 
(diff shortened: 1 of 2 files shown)


lhstrh (commented):
Can you please run `./gradlew spotlessApply` to fix the formatting test failure?

cmnrd (commented):
Looks good! While at it, I think it would be nice to add `name()` as well.

The test currently only checks if `fqn()` is available. It would be better to also assert the expected values.

OmerMajNition:
> Can you please run `./gradlew spotlessApply` to fix the formatting test failure?

OmerMajNition:
`./gradlew spotlessApply` Done

Added name() support as well and validation in test file

OmerMajNition:
reactor-cpp/lib/reactor.cc line # 50

`[[maybe_unused]] bool result = inputs_.insert(port).second;`

should become

`[[maybe_unused]] bool result = outputs_.insert(port).second;`

core/src/main/resources/lib/cpp/lfutil.hh
```diff
@@ -60,6 +60,8 @@ public:
   reactor::Duration get_elapsed_logical_time() const { return reactor->get_elapsed_logical_time(); }
   reactor::Duration get_elapsed_physical_time() const { return reactor->get_elapsed_physical_time(); }
   reactor::Environment* environment() const { return reactor->environment(); }
+  const std::string& fqn() const { return reactor->fqn(); }
```
tanneberger, line 63: please use trailing types: `auto fqn() const -> const std::string&`

tanneberger (changes requested):

tanneberger (approved):

PR #2432, opened 2024-11-05
Add cmake-init-include target property

erlingrj:
This PR adds a new target property called `cmake-init-include` (I am open to other naming suggestions). You can pass it a list of paths to cmake files, which will be included in the generated CMakeLists.txt at the very beginning, _before_ the `project(` statement. 

I am also removing the quotation marks on the LF_SOURCE_DIRECTORY compile def. The quotation mark makes it awkward to use it in the included cmake files, which is where they are likely to be used the most.

This means that the user can override the TOOLCHAIN_FILE cmake variable and can easily set up cross-compilation.

This PR enables STM32F4 support without piling on more platform and build logic into lfc but instead requiring the user to configure CMake correctly and add the right paths, compile defs and options. 

In this PR I use the target property to enable cross compiling for STM32: https://github.com/lf-lang/stm32-nucleo-f446re-template/pull/1

This PR is blocking #2185


edwardalee:
The change to `LF_SOURCE_DIRECTORY` and related definitions will break a number of demos in the playground. In fact, without the quotation marks, I'm not sure it is possible to use these definitions in the way they are used in the playground examples.  See for example [Uptime.lf](https://github.com/lf-lang/playground-lingua-franca/blob/e170f055cc84586e0a3f4158b65cbf93b7893182/examples/C/src/browser-ui/Uptime.lf#L33).  Maybe we actually need to have two variants of these constants, one with quotation marks and one without?  Seems ugly, but I can't think of any other solution.

erlingrj:
Hi Edward, thanks for your feedback. The problem you identify is solved with the companion PR here: https://github.com/lf-lang/reactor-c/pull/494

It adds back the quotation marks when they are defined as compiler flags. So the change is subtle, basically it only means that the CMake variable does not have the quotation marks and can thus conveniently be used in the `cmake-include` files, but the compiler defs will have quotation marks so that it can be conveniently used by the C code.

erlingrj:
@edwardalee I propose we merge this. Especially the changes to the CMake variables are very useful.

edwardalee (approved):
LGTM, modulo:
* get tests to pass
* merge https://github.com/lf-lang/reactor-c/pull/494#pullrequestreview-2489954035
* document the new target property in lf-lang.org.

PR #2435, opened 2024-11-14
bug fix in the Patmos platform

EhsanKhodadad:
Cmake uses -O3 when build type is "release" by default, while patmos-clang doesn't support this optimization. Therefore, I set CMAKE_C_FLAGS_RELEASE to -O2 in the cmakelist file to fix it.


erlingrj (approved):
Looks good!

erlingrj:
@EhsanKhodadad can you look into the Patmos CI failure? Dont mind the Zephyr failure is being addressed in a separate RP

Issue #2434, opened 2024-11-11
"export-dependency-graph" option is not working


keiichibando:
The LF application with `export-dependency-graph` option on C++ target generate dot file, but it doesn't contains a dependency graph.

```
target Cpp {
    timeout: 1sec,
    export-dependency-graph: true
}

reactor Source {
    output out:int;
    timer t(0, 100ms);
    reaction(t) -> out {=
        out.set(1);
    =}
}

reactor Sink {
    input in:int;
    reaction(in) {=
        std::cout << *in.get() << std::endl;
    =}
}

main reactor {
    source = new Source();
    sink = new Sink();
    source.out -> sink.in;
}
```

The compiled application generate the following dot file that doesn't contains a dependency graph.

```
digraph {
rankdir=LR;
}
```

PR #2436, opened 2024-11-14
Repaired export-dependency-graph property and dropped unsupported export-to-yaml property

cmnrd:
Fixes https://github.com/lf-lang/lingua-franca/issues/2434


erlingrj (approved):
Great that you could handle this. Looks like  `export_dependency_graph` is now called only after the program has completed? So you need to do Ctrl-C, or have timeout, or some other way of terminating the program before you get the graph?

cmnrd:
Yes, that is how it currently is and I think its good enough for now. Ideally, we would call the export before the execution starts, but currently the assembly of the port connections is tied in with the execute function. Separating it out is possible, but requires some more work.

PR #2437, opened 2024-11-18
Make macros available in methods

edwardalee:
Before this change, one could not call `lf_time_logical()` or `lf_tag()` in methods. This fixes that oversight. It also makes it possible (though not very attractive) to read inputs and produce outputs in methods by passing in the pointer to the input or output.  It's not too attractive to do this because you have to know the type of the input or output struct, but at least this makes it possible.  A test is added to verify these capabilities.


erlingrj (approved):
Looks good

erlingrj:
@edwardalee I merged it so that the Zephyr thing is in master, it seems to affect other PRs also

PR #2441, opened 2024-11-25
Python 11 and 12 support

edwardalee:
This PR makes Python versions 3.11, 3.12, and 3.13, at least, work with the Python target.
By default, the Python version is found by CMake with the constraint that it be at least 3.10.0.
This is specified by the following directive:

```
            find_package(Python 3.10.0 REQUIRED COMPONENTS Interpreter Development)
```

Because CMake does not work as advertised (see below), this PR adds a target parameter `python-version` to specify a precise version to use. For example, to use 3.13.0, you would specify:

```
target Python {
  python-version: "3.13.0"
}
```

This will change the above directive in the generated CMakeLists.txt file to read like this:

```
            find_package(Python 3.13.0 EXACT REQUIRED COMPONENTS Interpreter Development)
```

This PR fixes some memory issues, particularly with modal models, that were previously causing crashes with Python versions greater than 3.10.

---------
The rest of this text documents the issue with CMake.

There is a problem I need help with, however.  In this PR, the Python version is hardwired to 3.12.7 in the PythonGenerator.java file, which generates the CMake directives, and has this line:

```
            find_package(Python 3.12.7...<3.13.0 REQUIRED COMPONENTS Interpreter Development)
```
The problem is that if this line allows 3.10.0 up to 3.13.0, which is what we want, then lfc inevitably picks a different Python from one identified by CMake, and you get an error message like:

```
SystemError: initialization of LinguaFrancaActionDelay did not return an extension module
```

How can we force CMake to pick the Python version that is in the PATH?  Sadly, it does not do that.
This supersedes #1902, which I will close.

### DIFF
--- a/.github/workflows/lsp-tests.yml
+++ b/.github/workflows/lsp-tests.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.12"
       - name: Run language server Python tests without PyLint
         run: ./gradlew core:integrationTest --tests org.lflang.tests.lsp.LspTests.pythonValidationTestSyntaxOnly core:integrationTestCodeCoverageReport
       - name: Install pylint
--- a/.github/workflows/py-tests.yml
+++ b/.github/workflows/py-tests.yml
@@ -22,7 +22,6 @@ jobs:
     strategy:
       matrix:
         platform: ${{ (inputs.all-platforms && fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]')) || fromJSON('["ubuntu-latest"]') }}
-        python-version: ${{ (inputs.all-platforms && fromJSON('["3.10", "3.11", "3.12"]')) || fromJSON('["3.12"]') }}
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Check out lingua-franca repository
@@ -37,7 +36,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: '3.12'
       - name: Install dependencies OS X
         run: |
           brew install coreutils
--- a/.github/workflows/py-tests.yml
+++ b/.github/workflows/py-tests.yml
@@ -22,6 +22,7 @@ jobs:
     strategy:
       matrix:
         platform: ${{ (inputs.all-platforms && fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]')) || fromJSON('["ubuntu-latest"]') }}
+        python-version: ${{ (inputs.all-platforms && fromJSON('["3.10", "3.11", "3.12"]')) || fromJSON('["3.12"]') }}
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Check out lingua-franca repository
@@ -36,7 +37,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies OS X
         run: |
           brew install coreutils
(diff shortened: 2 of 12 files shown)


cmnrd:
It is not sufficient to modify your `PATH` to point to a different python version. Cmake does respect both venv and pyenv, and probably also other ways of managing Python environments.

edwardalee:
Unfortunately, it does not seem to be the case the cmake respects pyenv.  E.g., I have this:

```
 % pyenv versions
  system
* 3.10.13 (set by /Users/edwardlee/.pyenv/version)
  3.11.5
  3.12.0
  3.12.7
```

Nevertheless, cmake uses Python 3.12.7:

```
% cd ~/LinguaFranca/test/Python
% lfc-dev src/ActionDelay.lf
...
--- Executing command: cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/Users/edwardlee/LinguaFranca/test/Python -DCMAKE_INSTALL_BINDIR=bin -DLF_FILE_SEPARATOR='"/"' -DLF_SOURCE_DIRECTORY='"/Users/edwardlee/LinguaFranca/test/Python/src"' -DLF_PACKAGE_DIRECTORY='"/Users/edwardlee/LinguaFranca/test/Python"' -DLF_SOURCE_GEN_DIRECTORY='"/Users/edwardlee/LinguaFranca/test/Python/src-gen/ActionDelay"' /Users/edwardlee/LinguaFranca/test/Python/src-gen/ActionDelay
...
-- Found Python: /opt/homebrew/Frameworks/Python.framework/Versions/3.12/bin/python3.12 (found suitable version "3.12.7", required range is "3.10.0...<3.13.0") found components: Interpreter Development Development.Module Development.Embed
```

The same occurs if I run cmake manually in the src-gen directory, so this is not an lfc problem. So, empirical evidence shows that cmake does NOT respect pyenv.  Note that I'm running the latest cmake:

```
% cmake --version
cmake version 3.31.1
```

The older version before updating to this version exhibits the same behavior.

cmnrd:
Did you have a look at the [hints](https://cmake.org/cmake/help/latest/module/FindPython.html#hints) for ind_python? `Python_FIND_STRATEGY` should default to `LOCATION`. You can double check its value to make sure that cmake does not try to go for the highest version it can find. Also, relevant are `Python_FIND_VIRTUALENV` and `Python_FIND_FRAMEWORK`  which you could try to set to `FIRST` (or `ONLY`) and `LAST` respectively.

edwardalee:
Thanks for the pointer to HINTS!  However, this seems like a real mess.  Setting  `Python_FIND_STRATEGY` to `LOCATION` result in the build failing with the following message:

```
CMake Error at CMakeLists.txt:50 (find_package):
  Could not find a package configuration file provided by "Python" (requested
  version 3.10.0...<3.13.0) with any of the following names:

    PythonConfig.cmake
    python-config.cmake

  Add the installation prefix of "Python" to CMAKE_PREFIX_PATH or set
  "Python_DIR" to a directory containing one of the above files.  If "Python"
  provides a separate development package or SDK, be sure it has been
  installed.
```

According to the docs, `Python_FIND_VIRTUALENV` already defaults to `FIRST` (docs say that the behavior should be to use the virtual environment, which is not the behavior we are seeing).  Setting `Python_FIND_FRAMEWORK` to `LAST` gives the same error as above.  Hard wiring the PATH, as suggested in the error message, does not seem like the right solution.

MoezBHH:
A missing `Py_DECREF()` caused a segmentation fault in Python's C extensions. After the `PyObject_CallObject(callable, arglist)` call, we need to decrease the reference count of the `arglist` object.

Now, the required Python version is 3.10.0 and newer versions.

MoezBHH (approved):
All regression tests passed on my laptop with Ubuntu 24.10 and Python 3.12.7.

edwardalee:
All the tests also passed on my laptop with Python 3.13.0.  Any ideas why CI is stuck?

PR #2442, opened 2024-11-29
Use the NP scheduler when deadlines are present

lsk567:
This PR removes `pickScheduler()` in the C generator that picks the GEDF scheduler when deadlines are present. As [Pipeline.lf](https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/C/src/patterns/Pipeline.lf) exposes, the GEDF scheduler does not take advantage of the program's parallelism, making it more likely to cause deadline misses.

Instead, the NP scheduler is now used at all times unless other schedulers are explicitly specified.


erlingrj (approved):
Looks good!

PR #2444, opened 2024-12-06
Patmos: better compiler flags

schoeberl:


erlingrj (approved):
Looks unproblematic, will merge when CI passes

PR #2445, opened 2024-12-08
Fix FlexPRET CI

erlingrj:
This fixes a small issue in the FlexPRET CI (wonder why it worked before...). Also bumps reactor-c.


edwardalee (approved):
LGTM.  Maybe the `cd -` command was invalid?

PR #2452, opened 2024-12-31
Allow derived classes to override parameter default values

### LINKED ISSUE #2451: Can't override default parameter values in derived reactors
It would be useful to be able to override the default values of parameters when creating a derived reactor. For example,

```
target C

reactor A(x:int = 0) {
  reaction(startup) {=
    lf_print("A: %d\n", self->x);
  =}
}

reactor B(x:int = 1) extends A { }

main reactor {
  b = new B()
}
```
This gives a `duplicate member` error.  An alternative syntax would be:

```
target C

reactor A(x:int = 0) {
  reaction(startup) {=
    lf_print("A: %d\n", self->x);
  =}
}

reactor B extends A(x = 1)  { }

main reactor {
  b = new B()
}
```

edwardalee:
This fixes #2451. The validator does not currently check for some possible errors (like changing the type of a parameter), but the compiler should give reasonable error messages, so this is not a high priority.

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CParameterGenerator.java
@@ -6,6 +6,8 @@ import org.lflang.generator.ParameterInstance;
 import org.lflang.lf.Initializer;
 import org.lflang.lf.Parameter;
 
+import java.util.HashSet;
+
 /**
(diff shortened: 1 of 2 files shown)


edwardalee:
This is a small change with significant enhancement in capability for subclassing, so I'm merging after a self review.

PR #2454, opened 2025-01-05
CMake: Dont clear executable runtime path when installing

erlingrj:
While messing around with https://github.com/lf-lang/mujoco-c I was struggling with linking with an external shared object and the final LF binary not finding the shared library. It turns out that CMake by default clears the runtime path (RPATH) of an executable when it is installed (`cmake --target install`), so the binaries found in `src-gen/MyApp/build/MyApp` would run and find the shared object, but not the binary that was put in `bin`. 

This PR sets a CMake variable to disable this clearing of the RPATH. I cannot say I fully understand why this is the default behavior of CMake and if there are any pitfalls of disabling it. Maybe @tanneberger knows something?


edwardalee (approved):
I don't understand this change (as with most cmake magic), but I don't see any potential harm.

PR #2456, opened 2025-01-17
Draft: Debugging CI tests using ROS and SBT

byeonggiljun:


byeonggiljun:
@erlingrj @lsk567 @petervdonovan Hello, everyone. Does anyone know the reason for the errors below in the CI? I don't know why `rosdep` and `sbt` are suddenly not found (as seen in the master branch, it worked two weeks ago). This PR only changes the yml files to run only the failing tests.

```
sudo: rosdep: command not found
/home/runner/work/_temp/29c8f677-dc7e-4c95-b118-bf2289662091.sh: line 1: sbt: command not found
```

erlingrj:
My guess is that `ubuntu-latest`, which is what we are using all over, has been updated from Ubuntu 22.04 to 24.04. The rollout should be completed today: https://github.com/actions/runner-images/issues/10636

I don't think we should roll things back to 22.04 and instead start the process of porting all the CI jobs to 24.04.

byeonggiljun:
@erlingrj I see. Thank you so much! I also agree that we need to port the CI tests to 24.04.

PR #2458, opened 2025-01-20
Move CI to ubuntu-24.04

erlingrj:
This PR moves the CI workflows to `ubuntu-24.04`, we pin the version rather than using `ubuntu-latest`. This way we wont be caught by surprise when `ubuntu-latest` is changed. However, we need to manually move CI to future ubuntu versions.


core/src/main/java/org/lflang/LinguaFranca.xtext
```diff
@@ -237,6 +237,7 @@ Instantiation:
     ')' (('at' host=Host ';') | ';'?);
 
 Connection:
+    (attributes+=Attribute)*
```
byeonggiljun, line 240: What's the role of this change?

byeonggiljun (commented):
LGTM! Thank you so much for taking this up, @erlingrj. I just left one minor question.

core/src/main/java/org/lflang/LinguaFranca.xtext
erlingrj: Nice catch! That is a mistake and has nothing to do with this PR, thank you!

byeonggiljun, line 240:
```suggestion
```

byeonggiljun (approved):

PR #2459, opened 2025-01-21
Fix cycle checking when calculating min delay from nearest physical action

lsk567:
As the title suggests, currently, when actions for a deep loop, the recursion never terminates and leads to stack overflow. This PR addresses the issue by introducing a `visited` set.


core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -686,22 +686,44 @@ public String toString() {
...
+   *     otherwise
+   */
+  private TimeValue findNearestPhysicalActionTriggerRecursive(Set<ReactionInstance> visited, ReactionInstance reaction) {
+    System.out.println(reaction);
     TimeValue minDelay = TimeValue.MAX_VALUE;
     for (TriggerInstance<? extends Variable> trigger : reaction.triggers) {
       if (trigger.getDefinition() instanceof Action action) {
         ActionInstance actionInstance = (ActionInstance) trigger;
         if (action.getOrigin() == ActionOrigin.PHYSICAL) {
+          System.out.println("Physical action: " + action);
```
edwardalee, line 710: Suggest deleting this debug output.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -686,22 +686,44 @@ public String toString() {
...
           if (actionInstance.getMinDelay().isEarlierThan(minDelay)) {
             minDelay = actionInstance.getMinDelay();
           }
-        } else if (action.getOrigin() == ActionOrigin.LOGICAL) {
+        }
+        else if (action.getOrigin() == ActionOrigin.LOGICAL) {
+          System.out.println("Logical action: " + action);
+          System.out.println("Depends on: " + actionInstance.getDependsOnReactions());
```
edwardalee, line 717:
```suggestion
        } else if (action.getOrigin() == ActionOrigin.LOGICAL) {
```
You could also turn these into debug outputs.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -686,22 +686,44 @@ public String toString() {
...
           // Logical action
           // Follow it upstream inside the reactor
           for (ReactionInstance uReaction : actionInstance.getDependsOnReactions()) {
-            // Avoid a loop
-            if (!Objects.equal(uReaction, reaction)) {
+            // Avoid a potentially deep loop by checking the visited set.
+            if (!visited.contains(uReaction)) {
+              visited.add(uReaction); // Mark the upstream reaction as visited.
+              System.out.println("Upstream reaction: " + uReaction);
```
edwardalee, line 724: Delete or convert to debug output.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -713,7 +735,8 @@ public TimeValue findNearestPhysicalActionTrigger(ReactionInstance reaction) {
         // Outputs of contained reactions
         PortInstance outputInstance = (PortInstance) trigger;
         for (ReactionInstance uReaction : outputInstance.getDependsOnReactions()) {
-          TimeValue uMinDelay = findNearestPhysicalActionTrigger(uReaction);
+          visited.add(uReaction); // Mark the upstream reaction as visited.
```
edwardalee, line 738: Is this add premature? Won't the recursive call on the next line add it if it isn't there already?  And if added here, won't the recursive call actually not do anything?

edwardalee (changes requested):
Thanks for catching and fixing this! It would be nice to add a test case.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
lsk567: Yes, this is a good point. I've moved this to the front of the method.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -698,24 +719,64 @@ public TimeValue findNearestPhysicalActionTrigger(ReactionInstance reaction) {
...
+            TimeValue uMinDelay = findNearestPhysicalActionTriggerRecursive(visited, uReaction);
+            if (uMinDelay.isEarlierThan(minDelay)) {
+              minDelay = uMinDelay;
+            }
+          }
+        }
+        // The trigger is an input of a contained reactor. If it
+        // another upstream contained reactor connects to it, trace to
+        // the upstream contained reactor's output ports.
+        if (trigger.getDefinition() instanceof Input) {
```
lsk567, line 750: Is there a cleaner way to look at an input port (`PortInstance`) and get a connection delay from an upstream connection into the input port? @edwardalee @erlingrj

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -698,24 +719,64 @@ public TimeValue findNearestPhysicalActionTrigger(ReactionInstance reaction) {
...
+                    .filter(
+                        sRange ->
+                            sRange.destinations.stream()
+                                .map(it -> it.instance)
+                                .anyMatch(port::equals))
+                    .map(sRange -> ASTUtils.getDelay(sRange.connection.getDelay()))
+                    .filter(Objects::nonNull)
+                    .map(TimeValue::fromNanoSeconds)
+                    .findFirst()
+                    .orElse(TimeValue.ZERO);
```
lsk567, line 768: Is there a cleaner way to look at an input port (PortInstance) and get a connection delay from an upstream connection into the input port? @edwardalee @erlingrj

edwardalee: Wow that code is ugly. It's also not quite correct.  E.g., suppose you have hierarchy:
```
   p1 ----> p2 ---/D/---> p3
```
this will return TimeValue.ZERO, I think, ignoring D.
Are you trying to get the minimum tag offset along the path from the eventual source to the port?
This is pretty tricky.  Something like this was done by @byeonggiljun in CExtensionUtils.java, but there is a telling comment there:

```
        // The minimum delay calculation needs to be made in the C code because it
        // may depend on parameter values.
```
Also, if the port is a multiport, there could be multiple delays to it.

lsk567: > Are you trying to get the minimum tag offset along the path from the eventual source to the port?

Yes, that is precisely what `findNearestPhysicalActionTrigger` was designed to do, but it wasn't doing it thoroughly.

Here is the test case I designed. The current master branch would falsely report `10 sec` when tracing back from `out2`.

TriggerLoop

The fix in this PR now correctly identifies that the minimum path delay from `out2` to the physical action is `2 sec`, not `10 sec`:
```
 --> test/C/src/federated/TriggerLoop.lf:23:3
   |
22 | reactor Hard {
23 |   output out2: int
   |   ^^^^^^^^^^^^^^^^ Found a path from a physical action to output for reactor "Hard".
The amount of delay is 2 sec.
With centralized coordination, this can result in a large number of messages to the RTI.
Consider refactoring the code so that the output does not depend on the physical action,
or consider using decentralized coordination. To silence this warning, set the target
parameter coordination-options with a value like {advance-message-interval: 10 msec}
```

But it has to do some gymnastics since it seems hard to get the connection delay value from an input port (the `in` port of the left `Delay`, of type `RuntimeRange.Port`). So far I have only found a way to access the delay from `SendRange`, which requires tracing from the input port to the upstream output port, hence the ugliness.

Perhaps a long-term solution is to refactor `RuntimeRange` to make accessing connection delays easier.

> Something like this was done by @byeonggiljun in CExtensionUtils.java, but there is a telling comment there

If the parameters are not passed in on the command line, which should not be possible now, I think it should be okay?

> Also, if the port is a multiport, there could be multiple delays to it.

That's true. I will augment the test case to check for this.

lsk567:
The test case has been augmented to include multiports.

TriggerLoop

`lfc` correctly reports the minimal path delay from a physical action to an output port, which is 2 sec in this case.
```
24 | reactor Hard {
25 |   output out2: int
   |   ^^^^^^^^^^^^^^^^ Found a path from a physical action to output for reactor "Hard".
The amount of delay is 2 sec.
With centralized coordination, this can result in a large number of messages to the RTI.
Consider refactoring the code so that the output does not depend on the physical action,
or consider using decentralized coordination. To silence this warning, set the target
parameter coordination-options with a value like {advance-message-interval: 10 msec}
```

I think this PR is now ready for review.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
erlingrj: > > Something like this was done by @byeonggiljun in CExtensionUtils.java, but there is a telling comment there
> 
> If the parameters are not passed in on the command line, which should not be possible now, I think it should be okay?
> 

It doesn't work with the usual LF parameters, try making the connection delay a parameter, this algorithm will use the initializer of that parameter as the delay always. 

This should be solvable in LFC  because it doesn't depend on any target code.

erlingrj (dismissed):
Did not closely inspect the implementation, but verified that it doesn't work if delays are based on parameters

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -698,24 +719,69 @@ public TimeValue findNearestPhysicalActionTrigger(ReactionInstance reaction) {
...
           // Follow it upstream inside the reactor
           for (ReactionInstance uReaction : actionInstance.getDependsOnReactions()) {
-            // Avoid a loop
-            if (!Objects.equal(uReaction, reaction)) {
+            // Avoid a potentially deep loop by checking the visited set.
+            if (!visited.contains(uReaction)) {
               TimeValue uMinDelay =
-                  actionInstance.getMinDelay().add(findNearestPhysicalActionTrigger(uReaction));
+                  actionInstance
+                      .getMinDelay()
```
edwardalee, line 726: This doesn't look quite right because this minDelay of this action will get ignored if the uReaction has been previously visited via some other path.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -698,24 +719,69 @@ public TimeValue findNearestPhysicalActionTrigger(ReactionInstance reaction) {
...
+                                .map(it -> it.instance)
+                                .anyMatch(port::equals))
+                    .map(sRange -> ASTUtils.getDelay(sRange.connection.getDelay()))
+                    .filter(Objects::nonNull)
+                    .map(TimeValue::fromNanoSeconds)
+                    .findFirst()
+                    .orElse(TimeValue.ZERO);
+
+            for (ReactionInstance uReaction : upstreamPort.getDependsOnReactions()) {
+              if (!visited.contains(uReaction)) {
```
edwardalee, line 776: Again, I'm not sure this should be skipped just because the upstream has been visited. There may be multiple paths to the upstream.  I suggest replacing `visited` with a set that contains only the "in progress" reactions (those for which we are inside a recursive call to find their delay). Then also cache the values computed for each reaction (just for efficiency, not needed for correctness).

edwardalee (changes requested):
I think there are still corner cases that will cause this to fail, but it looks relatively easy to fix.

edwardalee (approved):
I have a made a pass on this PR and am approving it. @lsk567 please have a look.

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
edwardalee: This is resolved in 083c7081c9d1bebd6a815ff105eb24a3f616f21e

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
edwardalee: I have added support in PortInstance and ReactorInstance for calculating these delays. The use parameters for delays is resolved in 6639b32371520fa97e5297d3311f53d309747125

lsk567:
Many thanks, @edwardalee, for helping out with this PR🙏!

I added a test case `DelayToPhysicalActionParallelPaths.lf` to test if the algorithm can handle parallel paths, as we discussed in a February meeting, where my previous implementation terminates early. Your commits fix the issue.

It also seems that parameters are not yet working, as @erlingrj suggested. I tried replacing line 59 of `DelayToPhysicalAction.lf` to `r = new Wrapper(delay = 1 s)`, and the toolchain currently reports
```
 --> test/C/src/federated/DelayToPhysicalAction.lf:39:3
   |
38 | reactor Wrapper(delay: time = 10 ms) {
39 |   output out: int
   |   ^^^^^^^^^^^^^^^ Found a path from a physical action to output for reactor "Wrapper".
The amount of delay is 10 msec.
With centralized coordination, this can result in a large number of messages to the RTI.
Consider refactoring the code so that the output does not depend on the physical action,
or consider using decentralized coordination. To silence this warning, set the target
parameter coordination-options with a value like {advance-message-interval: 10 msec}
```
But fixing this might not be trivial...

edwardalee:
Ah, good catch @lsk567 ! Indeed, I did fix it so parameters were working, but not at the top level of the federate because the original code created a new ReactorInstance from scratch without using the Instantiation.  I have fixed that now by using the Instantiation.

Note that our tests are not checking. directly whether the calculated delay is correct. I'm not sure how to do that.  But if they are not correct, then STP violations become possible (though not ensured).

core/src/main/java/org/lflang/federated/generator/FederateInstance.java
```diff
@@ -702,49 +697,103 @@ public String toString() {
...
-              if (uMinDelay.isEarlierThan(minDelay)) {
-                minDelay = uMinDelay;
-              }
-            }
-          }
-        }
+   * @param output The output port.
+   */
+  private TimeValue minDelayFromPhysicalActionTo(
+      Map<ReactionInstance, TimeValue> visited, PortInstance output) {
```
lsk567, line 706:
```suggestion
   * @param port A port (either input or output) wrt which the min delay is a calculated.
   */
  private TimeValue minDelayFromPhysicalActionTo(
      Map<ReactionInstance, TimeValue> visited, PortInstance port) {
```

I suggest this name change because this function can be called on input ports a few lines down in `var uDelay = minDelayFromPhysicalActionTo(visited, upstreamPort.instance);`

core/src/main/java/org/lflang/generator/PortInstance.java
```diff
@@ -185,6 +198,50 @@ public List<RuntimeRange<PortInstance>> getDependsOnPorts() {
...
+    var connections = connectionsFrom(upstreamPort);
+    if (connections == null) return result;
+    for (var connection : connections) {
+      var delay = connection.getDelay();
+      if (delay != null) {
+        // The delay may reference a parameter defined in the parent of the connection,
+        // which may not be the same as the parent of the port, so resolve the delay using that
+        // parent.
+        var parent = upstreamPort.parent;
+        if (!parent.getDefinition().equals(connection.eContainer())) {
```
lsk567, line 230: This condition might be `true` _always_ because a port's parent reactor will not define an outside connection's parameter. But this can be a little subtle. I only realize this when seeing the diagram of `DelayToPhysicalAction.lf`.

lsk567:
> Note that our tests are not checking. directly whether the calculated delay is correct. I'm not sure how to do that. But if they are not correct, then STP violations become possible (though not ensured).

@edwardalee With the existing test harness, I also don't have good solutions for programmatically checking the correctness of compiler warnings. My workaround for now in `DelayToPhysicalActionWithLoops.lf` is adding a `@label` that says what kind of warning the compiler should produce.

I suggest we merge this PR for now and, in the future, support a developer-facing annotation in the LF program, such as `@_assert(<interval_var_name>, <expected_var_val>)`, that checks the values of intermediate variables (e.g., minDelayFromPhysicalAction in this case) during compilation.

core/src/main/java/org/lflang/generator/PortInstance.java
edwardalee: Actually, I think it can be either. `upstreamPort` may be an input, and then the `connection` will belong to the same parent as the port. But if `upstreamPort` is an output, then the connection belongs to the parent of the parent of the port. The problem is that we can't get the correct *instance* of the parent from the connection. The connection represents a syntactic object in the AST, but there can be many *instances* of this syntactic object. If we don't get the right instance, then we may not get the correct parameter values (because parameters can be overridden on a per-instance basis).

lsk567: I see. Yes, if `upstreamPort` is an input, then the connection will belong to the same parent.

PR #2463, opened 2025-01-27
[Ignore me] Revert "A target property for handling `downstream next event tag` (`DNET`) signal"

byeonggiljun:
Reverts lf-lang/lingua-franca#2400 as `reactor-c` is facing an error.


byeonggiljun:
As I mentioned in [reactor-c#511](https://github.com/lf-lang/reactor-c/pull/511), I could reproduce the same error even in this branch, either. Thus, I guess the problem is not because of the merge but just revealed due to its flakiness. I'll update the cause of the problem soon.

PR #2464, opened 2025-01-27
Revert "Set annotated layout options for modes."

erlingrj:
This reverts commit 02f965ae6e23db79807cb58528e3a0d46ad4969e.

Thanks @soerendomroes for identifying this. This PR simply reverts this commit. Do we lose any functionality by reverting this? I am gonna look into why this wasn't caught by the CI earlier. See #2465


soerendomroes:
Nothing important. By reverting this, you will not be able to set layout constraints on nodes.

PR #2466, opened 2025-02-02
Apply pluggable security (TLS, SST) for federated execution

Jakio815:
Please check descriptions in lf-lang/reactor-c#515


core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -456,17 +558,6 @@ private String getRemoteLaunchCode(
...
-        "    if ! command -v RTI &> /dev/null",
-        "    then",
-        "        echo \"" + ANSI_ERROR + "RTI could not be found." + ANSI_RESET + "\"",
-        "        echo \""
-            + ANSI_ERROR
-            + "The source code can be found in org.lflang/src/lib/core/federated/RTI"
-            + ANSI_RESET
-            + "\"",
-        "        exit 1",
-        "    fi",
```
Jakio815, line 469: Removed this part, because this assumes RTI binary is installed on the PATH of the system. The RTI is newly compiled everytime, so I deleted this part.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -441,9 +544,9 @@ private String getRemoteLaunchCode(
         "ssh " + target + " 'mkdir -p log; \\",
         "    echo \""
             + ANSI_INFO
-            + "-------------- Federation ID: \"${FEDERATION_ID}\""
+            + "-------------- Federation ID: "
             + ANSI_RESET
-            + "\" >> "
+            + "\"'$FEDERATION_ID' >> "
```
Jakio815, line 581: Wasn't working before, just a minor bug fix.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -584,13 +710,7 @@ private String getDistCode(Path remoteBase, String name, String user, String hos
             + "\" >> "
             + buildShellFileName
             + "; "
-            // The >> syntax appends stdout to a file. The 2>&1 appends stderr to the same file. tee
-            // sens to stdout and file.
-            + "echo \"rm -rf build && mkdir -p build && cd build && cmake .. && make 2>&1 | tee -a "
-            + remoteBuildLogFileName
-            + "\" >> "
-            + buildShellFileName
-            + "; "
```
Jakio815, line 593: Moved to block above.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -191,7 +206,7 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
                 "\n",
                 "echo \""
                     + ANSI_INFO
-                    + "RTI has exited. Wait for federates to exit."
```
Jakio815, line 194: At this point, the RTI has not exited yet. Fixed the printed statements.

.github/workflows/c-tests.yml
```diff
@@ -54,6 +54,16 @@ jobs:
       - name: Install RTI
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
+      - name: Clone SST repository and C library for SST federated tests
+        run: |
+          git clone https://github.com/iotauth/iotauth.git iotauth
+          echo "SST_ROOT=$GITHUB_WORKSPACE/iotauth" >> $GITHUB_ENV
+          git clone https://github.com/iotauth/sst-c-api.git sst-c-api
+          cd sst-c-api
```
hokeun: This seems problematic because it can lead to inconsistent versions between the Auth and the C API.
This should be 
```
git submodule update --init --recursive
cd entity/c
```
to be consistent with the version of iotauth, not depending on the submodule version individually.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -219,6 +221,15 @@ public boolean doGenerate(Resource resource, LFGeneratorContext context) throws
     // Compile an RTI for this federation.
     buildRtiLocally(context);
 
+    // If communication mode is SST, generate configurations for SST.
```
hokeun: This comment should mention both SST and TLS.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -249,12 +260,24 @@ private void buildRtiLocally(LFGeneratorContext context) {
...
-            "cmake",
-            List.of("-Bbuild", "-DCMAKE_INSTALL_PREFIX=" + fileConfig.getGenPath(), "."),
-            false,
-            fileConfig.getRtiSrcGenPath());
+
+    var configureArgs = new java.util.ArrayList<String>();
+    configureArgs.add("-Bbuild");
+    configureArgs.add("-DCMAKE_INSTALL_PREFIX=" + fileConfig.getGenPath());
+
+    // If communication mode is SST, the RTI must be built with -DCOMM_TYPE=SST.
```
hokeun: Ditto, both SST and TLS should be mentioned (see `else if`).

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
```diff
@@ -0,0 +1,832 @@
...
+import org.lflang.generator.LFGeneratorContext;
+import org.lflang.target.property.SSTProperty;
+import org.lflang.util.FileUtil;
+
+/**
+ * SST related methods.
+ *
+ * @author Dongha Kim
+ */
+public class SSTGenerator {
```
hokeun, line 30: This class seems to only include static methods and is not instantiated with `new`. However, looking at the usage of this class, it will make more sense to make this instantiable with configurations stored as member variables (e.g., `FederationFileConfig fileConfig`), rather than giving them every time you call each method. Please refactor both SSTGenerator and TLSGenerator classes so that they can be instantiated

core/src/main/java/org/lflang/federated/generator/TLSGenerator.java
```diff
@@ -0,0 +1,181 @@
...
+import org.lflang.generator.LFGeneratorContext;
+import org.lflang.util.FileUtil;
+
+/**
+ * TLS related methods (cert/key generation + copy to src-gen).
+ *
+ * - Generate cert+key for RTI and each federate into: fed-gen/[program]/credentials/[name]/
+ * - Copy only the needed pair into: src-gen/[entity]/credentials/
+ */
+public class TLSGenerator {
```
hokeun, line 20: Same as the SSTGenerator class, please make this instantiable and change methods to non-static.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -191,7 +206,7 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
...
+                    + "All components launched. Waiting for federates to exit..."
```
hokeun: This is confusing. Why are we waiting for federates to exit after all components have launched? Shouldn't we continue?

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -322,13 +358,15 @@ private String getSetupCode() {
         "# Use two distinct traps so we can see which signal causes this.",
         "cleanup() {",
         "    if [ \"$EXITED_SUCCESSFULLY\" = true ] ; then",
+        killAuthCommand,
```
hokeun: This architecture is confusing. Shouldn't this be added only when SST is used?

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -322,13 +358,15 @@ private String getSetupCode() {
...
         "        exit 0",
         "    else",
         "        printf \"Killing federate %s.\\n\" ${pids[*]}",
         "        # The || true clause means this is not an error if kill fails.",
         "        kill ${pids[@]} || true",
         "        printf \"#### Killing RTI %s.\\n\" ${RTI}",
         "        kill ${RTI} || true",
+        killAuthCommand,
```
hokeun: Also, the indentation seems off.

Jakio815: The formatting is done by the `./gradlew :core:spotlessApply` command...

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
Jakio815: `killAuthCommand` is initialized as an empty string ("") by default. So in other modes (TCP, TLS) it will be a blank line.

I done this because this was the most non-invasive way to handle this conditional Auth cleanup logic inside a static `String.join` block.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
Jakio815: Fixed on 0e0719c9b3d6ab7751c54ebf97170dc1ac3d6142.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
Jakio815: Fixed on ccc833c96d790fb0b614097a6e6a0d49035a35d0.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
Jakio815: I agree, it looked confusing. This part was where the launch script is waiting for the processes (federates) to finish. The processes were already launched before, I fixed it to

`Waiting for federates to complete...` Does this look better?

Or should I just delete this? Fixed in 4d670797a32bb70776783284c210cd08517c6e83.

.github/workflows/c-tests.yml
Jakio815: Fixed in b8efb1888fcd4e53b2ca3a89b24eff37f325ff8a

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
Jakio815: Fixed in 6e9c4808511e1c7195fad2bd7d775a1c5dbc47ba. This was a good suggestion!

core/src/main/java/org/lflang/federated/generator/TLSGenerator.java
Jakio815: Fixed in b5bbed7d4aa878e65cd5c1959911c9ca5627399b

I removed all unnecessary static functions, however still kept path-utility methods (e.g., `getLocalCertPath`) as `static` because they are referenced statically by external builder/launcher classes (such as `CBuildConfig` and `FedLauncherGenerator`) that do not possess a generator instance.

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
Jakio815: As suggested, I updated the generator to gracefully stop the Auth process by sending a 'quit' command instead of directly killing it.

Here is how the architecture works now:

1. When SST mode is enabled, the launcher creates a unique named pipe (FIFO) locally or on the remote machine (`auth_input_${FEDERATION_ID}.fifo`).
2. The standard input of the Auth process is redirected to read from this FIFO.
3. During the `cleanup()` trap phase (triggered on both successful and unsuccessful exits), the launcher writes 'quit' into the FIFO to signal the Auth process to exit gracefully, and then unlinks/removes the FIFO file.
4. The aggressive process `kill` commands have been completely removed.

hokeun: Ok, understood. I understand you will be using the new quit command for terminating Auth. Can you also rename the variable to `quitAuthCommand` or `terminateAuthCommand` rather than `kill`? `kill` sounds like an error exit.

hokeun: Also I suggest you do something like `isSST ? quitAuthCommand : ""` to make the code more readable, rather than implicitly setting quitAuthCommand to an empty string.

Jakio815: Fixed on 17404ec80ecb749b9715b0a29ee7e21518e34e3d and c133eb7b651577b1d334a1da8e0fde72f3e60872.

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
```diff
@@ -0,0 +1,826 @@
...
+          .nowhere()
+          .info("Auth necessary files copied into: " + fileConfig.getSSTAuthPath().toString());
+      updatePropertiesFile(
+          fileConfig.getSSTAuthPath().resolve("properties"),
+          fileConfig.getSSTAuthPath().toString());
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    // Generate SST config for the rti.
```
hokeun: Nit: rti -> RTI

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
```diff
@@ -0,0 +1,826 @@
...
+                  + getSSTConfig(fileConfig, federate.name).toString());
+    }
+
+    // Copy the configs and credentials of rti and federates, to the src-gen for tar deployments.
+    copyAuthAndConfigsAndKeys(federates, usePermanentDistKey);
+  }
+
+  public static Path getSSTConfig(FederationFileConfig fileConfig, String name) {
+    return fileConfig.getSSTConfigPath().resolve(name + ".config");
+  }
```
hokeun: Nit: should be "getSSTConfigPath" because the return type is the path.

hokeun (approved):
Looks good to me. I left some additional but minor comments. Please make sure to go over all comments and see if there are any missing/incorrect comments. Please get approval from @edwardalee before merging.

test/C/src/federated/DistributedCountTLS.lf
```diff
@@ -0,0 +1,46 @@
...
+    }
+    self->c++;
+  =}
+
+  reaction(shutdown) {=
+    if (self->c != 6) {
+      lf_print_error_and_exit("Expected to receive 5 items. Received %d.", self->c - 1);
+    }
+  =}
+}
```
edwardalee:
```suggestion
import Print from "DistributedCount.lf"
```

test/C/src/federated/DistributedCountDecentralizedTLS.lf
```diff
@@ -0,0 +1,60 @@
...
+  =}
+
+  reaction(shutdown) {=
+    if (self->c != 6) {
+      fprintf(stderr, "Expected to receive 5 items.\n");
+      exit(2);
+    }
+    printf("SUCCESS: Successfully received 5 items.\n");
+  =}
+}
```
edwardalee:
```suggestion
import Print from "DistributedCountDecentralized.lf"
```

test/C/src/federated/sst/DistributedCountSSTPermanentDistKey.lf
```diff
@@ -0,0 +1,52 @@
...
+    }
+    self->c++;
+  =}
+
+  reaction(shutdown) {=
+    if (self->c != 6) {
+      lf_print_error_and_exit("Expected to receive 5 items. Received %d.", self->c - 1);
+    }
+  =}
+}
```
edwardalee:
```suggestion
import Print from "../DistributedCount.lf"
```

test/C/src/federated/sst/DistributedCountSST.lf
```diff
@@ -0,0 +1,48 @@
...
+    }
+    self->c++;
+  =}
+
+  reaction(shutdown) {=
+    if (self->c != 6) {
+      lf_print_error_and_exit("Expected to receive 5 items. Received %d.", self->c - 1);
+    }
+  =}
+}
```
edwardalee:
```suggestion
import Print from "../DistributedCount.lf"
```

test/C/src/federated/sst/DistributedCountDecentralizedSST.lf
```diff
@@ -0,0 +1,63 @@
...
+  =}
+
+  reaction(shutdown) {=
+    if (self->c != 6) {
+      fprintf(stderr, "Expected to receive 5 items.\n");
+      exit(2);
+    }
+    printf("SUCCESS: Successfully received 5 items.\n");
+  =}
+}
```
edwardalee:
```suggestion
import Print from "../DistributedCountDecentralized.lf"
```

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -165,6 +173,7 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
                 getDistCode(rtiConfig.getDirectory(), federate.name, federate.user, federate.host))
             .append("\n");
         shCode
+            // TODO: Need to fix calling rtiConfig.getDirectory() here
```
edwardalee: Unclear "TODO". What is this about?

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -492,6 +617,27 @@ private String getDistCode(Path remoteBase, String name, String user, String hos
     String remoteBuildLogFileName = logDirectory + "/build.log";
     String buildShellFileName = "build_" + name + ".sh";
     String tarFileName = name + ".tar.gz";
+
+    // ---- Add this block ----
```
edwardalee: What is this "Add this block" comment for?

edwardalee (approved):
LGTM.  Some suggestions included.
Can we add some Python tests to make sure this works with Python federates?
I still need to look at the reactor-c PR.
I also requested a Copilot review, which will undoubtedly raise some issues.

.github/workflows/c-tests.yml
```diff
@@ -54,6 +54,17 @@ jobs:
       - name: Install RTI
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
+      - name: Clone SST repository and C library for SST federated tests
+        run: |
+          git clone https://github.com/iotauth/iotauth.git iotauth
+          echo "SST_ROOT=$GITHUB_WORKSPACE/iotauth" >> $GITHUB_ENV
+          cd iotauth
+          git submodule update --init --recursive
+          cd entity/c
```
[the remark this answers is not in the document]
Jakio815: Will be done in the next PR as soon as the next iotauth release happens.

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
```diff
@@ -0,0 +1,826 @@
...
+    String networkProtocol = "TCP";
+
+    // Create the configuration content
+    StringBuilder configContent = new StringBuilder();
+    configContent
+        .append("entityInfo.name=")
+        .append(entityName)
+        .append("\n")
+        .append("entityInfo.purpose={\"group\":\"RTI\"}\n")
+        .append("entityInfo.number_key=1\n")
```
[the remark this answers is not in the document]
Jakio815: Intended.

core/src/main/java/org/lflang/target/Target.java
```diff
@@ -550,8 +550,11 @@ public void initialize(TargetConfig config) {
               ClockSyncOptionsProperty.INSTANCE,
               CmakeArgsProperty.INSTANCE,
               CmakeIncludeProperty.INSTANCE,
+              CommunicationModeProperty.INSTANCE,
+              SSTProperty.INSTANCE,
               CmakeInitIncludeProperty.INSTANCE,
               CommunicationModeProperty.INSTANCE,
+              SSTProperty.INSTANCE,
```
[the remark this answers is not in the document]
Jakio815: Fixed 64d48c1aedc98c9d60e55d301b65f1ba9b9c3456.

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
```diff
@@ -0,0 +1,826 @@
...
+    entityList.add(rti);
+
+    // Federate entities
+    for (FederateInstance federate : federateInstances) {
+      String federateName = federate.name;
+      JsonObject entity =
+          createEntity(
+              "Federates", "net1." + federateName, "Net1." + federateName, usePermanentDistKey);
+      entityList.add(entity);
+    }
```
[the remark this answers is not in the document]
Jakio815: Fixed 23e80ce255d01a59454893cdd8e77c45aadad5a1

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -269,82 +284,144 @@ public void doGenerate(List<FederateInstance> federates, RtiConfig rtiConfig) {
...
+
+    String launchCodeWithLogging = String.join(" ", authLaunchCode, ">& auth.log &");
+    String launchCodeWithoutLogging = String.join(" ", authLaunchCode, "&");
+
+    return String.join(
+        "\n",
+        "\n# Prompt for the password before starting SST Auth",
+        "echo \"Executing Auth.\"",
+        "# Launch the SST Auth.",
+        "if [ \"$1\" = \"-l\" ]; then",
```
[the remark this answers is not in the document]
Jakio815: Fixed 1e3c2fee7acf39411dbd148e51b35689efdd6685

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -413,11 +414,28 @@ CodeBuilder generateCMakeCode(
...
+        // If communication mode is SST, find sst package.
+        cMakeCode.pr("# sst-c-api is already linked transitively via reactor-c.");
+      } else if (targetConfig.get(CommunicationModeProperty.INSTANCE) == CommunicationMode.TLS) {
+        // TLS requires OpenSSL only
+        cMakeCode.pr("# Find OpenSSL for TLS support");
+        cMakeCode.pr("find_package(OpenSSL REQUIRED)");
+        cMakeCode.pr(
+            "target_link_libraries(${LF_MAIN_TARGET} PRIVATE OpenSSL::SSL OpenSSL::Crypto)");
+        cMakeCode.newLine();
+      }
```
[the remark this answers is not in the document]
Jakio815: Fixed here. a20bca00bc7cc01dbdff4f39e73841eac6d67c90

Updated the whole check part of setting openssl_root_dir, when mac, including intel mac and apple mac.

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
Jakio815: Fixed 6357e31dff7379af1616c2ea90d2d55b63552345 with the comment below.

core/src/main/java/org/lflang/federated/generator/SSTGenerator.java
Jakio815: Fixed 6357e31dff7379af1616c2ea90d2d55b63552345

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
Jakio815: Removed this comment and the one below here. b48de9b56d01e1b12c7551c3c97bf0ec171f6b44

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
Jakio815: Same as above.

Jakio815:
I verified if TLS and SST mode was actually encrypting the communication.

I tested using the federation_id sent in plaintext in TCP mode.

image
You can check that it shows in plaintext as "test_federation_id_if_plaintext"

However, when I use SST or TLS mode,
image
All bytes are encrypted, and you can also check that the length of the payload is different showing in "Len".

Jakio815:
> LGTM. Some suggestions included. Can we add some Python tests to make sure this works with Python federates? I still need to look at the reactor-c PR. I also requested a Copilot review, which will undoubtedly raise some issues.

@edwardalee 
I added two more tests in the python test. It's just the same `DistributedCountSST.lf` and `DistributedCountTLS.lf` just python versions.

I added a block for the `python-tests.yml` to install required libraries.

Changes are ab48f55e1879f5b4142fc780304988180b783012.

PR #2468, opened 2025-02-02
Update rust runtime version.

Jakio815:
Update rust runtime version.


erlingrj (changes requested):
The commit hash in runtime-version.properties must match the commit hash of the reactor-rs submodule!

Jakio815:
@erlingrj Thanks for checking. Why don't we also merge lf-lang/reactor-rs#50?

erlingrj:
It was now just merged. Could you update the submodule pointer and then run `./gradlew updateRustRuntime`?

erlingrj (approved):
Thanks for fixing this!

PR #2471, opened 2025-02-07
Patmos can compile LF codes with its prev compiler

EhsanKhodadad:
Patmos can compile LF-generated c codes with its previous compiler version. We will change it again when the issue with the new compile is resolved.

### DIFF
--- a/.github/actions/setup-patmos/action.yml
+++ b/.github/actions/setup-patmos/action.yml
@@ -28,7 +28,7 @@ runs:
         mkdir ~/t-crest
         cd ~/t-crest
         git clone https://github.com/t-crest/patmos-misc.git misc
-        ./misc/build.sh
+        ./misc/build.sh toolchain1
         patmos-clang --version
         which patmos-clang
       shell: bash
--- a/.github/workflows/c-embedded.yml
+++ b/.github/workflows/c-embedded.yml
@@ -25,5 +25,5 @@ jobs:
 #     uses: ./.github/workflows/c-flexpret-tests.yml
 
       # Run the C Patmos integration tests.
-#  patmos:
-#    uses: ./.github/workflows/c-patmos-tests.yml
+  patmos:
+    uses: ./.github/workflows/c-patmos-tests.yml


.github/workflows/c-patmos-tests.yml
```diff
@@ -43,12 +43,20 @@ jobs:
...
             --tests org.lflang.tests.runtime.CPatmosTest.* \
             core:integrationTestCodeCoverageReport
           rm -rf test/C/src-gen
+      - name: Smoke tests
+        run: |
+          for file in $(find test/C/bin -type f -name '*Patmos*'); do
+            echo "Inspecting file: $file"
+            file "$file"
+            echo "Running pasim on $file"
+            pasim "$file" || echo "Error running pasim on $file"
```
lhstrh, line 58: Would it not be better to override `getExecCommand` in `CPatmosTest`? That way the errors get caught in the testing framework rather than in a separate CI step. To make this work, you'd have to rename `buildPatmosUnthreadedTests` to `runPatmosUnthreadedTests` and change the test level accordingly.

core/src/integrationTest/java/org/lflang/tests/runtime/CPatmosTest.java
```diff
@@ -40,11 +40,24 @@ public CPatmosTest() {
...
-  public void buildPatmosBasicTestsUnthreaded() {
+  public void buildPatmosUnthreadedTests() {
+    Assumptions.assumeTrue(isLinux(), "Patmos tests only run on Linux");
+    super.runTestsFor(
+        List.of(Target.C),
+        Message.DESC_PATMOS,
+        TestCategory.PATMOS::equals,
+        Transformers::noChanges,
+        Configurators::makePatmosCompatibleUnthreaded,
+        TestLevel.BUILD,
```
lhstrh, line 51: Consider running the tests instead of just building them and testing them later.

core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
```diff
@@ -237,13 +237,13 @@ CodeBuilder generateCMakeCode(
         break;
       case PATMOS:
         cMakeCode.newLine();
-        cMakeCode.pr("SET(CMAKE_SYSTEM_NAME patmos)");
-        cMakeCode.pr("SET(CMAKE_SYSTEM_PROCESSOR patmos)");
+        // cMakeCode.pr("SET(CMAKE_SYSTEM_NAME patmos)");
+        // cMakeCode.pr("SET(CMAKE_SYSTEM_PROCESSOR patmos)");
```
lhstrh, line 241: Why are these commented out? Could they just be removed?

lhstrh (commented):
Provided some comments.

erlingrj (changes requested):
Oops, my review was never submitted and left pending... Sorry about that, nice work with some improvements we can get it merged!

EhsanKhodadad:
Hi, 

@erlingrj and @lhstrh, could you review this PR? 

Thanks

lhstrh (approved):
LGTM! The only thing I suggest you change is the name/description of the PR because it's unclear. What is "prev compiler"?

edwardalee (approved):
I suggest changing the title to something like: Upgrade regression tests to compile and run PATMOS tests.

Is this an accurate description?

Issue #2475, opened 2025-03-04
Deadlock was occurred with hierarchical reactor


keiichibando:
The following source code will occurred deadlock.
The expected behavior is that B and C will be executed after A_Source reaction.
But the B and C reaction will not be executed after A_Source reaction. (There is no reaction after A_Source.)

It looks to me like that LF is deadlocked.

```
target C {
    timeout: 300ms,
    tracing: true,
    trace-plugin: telegraf_lf_trace_plugin,
    docker: {
        builder-base: "xronosinc/lf-trace-plugin-api:c",
    },
    workers: 1
}

reactor A_Source {
    timer t(0, 25ms);
    output out:int;

    reaction(t) -> out {=
        lf_print("[A_Source][Exec]");
        lf_set(out, 1);
    =}
}

reactor A_Sink {
    input in:int;

    reaction(in) {=
        lf_print("[A_Sink][Exec]");
    =}
}

reactor A {
    output out:int;
    input in:int;

    a_source = new A_Source();
    a_sink = new A_Sink();

    a_source.out -> out;
    in -> a_sink.in;
}

reactor B {
    input in:int;
    output out:int;

    reaction(in) -> out {=
        lf_print("[B][Exec]");
        lf_set(out, 1);
    =}
}

reactor C {
    input in1:int;
    input in2:int;
    output out:int;

    reaction(in1, in2) -> out {=
        lf_set(out, 1);
        lf_print("[C][Exec]");
    =}
}

federated reactor {
    a = new A();
    b = new B();
    c = new C();

    a.out -> b.in;
    a.out -> c.in1;
    b.out -> c.in2 after 25ms;
    c.out -> a.in;
}
```

Execution log:
```
 ✔ Container main1_devide-federate__a  Recreated                                                                                                                                                                                                                                                                   0.3s 
 ✔ Container main1_devide-federate__b  Recreated                                                                                                                                                                                                                                                                   0.3s 
 ✔ Container main1_devide-federate__c  Recreated                                                                                                                                                                                                                                                                   0.3s 
 ✔ Container main_pcs3_devide-rti      Recreated                                                                                                                                                                                                                                                                   0.2s 
Attaching to main1_devide-federate__a, main1_devide-federate__b, main1_devide-federate__c, main1_devide-rti
main1_devide-federate__a  | Federation ID for executable ./bin/federate__a: 1
main1_devide-federate__a  | ---- System clock resolution: 1 nsec
[...]

PR #2477, opened 2025-03-07
Fix federated ZDC detection algorithm

erlingrj:
Thanks to @byeonggiljun for debugging this. There was a bug in the ZDC detection algorithm where a node is added to the set of visited when we have a delayed connection to it. This made us skip this node later when we had an actual zero-delay connection. The fix is to only add nodes that you actually visit to the visited set.

Fixes https://github.com/lf-lang/lingua-franca/issues/2475


edwardalee (approved):
Wow, nice catch @byeonggiljun and @erlingrj !

PR #2479, opened 2025-03-11
Cache action bumped to v4

lhstrh:
Response to failing workflow run: https://github.com/lf-lang/lingua-franca/actions/runs/13735555422


erlingrj:
Hm, there might be some pip caching issue exposed by updating this action, see: https://github.com/lf-lang/benchmarks-lingua-franca/issues/68

This might fix it: https://github.com/lf-lang/benchmarks-lingua-franca/pull/69

PR #2482, opened 2025-03-23
Fed decentralized improvements

edwardalee:
This PR fixes some issues with federated code generation that make decentralized coordination harder to use:

1. Fixes an overflow error when an STA or STAA of `forever` is used.
2. Removes an unnecessary reaction to `startup` in the network receiver reactors. Such reactions are problematic for decentralized coordination because if a federate has no startup reactions of its own and no timers or logical actions, then it can safely set the `STA` to zero.  But not if it has `startup` reactions.

### DIFF
--- a/core/src/main/java/org/lflang/federated/extensions/CExtensionUtils.java
+++ b/core/src/main/java/org/lflang/federated/extensions/CExtensionUtils.java
@@ -100,7 +100,7 @@ public class CExtensionUtils {
    */
   public static String stpStructs(FederateInstance federate) {
     CodeBuilder code = new CodeBuilder();
-    federate.staaOffsets.sort((d1, d2) -> (int) (d1.time - d2.time));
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/lflang/federated/extensions/CExtensionUtils.java
```diff
@@ -100,7 +100,12 @@ public static String initializeTriggersForNetworkActions(
    */
   public static String stpStructs(FederateInstance federate) {
     CodeBuilder code = new CodeBuilder();
-    federate.staaOffsets.sort((d1, d2) -> (int) (d1.time - d2.time));
```
erlingrj, line 103: Nice catch, this was problematic if d2.time is FOREVER?

erlingrj (approved):
These changes all look good to me. Nice work!

core/src/main/java/org/lflang/federated/extensions/CExtensionUtils.java
edwardalee: Yes, this sort function would not work with forever. I've updated the new test to use forever.

PR #2484, opened 2025-03-28
Bumping up reactor-ts package version.

hokeun:
This should fix the issue with the unrecognized RTI message error in reactor-ts tests like this:
https://github.com/lf-lang/lingua-franca/actions/runs/14114135784/job/39540256256?pr=2482

New reactor-ts package has just been released: https://github.com/lf-lang/reactor-ts/releases/tag/v0.6.2


lhstrh:
This won't work because the package hasn't actually been released on [NPM](https://www.npmjs.com/package/@lf-lang/reactor-ts). I can help with that.

lhstrh (approved):
👍

PR #2487, opened 2025-04-07
Do not subtract network delays from STAA and refactor code

edwardalee:
This PR makes the following changes in the code generator:
1. Do not subtract network delays from STAA. It's not clear why this was done before.
2. Refactor the collection of unique STAA values using a SortedSet.

There is a companion reactor-c PR: https://github.com/lf-lang/reactor-c/pull/520


erlingrj (approved):
This, and the corresponding PR in reactor-c looks good to me!

Issue #2489, opened 2025-04-08
Federated example failed


zhuyipeng:
![Image](https://github.com/user-attachments/assets/c90b577b-31e8-451a-afa9-7dd3c6b6700a) 

in bin directory, there is no "bin/Federated_s" and "bin/Federated_d"

![Image](https://github.com/user-attachments/assets/b6c716bd-00bd-499d-82ae-d11b21248cd6)
in "fed-gen/federated" directory, it contains "federate__c" and "federate__p".

 Tried with 
```sh

RTI -n 2
./federate__c 
./federate__p

```
Encounter error like:
```sh

Fed 0 (c): FATAL ERROR: Received from RTI an unrecognized TCP message type: 1a.

```

![Image](https://github.com/user-attachments/assets/a40896e5-87df-4ccb-bf82-b4d3651d4538)

edwardalee:
This happens if the RTI was compiled from a different version of LF than the federates.  (I really regret our decision to separate out the RTI rather than code generating it along with the federation.)

zhuyipeng:
@edwardalee how should I fix it accordingly?

erlingrj:
What is your lfc version (`lfc --version`)

zhuyipeng:
@erlingrj  I am using vscode extension, 

![Image](https://github.com/user-attachments/assets/33d71bc7-1d5e-4920-85ce-1e9d17bfea76)

erlingrj:
OK, this is using v0.9.0 of `lfc`. You need to install the corresponding version of the RTI, this can be done by cloning reactor-c and checking out commit 3d7715c39fc40ad3c4c29918e724dc5b96738ca5

See here for a guide: https://www.lf-lang.org/docs/writing-reactors/distributed-execution#installation-of-the-rti

Note that this guide just clones the main branch, you must do `git checkout 3d7715c39fc40ad3c4c29918e724dc5b96738ca5` before building the RTI.

edwardalee:
Alternatively, you can switch to the Pre-Release version in the VSCode window above.  This will update the lfc code generator to the one in the nightly build.

PR #2490, opened 2025-04-08
Add semantic versioning to the RTI and verify it

erlingrj:
We have had several bug reports recently (#2489 ) because users have had the wrong version of the RTI installed. This PR addresses this issue by adding semantic versioning to the RTI and verifying that the correct version is installed. This requires us to manually bump the RTI version by updating the version in the CMakeLists.txt in `core/federated/RTI`. 

The output looks like this:
```sh
> RTI -v
RTI 0.1.19 e21869f9f 2025-04-08
```
First the version (in this case set to 0.1.19), then the commit hash, and finally the build date.

This is then checked in the federate launch script. Note that we only verify the major and the minor version (the first two numerals of the version). So this means that a particular version of `lfc` will require that the RTI version is e.g. 0.1.x, where x is any number. This allows us to add patches etc to the RTI without needing to update `lfc`. But if bigger change is introduced we can update the minor or the major version, then `lfc` will also have to be updated.

Companion reactor-c PR: https://github.com/lf-lang/reactor-c/pull/521

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -349,20 +349,40 @@ public class FedLauncherGenerator {
     return String.join("\n", commands);
   }
 
+  private String rtiRequiredVersion = "0.2.x";
+  private String rtiInstallInfo = "The RTI source code can be obtained from"
+            + " https://github.com/lf-lang/reactor-c/tree/main/core/federated/RTI";
+
+  private String getRtiVerifyVersionCode() {
+    return String.join(
+        "\n",
+        "# Verify that the correct version of the RTI is installed",
+        "RTI_REQUIRED_VERSION=\"" +  rtiRequiredVersion + "\"",
+        "RTI_REQUIRED_VERSION_NO_PATCH=$(echo $RTI_REQUIRED_VERSION | cut -d. -f1,2)",
+        "if ! which RTI > /dev/null; then",
+        "    echo \"ERROR: RTI executable not on PATH\"",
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -349,20 +349,40 @@ private String getRtiCommand(List<FederateInstance> federates, boolean isRemote)
     return String.join("\n", commands);
   }
 
+  private String rtiRequiredVersion = "0.2.x";
```
erlingrj, line 352: This string defines the RTI version required by lfc. Where should this be defined? THe TargetConfig, or maybe FileConfig?

edwardalee: I think version matching needs to be a little more sophisticated. Several versions of the RTI could be compatible with several versions of federates.  See https://github.com/lf-lang/reactor-c/pull/521#pullrequestreview-2749699933 for a suggestion about how to handle this.

edwardalee (approved):
This might be a bit too constraining, requiring an exact match.  Ideally, enhancements to the protocol could be made backward compatible.  I.e., a federate could join with new message type rather than `MSG_TYPE_FED_IDS` and get a response that includes the RTI version. The new message type would specify the protocol version desired by the federate. If the RTI responds with a lesser version, it may still be possible for the federate to work, if it has been designed to work with an older protocol if necessary.  Perhaps all of this could be a separate PR, but then I'm not sure how to handle the exact match required here. See https://github.com/lf-lang/reactor-c/pull/521#pullrequestreview-2749699933 .

PR #2491, opened 2025-04-08
Remove federate connections that have no effect

### LINKED ISSUE #2478: Federated causality-cycle check fails when input ports have no effect due to mistake in TPO assignment
This was reported here #2475 but I am moving it into a separate issue.

The problem occurs for programs like this:

Image

This will report an incorrect causality cycle when compiled. The reason can be found if we inspect the generated LF code for federate B:

```
@_fed_config()
main reactor  {
    b = new B()   
    @_tpoLevel(-3)
    ns_federate__a = new NetworkSender_0<{= 0 =}>()
    @_tpoLevel(2)
    nr_federate__b = new NetworkReceiver_0()
    _bin0 = new _bin()
    _bin0.port -> b.in
    nr_federate__b.msg -> _bin0.port0
    b.out -> ns_federate__a.msg
}
```

The problem here is that the NetworkSender is assigned a TPO (total port order) of -3, while the NetworkReceiver is assigned a TPO of 2. This creates a causality cycle because NetworkSender is an effect of NetworkReceiver. I have not looked into the TPO algorithms, but it does appear strange to me to both specify the TPO of network inputs and network outputs, I would think it would suffice to give the TPO of the inputs and let the outputs be derived from those?

Image

edwardalee:
This PR eliminates network connections if there are either no source reactions (nothing is ever sent on the connection) or there are no destination reactions (no received message ever triggers a reaction).

This PR also fixes an overflow error, where if a federated input is unconnected, an incorrect calculation of TPO (total port order) resulted from overflow when the level was given `Integer.MAX_VALUE`. This condition, however, should now never occur because the connection will be optimized away. Nevertheless, the code now cannot exhibit this overflow error.

Closes #2478.

### DIFF
--- a/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
+++ b/core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
@@ -397,7 +397,8 @@ public class FedASTUtils {
     // that it is connected to, which both have the same downstream reaction, have the correct
     // ordering wrt each other.
     var ub = p.getLevelUpperBound(index);
-    // Adjust the level so that input levels are even and output levels are odd, unless the level is Integer.MAX_VALUE,
+    // Adjust the level so that input levels are even and output levels are odd, unless the level is
+    // Integer.MAX_VALUE,
     // which occurs if a port has no dependent reactions.
     int level = Integer.MAX_VALUE;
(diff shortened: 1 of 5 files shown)


erlingrj (approved):
LGTM!

PR #2492, opened 2025-04-10
Compile RTI for each federation

erlingrj:
This PR addresses some recent issues we have had with incompatible versions of LFC and the RTI. This PR copies the RTI into the fed-gen folder and produces a trivial `CMakeLists.txt` there. This allows the RTI to be compiled just like a federate, the executable is copied into fed-gen/bin. The existing infrastructure for distributing src-gen and compiling remotely is used for the RTI.

A small change is needed in reactor-c: https://github.com/lf-lang/reactor-c/pull/523

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -130,13 +130,22 @@ public class FedLauncherGenerator {
     // Launch the RTI in the foreground.
     if (host.equals("localhost") || host.equals("0.0.0.0")) {
       // FIXME: the paths below will not work on Windows
-      shCode.append(getLaunchCode(getRtiCommand(federates, false))).append("\n");
+      shCode
+          .append(
+              getLaunchCode(getRtiCommand(fileConfig.getRtiBinPath().toString(), federates, false)))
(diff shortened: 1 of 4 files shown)


erlingrj:
@edwardalee is this the direction you were thinking?

edwardalee (approved):
Modulo the two FIXMEs:
```
    // FIXME: Add LOG_LEVEL to the RTI build.
    // FIXME: Add auth-stuff to the RTI build.
```
this looks good!

core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
```diff
@@ -424,56 +434,54 @@ private String getRemoteLaunchCode(
    * @param remoteBase The root directory on the remote machine.
    * @param federate The federate to distribute.
    */
-  private String getDistCode(Path remoteBase, FederateInstance federate) {
+  private String getDistCode(Path remoteBase, String name, String user, String host) {
```
erlingrj, line 437: The function is not changed. Only its arguments have changed so that it can be used to generate distribution code for the RTI as well

erlingrj:
@edwardalee this is ready to be merged now. Took me quite a while to get it working with remote execution of the RTI

PR #2497, opened 2025-04-26
Fix bug to inherit `preamble` on instantiated reactors.

Jakio815:


edwardalee:
Looks to me like the test failures are related to this change, since they involve redefinitions of symbols that are presumably defined in preambles.  It's interesting that only benchmark tests fail. These are older code. It may be that your fix is exposing a flaw in the benchmark code.  At some point, we changed things so that you could no longer define function bodies in file-level preambles. Those have to be defined within reactors, although their interface could be declared in a file-level preamble.  I think this requires a close look at the failing tests.

Jakio815:
I see. I'll try having a look on the benchmark-tests. Is there a suggestion? Should I see the `C/savina/src` ?

edwardalee:
I think the benchmark tests run programs in the benchmarks-lingua-franca repo: https://github.com/lf-lang/benchmarks-lingua-franca

edwardalee (approved):
With my revision, this looks good to go.

PR #2498, opened 2025-04-28
Show forever, never, and SI units in toString()

edwardalee:
This PR changes the toString() method of TimeValue to return "forever" and "never" when appropriate.
It also changes the default string representation in TimeUnit to use SI units, like "ms" instead of "msec".

### DIFF
--- a/core/src/main/java/org/lflang/TimeUnit.java
+++ b/core/src/main/java/org/lflang/TimeUnit.java
@@ -39,13 +39,13 @@ import java.util.stream.Collectors;
  */
 public enum TimeUnit {
   /** Nanoseconds. */
-  NANO("ns", "nsec", "nsecs"),
+  NANO("nsec", "ns", "nsecs"),
   /** Microseconds. */
-  MICRO("us", "usec", "usecs"),
+  MICRO("usec", "us", "usecs"),
   /** Milliseconds. */
-  MILLI("ms", "msec", "msecs"),
+  MILLI("msec", "ms", "msecs"),
   /** Seconds. */
-  SECOND("s", "sec", "secs", "second", "seconds"),
+  SECOND("sec", "s", "secs", "second", "seconds"),
   /** Minute. */
   // NOTE: Do not use MIN as the first entry. Common macro for minimum.
   MINUTE("minute", "min", "mins", "minutes"),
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/lflang/TimeUnit.java
```diff
@@ -58,10 +59,16 @@ public enum TimeUnit {
 
   private final Set<String> allNames;
   private final String canonicalName;
+  private final String siName;
 
-  TimeUnit(String canonicalName, String... aliases) {
+  TimeUnit(String canonicalName, String siName, String... aliases) {
```
lsk567, line 64: Note that this change would designate the `siName` of the `WEEK` enum to `weeks` (plural), which is kind of strange. Also, certain units here are not SI units, which technically make `siName` a misnomer.
In addition, `canonicalName` does not seem to be used anywhere, which is not good.

How about still calling the first field `canonicalName`, but with a preference for SI units and abbreviations?

```suggestion
  TimeUnit(String canonicalName, String... aliases) {
```

Then the time units above would be 
```java
  // Changes:
  // - Making canonical names SI units or abbreviations.
  // - Add a `wk` abbreviation for WEEK

  /** Nanoseconds. */
  NANO("ns", "nsec", "nsecs"),
  /** Microseconds. */
  MICRO("us", "usec", "usecs"),
  /** Milliseconds. */
  MILLI("ms", "msec", "msecs"),
  /** Seconds. */
  SECOND("s", "sec", "secs", "second", "seconds"),
  /** Minutes. */
  MINUTE("min", "minute", "mins", "minutes"),
  /** Hours. */
  HOUR("h", "hour", "hours"),
  /** Days. */
  DAY("d", "day", "days"),
  /** Weeks. */
  WEEK("wk", "week", "weeks"),
  ;
```

Then at the bottom, `toString` still returns canonical names.

```java
  @Override
  public String toString() {
    return this.canonicalName;
  }
```

I think this is a simpler and cleaner design.

edwardalee: This simpler design is what I tried first. It resulted in many test failures because of this function:

```
  public static String cMacroName(TimeUnit unit) {
    return unit.getCanonicalName().toUpperCase();
  }
```

lsk567: I see. The C runtime wants to use `MSEC(...)`, `NSEC(...)`, etc in macros.Then I am okay with the current design.

lsk567 (approved):

PR #2499, opened 2025-05-01
New `update` policy

lsk567:
This is a companion PR for https://github.com/lf-lang/reactor-c/pull/528, which renames the existing `replace` policy to `update` and introduces a new `replace` policy that drops the old event and schedules the new event.

In this PR, existing test cases that use the old `replace` policy have been updated to using the `update` policy.

### DIFF
--- a/core/src/main/java/org/lflang/TimeUnit.java
+++ b/core/src/main/java/org/lflang/TimeUnit.java
@@ -24,9 +24,8 @@
 
 package org.lflang;
 
-import static org.lflang.util.CollectionUtil.immutableSetOf;
-
 import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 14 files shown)


lsk567:
I don't know why GitHub says reactor-c still has conflicts to be resolved. The reactor-c submodule in this branch is already updated against the latest `main` branch.

edwardalee:
> I don't know why GitHub says reactor-c still has conflicts to be resolved. The reactor-c submodule in this branch is already updated against the latest `main` branch.

Just do `git add` on the reactor-c conflict and push the update to resolve the conflict.  This happens to me all the time.

edwardalee (changes requested):
On reflection, I'm not sure about the terminology change.  The term `update` literally suggests "increasing the date," which seems at least to give a good mnemonic for remembering the difference.  So maybe we should stay backward compatible and not change the meaning of `replace` and use `update` for the new policy?  This also has the advantage of being backward compatible and therefore not requiring changing any tests. I'm loath to make changes that are not backward compatible unless there is a compelling reason.

lsk567:
I didn't think of backward compatibility, and I agree this is a good idea.

lsk567:
This PR is not yet ready to merge. Test cases are still needed for the new `update` policy.

edwardalee (approved):
LGTM!

Issue #1748, opened 2023-05-18
Scope Visibility in Inheritance [C-Target]


mkhubaibumer:
In case of Inheritance, Reactor Arguments and Contained Reactors are not visible in Child Reactor citing "Couldn't resolve reference to Parameter"

```
target C;

reactor foo {
    input in:int;
    output out:int;
}

reactor Parent (n_triggers:int(1)) {
    input[n_triggers] in:int;
    output[n_triggers] out:int;

    f = new foo();
}

reactor Child extends Parent {
    // Not Working!
    input[n_triggers] another_in:int;

    // Not Working
    reaction(f.in) -> f.out {=
        // User Logic
    =}

    // Working
    reaction(in) -> out {=
        // User Logic
    =}
}
```

lhstrh:
Thanks for the report! The problem is that the [scope provider](https://github.com/lf-lang/lingua-franca/blob/master/org.lflang/src/org/lflang/scoping/LFScopeProviderImpl.java) hasn't been adjusted properly.

edwardalee:
In ASTUtils.java, there are a bunch of methods like `allParameters()` that return a list of parameters that include parameters of base classes, as opposed to the XText-generated function getParameters(), which returns only the local ones.  I suspect that there is lurking somewhere uses of getParameters() and similar functions where allParameters() should be used.

edwardalee:
PR #2500 fixes these issues.

PR #2500, opened 2025-05-08
Fix issues with inheritance

edwardalee:
This is a starting point for fixing issues with inheritance, including #1748.

### DIFF
--- a/core/src/main/java/org/lflang/ast/ASTUtils.java
+++ b/core/src/main/java/org/lflang/ast/ASTUtils.java
@@ -326,7 +326,8 @@ public class ASTUtils {
 
   /**
    * Produce a unique identifier within a reactor based on a given based name. If the name has not
-   * been used before, it is returned; if has been used, an index is appended that makes the name unique.
+   * been used before, it is returned; if has been used, an index is appended that makes the name
+   * unique.
    *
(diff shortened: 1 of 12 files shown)


Jakio815 (approved):

Jakio815:
Will the vs-code extension diagram generator need to be modified separately? Or will it be updated if we push it to main?

edwardalee:
> Will the vs-code extension diagram generator need to be modified separately? Or will it be updated if we push it to main?

It will be updated in the repo for the extension, but not in the marketplace.  We should do a release... We are ready...

Jakio815:
Still I think we can use the pre-lease version. Isn't it the nightly build updated everyday?

Jakio815:
I updated the vs-code extension to the nightly version, but still diagram does not show up properly. Do we also need to fix the vs-code generator?

However, `lfd` works correctly.

edwardalee:
> I updated the vs-code extension to the nightly version, but still diagram does not show up properly. Do we also need to fix the vs-code generator?
> 
> However, `lfd` works correctly.

I have pushed an update to the vscode repo to use the current master of lingua-franca.  I think this will result in the nightly version properly handling inheritance, but probably not until tomorrow.

PR #2503, opened 2025-05-12
Update async callback tests

lsk567:
Async callback tests only need a toggle when two back-to-back physical actions both live on the event queue. For test cases testing the min. spacing policies, using a toggle is not required. This PR removes the unnecessary toggles and fixes the misuse of logical action in `AsyncCallbackUpdate.lf`.


edwardalee (approved):
Wow, very subtle tests! LGTM!

PR #2517, opened 2025-06-08
Actually check federated test results

### LINKED ISSUE #2516: Python federated tests ignore errors
Unfortunately, none of the Python federated tests are actually testing anything except compilation. If a federate fails, the overall process reports success.  Here is a minimal example of a test that should always fail:

```
target Python
reactor Exit {
  reaction(startup) {=
    exit(1)
  =}
}
federated reactor {
  e = new Exit()
}
```
If you compile and run this, the process always exits with return status 0:
```
% lfc-dev src/Junk.lf
...
% bin/Junk
...
Fed 0 (e): ERROR: FATAL: Calling reaction _exit.reaction_function_0 failed.
ERROR: RTI: Socket to federate 0 is closed. Exiting the thread.
WARNING: Failed to accept the socket. Software caused connection abort.
RTI is exiting.
RTI has exited. Wait for federates to exit.
All done.
% echo $?
0
```
This means we haven't been testing Python federated for a while.

edwardalee:
This fixes #2516 by checking the exit status of federates.

### DIFF
--- a/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
+++ b/core/src/main/java/org/lflang/federated/launcher/FedLauncherGenerator.java
@@ -212,7 +212,7 @@ public class FedLauncherGenerator {
                 "# The errors are handled separately via trap.",
                 "for pid in \"${pids[@]}\"",
                 "do",
-                "    wait $pid || exit $?",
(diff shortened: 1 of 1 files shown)


edwardalee:
Oops, this fix is not correct because it causes the script to always exit with the first process.

Issue #1906, opened 2023-07-20
The Python test failure in Python's `PingPongDistributed` is the same as the one observed in `ChainWithDelay` where there is a segfault on line 427 of `gcmodule.c` in CPython due to the current `PyGC_Head` being `null` -- that is, the problem is that what should be a circular doubly linked list is actually not circular, because one of the nodes has a null pointer instead of pointing to either the next node or the sentinel node.


petervdonovan:
The Python test failure in Python's `PingPongDistributed` is the same as the one observed in `ChainWithDelay` where there is a segfault on line 427 of `gcmodule.c` in CPython due to the current `PyGC_Head` being `null` -- that is, the problem is that what should be a circular doubly linked list is actually not circular, because one of the nodes has a null pointer instead of pointing to either the next node or the sentinel node.

Furthermore, I have reproduced this segfault in `master`. To reproduce, it is sufficient to increase the number of iterations by increasing the `count` variable by a couple orders of magnitude; this increases the probability of running into the error. **So this bug shouldn't stop us from merging into `master`.**

The fact that only containers (Python lists, dictionaries, etc.) are included in the linked list of possibly GC-able objects narrows down the scope of what we have to search for. Furthermore, because we have multiple tests (`ChainWithDelay` and `PingPongDistributed`) that fail in the same way consistently, this makes it seem like this is related to what is happening in this GC list, rather than just being general undefined behavior from an unrelated part of the program. The only problem is that the segfault cannot be reproduced using `python3-dbg`, which makes it hard to view the state of the GC list in a debugger.

_Originally posted by @petervdonovan in https://github.com/lf-lang/lingua-franca/issues/1684#issuecomment-1644488770_

PR #2518, opened 2025-06-09
Do not register atexit to mask errors

### LINKED ISSUE #2516: Python federated tests ignore errors

edwardalee:
This fixes #2516 by moving the `atexit` handler that was masking error exits from Python tests to be registered only after all LF code has completed and exited. Thus, if tests fail and call `exit(1)` or `sys.exit(1)`, the failure will no longer be masked and will be reported in CI.

@petervdonovan gives an explanation below of why the `atexit` hook is being used at all.


petervdonovan:
> which disables Python garbage collection. Do we really want to do that? Why?

Probably the reason for the atexit hook was that I thought the GC would run on shutdown even if `gc.disable` was called. This seems to be mentioned [here](https://instagram-engineering.com/dismissing-python-garbage-collection-at-instagram-4dca40b29172). But I'm not sure about that.

Based on [this](https://github.com/lf-lang/lingua-franca/issues/1906), it looks like disabling GC was a workaround for an open issue. [This comment](https://github.com/lf-lang/lingua-franca/pull/1684#issuecomment-1644504822) indicates that GC (which is only for reference cycles) "shouldn't" be necessary, but the referenced commit was a workaround that does not fix the true underlying problem.

If #1906 cannot be reproduced anymore, probably the whole commit can be reverted.

edwardalee:
> If #1906 cannot be reproduced anymore, probably the whole commit can be reverted.

The only test failure in this PR seems completely unrelated, in "Test Rust Benchmarks".  ???
See https://github.com/lf-lang/lingua-franca/actions/runs/15528831488/job/43744165050?pr=2518

PR #2520, opened 2025-06-10
Update rust runtime

oowekyala:
Note: you need to merge https://github.com/lf-lang/reactor-rs/pull/53 first.


edwardalee:
These changes are already in master with a still more recent version of the rust runtime.

PR #2522, opened 2025-06-11
Update to KLighD 3.1.0 and Xtext 2.37.0

NiklasRentzCAU:
This version bumps the KLighD version up to 3.1.0.

KLighD builds on Xtext 2.37, so this dependency is updated as well.
KLighD 3.1.0 now also builds with Java 21, so this PR will need to update your build environment to use Java 21 as well if you want to merge this PR. I did not include this here, because I do not know your build infrastructure and how to properly test it.

Changing to 3.1.0 will have the following benefits for you:
- Eclipse 2025-03 will not throw an exception when opening the diagram widget anymore
- your Language Server will be usable outside of VS Code in a standalone application again, see eclipse-xtext/xtext#3391
- KLighD now supports theming, you may modify your synthesis to use the new API for dark themes. See kieler/semantics#96 and kieler/KLighD#175


edwardalee (approved):
LGTM!

edwardalee:
Unfortunately, the xText download fails... ???

NiklasRentzCAU:
It seems like this is not a problem with this PR, but with the main branch in general - that Itemis site that is used by every build does seem to be offline at the moment. I assume it will probably come online again this week when they notice the issue.

Even without that this PR would probably fail because KLighD now requires Java-21. This is up for discussion if you would want to adopt to the newer Java version.

PR #2531, opened 2025-07-25
Add lf_set_deadline() example as DeadlilneUpdate.lf

Jakio815:


test/C/src/DeadlineUpdate.lf
```diff
@@ -0,0 +1,64 @@
+// Test whether the lf_update_deadline() function works.
```
hokeun, line 1:
```suggestion
// Test of lf_update_deadline() API for updating deadlines dynamically.
```

test/C/src/DeadlineUpdate.lf
```diff
@@ -0,0 +1,64 @@
...
+
+  // Test Case 1: Verify that decreasing the deadline works.
+  // The reaction starts with a declared deadline of 100 ms, but we update it
+  // to 50 ms at runtime. A deadline violation should therefore be detected
+  // around physical time 50 ms . If no violation is detected
+  // after 50 ms, the test fails.
+  reaction(startup) -> a {=
+    self->i = 0;
+    lf_update_deadline(self, MSEC(50));
+    while(!lf_check_deadline(self, true)){
```
hokeun, line 25: I don't think this needs a while loop here, and can be simplified.
Also, how about you reduce the sleep to lf_sleep(MSEC(10)) and set the initial deadline to 500 msec, to make this test less flaky and take less time to execute?

test/C/src/DeadlineUpdate.lf
```diff
@@ -0,0 +1,64 @@
...
+
+  // Test Case 2: Verify that increasing the deadline works.
+  // The reaction is declared with a 50 ms deadline, but we update it
+  // to 100 ms at runtime. A violation should NOT be detected at 50 ms,
+  // because the updated deadline is 100 ms. Entering the deadline
+  // handler below would indicate failure.
+  reaction(a) {=
+    self->i = 0;
+    lf_update_deadline(self, MSEC(100));
+    while(!lf_check_deadline(self, true)){
```
hokeun, line 47: Similar comment here, this can be further simplified, no need to have a while loop.

test/C/src/DeadlineUpdate.lf
Jakio815: Fixed this.

test/C/src/DeadlineUpdate.lf
Jakio815: Fixed this.

PR #2539, opened 2025-08-08
Made injector singleton and register the ide module an startup of LS

### LINKED ISSUE #2367: No Content Assist in Editor
See https://github.com/lf-lang/vscode-lingua-franca/issues/166

VS Code has no content assist for lingua franca.

I recreated the ticket here since it is a problem in the this repository and to assign it to a milestone.

soerendomroes:
Fixes #2367

Since I did only test that the language server starts and that the VS Code now has content assist, someone should test whether this somehow affects the cli or point me to everything I should also test out.


core/src/testFixtures/java/org/lflang/tests/TestBase.java
```diff
@@ -285,8 +284,7 @@ protected static boolean isLinux() {
    */
   public static void runSingleTestAndPrintResults(
       LFTest test, Class<? extends TestBase> testClass, TestLevel level) {
-    Injector injector =
-        new LFStandaloneSetup(new LFRuntimeModule()).createInjectorAndDoEMFRegistration();
+    Injector injector = new LFStandaloneSetup().createInjectorAndDoEMFRegistration();
```
lhstrh, line 287: Looks like a similar change is also needed here: https://github.com/lf-lang/lingua-franca/blob/ab5e741d0445f3200c7977a159a3a9b4a6f882f2/cli/lfc/src/test/kotlin/org/lflang/cli/LfcIssueReportingTest.kt#L103C24-L103C93

edwardalee (commented):
What is the status of this? There is a test failure due to unresolved warnings and an open issue.

PR #2559, opened 2025-10-01
A sample code for comparing ForSyDe with LF

EhsanKhodadad:
This PR adds a simple test to the test/C/src/patmos folder. This test compares LF coding with ForSyDe, described in this paper: https://ieeexplore.ieee.org/document/11121303


edwardalee:
I don't understand the point of this test. It will always fail because it calls `exit(1)` after 10 iterations. It's also not clear in what way this compares LF to ForSyDe.  This is just a rather trivial feedback system.  To be a proper test, it should check the behavior, eg by testing the values at the sink, and exit in an orderly fashion with exit code 0 if the test passes.  Exiting with a non-zero exit code indicates a test failure.

PR #2576, opened 2026-01-01
Turn `trace-plugin` property into a boolean flag

lsk567:
The overall goal of this PR is to support a [new trace plugin](https://github.com/lf-lang/lf-trace-xronos) that streams data to the Xronos dashboard.

This PR lets `trace-plugin` accept a boolean flag rather than the name of the plugin. The string was passed to `find_library()` in the generated cmake file to locate the library. But since the new approach to link against plugins relies on `cmake-include`, this is no longer necessary.

A companion PR is: https://github.com/lf-lang/reactor-c/pull/554.

### DIFF
--- a/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
+++ b/core/src/main/java/org/lflang/generator/c/CCmakeGenerator.java
@@ -307,9 +307,8 @@ public class CCmakeGenerator {
             });
     // Add trace-plugin data
     var tracePlugin = targetConfig.getOrDefault(TracePluginProperty.INSTANCE);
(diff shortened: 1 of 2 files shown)


edwardalee (approved):
This looks good to me. Perhaps we could add a test in the lingua-franca repo that uses the default trace implementation as if it were an externally provided trace implementation? Otherwise, without an example, it's hard to see how a user could make use of the trace mechanism.

lhstrh:
Let me reference https://github.com/lf-lang/reactor-c/pull/554#issuecomment-3704392277, as I probably should have placed it here (and not in the companion reactor-c PR).

PR #2579, opened 2026-01-19
Fix bug with includes for custom port types

edwardalee:
In certain complicated combinations of import and extend, a bug would arise where a top-level preamble would not be included in files that need it to be.  This PR adds a test that previously failed to compile and is fixed by using a better mechanism for identifying when to include a top-level preamble.

This also fixes a bug where a base class might specify a `cmake-include` target property that might not get used even if a subclass is instantiated in the program.

This also aligns reactor-c with the latest main.


lsk567:
FYI, there is a parallel effort to fix the broken ROS CI in #2580.

edwardalee:
> FYI, there is a parallel effort to fix the broken ROS CI in #2580.

Yes, I know. I've been watching it. I'll reverse these changes if/when that succeeds.

PR #2590, opened 2026-02-05
Tests for the function `update_deadline` in Python target

byeonggiljun:
This PR adds Python tests for the `update_deadline` function in the Python target. Additionally, this test moves the flaky test `test/C/src/DeadlineUpdateIncrease.lf` into `test/C/src/failing` to prevent undeterministic failures.

There are two companion PRs: 
1. reactor-c PR: [reactor-c/pull/558](https://github.com/lf-lang/reactor-c/pull/558) and 
2. lf-python-support PR: [lf-python-support/pull/30](https://github.com/lf-lang/lf-python-support/pull/30)


edwardalee (approved):
The DeadlineUpdateDecrease test looks to me like it could be flaky. Does it sometimes fail in CI?  I'm not sure how to make it non-flaky, but perhaps calling `update_deadline` inside a deadline handler that is assured of being triggered (by an upstream reaction with a large sleep), and then calling `check_deadline` (with argument `false` to avoid a stack overflow) to verify that the deadline did in fact decrease.  Could you try this? It would be better to have a non-flaky test, even if failure is rare.

edwardalee:
> Could you please elaborate on the flakiness you found?

Oh, on closer examination, the test no longer looks flaky!  I'm not sure what I was seeing...  Actually, the test is pretty clever in avoiding flakiness.

byeonggiljun:
Thank you, @edwardalee, for suggesting the fix! I applied it, and the tests are passing as well.

PR #2623, opened 2026-05-18
Fix Python target on macOS with statically-linked Python (e.g., conda)

depetrol:
### Symptom

On macOS with a statically-linked host Python (such as conda), Python-target programs fail at import with:

```
SystemError: initialization of LinguaFranca<name> did not return an extension module
```

### Root cause

The Python target's generated CMake links each extension `.so` against `libpython` unconditionally. This is the failure mode described in [CPython issue 42616 — *"C Extensions on Darwin that link against libpython are likely to crash"*](https://bugs.python.org/issue42616): on macOS, when the host `python3` is statically linked (the conda default), an extension that links `libpython` causes dyld to load a second copy of `libpython.dylib` into the process. The two copies have independent `_PyRuntime` state, and when the extension's init function calls into the Python C API the call lands in the uninitialized second runtime — which then re-runs the full interpreter startup and typically dies on `site` / `.pth` processing.

## The fix

On macOS, build the extension with `-undefined,dynamic_lookup` instead of linking `${Python_LIBRARIES}`, so Python's symbols resolve against the already-running host interpreter at import time and no second `libpython` is loaded.

### Implications

The fix seems to work fine also in non-conda environments on my macOS, but I am unsure if this would break other environments.


depetrol:
I don't think the failing ts-test is related to this change...

edwardalee:
> I don't think the failing ts-test is related to this change...

I have pushed a fix for this failing test to master, so if you merge in master, the failure should go away.

edwardalee (approved):
This looks reasonable to me.

PR #2627, opened 2026-05-23
Added support for extends keyword in Rust target

Eggsrael:
Added support for `extends` keyword in the Rust target. This was done to move the Rust target closer to feature parity with the other language targets. 

Key Changes made 
* Refactored `RustModel::makeReactorInfors()` 
    * created a function from the  reactor -> ReactorInfo lambda
    * three extension functions to handle \<type\>Info creating
    * changed `reactor.parameters` call to the `ASTUtils.allParameters()`
* Four tests for `extends`
    * `test/Rust/src/InheritanceAction.lf`
    * `test/Rust/src/InheritedInstance.lf`
    * `test/Rust/src/InheritedParameter.lf`
    * `test/Rust/src/InheritedReaction.lf`
* Added a Rust Collections test
    * `test/Rust/src/Collections.lf`


Eggsrael:
I messed up the branch with a fast-forward merge so I will be making a different PR

Issue #2626, opened 2026-05-21
Bug on decentralized coordination for Banks?


Jakio815:
I was testing [BankIndex.lf](https://github.com/lf-lang/lingua-franca/blob/master/test/C/src/federated/BankIndex.lf) in the C target.

However, the test failed if I used the `coordination:decentralized` option.

<img width="1809" height="1051" alt="Image" src="https://github.com/user-attachments/assets/078cb827-a47d-414b-a3a3-ab9f37a644ff" />

I verified that the `bank_index` feature was properly printed, the problem is somewhere on the network side...

edwardalee:
This is an interesting example that underscores that we probably don't have the right defaults for `maxwait` for decentralized coordination.  To work in the current master, this example needs to set `absent_after` to `forever` on the connection. It currently doesn't work to just set `maxwait` to `forever` because the decentralized treats the start time as a special case: It commits to the start time regardless of the status of inputs and regardless of the `maxwait` value.  Arguably, it should perform any specified wait before committing to the start time so that the start time operates like any other time.  Also, arguably, the default `maxwait` should be `forever`, not zero as it currently is.

These two changes are made in [this PR](https://github.com/lf-lang/reactor-c/pull/583), but these are not backward-compatible changes.  I'm evaluating the consequences.

PR #2629, opened 2026-05-24
Make default maxwait forever

edwardalee:
This PR changes decentralized coordination so that the default `maxwait` is `forever`, not zero. This is not a backward-compatible change, but it means that switching from centralized to decentralized coordination is much more likely to yield the same behavior by default. I.e., it makes the default more deterministic.

When reviewing, note particularly the tests that had to change because of the new default for `maxwait`. In particular, the following change is sometimes needed for federated applications that use the decentralized coordinator:

* If no `maxwait` was previously given on instantiations (and no legacy `STP_offset` or `STA` parameter was given in the class definition), then the default of zero was previously used. Now the default is `forever`. If your application needs a `maxwait` of zero, then add the attribute `@maxwait(0)` just prior to the instantiation of the federate.

The default value for `absent_after` (legacy `STAA`) remains zero because any other value will have the same effect when the `maxwait` is forever.

This also changes the diagram rendering so that any `maxwait` that is explicitly specified is shown in the diagram, even if it equals the default.

Companion PR: https://github.com/lf-lang/reactor-c/pull/583

This addresses https://github.com/lf-lang/lingua-franca/issues/2626.


edwardalee:
@copilot resolve the merge conflicts in this pull request. My local copy of the repo shows no conflicts.

edwardalee:
@copilot resolve the merge conflicts in this pull request

core/src/main/java/org/lflang/AttributeUtils.java
```diff
@@ -340,12 +351,15 @@ public static TimeValue getMaxWait(EObject node) {
     if (attr != null) {
       // The attribute is expected to have a single argument of type Time or the literal "0".
       // The validator checks this.
-      final var time = attr.getAttrParms().get(0).getTime();
+      final var parm = attr.getAttrParms().get(0);
+      final var time = parm.getTime();
```
hokeun, line 355: Nit: I don't think this change (adding `var parm`) is necessary.

hokeun (approved):
Looks good to me! Just left a minor comment. 

By the way, would this change affect the playground examples like these?:
- https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/resettable-counter/ResettableCounter.lf
- https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/resettable-counter/ResettableCounterLF.lf
- https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/traffic-lights-control/trafficLights.lf

core/src/main/java/org/lflang/AttributeUtils.java
edwardalee: Actually, `parm` is used three lines below.

test/C/src/federated/Dataflow.lf
```diff
@@ -1,4 +1,5 @@
 target C {
+  build-type: debug,
```
Jakio815: Are we planning to keep this?

Jakio815 (approved):
LGTM, but as prof Kim mentioned, we should also check the playground, and other repos also.

test/C/src/federated/Dataflow.lf
edwardalee: Removed.

edwardalee:
> Looks good to me! Just left a minor comment.
> 
> By the way, would this change affect the playground examples like these?:
> 
> * https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/resettable-counter/ResettableCounter.lf
> * https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/resettable-counter/ResettableCounterLF.lf
> * https://github.com/lf-lang/playground-lingua-franca/blob/main/examples/Python/src/traffic-lights-control/trafficLights.lf

Good point! I will update those next.

PR #2632, opened 2026-05-28
Implement protobuf serialization between LF federates in Python runtime and add a test

hokeun:
> NOTE: This PR passes all Python tests, except for "Report to CodeCov", which failed due to a network connection error, not because of the changes in this PR. Please see https://github.com/lf-lang/lingua-franca/issues/2633 for more details about this flaky test issue.

This PR implements the protobuf serialization between LF federates in the Python target. This PR also adds a simple test to make sure that protobuf serialization works.

### DIFF
--- a/core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
+++ b/core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
@@ -10,6 +10,7 @@ import org.lflang.federated.generator.FederationFileConfig;
 import org.lflang.federated.launcher.RtiConfig;
 import org.lflang.federated.serialization.FedCustomPythonSerialization;
 import org.lflang.federated.serialization.FedNativePythonSerialization;
+import org.lflang.federated.serialization.FedProtoPythonSerialization;
 import org.lflang.federated.serialization.FedSerialization;
 import org.lflang.federated.serialization.SupportedSerializers;
 import org.lflang.generator.CodeBuilder;
@@ -70,7 +71,7 @@ public class PythonExtension extends CExtension {
 
   @Override
   public String getNetworkBufferType() {
-    return "void*";
+    return "";
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


hokeun:
NOTE: This PR passes all Python tests, except for "Report to CodeCov", which failed due to a network connection error, not because of the changes in this PR.

Jakio815:
Is this actually a serializer?

In the C target, serialization is fully automated by the compiler. However, in this Python implementation, the user still has to manually call `SerializeToString()` and `ParseFromString()` inside their reactions. 
Looking at `FedProtoPythonSerialization.java`, the compiler doesn't actually perform any protobuf serialization/parsing. 
* **Sender:** It simply calls `PyBytes_AsStringAndSize` to extract raw bytes from the already-serialized Python `bytes` object:
  ```java
  if (PyBytes_AsStringAndSize(varName, &BUF_VAR, &LEN_VAR) == -1) { ... }

* **Receiver**: It just wraps the received raw bytes back into a Python bytes object via `PyBytes_FromStringAndSize`.
Since the compiler is not doing any packing under the hood, serializer "proto" is functionally acting as a raw bytes serializer (which bypasses pickle to allow raw binary payloads).

I don't have a better idea than this implementation right now, but is this what we want?

edwardalee:
I agree. The reaction should not call `serializeToString`.   The `serializeToString` method should be called by the serializer infrastructure.  I'm not sure how to do that because the serializer infrastructure is C functions, not Python functions. But it should be possible.

hokeun:
@Jakio815 @edwardalee Thank you both! Sorry, it was clearly an oversight. I made changes accordingly.

edwardalee (approved):
LGTM!  One thing worth checking before merging: Do you get a reasonable error message if you specify `serializer "proto"` but then try to send an ordinary Python object? I think you will, but worth checking.

hokeun:
> LGTM! One thing worth checking before merging: Do you get a reasonable error message if you specify `serializer "proto"` but then try to send an ordinary Python object? I think you will, but worth checking.

Thanks! I tried sending an int like
```
target Python {
  protobufs: ProtoHelloWorld.proto,
  timeout: 2 sec
}

reactor Sender {
  output out

  state count = 0

  timer t(0, 1 sec)

  reaction(t) -> out {=
    self.count += 1
    protoHelloWorld = ProtoHelloWorld.ProtoHelloWorld()
    protoHelloWorld.name = "Hello World"
    protoHelloWorld.number = self.count
    out.set(self.count)
  =}
}
```

And as you can expect, no warning was generated during `lfc-dev`, but during runtime I got an error like this:
```
AttributeError: 'int' object has no attribute 'SerializeToString'
Fed 0 (sender_main): FATAL ERROR: Failed to call SerializeToString on proto object.
```

I believe this error is reasonable and understandable. Would you have any further suggestions?

Jakio815:
LGTM also!

Jakio815 (approved):

edwardalee:
> I believe this error is reasonable and understandable. Would you have any further suggestions?

LGTM.

PR #2635, opened 2026-05-29
Implement Polyglot target supporting a mixture of C and Python targets.

hokeun:
This PR implements the Polyglot target to support a mixture of C and Python target federates in Lingua Franca (LF). The LF syntax to specify the Polyglot target is `target Polyglot`, where you can include common target configs. The new Polyglot target only supports federated reactors, not main reactors. To specify the target language of each federate (top-level reactor) in the Polyglot target LF program, we can use annotations such as `@language(C)` and `@language(Python)`. Currently, we only allow C and Python as target languages for the federates in the Polyglot target. Below is an example program where the Sender reactor is written in C, and the Receiver reactor is written in Python. For network communication, we use protobufs.

```
target Polyglot {
  protobufs: [ProtoHelloWorld.proto],
  timeout: 2 sec
}

@language(C)
reactor Sender {
  output out: ProtoHelloWorld*

  state count: int = 0

  timer t(0, 1 sec)

  reaction(t) -> out {=
    self->count++;
    ProtoHelloWorld* msg = (ProtoHelloWorld*)malloc(sizeof(ProtoHelloWorld));
    proto_hello_world__init(msg);
    msg->name = "Hello World";
    msg->number = self->count;
    lf_set(out, msg);
  =}
}

@language(Python)
reactor Receiver {
  input inp

  state count = 0

  reaction(inp) {=
    print(f"Received: name=\"{inp.value.name}\", number={inp.value.number}.")
    if inp.value.number != self.count + 1:
      sys.stderr.write("Expected number " + str(self.count + 1) + ".\n")
    self.count += 1
  =}
}

federated reactor {
  sender = new Sender()
  receiver = new Receiver()

  sender.out -> receiver.inp after 100 msec serializer "proto"
}
```

In addition, the Polyglot target also works via imports, where the reactors in different targets are imported from library reactors. Below is an example program where the sender reactor is imported from the Python target and the receiver reactor is imported from the C target.

```
target Polyglot {
  protobufs: [ProtoHelloWorld.proto],
  timeout: 2 sec
}

import PythonSender from "../lib/PythonProtoSenderReceiver.lf"
import CReceiver from "../lib/CProtoSenderReceiver.lf"

federated reactor {
  sender = new PythonSender()
  receiver = new CReceiver()

  sender.out -> receiver.in after 100 msec serializer "proto"
}

```

This PR also adds the CI test for the new Polyglot target, as well as enabling the serialization tests in C and Python.


core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
```diff
@@ -206,8 +207,84 @@ protected void deserialize(
...
+                  + "->token->value, "
+                  + value
+                  + "->token->length);\n"
+                  + "PyObject* _lf_parse_result = PyObject_CallMethod("
+                  + FedSerialization.deserializedVarName
+                  + ", \"ParseFromString\", \"O\", _lf_proto_bytes_raw);\n"
+                  + "Py_XDECREF(_lf_proto_bytes_raw);\n"
+                  + "if (_lf_parse_result == NULL) {\n"
+                  + "    if (PyErr_Occurred()) PyErr_Print();\n"
+                  + "    lf_print_error_and_exit(\"Failed to parse proto message from"
```
edwardalee: Should this be a fatal error?  Maybe fatal errors should only be triggered by improper setup, not by improper messages arriving over the network?  Maybe a warning and drop the message?  I'm not sure.  The problem here might be a security risk: an attacker could take down a federate by sending a bogus message?  Similar question for the above fatal errors.

core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
```diff
@@ -259,13 +337,50 @@ protected void serializeAndSend(
...
+                && !dstPort.getType().getId().isEmpty();
+        if (dstPortIsTyped) {
+          // Destination port is typed (C receiver): send raw protobuf bytes via SerializeToString.
+          result.pr(
+              "PyObject* serialized_pyobject = PyObject_CallMethod("
+                  + variableToSerialize
+                  + ", \"SerializeToString\", NULL);\n"
+                  + "if (serialized_pyobject == NULL) {\n"
+                  + "    if (PyErr_Occurred()) PyErr_Print();\n"
+                  + "    lf_print_error_and_exit(\"Failed to call SerializeToString on proto"
```
edwardalee: Again, not clear this should be a fatal error.  Same for below.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
```diff
@@ -737,11 +739,36 @@ private List<FederateInstance> getFederateInstances(
...
+          (langName != null) ? Target.forName(langName) : java.util.Optional.<Target>empty();
+      if (langOpt.isEmpty()) {
+        messageReporter
+            .at(instantiation)
+            .error(
+                "Reactor '"
+                    + reactorDef.getName()
+                    + "' must have a @language(C) or @language(Python) annotation in Polyglot"
+                    + " mode.");
+        federateTarget = Target.C; // fallback to avoid NPE
```
edwardalee, line 771: For an imported target, will it work for the imported file to have a `target C` or `target Python` designation? I think this would usually give a better design than mixing languages in the same file, and it would be particularly useful, maybe, for library reactors (although I guess they would have to be designed to use compatible serializers anyway).

core/src/main/java/org/lflang/target/Target.java
```diff
@@ -335,7 +335,14 @@ public enum Target {
       // are those that are a valid expression. Others may be escaped
       // with the syntax r#keyword.
       Arrays.asList("self", "true", "false")),
-  UC("uC", true, Target.C.keywords);
+  UC("uC", true, Target.C.keywords),
+  /**
+   * A meta-target for multi-language federated programs. Each reactor must carry a
+   * {@code @language(C)} or {@code @language(Python)} annotation. The main reactor must be a
+   * {@code federated reactor}. The lfc resolves each federate's actual target from its annotation
+   * before compiling.
```
edwardalee: Again, it seems that importing files with `target C` and `target Python` will likely be cleaner, except for small programs.

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -1206,6 +1206,27 @@ public void checkReactor(Reactor reactor) throws IOException {
...
+      }
+      if (!reactor.isFederated() && !reactor.isMain()) {
+        // Every non-federated reactor must declare its language via @language.
+        String lang = AttributeUtils.getAttributeValue(reactor, "language");
+        if (lang == null) {
+          error(
+              "Reactor '"
+                  + reactor.getName()
+                  + "' must have a @language(C) or @language(Python) annotation in Polyglot mode.",
+              Literals.REACTOR_DECL__NAME);
```
edwardalee: Again, support imports?

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -1359,6 +1391,46 @@ public void checkAttributes(Attribute attr) {
     }
   }
 
+  private void checkLanguageAttribute(Attribute attr) {
```
edwardalee, line 1448: Should there be a check that every top-level instantiation has either an `@language` attribute or is imported?  Perhaps instantiations should be checked also so that you can't instantiate a Python reactor deep inside a C reactor.

edwardalee (approved):
Wow, very cool!  One big issue is that I think in many cases it would be better to be able to import a `target C` class definition and a `target Python` class definition without having to include in those files the `@language` attribute.  This could be done in a subsequent PR if you want to break down the pieces, but in the long run, I think this will be cleaner than mixing languages in one file (though I don't see any good reason to not support that).

core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
hokeun: Great point! I addressed this in https://github.com/lf-lang/lingua-franca/pull/2635/changes/e0e84184898ab064980be241b2f9f4d1f9c2cbda where I changed the fatal errors to warnings + droping messages for the improper messages over the network.

core/src/main/java/org/lflang/federated/extensions/PythonExtension.java
hokeun: Also addressed this in https://github.com/lf-lang/lingua-franca/commit/e0e84184898ab064980be241b2f9f4d1f9c2cbda

core/src/main/java/org/lflang/federated/generator/FedASTUtils.java
```diff
@@ -258,8 +258,11 @@ private static void addNetworkReceiverReactor(
       Resource resource,
       MessageReporter messageReporter) {
     LfFactory factory = LfFactory.eINSTANCE;
+    // Use the destination federate's target extension for the receiver reactor.
+    // (Using srcFederate here was a latent bug that only surfaces in Polyglot where
+    // src and dst federates may have different targets.)
     var extension =
-        FedTargetExtensionFactory.getExtension(connection.srcFederate.targetConfig.target);
+        FedTargetExtensionFactory.getExtension(connection.dstFederate.targetConfig.target);
```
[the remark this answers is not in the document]
hokeun: This has been addressed in https://github.com/lf-lang/lingua-franca/pull/2635/changes/f9c2b932415aa20521caaa1c410e3ab4a0f2a65e.

core/src/main/java/org/lflang/federated/generator/FedGenerator.java
hokeun: Thank you for the great suggestion. I implemented this in https://github.com/lf-lang/lingua-franca/pull/2635/changes/627342db5ecd14d8021c3e83cc5a54c3c667a27b and subsequent commits.

core/src/main/java/org/lflang/target/Target.java
hokeun: Fixed this in https://github.com/lf-lang/lingua-franca/pull/2635/changes/d3fcada7ad2ffbc2fb56660fa90c2b54e08152b6!

core/src/main/java/org/lflang/validation/LFValidator.java
hokeun: Thanks, LFValidator has been fixed in https://github.com/lf-lang/lingua-franca/pull/2635/changes/d3fcada7ad2ffbc2fb56660fa90c2b54e08152b6.

core/src/main/java/org/lflang/validation/LFValidator.java
hokeun: Has been fixed in https://github.com/lf-lang/lingua-franca/commit/d3fcada7ad2ffbc2fb56660fa90c2b54e08152b6.

PR #2637, opened 2026-06-03
Support changing Rust Editions in Rust target

Eggsrael:
Allow for users to choose which edition of Rust they would like to use. Currently the Rust edition is hard coded with the 2018 edition. Allowing users to specify what edition should allow for greater control over the code.

 If users do not specify a target it will continue to default to 2018.

Example of specifying a rust edition: 
```rust
Rust  {
    edition: "2021"
}

main reactor Foo { 
...
}
```
This will be the output for the reactor's `Cargo.toml`
```toml
[package]
name = "Foo"
version = "1.0.0"
authors = ["Author"]
edition = "2021"
```

If an invalid edition is chosen, such as `"2022"`,  an error will be outputted as follows: 
```
target Rust {
     edition: "2022"
              ^^^^^^ Edition: 2022 is not a valid edition. Please use one of the following:  2018, 2021, 2024
}
```

These versions are specified in `.../target/property/RustEditionProperty.java`


hokeun (approved):
It looks like a useful and probably also necessary target config for Rust!
Do you have any use case where specifying the Rust edition other than 2018 is necessary? It may be worth adding an .lf test using this feature as an example.

Eggsrael:
With regards to editions I haven't seen many errors with the existing test cases except with ActionValuesCleanup.lf failing on 2024. This is due to the test case using `static mut` on an `AtomicBool`. Pre 2024 the compiler would just send a warning that it can lead to UB because `AtomicBool` has [Interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html#interior-mutability). 

So instead of allowing this: 
```rust
  preamble {=
    use std::sync::atomic::AtomicBool;
    use std::sync::atomic::Ordering;
    // set to true when destructor is called
    static mut DROPPED: AtomicBool = AtomicBool::new(false);

    #[derive(Clone, Debug)]
    struct FooDrop { }
    impl std::ops::Drop for FooDrop {
      fn drop(&mut self) {
        unsafe {
          DROPPED.store(true, Ordering::SeqCst);
        }
      }
    }
  =}
```

2024 edition makes you drop `static mut` which is considered safe: 
```rust
  preamble {=
    use std::sync::atomic::AtomicBool;
    use std::sync::atomic::Ordering;
    // set to true when destructor is called
    static DROPPED: AtomicBool = AtomicBool::new(false);

    #[derive(Clone, Debug)]
    struct FooDrop { }
    impl std::ops::Drop for FooDrop {
      fn drop(&mut self) {
          DROPPED.store(true, Ordering::SeqCst);
      }
    }
  =}
```

One addition I can think of right now is if-let chains 
```rust
// 2024
if let Some(foo) = foo_in && foo.has_something() && let Some(bar) = bar_in { 
    ... // <- foo and bar are in scope
}

// pre 2024
if let Some(foo) = foo_in { 
  if foo.has_something() {
    if let Some(bar) = bar_in { 
       ... // <- foo and bar are in scope 
    }
  }
  // <- foo is in scope
}

```
Setting the rust version to be newer than 2018 doesn't have many issues but looking at it again 2015 is causing issues. Ill drop that year.

edwardalee:
This looks like a nice enhancement.
Ideally, we would check that all the tests pass in all the supported editions.
Can we modify the CI to run the tests with multiple editions?
This does have the downside of slowing down CI.

Eggsrael:
I don't think that is necessary as the compiler does not change when changing the editions. The compiler only changes when you update `rustc`.  When you change the edition it tells the compiler what AST to use and once that gets translated to the intermediate languages edition does not matter. I can update all of the tests and Benchmarks to be edition agnostic (2018-2024)  it should not be that hard as it is mostly version bumping.

edwardalee:
Sounds good!

edwardalee (approved):

PR #2642, opened 2026-06-11
Raising error for unsupported decentralized coordination for TypeScript target.

hokeun:
Problem: Currently, `lfc` compiles a TypeScript LF program with decentralized coordination in the target config without any error, even though decentralized coordination is not supported for TypeScript. Federated execution is supported with TypeScript, but only with centralized coordination. 

- Relevant PR for fixing the handbook documentation: https://github.com/lf-lang/lf-lang.github.io/pull/340

### DIFF
--- a/core/src/main/java/org/lflang/validation/LFValidator.java
+++ b/core/src/main/java/org/lflang/validation/LFValidator.java
@@ -119,6 +119,8 @@ import org.lflang.lf.WidthSpec;
 import org.lflang.lf.WidthTerm;
 import org.lflang.target.Target;
 import org.lflang.target.TargetConfig;
+import org.lflang.target.property.CoordinationProperty;
+import org.lflang.target.property.type.CoordinationModeType.CoordinationMode;
 import org.lflang.util.FileUtil;
 
 /**
@@ -1328,6 +1330,19 @@ public class LFValidator extends BaseLFValidator {
     String lfFileName = FileUtil.nameWithoutExtension(target.eResource());
     if (Character.isDigit(lfFileName.charAt(0))) {
       errorReporter.nowhere().error("LF file names must not start with a number");
+    }
+    if (targetOpt.isPresent() && targetOpt.get() == Target.TS && target.getConfig() != null) {
+      for (var pair : target.getConfig().getPairs()) {
+        if (CoordinationProperty.INSTANCE.name().equals(pair.getName())) {
+          String value = ASTUtils.elementToSingleString(pair.getValue());
+          if (CoordinationMode.DECENTRALIZED.toString().equals(value)) {
+            error(
+                "The TypeScript target does not support decentralized coordination.",
+                Literals.TARGET_DECL__CONFIG);
+          }
+          break;
+        }
+      }
     }
   }
 
--- a/core/src/main/java/org/lflang/validation/LFValidator.java
+++ b/core/src/main/java/org/lflang/validation/LFValidator.java
@@ -119,8 +119,6 @@ import org.lflang.lf.WidthSpec;
 import org.lflang.lf.WidthTerm;
 import org.lflang.target.Target;
 import org.lflang.target.TargetConfig;
-import org.lflang.target.property.CoordinationProperty;
-import org.lflang.target.property.type.CoordinationModeType.CoordinationMode;
 import org.lflang.util.FileUtil;
 
 /**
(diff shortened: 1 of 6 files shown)


core/src/main/resources/lib/ts/pnpm-workspace.yaml
```diff
@@ -1,3 +1 @@
-allowBuilds:
-  '@lf-lang/reactor-ts': true
-  'microtime': true
+dangerouslyAllowAllBuilds: true
```
hokeun, line 1: Fixing the error of "[ERR_PNPM_GIT_DEP_PREPARE_NOT_ALLOWED] Failed to prepare git-hosted package fetched from "https://codeload.github.com/lf-lang/reactor-ts/tar.gz/87621f4c5a50b6a0c80cff43a1830938d83b340b": The git-hosted package "@lf-lang/reactor-ts@0.6.2" needs to execute build scripts but is not in the "allowBuilds" allowlist."
The root cause is clear: pnpm 10.x changed allowBuilds matching — for git-hosted packages it now requires the full resolved URL (with commit hash), not just the package name. Since the hash is only known after pnpm downloads the package, the static pnpm-workspace.yaml template can't pre-authorize it by name alone.

The reliable fix is `dangerouslyAllowAllBuilds: true`. Despite the name, it's safe here because these are generated, isolated project directories where we explicitly control every installed package, there's no risk of a malicious package sneaking in.

hokeun: Why `allowBuilds: { '@lf-lang/reactor-ts': true }` fails: pnpm 10 (specifically 10.0–10.1) only matches git-hosted packages against their resolved specifier, i.e., the codeload URL with the specific commit hash (@lf-lang/reactor-ts@https://codeload.github.com/.../tar.gz/<sha>). The package name alone doesn't match. Since the hash is only known after pnpm fetches the package, we can never put the correct key in a static template file.

Why `dangerouslyAllowAllBuilds: true` works: This is a first-class pnpm 10 setting (defined in pnpm's config schema as dangerouslyAllowAllBuilds?: boolean) that bypasses the allowBuilds check entirely. It's safe to use here because these are generated, isolated project directories, every installed package is explicitly listed in a package.json we control, so there's no risk of a malicious package sneaking in.

core/src/main/java/org/lflang/validation/LFValidator.java
```diff
@@ -1329,6 +1331,19 @@ public void checkTargetDecl(TargetDecl target) throws IOException {
...
+          String value = ASTUtils.elementToSingleString(pair.getValue());
+          if (CoordinationMode.DECENTRALIZED.toString().equals(value)) {
+            error(
+                "The TypeScript target does not support decentralized coordination.",
+                Literals.TARGET_DECL__CONFIG);
+          }
+          break;
+        }
+      }
+    }
```
Jakio815: I think checking the coordination mode doesn't belong here in `LFValidator.java`. 

`LFValidator` is responsible for general AST checks (like verifying the target name or file naming rules). Hardcoding target-specific rules for the `coordination` property here is not ideal. 

A much more appropriate place is `CoordinationProperty.validate()`. Other target properties (like `workers` or `single-threaded`) already validate their target-specific constraints in their own property classes.

Even if we move the logic there, the editor will still give a red line, because it's checked in the code block below (`public void checkTargetProperties(KeyValuePairs targetProperties) {`.

I'll push a quick fix on this.

Jakio815: I just pushed it https://github.com/lf-lang/lingua-franca/pull/2642/commits/017ca881b8856aa4d4b834631960caeaac9adc8c, let's see if the CI passes.

hokeun: Ah, you're right. Thank you for the fix!

edwardalee (approved):
Looks OK to me.  The `dangerouslyAllowAllBuilds` setting makes me nervous, but the explanation sounds reasonable.

hokeun:
> Looks OK to me. The `dangerouslyAllowAllBuilds` setting makes me nervous, but the explanation sounds reasonable.

Thank you! Using `allowBuilds `fails due to a known limitation in pnpm 10 that prevents name-based allowlisting for git-hosted tarball dependencies, making `dangerouslyAllowAllBuilds: true` the only viable workaround for our CI tests. This flag should be safe here since our CI environment uses programmatically isolated, strictly controlled project directories with no exposure to unvetted third-party code.

PR #2669, opened 2026-07-23
Embedded device support in Rust target

Eggsrael:
I'm putting this in drafts so eyes can be put on what I have been working on and get opinions.

In the embedded Rust ecosystem there are two main approaches, the blocking [embedded-hal](https://github.com/rust-embedded/embedded-hal) driver ecosystem, and [Embassy](https://embassy.dev/)'s async-first model. There are pros and cons to both but I decided to just work with the Embedded-Hal based drivers.  Embassy makes `async` a first class citizen so everything must be a task. This would need a complete rewrite of how code generation is handled and Embassy does not support OS deployment. Everything provided by Embedded-hal is implemented as a `trait` so you can easily take a crate or driver written into it and deploy it with little to no friction. Embedded-hal is also just a more mature and fleshed out solution. 

Currently I'm starting to work on porting the reactor-rs crate to be `no_std` so it can be detached from a libc or OS. The goal is to move to `core` which is dependency free primitives library that std is built off of. There is also `alloc` which is heap allocated data structures which may need to be used as well. 

my goal is to make adding board as painless as possible but, due to the nature of embedded systems that just might not be possible. I have been working on two boards that I have on hand, esp32 and rp2350 and I'm basing it off of the templates for those that are provided by the [esp-hal](https://github.com/esp-rs/esp-hal) and [rp-hal](https://github.com/rp-rs/rp-hal) teams.

This is mainly a prototype to gauge interest as I still don't know if this approach will be viable once reactor-rs is working.

### DIFF
--- a/core/src/main/java/org/lflang/target/Target.java
+++ b/core/src/main/java/org/lflang/target/Target.java
@@ -642,6 +642,7 @@ public enum Target {
               SingleThreadedProperty.INSTANCE,
               WorkersProperty.INSTANCE,
               RustEditionProperty.INSTANCE,
+              PlatformProperty.INSTANCE,
               DockerProperty.INSTANCE);
       case TS ->
           config.register(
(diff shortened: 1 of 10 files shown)


Eggsrael:
Possible `time::Instant` replacement for `no_std` target.
https://docs.rs/fugit/latest/fugit/

PR #2677, opened 2026-08-12
Support numbers in import package names

edwardalee:
This PR allows imported package names to have numbers so that package names like `pololu-3pi-c` are supported. This required adjusting the grammar so that such names work after tokenizing.


hokeun (approved):
Thank you for the fix! Looks good to me!

